PR #2006, opened 2016-07-11
noms-based fuse implementation

ahl:
I was hoping folks might take a look at the state of the noms-based fused stuff. It's not 100% ready to go, but I'd appreciate your feedback. If there are big, structural things you'd like me to take care of, feel free to point those out and I'll make those changes before you look in finer detail.

Left to do:
* Work out some issues I ran into regarding Cycle types
* Deal with the condition of head being out of sync (i.e. if the dataset is mounted somewhere else)
* A couple more fuse methods (chmod, chown, truncate)
* *testing* -- been testing with some scripts, but I think there's a neat way to test without needing fuse installed
* Reorganize the code to be a little more sensical

If you want to run this I think it's as simple as grabbing github.com/hanwen/go-fuse and FUSE (I'm using osxfuse.github.io on my Mac which available in Mac Ports).


ahl:
re-opening with my new, post-launch fork

PR #2207, opened 2016-07-30
Support reading from stdin in url-fetch

kalman:
I need this in the short term because CSV raw files are too large to
check into github, so they need to be split up. It's easier to just do
"cat * | url-fetch -stdin" than recombining them on the file system.


kalman:
this PR is busted because of the repo changing

PR #2208, opened 2016-07-30
Nomdex

willhite:
Two things that need to be improved:
   testing of complex and/or operations on ranges.
   print status during index update.


willhite:
A first version of nomdex is available as a PR. I can build an index on the ny-vehicle-registrations dataset (~120 keys and ~11M records) in about 11 minutes and using about 11Gb of memory. So I'm wondering where to go from here. There seem to be 2 immediate issues that are worth thiinking about.
1) performance -- it would be nice to be less susceptible to running out of memory based on size. Currently, I have to build all the value sets of the index in memory. I could reduce this vulnerability if we had streaming sets.
2) generality -- indexes that support < and > queries need to be ordered in some meaningful way. Currently the only indexes that can support these operations are strings, numbers, and booleans.

3) queries with multiple indexes -- currently, the query language can only use one index. In order to support multiple indexes we need to have a function that selects items using a key-based cursor and returns them in order by Ref.

samples/go/nomdex/nomdex_command.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	flag "github.com/tsuru/gnuflag"
+)
+
+type nomdexCommand struct {
```
aboodman: Much of this infrastructure for subcommands is shared with the `noms` command -- Did you consider factoring it out?

samples/go/nomdex/nomdex_update.go
```diff
@@ -0,0 +1,126 @@
...
+	typeCache := map[*types.Type]bool{}
+
+	index := Index{m: IndexMap{}}
+
+	walk.AllP(rootObject, db, func(v types.Value, r *types.Ref) {
+		typ := v.Type()
+		typeCacheMutex.Lock()
+		hasPath, ok := typeCache[typ]
+		typeCacheMutex.Unlock()
+		if !ok || hasPath {
```
aboodman: This is cool, but does it always work? Say that the input path was `[42]`. The first `List<Foo>` we see might be only 40 elements long, and so this would fail. That doesn't mean that the next list we see won't have 43 elements.

aboodman: I think in order to do this kind of trickery we need to look at the path more carefully. It might be possible to eliminate certain paths, but not this easily.

@kalman is doing work right now to make it easier to inspect the path components individually.

samples/go/nomdex/nomdex_find.go
```diff
@@ -0,0 +1,112 @@
...
+		if !ok {
+			return nil, types.Map{}, fmt.Errorf("%s is not a valid index", outDsArg)
+		}
+	}
+
+	typ := types.MakeMapType(
+		types.MakeUnionType(types.StringType, types.NumberType),
+		types.MakeSetType(types.ValueType))
+
+	if !types.IsSubtype(typ, index.Type()) {
```
aboodman: You should be able to express the type of the Commit itself, so you don't need to do the checking of whether the commit is a struct or if it is the commit type above.

go/types/map.go
```diff
@@ -218,6 +218,25 @@ func (m Map) IterAll(cb mapIterAllCallback) {
 	})
 }
 
+func (m Map) IterAllRange(start, end Value, includeStart, includeEnd bool, cb mapIterAllCallback) {
```
aboodman: @kalman should review.

aboodman:
This is really cool.

go/types/map.go
kalman: "All" vs "Range" seems contradictory. Can this be `IterRange()`?

Also could the caller just decide whether to include the start or end at runtime? It would make this interface simpler. `IterRange` can - and should document - that the range on either end is inclusive.

kalman: Also this needs a test.

aboodman:
I'm seeing a problem when running find on some datasets:

```
!? nomdex find --index='http://demo.noms.io/aa::repros/pull-2208.value' '_ > 40 and _ < 42'
Unable to open database/index
	error: http://demo.noms.io/aa::repros/pull-2208.value does not point to a suitable index type:
```

aboodman:
I think it is because that commit type is pretty complicated.

willhite:
@aboodman You're right the type is complex, lots of unions. One of the types is map<string, struct> and it needs to be a map<string, set<value>>. Can you point me to the dataset and build command that you were using? I'ld like to try and rebuild this bad index.

aboodman:
Sure:

```
noms sync http://demo.noms.io/aa::flickr/photos http://localhost:8000::aa/flickr/photos
nomdex up --in-path=http://localhost:8000/aa::flickr/photos --by='.geoposition.latitude' --out-ds='aa/flickr/photos/by-lat'
```

PR #2232, opened 2016-08-01
Validate type of new Root in root/ POST handler

### LINKED ISSUE #2116: noms server does not validate new Root type when POSTing to root/
Dataset commit validates that the commit struct is a valid commit struct type but the noms server does not validate incoming data type so it is possible for a client to commit something that isn't a valid commit struct type.

For example (pseudo):

```
// broken-client.go
notACommitStruct := tyoes.Number(42)
db.Commit(notACommitStruct, types.NewSet())
```

The code for Commit validates that we only commit a Commit struct that is a subtype of commit (`datas.IsCommitType`). However, a client might be buggy (or incomplete) and not do this check. It is therefore possible that the noms server gets something that should be a commit struct but really isn't.

cmasone-attic:
In the server side of the Remote Databse, the handler for
UpdateRoot now verifies that the new proposed Root is of a
legal type: empty map OR Map\<String, Ref\<Commit-like>>

Fixes #2116

### DIFF
--- a/go/datas/remote_database_handlers.go
+++ b/go/datas/remote_database_handlers.go
@@ -223,7 +223,7 @@ func handleRootGet(w http.ResponseWriter, req *http.Request, ps URLParams, rt ch
 	w.Header().Add("content-type", "text/plain")
 }
 
-func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, rt chunks.ChunkStore) {
+func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, cs chunks.ChunkStore) {
 	d.PanicIfTrue(req.Method != "POST", "Expected post method.")
 
 	params := req.URL.Query()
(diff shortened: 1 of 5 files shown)


go/datas/remote_database_handlers.go
```diff
@@ -234,8 +234,38 @@ func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, rt c
...
+	if !cs.UpdateRoot(current, last) {
 		w.WriteHeader(http.StatusConflict)
 		return
 	}
 }
+
+func isMapOfStringToRefOfCommit(m types.Map) bool {
+	mapTypes := m.Type().Desc.(types.CompoundDesc).ElemTypes
+	keyType, valType := mapTypes[0], mapTypes[1]
+	return keyType.Kind() == types.StringKind && (isRefOfCommitType(valType) || isUnionOfRefOfCommitType(valType))
```
rafael-atticlabs: @arv, can you look at this? I kind of feel like `isRefOfCommitType` should handle this case, no?

rafael-atticlabs:
lgtm. @arv should weigh in on checking commit type.

go/datas/remote_database_handlers.go
cmasone-attic: I tried that initially, and it does not handle this case. I think it's because this is actually a Union of Refs to something Commit-like, as opposed to a Ref to a Union of Commit-like things, and `isRefOfCommitType` handles only the latter.

go/datas/remote_database_handlers.go
```diff
@@ -234,8 +234,38 @@ func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, rt c
...
+	// Ensure that proposed new Root is present in cs
+	c := cs.Get(current)
+	d.PanicIfTrue(c.IsEmpty(), "Can't set Root to a non-present Chunk")
+
+	// Ensure that proposed new Root is a Map and, if it has anything in it, that it's <String, RefOfCommit>
+	v := types.DecodeValue(c, nil)
+	d.PanicIfTrue(v.Type().Kind() != types.MapKind, "Root of a Database must be a Map")
+	m := v.(types.Map)
+	if !m.Empty() && !isMapOfStringToRefOfCommit(m) {
+		panic(d.Wrap(fmt.Errorf("Root of a Database must be a Map<String, RefOfCommit>, not %s", m.Type().Describe())))
```
arv: `Root of a Database must be a Map<String, Ref<Commit>>`

arv:
LGTM

PR #2239, opened 2016-08-02
Rate-limit LevelDBStore read operations as well

### LINKED ISSUE #2227: Rate-limit Get/Has in LevelDBStore
Without doing this, we eventually run out of FDs when doing huge numbers of simultaneous Gets.

cmasone-attic:
Under load, our server can exhaust the number of file descriptors it's
allowed to have open at one time. Part of this is because of how many
incoming connections it's handling, but we believe that handling lots
of simultaneous reads to leveldb is the larger part of the issue.

This patch applies the rate limit we were using for writing to both
read and write operations.

Fixes #2227

### DIFF
--- a/go/chunks/leveldb_store.go
+++ b/go/chunks/leveldb_store.go
@@ -147,7 +147,7 @@ func (l *LevelDBStore) setVersIfUnset() {
 
 type internalLevelDBStore struct {
 	db                                     *rateLimitedLevelDB
-	mu                                     sync.RWMutex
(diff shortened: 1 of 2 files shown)


cmasone-attic:
PTAL

rafael-atticlabs:
locking around root doesn't look right.

cmasone-attic:
PTAL

rafael-atticlabs:
lgtm

PR #2242, opened 2016-08-02
add Blob.Splice()

ahl:
attn @arv

### DIFF
--- a/go/types/blob.go
+++ b/go/types/blob.go
@@ -34,6 +34,27 @@ func (b Blob) Reader() io.ReadSeeker {
 	return &BlobReader{b.seq, cursor, nil, 0}
 }
 
+func (b Blob) Splice(idx uint64, deleteCount uint64, data []byte) Blob {
+	if deleteCount == 0 && len(data) == 0 {
+		return b
+	}
+
+	d.Chk.True(idx <= b.Len())
+	d.Chk.True(idx+deleteCount <= b.Len())
+
+	cur := newCursorAtIndex(b.seq, idx)
+	ch := newSequenceChunker(cur, nil, makeBlobLeafChunkFn(b.seq.valueReader()), newIndexedMetaSequenceChunkFn(BlobKind, b.seq.valueReader()), hashValueByte)
+	for deleteCount > 0 {
+		ch.Skip()
+		deleteCount--
+	}
+
+	for _, v := range data {
+		ch.Append(v)
+	}
+	return newBlob(ch.Done(nil).(indexedSequence))
+}
+
 // Collection interface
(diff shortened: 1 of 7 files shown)


arv:
add to whitelist

arv:
LGTM

Do you want to do the JS too? Otherwise we should add a bug.

ahl:
I'll take a swing a the js stuff; fingers crossed that babel doesn't mess me up again.

ahl:
@arv can you take a look at the js side when you have a sec?

js/src/blob.js
```diff
@@ -39,6 +41,14 @@ export default class Blob extends Collection<IndexedSequence> {
   get length(): number {
     return this.sequence.numLeaves;
   }
+
+  splice(idx: number, deleteCount: number, ...insert: number[]): Promise<Blob> {
```
arv: I'm not sure we want `...insert: number[]` here. Maybe we should pass a `Uint8Array` here instead.

arv: `npm version minor` to let npm modules pick up the new API.

arv:
LGTM

@kalman Can you also take a look (TBR)?

go/types/blob.go
```diff
@@ -34,6 +34,27 @@ func (b Blob) Reader() io.ReadSeeker {
 	return &BlobReader{b.seq, cursor, nil, 0}
 }
 
+func (b Blob) Splice(idx uint64, deleteCount uint64, data []byte) Blob {
+	if deleteCount == 0 && len(data) == 0 {
```
kalman: why the `&& len(data) == 0`?

kalman: wait dumb question never mind

go/types/blob.go
```diff
@@ -34,6 +34,27 @@ func (b Blob) Reader() io.ReadSeeker {
...
+		return b
+	}
+
+	d.Chk.True(idx <= b.Len())
+	d.Chk.True(idx+deleteCount <= b.Len())
+
+	cur := newCursorAtIndex(b.seq, idx)
```
kalman: 2 main comments:
1. A more Go-ish interface here would be to implement Splice using the Go `io.*` libraries. I can't think of a good way to do it though :\
2. We should file an issue to make this look like NewStreamingBlob. Boxing/unboxing each byte is super inefficient, which is why that custom chunking code exists. If Splice becomes performance sensitive, especially if `data` becomes large, then we should fix it.

kalman: But lgtm.

kalman: (it would be good to file that issue then reference it here - note that JS doesn't do the fancy blob writing)

PR #2246, opened 2016-08-02
Change "Amazon Dynamo" to "Amazon DynamoDB"

willglynn:
Nitpick: "Amazon Dynamo" mention in the intro should be "Amazon DynamoDB".

[Amazon Dynamo](https://en.wikipedia.org/wiki/Dynamo_(storage_system)) is a particular system with which `DynamoStore` does not interact.


aboodman:
Nice catch. Sorry to be pedantic, but in order to keep the IP clean, we need to have you sign a CA in order to accept a patch (even for something trivial like this):

http://noms.io/ca_individual.html
http://noms.io/ca_corporation.html

willglynn:
Done!

willglynn:
And I rebased to resolve the merge conflict.

aboodman:
Thanks for the contribution!

Issue #2247, opened 2016-08-02
HRS: Empty map could be cleaner


arv:
We currently write:

```
Map<, >({})
```

This would look better as `Map<>` or `Map<,>`.

PR #2252, opened 2016-08-02
Clean up writing of empty sets

aaparella:
Change the human readable encoding of sets when empty.

Issue #2247

### DIFF
--- a/go/types/encode_human_readable.go
+++ b/go/types/encode_human_readable.go
@@ -205,6 +205,9 @@ func (w *hrsWriter) WriteTagged(v Value) {
 		w.Write(v)
 	case BlobKind, ListKind, MapKind, RefKind, SetKind, TypeKind, CycleKind:
 		w.writeType(t, nil)
+		if l, ok := v.(lenable); ok && l.Len() == 0 {
+			break
+		}
 		w.write("(")
 		w.Write(v)
 		w.write(")")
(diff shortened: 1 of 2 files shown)


arv:
We need a CLA to be signed: http://noms.io/ca_individual.html

rafael-atticlabs:
@aaparella, thanks so much for the patch. Can you sign the CLA: http://noms.io/ca_individual.html?

go/types/encode_human_readable_test.go
```diff
@@ -406,9 +406,9 @@ func TestEmptyCollections(t *testing.T) {
 	c := MakeMapType(BlobType, NumberType)
 	assertWriteTaggedHRSEqual(t, "Type(Map<Blob, Number>)", c)
 	d := NewMap()
-	assertWriteTaggedHRSEqual(t, "Map<, >({})", d)
+	assertWriteTaggedHRSEqual(t, "Map<, >", d)
```
arv: This was the output I didn't like. I would prefer it to print `Map<>({})` or `Map<,>({})`.

I think we should always print the value.

aaparella: Oh I misunderstood the issue. So you're saying it was the spacing of the types that was the problem?

arv: Yes, specifically the space after the comma before the empty union.

aaparella:
Signed the CLA and updated the branch. Let me know if this is more of what you had in mind.

go/types/encode_human_readable.go
```diff
@@ -246,7 +246,10 @@ func (w *hrsWriter) writeType(t *Type, parentStructTypes []*Type) {
 		w.write("<")
 		for i, et := range t.Desc.(CompoundDesc).ElemTypes {
 			if i != 0 {
-				w.write(", ")
+				w.write(",")
+				if et.Kind() != UnionKind {
```
arv: I don't think this will do the right thing in case of non empty unions. Desired output:

```
Map<T, V | W>
Map<T,>
```

Maybe add a test for this too?

PR #2254, opened 2016-08-02
Blob stdin

kalman:


kalman:
@willhite could you PTAL at this patch? You reviewed the same thing earlier, but the PR got busted or orphaned or something.

I went back to the --stdin flag and added a test for files.

willhite:
LGTM

PR #2261, opened 2016-08-03
Dataset cleanup

cmasone-attic:
1) ValueStore now maintains validation cache info across Commit() calls,
   so we can remove a workaround in Dataset.Commit()
2) Update comment for setNewHead
3) setNewHead was decomposing a Commit object just so it could pass the pieces to Dataset.Commit()...which put them all back together again. Seemed silly.

### DIFF
--- a/go/dataset/dataset.go
+++ b/go/dataset/dataset.go
@@ -87,7 +87,6 @@ func (ds *Dataset) Commit(v types.Value, opts CommitOptions) (Dataset, error) {
 	if (parents == types.Set{}) {
 		parents = types.NewSet()
 		if headRef, ok := ds.MaybeHeadRef(); ok {
-			headRef.TargetValue(ds.Database()) // TODO: This is a hack to deconfuse the validation code, which doesn't hold onto validation state between commits.
 			parents = parents.Insert(headRef)
 		}
 	}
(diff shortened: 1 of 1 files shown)


cmasone-attic:
PTAL

go/dataset/dataset.go
```diff
@@ -135,15 +134,9 @@ func (ds *Dataset) validateRefAsCommit(r types.Ref) types.Struct {
...
-// returned WITHOUT newHeadRef in it. The caller should take any necessary corrective action and try
-// again using this new Dataset.
+// setNewHead attempts to make the object pointed to by newHeadRef the new Head of ds. First, it checks that the object exists in ds and validates that it decodes to the correct type of value. Next, it attempts to commit the object to ds.Database(). This may fail if, for instance, the Head of ds has been changed by another goroutine or process. In the event that the commit fails, the error from Database().Commit() is returned along with a new Dataset that's at it's proper, current Head. The caller should take any necessary corrective action and try again using this new Dataset.
 func (ds *Dataset) setNewHead(newHeadRef types.Ref) (Dataset, error) {
 	commit := ds.validateRefAsCommit(newHeadRef)
-	p := commit.Get(datas.ParentsField).(types.Set)
-	m := commit.Get(datas.MetaField).(types.Struct)
-	return ds.Commit(commit.Get(datas.ValueField), CommitOptions{Meta: m, Parents: p})
+	store, err := ds.Database().Commit(ds.id, commit)
+	return Dataset{store, ds.id}, err
```
arv: I'm worried by this. The old code computed the commit type. Are we sure that the newHeafRef is going to have the same commit type?

cmasone-attic: Where did the old code do this computation? Both the old code and the new code call `validateRefAsCommit`, so we still know that it is, in fact, a Commit object. Is that what you mean?

rafael-atticlabs: Let's discuss when you get in. Logically, sync is setting head to a new ref here. If the resulting ref that's gets set is different, that seems --- bad. (I.e. This code looked strange to me as well, and I would like to think that @cmasone-attic's change is correct).

rafael-atticlabs:
lgtm. let's wait on response from @arv

arv:
I walked through all the code here and NewCommit with the same value, parents and meta produces an identical Commit object.

LGTM

PR #2262, opened 2016-08-03
Simplify Path construction in Go

kalman:
There are 3 ways to construct a Path: via methods like AddIndex, via
parsing/AddPath, and simply via using append() (since Path is a slice).
This patch deletes the former and leaves only (Maybe)ParsePath and
append() with the ability to manually create the various path parts.

### DIFF
--- a/cmd/noms/diff/diff.go
+++ b/cmd/noms/diff/diff.go
@@ -74,7 +74,7 @@ func diffLists(w io.Writer, p types.Path, v1, v2 types.List) (err error) {
 				if shouldDescend(lastEl, newEl) {
 					idx := types.Number(splice.SpAt + i)
 					writeFooter(w, &wroteHdr)
-					err = diff(w, append(p, types.NewIndexPathPart(idx)), idx, lastEl, newEl)
+					err = diff(w, append(p, types.NewIndexPath(idx)), idx, lastEl, newEl)
 				} else {
 					writeHeader(w, p, &wroteHdr)
 					line(w, DEL, nil, v1.Get(splice.SpAt+i))
@@ -165,7 +165,7 @@ func diffOrdered(w io.Writer, p types.Path, lf lineFunc, df diffFunc, kf, v1, v2
 			c1, c2 := v1(change.V), v2(change.V)
 			if shouldDescend(c1, c2) {
 				writeFooter(w, &wroteHdr)
-				err = diff(w, append(p, types.NewIndexPathPart(k)), change.V, c1, c2)
+				err = diff(w, append(p, types.NewIndexPath(k)), change.V, c1, c2)
 			} else {
 				writeHeader(w, p, &wroteHdr)
 				lf(w, DEL, k, c1)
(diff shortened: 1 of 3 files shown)


cmd/noms/diff/diff.go
```diff
@@ -23,7 +23,7 @@ func shouldDescend(v1, v2 types.Value) bool {
 }
 
 func Diff(w io.Writer, v1, v2 types.Value) error {
-	return diff(w, types.NewPath(), nil, v1, v2)
+	return diff(w, types.Path{}, nil, v1, v2)
```
kalman: I want to think of Paths now as more literally []PathPart, with some helper functions. So removing NewPath and just using Path{} is more straightforward.

go/types/path.go
```diff
@@ -145,96 +123,121 @@ func (p Path) String() string {
...
+}
+
+// Indexes into Maps and Lists by key or index.
+type IndexPathPart struct {
+	// The value of the index, e.g. `[42]` or `["value"]`.
+	Index Value
+	// Whether this index should resolve to the key of a map, given by a `@key` annotation.
+	// Typically IntoKey is false, and indices would resolve to the values. E.g. given `{a: 42}` then `["a"]` resolves to `42`.
+	// If IntoKey is true, then it resolves to `"a"`. For IndexPathPart this isn't particularly useful - it's mostly provided for consistency with HashIndexPathPart - but note that given `{a: 42}` then `["b"]` resolves to nil, not `"b"`.
+	IntoKey bool
```
kalman: Now these are part of the public API, I want to give them better names.

go/types/path.go
```diff
@@ -19,56 +19,23 @@ import (
 
 var annotationRe = regexp.MustCompile("^@([a-z]+)")
 
-type Path []pathPart
+// A Path is an address to a Noms value - and unlike refs (i.e. #abcd...) they can address inlined values.
```
aboodman: This is a great comment! You are getting better :).

aboodman: Nit: `#abcd...` is a _hash_ in our nomenclature. `Ref` is the object that also includes type information.

aboodman:
lgtm

aboodman:
Double-checking: does this _remove_ any existing public API? I'm not sure how we should handle that. But the first time we do it post-launch, we should think about it and talk to folks like @vinibaggio to figure out a strategy.

kalman:
> does this remove any existing public API

Yes, not a good habit to get into. But it seems pretty innocuous in this case (especially since part of the reason I'm doing this is to reduce burden in porting to JS).

go/types/path.go
kalman: Done

kalman:
I'm also considering renaming `FieldPathPart` to just `FieldPath` and so on (drop the `Path` from all the Path implementations). More concise and it makes as much sense IMO.

kalman:
Like so

kalman:
I've added the APIChange. LGTY? (I'll email the group)

PR #2266, opened 2016-08-04
Fix command line flag

suryagaddipati:


mikegray:
@suryagaddipati We need a CLA to be signed: http://noms.io/ca_individual.html before we can accept your patch. Thanks for the help!

suryagaddipati:
signed!

mikegray:
thanks again!

PR #2267, opened 2016-08-04
JS: Fix encoding bugs related to signed varints

### LINKED ISSUE #2104: JS csv example appears to be broken
If i run it, it imports the csv and then errors with a `400` error from *somewhere*.

apropos: #1888

### LINKED ISSUE #2234: number encoding broken for js
referencing #2104, part of that issue is related to trying to encode `-122.411912027329`, which via the `floatToIntExp` will expand to` [-8613972521749825, 46]`, but then the `svarint` package tries to encode `-8613972521749825` to a `varint` using [zigzag](https://developers.google.com/protocol-buffers/docs/encoding#types) encoding by `multiplying by -2 and then subtracting 1`.  Unfortunately `(-8613972521749825 * -2)` overflows the safe integer range for JS.

arv:
The library that we were using multiplied (and divided) by 2 to
do the zigzag encoding. However, if we are already close to the
precision limit then we lose precision and the encoding/decoding
does not round trip.

Instead, we split the float64 into two Uint32 numbers and do the
operations on those.

Fixes #2104
Fixes #2234


js/src/signed-varint.js
```diff
@@ -0,0 +1,127 @@
...
+      }
+      arr[j++] = (num & (0x7f << i)) >>> i;
+    }
+    return j;
+  }
+
+  if (hi === 0) {
+    const sigbits = 32 - Math.clz32(lo);
+    const byteLength = Math.max(1, Math.ceil(sigbits / 7));
+    const arr = Bytes.alloc(byteLength);
```
arv: This temporary buffer is not needed. I'll fix that tomorrow.

go/types/xp_test.go
```diff
@@ -29,6 +29,13 @@ func newTestSuite() *testSuite {
 		&testValue{Number(-1), "hq0jvv1enraehfggfk8s27ll1rmirt96", "num - -1"},
 		&testValue{Number(0), "elie88b5iouak7onvi2mpkcgoqqr771l", "num - 0"},
 		&testValue{Number(1), "6h9ldndhjoq0r5sbn1955gaearq5dovc", "num - 1"},
+		&testValue{Number(-122.411912027329), "hcdjnev3lccjplue6pb0fkhgeehv6oec", "num - -122.411912027329"},
+		// JS Number.MAX_SAFE_INTEGER
+		&testValue{Number(9007199254740991), "3fpnjghte4v4q8qogl4bga0qldetlo7b", "num - 9007199254740991"},
+		// JS Number.MIN_SAFE_INTEGER
+		&testValue{Number(-9007199254740991), "jd80frddd2fs3q567tledcgmfs85dvke", "num - -9007199254740991"},
+		// JS Number.EPSILON
+		&testValue{Number(2.220446049250313e-16), "qapetp8502l672v2vie52nd4qjviq5je", "num - 2.220446049250313e-16"},
```
mikegray: good to have these special cases in here.

go/types/xp_test.go
```diff
@@ -72,6 +80,6 @@ func (suite *testSuite) TestTypes() {
 }
 
 // Called from "go test"
-func TestSuite(t *testing.T) {
+func TestPlatformSuite(t *testing.T) {
```
mikegray: why rename?

js/src/signed-varint.js
```diff
@@ -0,0 +1,127 @@
...
+
+function toUint32(n: number): number {
+  return n >>> 0;
+}
+
+/**
+ * Encodes `val` as signed varint and writes that into `buf` at `offset`. This returns the number
+ * of bytes written.
+ */
+export function encode(val: number, buf: Uint8Array, offset: number): number {
```
mikegray: since you are importing this `as encodeVarint` above, why not just make that the function name?

js/src/signed-varint.js
```diff
@@ -0,0 +1,127 @@
...
+
+  return [(mathPowTwoThirtyTwo * hi + lo) / sign, count];
+}
+
+/**
+ * The number of bytes needed to encode `n` as a signed varint.
+ */
+export function encodingLength(n: number): number {
+  if (n <= 0) {
+    return 1;
```
mikegray: is this right for n < 0 but that is larger than 1 byte encoding?  e.g. n === -312?

js/src/xp-test.js
```diff
@@ -49,6 +51,16 @@ suite('cross platform test', () => {
...
+      new TestValue(-122.411912027329, 'hcdjnev3lccjplue6pb0fkhgeehv6oec',
+        'num - -122.411912027329'),
+      // Number.MAX_SAFE_INTEGER
+      new TestValue(9007199254740991, '3fpnjghte4v4q8qogl4bga0qldetlo7b', 'num - 9007199254740991'),
+      // Number.MIN_SAFE_INTEGER
+      new TestValue(-9007199254740991, 'jd80frddd2fs3q567tledcgmfs85dvke',
+        'num - -9007199254740991'),
+      // Number.EPSILON
+      new TestValue(2.220446049250313e-16, 'qapetp8502l672v2vie52nd4qjviq5je',
+        'num - 2.220446049250313e-16'),
```
mikegray: might be good to put golang math.MaxFloat64 here too - 1.7976931348623157e+308?

go/types/number_util.go
```diff
@@ -4,9 +4,7 @@
 
 package types
 
-import (
-	"math"
-)
+import "math"
```
mikegray: all changes in this file are just formatting really

go/types/encoding_test.go
```diff
@@ -168,7 +168,7 @@ func TestRoundTrips(t *testing.T) {
 		f = math.Copysign(f, -1)
 		assertRoundTrips(Number(f))
 	}
-	floatTest := []float64{1.01, 1.001, 1.0001, 1.00001, 1.000001, 100.01, 1000.000001}
+	floatTest := []float64{1.01, 1.001, 1.0001, 1.00001, 1.000001, 100.01, 1000.000001, 122.411912027329, 0.42}
```
mikegray: should the special case numbers below (JS Max/Min safe int, machine epsilon) go here maybe instead of the xp-test? makes sense for them to be here to be sure we can encode/decode roundtrip in golang properly, but also makes sense for them to exist in xp-test to be sure we get same hash across golang vs. JS too.

js/src/signed-varint.js
```diff
@@ -0,0 +1,127 @@
...
+      lo = 0xffffffff;
+    }
+  }
+
+  function append(num, start, end, arr, j) {
+    for (let i = start; i < end; i += 7) {
+      if (j !== 0) {
+        arr[j - 1] |= 0x80;
+      }
+      arr[j++] = (num & (0x7f << i)) >>> i;
```
mikegray: i really dislike bit manipulation code. thanks for working on this!

mikegray:
back to you

go/types/xp_test.go
arv: TestSuite is too generic... The name of this is needs to be `Test.+` but since the types package is so large calling it TestSuite seems not specific enough.

js/src/signed-varint.js
arv: The file name already names this signed varint. The reason I renamed it in the importer is to make it clearer in the code since the importer already has encode/decode.

js/src/signed-varint.js
arv: Another option is to assert `>= 0`. This is not supposed to be used on negative numbers. Maybe I should make that clearer by renaming this?

js/src/xp-test.js
arv: Good idea.

js/src/rolling-value-hasher.js
```diff
@@ -112,13 +112,12 @@ export default class RollingValueHasher {
   }
 
   hashVarint(n: number) {
-    const count = svarint.encodingLength(n);
     if (this.lengthOnly) {
-      this.bytesHashed += count;
+      this.bytesHashed += varintEncodingLength(n);
```
mikegray: n is always positive here?

go/types/xp_test.go
mikegray: sounds good

go/types/number_util.go
arv: Yeah... `go fmt` fixed them for me. I guess someone forgot to run `go fmt` at some point?

js/src/signed-varint.js
mikegray: yeah, thought about that last night - seems good to me.

go/types/encoding_test.go
arv: I'll add them in both.

The values in these slices are negated so I need to add them outside the slices.

js/src/signed-varint.js
arv: I think there is still a lot of room for improvement here but at least it works now.

js/src/signed-varint.js
arv: Looking at the callsites... they may pass negative numbers in. Let me change this to do Math.abs on the number first.

arv:
PTAL

js/src/rolling-value-hasher.js
arv: No. I changed encodingLength to handle this.

js/src/signed-varint.js
```diff
@@ -0,0 +1,107 @@
...
+    lo++;
+  }
+
+  return [(mathPowTwoThirtyTwo * hi + lo) / sign, count];
+}
+
+/**
+ * The number of bytes needed to encode `n` as a signed varint.
+ */
+export function encodingLength(n: number): number {
```
mikegray: I think the encoding length of 64 is 2 bytes, but -64 is only 1.

arv: Added tests and as you said, -64 fails. Working on a fix.

arv:
Added the ugliest commit ever. I should really figure out how to make this not suck so much.

PTAL

arv:
@mikegray PTAL. The encodingLength function is now pretty clear.

js/src/signed-varint.js
```diff
@@ -0,0 +1,128 @@
...
+  // Signed varints are encoded using zigzag encoding, where the number is multiplied by 2 and if
+  // negative then the sign is removed and we subtract 1 (v >= 0 ? 2 * v : -2 * v - 1)
+
+  if (n === 0) {
```
mikegray: will never be called with -0, correct?

mikegray:
lgtm

js/src/signed-varint.js
arv: `0 === -0` is always `true`

PR #2271, opened 2016-08-04
Allow Dataset HEAD to be set to non-fastforward commits

### LINKED ISSUE #2240: noms sync to roll back 1 commit when there are many is slown
to repro:

`noms sync http://demo.noms.io/ahl::bug2213 /var/tmp/test::biggie`
(wait for a long time)
`noms sync /var/tmp/test::o38hugtf3l1e8rqtj89mijj1dq57eh4m /var/tmp/test::biggie`

This should be fast, but is a lot slower than this:
`noms sync /var/tmp/test::o38hugtf3l1e8rqtj89mijj1dq57eh4m /var/tmp/test::entirelynew`

cmasone-attic:
The Dataset.Commit() code pathway still enforces fast-forward-only
behavior, but a new SetHead() method allows the HEAD of a Dataset to
be forced to any other Commit.

noms sync detects the case where the source Commit is not a descendent
of the provided sink Dataset's HEAD and uses the new API to force the
sink to the desired new Commit, printing out the now-abandoned old
HEAD.

Fixes #2240

### DIFF
--- a/cmd/noms/noms_sync.go
+++ b/cmd/noms/noms_sync.go
@@ -77,15 +77,20 @@ func runSync(args []string) int {
 		lastProgressCh <- last
 	}()
 
+	sourceRef := types.NewRef(sourceObj)
+	sinkRef, _ := sinkDataset.MaybeHeadRef()
+	nonFF := false
 	err = d.Try(func() {
 		defer profile.MaybeStartProfile().Stop()
-		sourceRef := types.NewRef(sourceObj)
+		sinkDataset.Pull(sourceStore, sourceRef, p, progressCh)
+
 		var err error
-		if rewind := sinkDataset.Pull(sourceStore, sourceRef, p, progressCh); rewind {
-			fmt.Printf("Abandoning %s; new head is %s\n", sinkDataset.HeadRef().TargetHash(), sourceRef.TargetHash())
-			sinkDataset, err = sinkDataset.SetHead(sourceRef)
-		} else {
+		if sourceRef.Height() > sinkRef.Height() {
 			sinkDataset, err = sinkDataset.FastForward(sourceRef)
+		}
+		if sourceRef.Height() <= sinkRef.Height() || err == datas.ErrMergeNeeded {
+			sinkDataset, err = sinkDataset.SetHead(sourceRef)
+			nonFF = true
 		}
 		d.PanicIfError(err)
 	})
@@ -98,8 +103,10 @@ func runSync(args []string) int {
 	if last := <-lastProgressCh; last.DoneCount > 0 {
 		status.Printf("Done - Synced %s in %s (%s/s)", humanize.Bytes(last.DoneBytes), since(start), bytesPerSec(last, start))
 		status.Done()
+	} else if nonFF && !sourceRef.Equals(sinkRef) {
+		fmt.Printf("Abandoning %s; new head is %s\n", sinkRef.TargetHash(), sourceRef.TargetHash())
 	} else {
-		fmt.Println(args[0], "is up to date.")
+		fmt.Println(args[1], "is up to date.")
 	}
 
 	return 0
--- a/cmd/noms/noms_sync.go
+++ b/cmd/noms/noms_sync.go
@@ -79,8 +79,14 @@ func runSync(args []string) int {
 
 	err = d.Try(func() {
 		defer profile.MaybeStartProfile().Stop()
+		sourceRef := types.NewRef(sourceObj)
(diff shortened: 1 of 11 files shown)


cmasone-attic:
Before I finish this out by adding comments and a test or two, what do you think of the approach?

rafael-atticlabs:
@aboodman I think the UX question is to you. If you are lgtm on that, I'll take the code review.

cmd/noms/noms_sync.go
```diff
@@ -79,8 +79,14 @@ func runSync(args []string) int {
 
 	err = d.Try(func() {
 		defer profile.MaybeStartProfile().Stop()
+		sourceRef := types.NewRef(sourceObj)
 		var err error
-		sinkDataset, err = sinkDataset.Pull(sourceStore, types.NewRef(sourceObj), p, progressCh)
+		if rewind := sinkDataset.Pull(sourceStore, sourceRef, p, progressCh); rewind {
+			fmt.Printf("Abandoning %s; new head is %s\n", sinkDataset.HeadRef().TargetHash(), sourceRef.TargetHash())
```
aboodman: Warning: Sync was not a fast-forward. Commit %s was abandoned.

aboodman: Nevermind, I prefer yours.

aboodman:
woohoo. lgtm

cmasone-attic:
@rafael-atticlabs do you still want to look at this? I added some comments for the new functions I added as well.

rafael-atticlabs:
looking.

cmasone-attic:
ok, @rafael-atticlabs  PTAL

go/datas/pull.go
```diff
@@ -149,7 +149,7 @@ func Pull(srcDB, sinkDB Database, sourceRef, sinkHeadRef types.Ref, concurrency
 		}
 	}
 	sinkDB.validatingBatchStore().AddHints(hints)
-	return false
+	return
```
rafael-atticlabs: don't need this return anymore.

cmd/noms/noms_sync.go
```diff
@@ -77,10 +77,21 @@ func runSync(args []string) int {
...
+	sourceRef := types.NewRef(sourceObj)
+	sinkRef, _ := sinkDataset.MaybeHeadRef()
+	nonFF := false
 	err = d.Try(func() {
 		defer profile.MaybeStartProfile().Stop()
+		sinkDataset.Pull(sourceStore, sourceRef, p, progressCh)
+
 		var err error
-		sinkDataset, err = sinkDataset.Pull(sourceStore, types.NewRef(sourceObj), p, progressCh)
+		if sourceRef.Height() > sinkRef.Height() {
```
rafael-atticlabs: it's kind of a bummer that the only way to test this case is by testing noms_sync. What about moving this logic back into dataset.Pull() or if you think we should keep something like Pull to just move objects, then adding api to Dataset?

cmasone-attic: On one hand, it's seemed a bit inconsistent to me that `Dataset.Pull()` handled updating the Head of the dataset and `datas.Pull()` only moved objects around. And I think that we need that flexibility at the `datas` layer. So, for that reason I'd rather make the two Pull methods be more similar.

That said, I think that adding more API to Dataset beyond FastForward() and SetHead() doesn't make much sense. It should have primitives to allow people to implement the policies they want, but not have a helper function for all the different variations of head-changing behavior people might want.

It probably makes sense to add this ref-height check to FastForward(), though. I don't think it makes sense to, for example, have SetHead() always call FastForward() and then follow its own logic if it gets back ErrMergeNeeded.

I think the problem is that we still don't have a great way of testing logic in our command line tools, which is something we need to fix

cmasone-attic: Also, tucking the checks away inside Dataset API would make it hard to get out the info I need to print out the right user feedback. I suppose that whatever functions we changed/added could return some kind of indicator, but that would seem weird to me. It'd be returning a new Dataset, an error, and then some kind of return code as well to indicate what manner of head-setting it did.

rafael-atticlabs: ok

rafael-atticlabs:
lgtm

cmd/noms/noms_sync.go
cmasone-attic: Actually, we do have some tests for noms sync that use our ClientTestSuite
thing. Maybe I can leverage that

PR #2286, opened 2016-08-05
Remove 2nd slack badge from README.md

kalman:


kalman:
Maybe having 2 badges is doubling server load :-)

mikegray:
i don't think it was server load - the process just wasn't running. maybe it stopped because of load, not sure - I didn't go looking at the log files - was just concerned with getting it running again. lgtm though.

kalman:
Thanks. Yeah, server load or concurrency or something. Either way I can argue that having 2 badges on the same page looks weird.

PR #2288, opened 2016-08-05
JS: Partial fix for Safari

arv:
getAllResponseHeaders may contain empty rows


arv:
@kalman There are more Safari issues but the Safari debugger is a POS and crashes when it is opened.

arv:
I reverted back to using alerts!!!

kalman:
lgtm

I guess the line number was bogus then :-)

arv:
There is more things going on here but this takes care of one bug.

PR #2289, opened 2016-08-05
Add streaming set and improve streaming collection performance.

willhite:
Streaming collections now share one ldb instance rather than having
1 ldb instance per collection.

### DIFF
--- a/go/types/opcache.go
+++ b/go/types/opcache.go
@@ -8,15 +8,55 @@ import (
 	"encoding/binary"
 	"io/ioutil"
 	"os"
+	"sync/atomic"
 
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/iterator"
 	"github.com/syndtr/goleveldb/leveldb/opt"
+	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
-func newOpCache(vrw ValueReadWriter) *opCache {
+const uint32Size = 4
+
+type opCacheStore interface {
+	opCache() opCache
+	destroy() error
+}
+
+type opCache interface {
+	MapSet(mapKey Value, mapVal Value)
+	SetInsert(val Value)
+	NewIterator() opCacheIterator
+}
+
+type opCacheIterator interface {
+	MapOp() sequenceItem
+	SetOp() sequenceItem
+	Next() bool
+	Release()
+}
+
+type ldbOpCacheStore struct {
+	ldb          *leveldb.DB
+	dbDir        string
+	collectionId uint32
+	vrw          ValueReadWriter
+}
+
+type ldbOpCache struct {
+	vrw   ValueReadWriter
+	colId uint32
+	ldb   *leveldb.DB
+}
+
+type ldbOpCacheIterator struct {
+	iter iterator.Iterator
+	vr   ValueReader
+}
+
+func newLdbOpCacheStore(vrw ValueReadWriter) *ldbOpCacheStore {
 	dir, err := ioutil.TempDir("", "")
 	d.Chk.NoError(err)
 	db, err := leveldb.OpenFile(dir, &opt.Options{
@@ -27,62 +67,56 @@ func newOpCache(vrw ValueReadWriter) *opCache {
 		WriteBuffer:            1 << 27, // 128MiB
 	})
 	d.Chk.NoError(err, "opening put cache in %s", dir)
-	return &opCache{ops: db, dbDir: dir, vrw: vrw}
+	return &ldbOpCacheStore{ldb: db, dbDir: dir, vrw: vrw}
 }
 
-type opCache struct {
-	ops           *leveldb.DB
-	dbDir         string
-	vrw           ValueReadWriter
-	ldbKeyScratch [1 + hash.ByteLen]byte
-	keyScratch    [initialBufferSize]byte
-	valScratch    [initialBufferSize]byte
+func (store *ldbOpCacheStore) destroy() error {
+	d.Chk.NoError(store.ldb.Close())
+	return os.RemoveAll(store.dbDir)
 }
 
-type opCacheIterator struct {
-	iter iterator.Iterator
-	vr   ValueReader
+func (store *ldbOpCacheStore) opCache() opCache {
+	colId := atomic.AddUint32(&store.collectionId, 1)
+	return &ldbOpCache{vrw: store.vrw, colId: colId, ldb: store.ldb}
 }
-
-var uint32Size = binary.Size(uint32(0))
 
 // Set can be called from any goroutine
-func (p *opCache) Set(mapKey Value, mapVal Value) {
+func (opc *ldbOpCache) MapSet(mapKey Value, mapVal Value) {
+	mapKeyArray := [initialBufferSize]byte{}
+	mapValArray := [initialBufferSize]byte{}
+
 	switch mapKey.Type().Kind() {
 	default:
-		// This is the complicated case. For non-primitives, we want the ldb key to be the hash of mapKey, but we obviously need to get both mapKey and mapVal into ldb somehow. The simplest thing is just to do this:
+		ldbKey := ldbKeyFromValueHash(mapKey, opc.colId)
+
+		// Since we've used the ref of keyValue as our ldbKey. We need to store mapKey and mapVal in the ldb value. We use the following format for that:
 		//
 		//     uint32 (4 bytes)             bytes                 bytes
 		// +-----------------------+---------------------+----------------------+
 		// | key serialization len |    serialized key   |   serialized value   |
 		// +-----------------------+---------------------+----------------------+
 
-		// Note that, if mapKey and/or mapVal are prolly trees, any in-memory child chunks will be written to vrw at this time.
-		p.ldbKeyScratch[0] = byte(mapKey.Type().Kind())
-		copy(p.ldbKeyScratch[1:], mapKey.Hash().DigestSlice())
-		mapKeyData := encToSlice(mapKey, p.keyScratch[:], p.vrw)
-		mapValData := encToSlice(mapVal, p.valScratch[:], p.vrw)
+		encodedMapKey := encToSlice(mapKey, mapKeyArray[:], opc.vrw)
+		encodedMapVal := encToSlice(mapVal, mapValArray[:], opc.vrw)
+		ldbValueArray := [initialBufferSize * 2]byte{}
+		binary.LittleEndian.PutUint32(ldbValueArray[:], uint32(len(encodedMapKey)))
+		ldbValue := ldbValueArray[0:4]
+		ldbValue = append(ldbValue, encodedMapKey...)
+		ldbValue = append(ldbValue, encodedMapVal...)
 
-		mapKeyByteLen := len(mapKeyData)
-		data := make([]byte, uint32Size+mapKeyByteLen+len(mapValData))
-		binary.LittleEndian.PutUint32(data, uint32(mapKeyByteLen))
-		copy(data[uint32Size:], mapKeyData)
-		copy(data[uint32Size+mapKeyByteLen:], mapValData)
-
-		// TODO: Will manually batching these help?
-		err := p.ops.Put(p.ldbKeyScratch[:], data, nil)
+		// TODO: Will manually batching calls to ldb.Put() help?
+		err := opc.ldb.Put(ldbKey, ldbValue, nil)
 		d.Chk.NoError(err)
 
 	case BoolKind, NumberKind, StringKind:
-		// In this case, we can just serialize mapKey and use it as the ldb key, so we can also just serialize mapVal and dump that into the DB.
-		keyData := encToSlice(mapKey, p.keyScratch[:], p.vrw)
-		valData := encToSlice(mapVal, p.valScratch[:], p.vrw)
-		// TODO: Will manually batching these help?
-		err := p.ops.Put(keyData, valData, nil)
+		ldbKey := ldbKeyFromValue(mapKey, opc.colId, opc.vrw)
+		encodedMapVal := encToSlice(mapVal, mapValArray[:], opc.vrw)
+		err := opc.ldb.Put(ldbKey, encodedMapVal, nil)
 		d.Chk.NoError(err)
 	}
 }
 
+// Note that, if 'v' are prolly trees, any in-memory child chunks will be written to vw at this time.
 func encToSlice(v Value, initBuf []byte, vw ValueWriter) []byte {
 	// TODO: Are there enough calls to this that it's worth re-using a nomsWriter and valueEncoder?
 	w := &binaryNomsWriter{initBuf, 0}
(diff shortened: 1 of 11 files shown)


go/types/opcache.go
```diff
@@ -8,15 +8,33 @@ import (
...
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/iterator"
 	"github.com/syndtr/goleveldb/leveldb/opt"
+	"github.com/syndtr/goleveldb/leveldb/util"
 )
 
-func newOpCache(vrw ValueReadWriter) *opCache {
+var (
+	collectionId = uint32(0)
+	ldb          *opCacheDb
```
cmasone-attic: I'm generally not a fan of the 'Singleton' pattern. No clear ownership tends to lead to failures to clean up and/or shutdown ordering weirdness. What was the issue with each mutator having its own opcache? Too many files open or something?

go/types/opcache.go
```diff
@@ -27,25 +45,65 @@ func newOpCache(vrw ValueReadWriter) *opCache {
...
-	ldbKeyScratch [1 + hash.ByteLen]byte
-	keyScratch    [initialBufferSize]byte
-	valScratch    [initialBufferSize]byte
+	vrw              ValueReadWriter
+	ldbKeyScratch    [1 + hash.ByteLen]byte
+	keyScratch       [initialBufferSize]byte
+	prefixKeyScratch [initialBufferSize]byte
+	valScratch       [initialBufferSize]byte
+	colId            uint32
+	prefix           []byte
```
cmasone-attic: `prefix` has a known size too, right? it's `[prefixByteSize]byte`. You could avoid a heap allocation by having this be an array and just `PutUint32()`ing into it directly

go/types/opcache.go
```diff
@@ -27,25 +45,65 @@ func newOpCache(vrw ValueReadWriter) *opCache {
...
 	})
 	d.Chk.NoError(err, "opening put cache in %s", dir)
-	return &opCache{ops: db, dbDir: dir, vrw: vrw}
+	return &opCacheDb{ops: db, dbDir: dir}
+}
+
+func newOpCache(vrw ValueReadWriter) *opCache {
+	incrementActiveColCount()
+	prefix := [prefixByteSize]byte{}
+	colId := atomic.AddUint32(&collectionId, 1)
```
cmasone-attic: I feel like you don't need this atomic op here. You could jut increment `collectionId` in `incrementActiveColCount()` and return it, no? It just seems nice to reduce the number of synchronization primitives you need.

go/types/opcache.go
```diff
@@ -27,25 +45,65 @@ func newOpCache(vrw ValueReadWriter) *opCache {
...
+		d.Chk.NoError(ldb.ops.Close())
+		err = os.RemoveAll(ldb.dbDir)
+		ldb = nil
+	}
+	activeOpcCnt--
+	return
+}
+
+func (p *opCache) PutOp(key, val []byte) error {
+	copy(p.prefixKeyScratch[:], p.prefix)
```
cmasone-attic: You know...I say that Set() can be called from any goroutine, but it definitely isn't safe to call concurrently, because every call to Set() uses the same scratch space buffer. Maybe it's OK to just update the comment for now, but it may be something we need to address later.

go/types/opcache.go
```diff
@@ -27,25 +45,65 @@ func newOpCache(vrw ValueReadWriter) *opCache {
...
+	ldbKeyScratch    [1 + hash.ByteLen]byte
+	keyScratch       [initialBufferSize]byte
+	prefixKeyScratch [initialBufferSize]byte
+	valScratch       [initialBufferSize]byte
+	colId            uint32
+	prefix           []byte
 }
 
 type opCacheIterator struct {
+	opc  *opCache
```
cmasone-attic: I don't see anywhere in the `opCacheIterator` methods that this gets used. Leftover from a previous version of the patch?

go/types/opcache.go
```diff
@@ -122,6 +180,45 @@ func (i *opCacheIterator) Op() sequenceItem {
 	return entry
 }
 
+// Insert can be called from any goroutine
+func (p *opCache) Insert(val Value) {
+	switch val.Type().Kind() {
+	default:
+		// This is the complicated case. For non-primitives, we want the ldb key to be the hash of mapKey, but we obviously need to get both mapKey and mapVal into ldb somehow. The simplest thing is just to do this:
```
cmasone-attic: Fix up this comment.

go/types/opcache.go
willhite: This change was inspired by my nomdex work. We can have thousands of these streaming collections going simultaneously. Having an ldb instance for each one of them is problematic for at least two reasons: 
1) the number of directories created and potentially leaked in the case of a crash
2) the amount of memory overhead for each ldb instance. 200-300 instances was causing ~10Gb of memory usage that completely disappeared with this change.

I agree about singletons but in this case the ownership issues don't seem to be much more complicated than what we had before.

go/types/opcache.go
```diff
@@ -122,6 +180,45 @@ func (i *opCacheIterator) Op() sequenceItem {
...
+		//
+		//     uint32 (4 bytes)             bytes
+		// +-----------------------+----------------------+
+		// | key serialization len |   serialized value   |
```
cmasone-attic: I don't think this is what you're doing, here? You're just setting the LDB key to be `val.Hash()`, essentially, and then encoding val to a byte slice and setting that to be the LDB value

go/types/opcache.go
```diff
@@ -122,6 +180,45 @@ func (i *opCacheIterator) Op() sequenceItem {
...
+		// +-----------------------+----------------------+
+
+		// Note that, if val is a prolly trees, any in-memory child chunks will be written to vrw at this time.
+		p.ldbKeyScratch[0] = byte(val.Type().Kind())
+		copy(p.ldbKeyScratch[1:], val.Hash().DigestSlice())
+		mapValData := encToSlice(val, p.valScratch[:], p.vrw)
+
+		data := make([]byte, len(mapValData))
+		copy(data, mapValData)
```
cmasone-attic: Why re-copy this slice?

go/types/opcache.go
cmasone-attic: Before, each opCache had a defined lifetime: it was bound to a given mutator instance. Now, if any instance is alive, then this whole thing lives. It also means you can't inject some other opCache implementation for testing.

Also, if your 10GB number is accurate, it means that each open levelDB instance adds 33MB of unique memory overhead? That seems like a lot. I know that there are some caches that each instance uses. Perhaps we could reduce those in size for these opcaches and see if that changes things. There's `DefaultBlockCacheCapacity`, which defaults to 8MB. We think that's for reads, but we can't actually decipher the code and comments. There's also `DefaultWriteBuffer`, which starts at 4MB. Other stuff: https://godoc.org/github.com/syndtr/goleveldb/leveldb/opt

go/types/opcache.go
willhite: Is there a reason you didn't use a scratch buffer on the stack for this work?

cmasone-attic: I was concerned about allocating every time Set() was called, even on the stack. I think that, depending on what happens down in levelDB.Put(), doing that can wind up getting heap allocated anyway. I was probably being overzealous. If we want to allow Set and Insert to be called concurrently, stack allocation is likely best for now. We could use a Pool of buffers later if we discover performance issues.

willhite:
@cmasone-attic PTAL. Please focus on the changes to make Database the owner of OpCache. I also made changes to the byte mashing that goes on in op_cache.go. Once we're in agreement about the changes to Database interface, I'll add more tests. Thanks.

go/types/opcache.go
```diff
@@ -58,27 +105,33 @@ func (p *opCache) Set(mapKey Value, mapVal Value) {
...
-		mapKeyByteLen := len(mapKeyData)
-		data := make([]byte, uint32Size+mapKeyByteLen+len(mapValData))
-		binary.LittleEndian.PutUint32(data, uint32(mapKeyByteLen))
-		copy(data[uint32Size:], mapKeyData)
-		copy(data[uint32Size+mapKeyByteLen:], mapValData)
+		ldbKeyArray := [uint32Size + 1 + hash.ByteLen]byte{}
+		binary.LittleEndian.PutUint32(ldbKeyArray[:], opc.colId)
+		ldbKeyArray[uint32Size] = byte(mapKey.Type().Kind())
+		ldbKey := ldbKeyArray[0 : uint32Size+1]
+		ldbKey = append(ldbKey, mapKey.Hash().DigestSlice()...)
```
cmasone-attic: Using `copy()` here reads much more clearly, to me. Possibly busting this out into a helper method like `constructLdbKey(scratch []byte, colId uint32, mapKey Value)` would make the code here in MapSet more readable.

go/types/opcache.go
```diff
@@ -8,15 +8,65 @@ import (
...
+	ldb          *leveldb.DB
+	dbDir        string
+	collectionId uint32
+	vrw          ValueReadWriter
+}
+
+type LdbOpCache struct {
+	vrw   ValueReadWriter
+	colId uint32
+	store *LdbOpCacheStore
```
cmasone-attic: I _think_ the only thing you do with `store` in `LdbOpCache` methods is access the `ldb` field. Why not just make this be `ldb *leveldb.DB`?

go/types/opcache.go
```diff
@@ -58,27 +105,33 @@ func (p *opCache) Set(mapKey Value, mapVal Value) {
...
 		d.Chk.NoError(err)
 
 	case BoolKind, NumberKind, StringKind:
 		// In this case, we can just serialize mapKey and use it as the ldb key, so we can also just serialize mapVal and dump that into the DB.
-		keyData := encToSlice(mapKey, p.keyScratch[:], p.vrw)
-		valData := encToSlice(mapVal, p.valScratch[:], p.vrw)
+		ldbKeyArray := [initialBufferSize]byte{}
+		binary.LittleEndian.PutUint32(ldbKeyArray[:], opc.colId)
+		kvData := ldbKeyArray[0:uint32Size]
+		encodedMapKey := encToSlice(mapKey, mapKeyArray[:], opc.vrw)
```
cmasone-attic: I don't see why you don't just use `mapKeyArray` for all of this work, like so:
```
		binary.LittleEndian.PutUint32(mapKeyArray[:], opc.colId)
		encodedMapKey := encToSlice(mapKey, mapKeyArray[uint32Size:], opc.vrw)
```

cmasone-attic: wait...now I do :-P

go/types/opcache.go
```diff
@@ -122,6 +173,45 @@ func (i *opCacheIterator) Op() sequenceItem {
 	return entry
 }
 
-func (i *opCacheIterator) Release() {
+// Insert can be called from any goroutine
+func (opc *LdbOpCache) SetInsert(val Value) {
```
cmasone-attic: This should be up with the other LdbOpCache methods

go/types/opcache.go
```diff
@@ -122,6 +173,45 @@ func (i *opCacheIterator) Op() sequenceItem {
...
+	valArray := [initialBufferSize]byte{}
+
+	switch val.Type().Kind() {
+	default:
+		// Set entries are fairly simple to store, simply store the element's hash (or the element itself if bool, number, or string) in the ldb key and the value in ldb value.
+
+		// Note that, if val is a prolly trees, any in-memory child chunks will be written to vrw at this time.
+		ldbKeyArray := [uint32Size + 1 + hash.ByteLen]byte{}
```
cmasone-attic: Similar comment as above, with using a helper method to make the flow of this code here more readable.

go/types/opcache.go
```diff
@@ -122,6 +173,45 @@ func (i *opCacheIterator) Op() sequenceItem {
...
+
+	case BoolKind, NumberKind, StringKind:
+		// In this case, the ldbKey is just the colId + encodedVal, and the ldbValue is just the encodedVal
+		ldbKeyArray := [initialBufferSize]byte{}
+		binary.LittleEndian.PutUint32(ldbKeyArray[:], opc.colId)
+		ldbKeySlice := ldbKeyArray[0:uint32Size]
+		encodedVal := encToSlice(val, valArray[:], opc.vrw)
+		ldbKeySlice = append(ldbKeySlice, encodedVal...)
+		// TODO: Will manually batching these help?
+		err := opc.store.ldb.Put(ldbKeySlice, encodedVal, nil)
```
cmasone-attic: `encodedVal` winds up in both the key and value in the level db, then? I figure you could just put it in the key, and then extract it from there when `SetOp` is called, and save storage. Probably only really meaningful savings in the case of long strings, but still.

cmasone-attic:
Ok, I know we can have an unexported method in the ValueReadWriter interface, which ValueStore can then implement. My goal is to make that method be available on a Database when it's passed as a ValueReadWriter into the types package, but not otherwise. I'm going to see if that is a thing, and then get back to you.

cmasone-attic:
Ah ha! I've got a patch that changes things around a tiny bit so that ValueStore can implement package-private methods belonging to the ValueReadWriter interface, and they will be available when a Database is passed into methods and functions in the `types` package. Let me upload it so you can take a look

willhite:
@cmasone-attic PTAL

go/types/opcache.go
```diff
@@ -8,15 +8,55 @@ import (
...
+	ldb          *leveldb.DB
+	dbDir        string
+	collectionId uint32
+	vrw          ValueReadWriter
+}
+
+type ldbOpCache struct {
+	vrw   ValueReadWriter
+	colId uint32
+	store *ldbOpCacheStore
```
cmasone-attic: As far as I can tell, the only thing you use `store` for is to get `store.ldb`, so why not just give ldbOpCache a `*leveldb.DB` instead?

go/types/opcache.go
```diff
@@ -91,37 +125,100 @@ func encToSlice(v Value, initBuf []byte, vw ValueWriter) []byte {
...
 	entry := mapEntry{}
 	ldbKey := i.iter.Key()
-	data := i.iter.Value()
-	dataOffset := 0
-	switch NomsKind(ldbKey[0]) {
+	ldbValue := i.iter.Value()
+	switch NomsKind(ldbKey[uint32Size]) {
 	case BoolKind, NumberKind, StringKind:
-		entry.key = DecodeFromBytes(ldbKey, i.vr, staticTypeCache)
+		entry.key = DecodeFromBytes(ldbKey[4:], i.vr, staticTypeCache)
```
cmasone-attic: can this `4` be a constant of some kind? So if we change the length of the prefix we use at some point, we don't need to try to find all the places we assumed it was 4?

go/types/opcache.go
```diff
@@ -91,37 +125,100 @@ func encToSlice(v Value, initBuf []byte, vw ValueWriter) []byte {
...
 	}
 
-	entry.value = DecodeFromBytes(data[dataOffset:], i.vr, staticTypeCache)
 	return entry
 }
 
-func (i *opCacheIterator) Release() {
+// Insert can be called from any goroutine
+func (opc *ldbOpCache) SetInsert(val Value) {
+
```
cmasone-attic: blank line

go/types/opcache.go
```diff
@@ -91,37 +125,100 @@ func encToSlice(v Value, initBuf []byte, vw ValueWriter) []byte {
...
+		d.Chk.NoError(err)
+	}
+}
+
+func (i *ldbOpCacheIterator) SetOp() sequenceItem {
+	ldbKey := i.iter.Key()
+
+	switch NomsKind(ldbKey[uint32Size]) {
+	case BoolKind, NumberKind, StringKind:
+		return DecodeFromBytes(ldbKey[uint32Size:], i.vr, staticTypeCache)
```
cmasone-attic: is `ldbKey[uint32Size:]` here supposed to be `ldbKey[prefixLen:]` or something? Unless I misunderstand, I think you're chopping off the prefix here.

go/types/opcache.go
```diff
@@ -91,37 +125,100 @@ func encToSlice(v Value, initBuf []byte, vw ValueWriter) []byte {
...
+	ldbKey := writeLdbKeyHeaderBytes(ldbKeyArray[:], colId, val)
+	return append(ldbKey, val.Hash().DigestSlice()...)
+}
+
+// ldbKeys for primitive Noms Values (e.g. bool, number, & string) consist of a byte string that encodes:
+// colId(4 bytes) + serialized value(n bytes)
+// Note: the first byte of the serialized value is the NomsKind.
+func ldbKeyFromValue(val Value, colId uint32, vrw ValueReadWriter) []byte {
+	valArray := [initialBufferSize]byte{}
+	ldbKeyArray := [initialBufferSize]byte{}
```
cmasone-attic: You know that this is going to be at most 5 bytes, right? So you can just allocate that much space here.

cmasone-attic: ahh...wait. But you're going to append the whole encoded value into it. Hm.

cmasone-attic: nvm

cmasone-attic:
LGTM w/nits

go/types/opcache.go
willhite: fyi, this particular issue caused me a few hours in debugging time. There's a comment about this on ldbKeyFromValue(). This is the way things were encoded originally and I opted to maintain compatibility with opCacheComparer rather than change the encoding by adding an additional byte here.

cmasone-attic: Yeah, Raf and I were both uncomfortable by how much this relies on details of encoding -- such as the first byte of primitives being the NomsKind :-/

Still, though, you're chopping off the prefix here, right? Which is always 4 bytes, but not actually a uint32? That's all that I was getting at. While both are 4 bytes (right now), putting `uint32Size` here kind of masks what you're actually doing

willhite: At some point I jettisoned the notion that prefix was a constant since in some cases it's 4 bytes and in others it's 5 bytes. All ldbKeys start with a uint32(4 bytes). ldbKeys that use refs have an addition nomsKind byte on them.

PR #2291, opened 2016-08-05
Splore: Use Struct as label for non named structs

arv:
If a struct does not have a name we used to have no label for the node.
It looks kind of funky so instead it will use the text "Struct" as the
label.


aboodman:
Is there room in the UI to generalize this to Struct (Commit) in the case where the thing does have a name?

aboodman:
Hm, on second thought i think that will look weird. nm

aboodman:
lgtm

arv:
What we could do is to add different icons for the nodes... similarly to how Visual Studio uses different icons for different types

aboodman:
I actually tried that the first time and couldn't come up with shapes I was
happy with in svg.

PR #2295, opened 2016-08-05
JS: Safari 9 support

### LINKED ISSUE #2265: Splore in Safari crashes with "TypeError: Invalid attempt to destructure non-iterable instance"
```
TypeError: Invalid attempt to destructure non-iterable instance
uout.js:13:10384
onloadendout.js:13:9380
```
The code is
```
return (0, f.invariant)(u instanceof d.OrderedSequenceCursor), (0, f.invariant)(c instanceof d.OrderedSequenceCursor), y.next = 9, s.default.awrap(a(!1, u, c));
```
which in un-obfuscated land I'm pretty sure is https://github.com/attic-labs/noms/blob/master/js/src/ordered-sequence-diff.js#L81 i.e.
```
[aHasMore, bHasMore] = await doFastForward(false, aParent, bParent);
```
wuh? @arv

arv:
- Lower case header names in fetch function
- fetchUint8Array should return an Uint8Array and not an ArrayBuffer
- Safari does not have slice on typed arrays

Fixes #2265

### DIFF
--- a/js/src/browser/bytes.js
+++ b/js/src/browser/bytes.js
@@ -58,7 +58,9 @@ export function copy(source: Uint8Array, target: Uint8Array, targetStart: number
 
 export function slice(buff: Uint8Array, start: number, end: number): Uint8Array {
   // Safari does not have slice on typed arrays.
-  return new Uint8Array(buff.buffer, buff.byteOffset + start, end - start);
+  const v = alloc(end - start);
+  copy(buff.subarray(start, end), v, 0);
+  return v;
 }
 
 export function subarray(buff: Uint8Array, start: number, end: number): Uint8Array {
--- a/js/src/browser/bytes.js
+++ b/js/src/browser/bytes.js
@@ -56,11 +56,12 @@ export function copy(source: Uint8Array, target: Uint8Array, targetStart: number
   target.set(source, targetStart);
 }
 
+/**
+ * Slice returns a copy of parts of the bytes starting at `start` ending at `end` (exclusive).
+ */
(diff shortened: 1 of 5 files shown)


arv:
@kalman Merge if LGTY

kalman:
lgtm

kalman:
flow isn't happy though :-(

arv:
Oh. I'll look at it next week.

arv:
I got some minutes and now Flow and tests pass locally.

js/src/browser/bytes.js
```diff
@@ -57,7 +57,10 @@ export function copy(source: Uint8Array, target: Uint8Array, targetStart: number
 }
 
 export function slice(buff: Uint8Array, start: number, end: number): Uint8Array {
-  return buff.slice(start, end);
+  // Safari does not have slice on typed arrays.
```
kalman: Did `slice` copy? Obviously this new code does. My point being, is it worth changing this code to:
```
if (buff.slice) {
  return buff.slice(start, end);
}
// Safari does not have slice on typed arrays.
const v = ...;
return v;
```

kalman:
lgtm

js/src/browser/bytes.js
arv: Yeah, slice copies. There is even a test for it.

arv: But now that I look at the spec we can use `ArrayBuffer.prototype.slice` which does work in Safari 9.

PR #2314, opened 2016-08-08
Allow RegisterOutputFlag to register more than once. Fixes #2312.

### LINKED ISSUE #2312: --no-pager flag not showing up in help doc
```
!? noms diff -h
usage: diff [--summarize] <object1> <object2>

See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the object arguments.

options:
--summarize  (= false)
    Writes a summary of the changes instead
```

however, you can still pass `--no-pager` and it works. It's just hidden.

@arv, this appears to be related to your gnuflags change.

willhite:


mikegray:
lgtm, but not sure I understand why it is needed though.

willhite:
@cmasone-attic I made this change because this flag needs to get registered multiple times in different flagsets by the noms programs. @mikegray originally copied this pattern from your code in RegisterLevelDBFlags() in leveldb_store.go. I think you might want to consider removing the flagsRegistered check from that code as well.

cmasone-attic:
I didn't write that code, @willhite -- you did :-)

willhite:
Ha! I thought, I remembered writing that but I traced it back in the noms code rather than leveldb code.

Issue #2292, opened 2016-08-05
csv-import to map should not remove indexed column from struct


rafael-atticlabs:
This probably isn't much of a space savings, given chunk-level compression and it makes exporting correctly harder.

mikegray:
fixed as of a81203418272d575f8ebcad945048060189826cf

PR #2316, opened 2016-08-08
changing csv-import of map to include pk in struct

mikegray:
fixing issue #2292

### DIFF
--- a/samples/go/csv/read.go
+++ b/samples/go/csv/read.go
@@ -129,20 +129,20 @@ func ReadToMap(r *csv.Reader, headersRaw []string, pkIdx int, kinds KindSlice, v
 			panic(err)
 		}
 
-		fieldIndex := 0
 		var pk types.Value
 		fields := make(types.ValueSlice, len(headersRaw))
-		for x, v := range row {
-			if fieldIndex < len(headersRaw) {
-				fieldOrigIndex := fieldOrder[fieldIndex]
+		for i, v := range row {
+			if i < len(headersRaw) {
+				fieldOrigIndex := fieldOrder[i]
 				fields[fieldOrigIndex], err = StringToValue(v, kindMap[fieldOrigIndex])
-				if x == pkIdx {
+				if i == pkIdx {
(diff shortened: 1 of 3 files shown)


samples/go/csv/read.go
```diff
@@ -146,17 +131,18 @@ func ReadToMap(r *csv.Reader, headersRaw []string, pkIdx int, kinds KindSlice, v
 
 		fieldIndex := 0
 		var pk types.Value
-		fields := make(types.ValueSlice, len(headers))
+		fields := make(types.ValueSlice, len(headersRaw))
 		for x, v := range row {
-			if x == pkIdx {
-				pk, err = StringToValue(v, pkKind)
-			} else if fieldIndex < len(headers) {
+			if fieldIndex < len(headersRaw) {
```
aboodman: Can `fieldIndex` ever be less than `len(headersRaw)` now?

mikegray: should always be less, except if the csv file has a different number of fields that the header row indicated.

willhite: It looks like we're ignoring any fields that exceed the number of headers. Shouldn't we print a warning or something? Also, I don't see why we need the fieldIndex var here. Can't you just use 'x'?

mikegray: fixed

mikegray:
PTAL again

willhite:
LGTM. One point though, In the past, we were ignoring the case where there were more fields in a row than headers which means that we will now fail on .csv files that we were able to process. If that's the case we might need to turn the error for this into a warning.

mikegray:
@aboodman what is your thought here - do we panic and stop parsing? just give a warning? or do we silently ignore the issue and not import those fields (this was the default before)?

aboodman:
We actually rely on the old behavior of silent ignore, so we need to keep it.

mikegray:
Ok @willhite I went back to the original notion of ignoring extra fields. PTAL again, minor changes since last review.

PR #2317, opened 2016-08-09
Port path implementation to JS

kalman:
JS had an incomplete implementation, most notably missing: string
parsing, hash # syntax, and support for hash indices.

### DIFF
--- a/js/.eslintrc.js
+++ b/js/.eslintrc.js
@@ -3,6 +3,3 @@
 // http://www.apache.org/licenses/LICENSE-2.0
 
 module.exports = require('@attic/eslintrc');
-
-// TODO: Include this in @attic/eslintrc.
-module.exports.rules['eqeqeq'] = ['error', 'allow-null'];
--- a/js/.eslintrc.js
+++ b/js/.eslintrc.js
@@ -3,3 +3,6 @@
 // http://www.apache.org/licenses/LICENSE-2.0
 
 module.exports = require('@attic/eslintrc');
+
+// TODO: Include this in @attic/eslintrc.
+module.exports.rules['eqeqeq'] = ['error', 'allow-null'];
--- a/js/src/hash.js
+++ b/js/src/hash.js
@@ -42,7 +42,7 @@ export default class Hash {
     return encode(this._digest);
   }
 
-  static parse(s: string): ?Hash {
+  static parse(s: string): Hash | null {
     if (pattern.test(s)) {
       return new Hash(decode(s));
     }
--- a/js/src/path-test.js
+++ b/js/src/path-test.js
@@ -8,26 +8,36 @@ import {assert} from 'chai';
 import {suite, test} from 'mocha';
 import {equals} from './compare.js';
 
-import Path from './path.js';
-import type Value from './value.js';
+import {getHash} from './get-hash.js';
 import List from './list.js';
 import Map from './map.js';
+import {default as Path} from './path.js';
+import Ref from './ref.js';
+import Set from './set.js';
+import type Value from './value.js';
 import {newStruct} from './struct.js';
 
-suite('Path', () => {
+function strify(s) {
+  return JSON.stringify(s);
+}
 
-  async function assertPathEqual(expect: any, ref: Value, path: Path):
-      Promise<void> {
-    // $FlowIssue: need to be able to pass in null for ref
-    const actual = await path.resolve(ref);
-    if (actual === undefined || expect === undefined) {
-      assert.strictEqual(expect, actual);
-      return;
-    }
+function hashIdx(v: Value): string {
+  return `[#${getHash(v).toString()}]`;
+}
 
-    assert.isTrue(equals(expect, actual));
+async function assertResolvesTo(expect: ?Value, ref: Value, str: string) {
+  const p = Path.parse(str);
+  const actual = await p.resolve(ref);
+  if (expect == null) {
+    assert.isTrue(actual == null, `Expected null, but got ${strify(actual)}`);
+  } else if (actual == null) {
+    assert.isTrue(false, `Expected ${strify(expect)}, but got null`);
+  } else {
+    assert.isTrue(equals(expect, actual), `Expected ${strify(expect)}, but got ${strify(actual)}`);
   }
+}
 
+suite('Path', () => {
   test('struct', async () => {
     const v = newStruct('', {
       foo: 'foo',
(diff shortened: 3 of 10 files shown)


kalman:
This is a fairly literal port of Go --> JS. I've changed Go a tiny bit in places to make it more consistent with JS.

js/.eslintrc.js
```diff
@@ -3,3 +3,6 @@
 // http://www.apache.org/licenses/LICENSE-2.0
 
 module.exports = require('@attic/eslintrc');
+
+// TODO: Include this in @attic/eslintrc.
+module.exports.rules['eqeqeq'] = ['error', 'allow-null'];
```
kalman: WDYT?

kalman: I recall some reason we intentionally didn't include it - but I can't remember

arv: In spirit we are following AirBnb js style guide which does not have this exception.

js/src/path-test.js
```diff
@@ -8,100 +8,277 @@ import {assert} from 'chai';
 import {suite, test} from 'mocha';
 import {equals} from './compare.js';
 
-import Path from './path.js';
-import type Value from './value.js';
+import {getHash} from './get-hash.js';
 import List from './list.js';
 import Map from './map.js';
+import {default as Path} from './path.js';
```
arv:
```js
import Path from './path.js';
```

js/src/path-test.js
```diff
@@ -8,100 +8,277 @@ import {assert} from 'chai';
...
+import Ref from './ref.js';
+import Set from './set.js';
+import type Value from './value.js';
 import {newStruct} from './struct.js';
 
-suite('Path', () => {
+function strify(s) {
```
arv: strify?

js/src/path-test.js
```diff
@@ -8,100 +8,277 @@ import {assert} from 'chai';
...
-  async function assertPathEqual(expect: any, ref: Value, path: Path):
-      Promise<void> {
-    // $FlowIssue: need to be able to pass in null for ref
-    const actual = await path.resolve(ref);
-    if (actual === undefined || expect === undefined) {
-      assert.strictEqual(expect, actual);
-      return;
-    }
+function hashIdx(v: Value): string {
+  return `[#${getHash(v).toString()}]`;
```
arv: no need for toString here. Template literals calls toString already.

arv: Unless flow is more strict than js?

js/src/path-test.js
```diff
@@ -8,100 +8,277 @@ import {assert} from 'chai';
...
+}
 
-    assert.isTrue(equals(expect, actual));
+async function assertResolvesTo(expect: ?Value, ref: Value, str: string) {
+  const p = Path.parse(str);
+  const actual = await p.resolve(ref);
+  if (expect == null) {
```
arv: Tests should be more explicit. Use `null` or `undefined` since we should ensure we get the right one.

js/src/path-test.js
```diff
@@ -8,100 +8,277 @@ import {assert} from 'chai';
...
+      await assertResolvesTo(exp, col, hashIdx(val));
+      // Keys resolves to themselves.
+      if (exp != null) {
+        exp = val;
+      }
+      await assertResolvesTo(exp, col, hashIdx(val) + '@key');
+    };
+
+    // Primitives are only addressable by their values.
+    await resolvesTo(m, null, b);
```
arv: Maps return undefined and not null. Let's assert that.

js/src/path-test.js
```diff
@@ -8,100 +8,277 @@ import {assert} from 'chai';
...
+    await assertResolvesTo('fire', s, `.foo[1]${hashIdx(m1)}`);
+    await assertResolvesTo(m1, s, `.foo[1]${hashIdx(m1)}@key`);
+    await assertResolvesTo('car', s, `.foo[1]${hashIdx(m1)}@key["c"]`);
   });
 
-  function assertToString(expect: string, path: Path) {
-    assert.strictEqual(expect, path.toString());
-  }
+  test('parse success', () => {
+    const test = (s: string) => {
```
arv: Can you call this something else than test? It is shadowing the outer one which might be a bit confusing.

arv: maybe testPars

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+import {OrderedKey} from './meta-sequence.js';
+import {OrderedSequence} from './ordered-sequence.js';
+import {fieldNameComponentRe} from './struct.js';
 import {getTypeOfValue, StructDesc} from './type.js';
 
-interface Part {
-  resolve(v: Promise<?Value>): Promise<?Value>;
+const annotationRe = /^@([a-z]+)/;
+
+export interface Part {
```
arv: JSDoc for all exported bindings please.

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+/**
+ * A Path is an address to a Noms value - and unlike hashes (i.e. #abcd...) they can address inlined
+ * values. See https://github.com/attic-labs/noms/blob/master/doc/spelling.md.
+ */
+export default class Path {
+  _parts: Array<Part>;
 
-  constructor(name: string) {
-    this.name = name;
+  static parse(str: string): Path {
```
arv: I would prefer this to be an exported function of this module instead of a static method of the class... or make it work with sub classing...

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+    const p = new Path();
+    p._parse(str);
+    return p;
   }
 
-  resolve(v: Promise<?Value>): Promise<?Value> {
-    return v.then(value => {
-      if (value === null || value === undefined) {
-        return;
+  constructor(parts: Array<Part> = []) {
```
arv: Or `...parts: Array<Path>`

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+    this._parts = parts;
+  }
+
+  append(part: Part): Path {
+    return new Path(this._parts.concat(part));
+  }
+
+  toString(): string {
+    return this._parts.map(p => p.toString()).join('');
```
arv: join already does toString

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+  }
+
+  async resolve(v: Value): Promise<?Value> {
+    let res = v;
+    for (const part of this._parts) {
+      if (res == null) {
```
arv: Use undefined. Testing for both is a sign that we cannot decide which one we want to use.

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
       }
+      res = await part.resolve(res);
+    }
+    return res;
+  }
 
-      const t = getTypeOfValue(value);
-      if (t.kind !== Kind.Struct) {
-        return;
+  _parse(str: string) {
```
arv: I think it might be cleaner to treat Paths as immutable and have parse return a new Path.

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+    if (str === '') {
+      return;
+    }
+
+    const op = str[0], tail = str.slice(1);
```
arv:
```js
const [op, ...tail] = str;
```

js/.eslintrc.js
kalman: That's what I thought. Shame.

Right below they also say "use shortcuts in `if` statements, e.g. instead of `if (foo === '') use `if (foo)`", which is dangerous advice in this context.

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+        }
+        intoKey = true;
+        rem = rem2;
       }
 
-      // $FlowIssue: Flow doesn't know that it's safe to just access the field name here.
-      return value[this.name];
-    });
+      let part: Part;
+      if (idx != null) {
```
arv: What does _parsePathIndex return? null or undefined? It should not return both.

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+  /**
+   * The name of the field, e.g. `.Name`.
+   */
+  name: string;
+
+  constructor(name: string) {
+    this.name = name;
+  }
+
+  resolve(value: Value): Promise<?Value> {
```
arv: We should also start documenting all public methods but it can wait.

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+    const t = getTypeOfValue(value);
+    if (t.kind !== Kind.Struct) {
+      return Promise.resolve();
+    }
+
+    const f = (t.desc: StructDesc).getField(this.name);
```
arv: Nice. I always forget I can do this and I end up doing instanceof which requires a hard dependency.

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+    if (!f) {
+      return Promise.resolve(); // non-present field
```
arv: resolve(undefined)

js/src/path.js
```diff
@@ -4,43 +4,220 @@
...
+    }
+
+    // $FlowIssue: Flow doesn't know that it's safe to just access the field name here.
```
arv: I wonder if we could do `(value: {[key: string]: ?Value})[this.name]`... not sure if it is better though.

js/src/path.js
```diff
@@ -51,86 +228,141 @@ class FieldPart {
...
+/**
+ * Indexes into Maps and Lists by key or index.
+ */
+export class IndexPath {
+  /**
+   * The value of the index, e.g. `[42]` or `["value"]`.
+   */
+  index: indexType;
+
+	/**
```
arv: tab?

js/src/path.js
```diff
@@ -51,86 +228,141 @@ class FieldPart {
...
+   * Whether this index should resolve to the key of a map, given by a `@key` annotation.
+   *
+   * Typically IntoKey is false, and indices would resolve to the values. E.g.  given `{a: 42}`
+   * then `["a"]` resolves to `42`.
+   *
+   * If IntoKey is true, then it resolves to `"a"`. For IndexPath this isn't particularly useful
+   * - it's mostly provided for consistency with HashIndexPath - but note that given `{a: 42}`
+   *   then `["b"]` resolves to nil, not `"b"`.
```
arv: s/nil/undefined|null/?

js/src/path.js
```diff
@@ -51,86 +228,141 @@ class FieldPart {
...
+  /**
+   * The hash of the key or value to search for. Maps and Set are ordered, so this in
+   * O(log(size)).
+   */
+  hash: Hash;
 
-  constructor() {
-    this._parts = [];
-  }
+	/**
```
arv: tab?

js/src/path.js
```diff
@@ -51,86 +228,141 @@ class FieldPart {
...
+      getCurrentValue = cur => cur.getCurrent();
+    } else if (value instanceof Map) {
+      seq = value.sequence;
+      if (this.intoKey) {
+        getCurrentValue = cur => cur.getCurrent()[0]; // key
+      } else {
+        getCurrentValue = cur => cur.getCurrent()[1]; // value
+      }
+    } else {
+      return;
```
arv: `return null`

js/src/struct.js
```diff
@@ -18,6 +18,9 @@ import * as Bytes from './bytes.js';
 
 type StructData = {[key: string]: Value};
 
+export const fieldNameComponentRe = /^[a-zA-Z][a-zA-Z0-9_]*/;
```
arv: jsdoc

js/.eslintrc.js
arv: After reading the rest of the PR it is also clear to me that using `== null` is the cowards way out. We should know what we want.

arv:
Back to you

js/src/path-test.js
arv: nm. I see that we do this in other places.

js/src/path-test.js
kalman: Alright, take $2.

js/src/path-test.js
kalman: Done

js/.eslintrc.js
kalman: Done

js/src/path-test.js
kalman: Flow is fine, done.

js/src/path-test.js
kalman: The problem with this (and you will see this in the patchset) is that I need to add in a bunch of places:
```
let p = ...;
invariant(p !== undefined); // flow happy
if (p === null) {
  // actual null case
}
```
because without the `invariant` flow thinks - correctly, I suppose - that within the `=== null` block it can be `undefined`.

oh well.

arv: We do have a `notNull` function to make this cleaner.

arv: but I guess that does not work in this case

js/src/path-test.js
kalman: The correct expectation is `null` because I want `resolve` to always return null if it's not found, not undefined.

js/src/path-test.js
kalman: I'll call it `t` :-)

js/src/path.js
kalman: done

js/src/path.js
kalman: A static member is consistent with other parseable things like Hash (which has Hash.parse)... it's also nicer for namespacing IMO. I don't ever expect subclassing to play into things. I can change it if you really want though.

js/src/path.js
kalman: Better

js/src/path.js
kalman: done

js/src/path.js
kalman: Using `null`

js/src/path.js
kalman: I've moved this out to a (non-exported) static function - the idea is actually to be mutable during construction, but only expose immutable methods post contruction.

js/src/path.js
kalman: neat

js/src/path.js
kalman: should be clearer now

js/src/path.js
kalman: I've documented these on the interface, should be enough for no

js/src/path.js
kalman: done

js/src/path.js
kalman: Yeah I'm not sure either

js/src/path.js
kalman: done

js/src/path.js
kalman: done

js/src/path.js
kalman: done

js/src/path.js
kalman: done

js/src/path.js
kalman: though the output would be an array sadly :-(

kalman:
PTAL

js/src/path.js
```diff
@@ -35,161 +45,161 @@ export default class Path {
       throw new SyntaxError('Empty path');
     }
     const p = new Path();
-    p._parse(str);
+    constructPath(p._parts, str);
     return p;
   }
 
-  constructor(parts: Array<Part> = []) {
+  constructor(...parts: Array<Part>) {
```
arv: This is not backwards compatible => `npm version major`

js/src/path.js
```diff
@@ -35,161 +45,161 @@ export default class Path {
...
-        throw new SyntaxError('Invalid field: ' + tail);
-      }
-      const idx = match[0].length;
-      this._parts.push(new FieldPath(tail.slice(0, idx)));
-      this._parse(tail.slice(idx));
-      return;
+  if (op === '.') {
+    const match = tail.match(fieldNameComponentRe);
+    if (!match) {
+      throw new SyntaxError('Invalid field: ' + tail);
```
arv:
```js
`Invalid field: ${tail}`
```

js/src/path.js
kalman: I only exposed Path as part of this patch in noms.js, so it's fine.

js/src/path.js
```diff
@@ -366,3 +374,7 @@ export class HashIndexPath {
     return `[#${this.hash.toString()}]${ann}`;
   }
 }
+
+function undefinedToNull(v: ?Value): ?Value {
```
arv: Looks like Flow got a way to distinguish these now:

https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzAFcAuY7AEwFMoBLbKisAXjGyJhgG5McC2y7Ti3LU6DCjyA

```js
function undefinedToNull(v: ?Value): Value | null {
```

js/src/path.js
arv: Yup. I saw that too now...

js/src/path.js
kalman: That's awesome

js/src/hash.js
```diff
@@ -42,7 +42,7 @@ export default class Hash {
     return encode(this._digest);
   }
 
-  static parse(s: string): ?Hash {
+  static parse(s: string): Hash | null {
```
kalman: I've changed this, so bumped the major version.

arv:
LGTM

js/src/struct.js
kalman: ah sorry I forgot this. Adding to next patch.

PR #2320, opened 2016-08-09
adding info to faq

mikegray:
@aboodman PTAL, feel free to rewrite, but these are a few of the questions I have seen several times that would be good to have here to point toward in the future

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -28,6 +28,14 @@ That said, it is definitely possible to have write patterns that defeat this. De
 
 Theoretically, definitely. In Git, for example, the concept of "shallow clones" exists, and we could do something similar in Noms. This has not been implemented yet.
 
+### Is this another one of those NoSQL database thingies?
+
+It depends on the specific features and/or functionality you use to define `NoSQL`. Does Noms support similar type of data? Yes. Does Noms support a flexible schema? Yes.  Does Noms "compromise consistency"? No. Is Noms “eventually consistent”? No, it is always consistent.
+
+### Git can store files, how is this different?
+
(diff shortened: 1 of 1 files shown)


doc/faq.md
```diff
@@ -28,6 +28,14 @@ That said, it is definitely possible to have write patterns that defeat this. De
 
 Theoretically, definitely. In Git, for example, the concept of "shallow clones" exists, and we could do something similar in Noms. This has not been implemented yet.
 
+### Is this another one of those NoSQL database thingies?
+
+It depends on the specific features and/or functionality you use to define `NoSQL`. Does Noms support similar type of data? Yes. Does Noms support a flexible schema? Yes.  Does Noms "compromise consistency"? No. Is Noms “eventually consistent”? No, it is always consistent.
```
aboodman: The thing is that usually when people are asking about this they are thinking about something like mongo. Noms is not a distributed database, so these questions kind of don't make sense. When people talk about 'eventual consistency' they are talking about a multimaster situation with a distributed database -- you have writes at multiple servers and they need to resolve somehow. This is asking a question at a higher level of abstraction than we are currently playing at, so it's kind of hard to answer.

mikegray: Agreed - it's kind of hard to answer which is why I would rather have it answered here and point to it than try to answer it in ad hoc fashion. Can we answer it in a similar way to what you just described above:

`Noms is not currently a distributed database with a multi-master architecture (e.g. Mongo). Hey wait, you said it was distributed - see clarification above. Hey wait, you said it was a database - see clarification above. Is it _anything_ like Mongo? Noms is an open-source database that supports data with a flexible schema, including queries and indices.` 

Or if you would prefer to leave it unanswered that's fine too. What about the Git question?

aboodman:
We should definitely answer them. I'm not excited about these answers. Please keep the PR open and if I don't get back to it keep pestering me.

aboodman:
@rafael-atticlabs is also good at this kind of thing.

mikegray:
@rafael-atticlabs PTAL and advise on wording when you can; or @aboodman if you have had more thoughts feel free to respond.  I am okay if you say to just throw this PR away too at this point.

Issue #1399, opened 2016-05-03
Implement paths in JS


aboodman:
Summary: this is done in Go, just needs to be done in JS

===

# Problem

We say that Noms is a "content-addressed" database as a shorthand, and that in Noms every value has a hash. In truth, though Noms is centered around content-addressing, and one can compute a hash from every value, it is not true that one can _address_ every value in Noms using its hash alone.

For example:

```
let db = new Database(new MemoryStore());
const a = 42;
const b = {bar:a};
db = db.writeValue({foo:b});
```

At this point, you cannot do `db.readValue(<hash of a>)` or `db.readValue(<hash of b>)`. Those values do not exist as independent chunks in noms, and thus are not _addressable_.

This might not seem like a problem, but it comes up from time to time that we would like a way to talk about values in noms which are not within their own chunks:

* For the output of `noms-diff`, we want to talk about where in the tree a change happened. This requires a way to talk about arbitrary values independent of how they are chunked.

* It would be nice to be able to permalink to any node in splore, but we can't -- we can only link to refs.

* We commonly write software that accepts some noms values as input, but we can only pass values that are at the root of a chunk.

A challenging aspect about coming up with such a mechanism for noms is that our collections can contain a mix of types. So a simple path syntax like `foo/sha1-00000/42/baz` can be ambiguous: an map can have the numeric key `42` and the string key `"42"`. A set can have a string value `"sha-00000"` and a ref value `sha1-00000`.

Another challenge: we need to be able to address complex values that are only used as keys in sets or maps.

# Proposal

A _Noms Path_ (or _path_ for short) is a compact string which is a way to efficiently address any value in Noms.

Here are some example paths in a proposed syntax:

```
// root of dataset "foo"
// field "value"
foo.value

// root of dataset "foo"
// field "value"
// value associated with key 42
foo.value[42]

// root of dataset "foo"
// field "value"
// value associated with key having ref sha1-dc09...
foo.value[#sha1-dc094aa17fdaeb9f95b9b5e2f266a1542fba5469]

// root of dataset "foo"
// field "value"
// key having hash sha1-dc09...
// field "monkey"
// value associated with key "foo"
foo.value[#sha1-dc094aa17fdaeb9f95b9b5e2f266a1542fba5469]@key.monkey["foo"]

// root of chunk with hash sha1-00bd
#sha1-00bd6f8012d09ccbeb08a72e74def78a6d3f3b11

// root of chunk with hash sha1-00bd
[...]

PR #2325, opened 2016-08-10
Port AbsolutePath implementation to JS

kalman:


kalman:
2nd part of https://github.com/attic-labs/noms/issues/1399

js/src/absolute-path-test.js
```diff
@@ -0,0 +1,98 @@
+// @flow
```
kalman: Again, this and absolute-path.js are literal translations of the Go implementation.

js/src/path.js
```diff
@@ -40,13 +40,16 @@ export interface Part {
 export default class Path {
   _parts: Array<Part>;
 
-  static parse(str: string): Path {
+  /**
+   * Returns `str` parsed as Path if successful, or a null path with an error message if not.
+   */
+  static parse(str: string): [Path | null, string /* error */] {
```
kalman: I decided to change this to not throw errors... it's a bit weird for a JS idiom, but it's either that or adding a separate way to validate paths (and likewise absolute paths). Validation via try to parse --> catch exception isn't good.

kalman: The other option would be to simply remove the error message - or make Path contain its own error message - but ugh.

js/src/absolute-path.js
```diff
@@ -0,0 +1,114 @@
+// @flow
+
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+import {invariant, notNull} from './assert.js';
+import {datasetRe} from './dataset.js';
+import Database from './database.js';
+import {default as Hash, stringLength} from './hash.js';
```
arv: import Hash, {stringLength} from ...

js/src/absolute-path.js
```diff
@@ -0,0 +1,114 @@
...
+      invariant(parts.length === 2);
+      dataset = parts[1];
+      pathStr = str.slice(parts[0].length);
+    }
+
+    if (pathStr.length === 0) {
+      return [new AbsolutePath(dataset, hash, new Path()), ''];
+    }
+
+    const [path, err] = Path.parse(pathStr);
```
arv: Why not throw an exception instead?

js/src/absolute-path.js
```diff
@@ -0,0 +1,114 @@
...
+      throw new Error('unreachable');
+    }
+
+    if (val === undefined) {
+      val = null;
+    }
+
+    if (val !== null) {
+      return this.path.resolve(val);
+    } else {
```
arv: No else after return

arv:
The go style error reporting is a bit strange. Let's talk about it tomorrow.

js/src/absolute-path.js
kalman: done

js/src/absolute-path.js
```diff
@@ -0,0 +1,114 @@
...
+import Path from './path.js';
+import type Value from './value.js';
+
+const datasetCapturePrefixRe = new RegExp('^(' + datasetRe.source + ')');
+
+/**
+ * An AbsolutePath is a Path relative to either a dataset head, or a hash.
+ */
+export default class AbsolutePath {
+  /** The dataset ID that `_path` is in, or `''` if none. */
```
kalman: typo, should be `path`, fixed

js/src/absolute-path.js
kalman: done

js/src/absolute-path.js
kalman: done

kalman:
PTAL

js/src/absolute-path.js
```diff
@@ -0,0 +1,107 @@
...
+ * `http://demo.noms.io::#abcdef.bar` the `#abcdef.bar` component.
+ */
+export default class AbsolutePath {
+  /** The dataset ID that `path` is in, or `''` if none. */
+  dataset: string;
+
+  /** The hash the that `path` is in, if any. */
+  hash: Hash | null;
+
+  /** Path relative to either `_dataset` or `_hash`. */
```
arv: _dataset? _hash?

js/src/absolute-path.js
```diff
@@ -0,0 +1,107 @@
...
+  path: Path;
+
+  /**
+   * Returns `str` parsed as an AbsolutePath. Throws a `SyntaxError` is `str` isn't a valid path.
```
arv: s/is/if/

js/src/absolute-path.js
```diff
@@ -0,0 +1,107 @@
...
+
+    if (pathStr.length === 0) {
+      return new AbsolutePath(dataset, hash, new Path());
+    }
+
+    const path = Path.parse(pathStr);
+    return new AbsolutePath(dataset, hash, path);
+  }
+
+  constructor(dataset: string, hash: Hash | null, path: Path) {
```
arv: Constructor first?

js/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@attic/noms",
   "license": "Apache-2.0",
-  "version": "57.0.0",
+  "version": "58.0.0",
```
arv: Do you still need a major version inc?

arv:
Back to you.

js/package.json
kalman: Good point, no

js/src/absolute-path.js
kalman: I like having `parse` first because it's the most important function in the class (these aren't meant to be constructed directly).

Also most language style guides I've used over the years put public static methods first (and airbnb doesn't have anything to say about it).

js/src/absolute-path.js
kalman: done

js/src/absolute-path.js
kalman: oops

kalman:
ptal

js/src/absolute-path-test.js
```diff
@@ -0,0 +1,98 @@
...
+    await resolvesTo(null, 'foo.parents');
+    await resolvesTo(null, 'foo.value');
+    await resolvesTo(null, 'foo.value[0]');
+    await resolvesTo(null, `#${getHash('baz').toString()}`);
+    await resolvesTo(null, `#${getHash('baz').toString()}[0]`);
+  });
+
+  test('parse errors', () => {
+    const t = (path: string, exp: string) => {
+      try {
```
arv: This would pass even if there is no exception. Use the following pattern:

```js
let e;
try {
  ...
} catch (ex) {
  e = ex;
}
assert.isTrue(e ...)
...
```

or use the isOk pattern you used below.

arv:
LGTM with nits

js/src/absolute-path-test.js
kalman: Good catch

Issue #2263, opened 2016-08-03
csv-analyze to determine types and PK


mikegray:
It would be better if csv-import didn't always import everything as strings by default. One possible step of getting to more specific types that I have been considering is to build `csv-analyze` (or similar) - command that can either look at the first N lines or perform a more random sampling  throughout the given csv file. The output of csv-analyze would be to answer the questions:
- is any particular field unique across all of the lines used - identify the PK
- for each field, can the value be considered a boolean or a number (or other type) rather than a string?

Output could be generated to match input args needed by csv-import to specify field types and PK (for set/map imports).

This bug is related to bug #1610 - but I believe this specific bug could be a first step which then leads to a more pipelined approach that runs automatically as part of csv-import in the future.

Issue #1610, opened 2016-05-23
I feel like csv_import could be a cool demo of implicit typing


ghost:
If we have it type detect each row as a struct which has fields which are the "smallest" types it fits into.

The demo is import a csv, do noms show of the root and it outputs the schema it created. you can do two examples, one where there's perfect fit and one where there's not.

rafael-atticlabs:
I'm worried now that doing this is at odds with the goal of importing CSV data at maximum speed, but perhaps it could be an option like (--detect-types).

Probably not a near-term priority, though

Issue #2100, opened 2016-07-19
`noms commit`


aboodman:
# Summary

Factor out a `noms commit` command from all existing importers. Take responsibility for commit out of those programs and delegate it to this new command.

# Problem(s)

1. We frequently want the ability to add metadata (#2091) to some importers, but then we end up replicating logic in multiple places.
2. We frequently want the ability to turn duplicate commits into no-ops, but again, this logic would be replicated.
3. We frequently want the ability to aggregate the output of importers into more complex structures. For example, we've had requests to csv-import multiple files as one atomic commit.

# Proposal

```
noms commit [flags] [<path>] <dataset> [--]

Commits `path` to `dataset`. If `--` is present, then the path is read from stdin.

Flags

-allow-dupe             Creates a new commit, even if it would be identical (modulo metadata and parents) to the existing HEAD. Otherwise, such commits are skipped.
-meta key:value         Creates a metadata field called "key" with "value". Value should be human-readable encoded.
-meta-p key:path        Creates a metadata field called "key" set to the value at <path>
-comment value          Alias for -meta 'comment:value'
-date [iso8601-string]  Alias for -meta 'date:str'. 'str' must be iso8601-formatted. If 'str' is empty, it defaults to the current date.
```

Along with implementing this, we'd change things like csv-import to write a root chunk, but not commit it anywhere. We'd also have a convention with all importers to take a `-q[uiet]` flag to suppress ouput except for the final hash, so that you can do things like:

```
csv-import -q ... | noms commit <path> --
```

Note that this proposal alone wouldn't solve problem 3 above, but it would be a step closer. All you'd need is some way to group a bunch of csv imports together into a struct of some sort. Either by adding more flags to `noms commit`, or else implementing something like `noms set make <paths>` (which would construct a set from input paths).

mikegray:
How about instead of `-comment` we use `-message|-m` like `git commit -m "commit message here"`?

aboodman:
Makes sense!

mikegray:
[...]

Issue #1972, opened 2016-07-05
Use all available cores to import csv


aboodman:
Right now on machines with lots of cores, we don't use hardly any of them.

rafael-atticlabs:
Alot of the ground work is now in place for this. The remaining work is to implement a way of splitting the input file into `n` pieces (where n == number of cores), and have each piece, parse, send to `GraphBuilder` and then produce a final p-tree at the end.

PR #2329, opened 2016-08-10
add csv-import --detect-column-types option

mikegray:
a piece of #2263 and a portion of #1610 - avoiding the performance issues of doing this on the actual import by instead making this option only do the detection and then writing the results to stdout (e.g. not performing the import)

### DIFF
--- a/samples/go/csv/csv-import/importer.go
+++ b/samples/go/csv/csv-import/importer.go
@@ -47,6 +47,7 @@ func main() {
 	noProgress := flag.Bool("no-progress", false, "prevents progress from being output if true")
 	destType := flag.String("dest-type", "list", "the destination type to import to. can be 'list' or 'map:<pk>', where <pk> is the index position (0-based) of the column that is a the unique identifier for the column")
 	skipRecords := flag.Uint("skip-records", 0, "number of records to skip at beginning of file")
+	detectColumnTypes := flag.Bool("detect-column-types", false, "detect column types by analyzing a portion of csv file")
 	destTypePattern := regexp.MustCompile("^(list|map):(\\d+)$")
 
 	spec.RegisterDatabaseFlags(flag.CommandLine)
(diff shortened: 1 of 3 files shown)


samples/go/csv/csv-import/importer.go
```diff
@@ -63,7 +64,7 @@ func main() {
 	switch {
 	case flag.NArg() == 0:
 		err = errors.New("Maybe you put options after the dataset?")
-	case flag.NArg() == 1 && *path == "":
+	case flag.NArg() == 1 && *path == "" && !*detectColumnTypes:
```
willhite: I'm trying to understand why this last " && !*detectColumnTypes". It doesn't seem like you want this check here.

mikegray: Because if I am just detecting column types, I don't actually need the dataset arg.

samples/go/csv/csv-import/importer.go
```diff
@@ -146,15 +151,19 @@ func main() {
...
 
-	ds, err := spec.GetDataset(flag.Arg(dataSetArgN))
-	d.CheckError(err)
-	defer ds.Database().Close()
-
 	kinds := []types.NomsKind{}
 	if *columnTypes != "" {
 		kinds = csv.StringsToKinds(strings.Split(*columnTypes, ","))
+	} else if *detectColumnTypes {
+		kinds = csv.GetSchema(cr, headers)
```
willhite: I think it would be slightly simpler if you just moved this code up to line 117. I think it would make it more clear that this detectColumnTypes arg causes something fundamentally different to happen and it would keep this latter code a little simpler. That's just my opinion though, so only make the change if you agree.

willhite:
LGTM with suggestion.

samples/go/csv/csv-import/importer.go
mikegray: If I move it up to line 117 then I also need to move up (copy) or reimplement the creation of the actual csv.NewCSVReader, handling the skip records parameter, and the reading of the header line(s) depending on the params.  I do feel like this file is getting bigger and more complicated with multiple paths, but I also know that @kalman is going to be doing even more work in here to enable more cores/parallelization #1972 in the next few days/weeks and I will be coming back to remove the commit options #2100 and nested maps #1956.  This code will be refactored more soon.  It is good you saw the complexity and pointed out another option, I am just picking one of two evils and will go with the current version for now. It is bugging me so I might come back and refactor this file as a separate PR now.

PR #2334, opened 2016-08-10
Tweak interface/type embedding in databaseCommon

cmasone-attic:
In service of streaming set and map mutators, it'd be helpful for
ValueStore to implement some methods that are exposed only to other
code within the types package, even when a ValueStore is embedded
in a Database. That said, we don't want those methods to become a
part of the public Database API.

To enable this, this patch changes databaseCommon to embed a
*types.ValueStore, instead of composing one. This causes all the
ValueStore methods to be a part of the databaseCommon API, but
since all the Database construction methods return a Database by
interface, stuff like BatchStore() all gets masked. Thus, callers
still can't access ValueStore methods. Databases are passed into
the types package as ValueReadWriters, though, so code in there
can call any method on that interface -- including unexported stuff.

### DIFF
--- a/go/datas/database.go
+++ b/go/datas/database.go
@@ -15,8 +15,7 @@ import (
 // Database provides versioned storage for noms values. Each Database instance represents one moment in history. Heads() returns the Commit from each active fork at that moment. The Commit() method returns a new Database, representing a new moment in history.
 type Database interface {
 	// To implement types.ValueWriter, Database implementations provide WriteValue(). WriteValue() writes v to this Database, though v is not guaranteed to be be persistent until after a subsequent Commit(). The return value is the Ref of v.
-	types.ValueWriter
-	types.ValueReader
+	types.ValueReadWriter
 	io.Closer
 
 	// MaybeHead returns the current Head Commit of this Database, which contains the current root of the Database's value tree, if available. If not, it returns a new Commit and 'false'.
--- a/go/datas/database_common.go
+++ b/go/datas/database_common.go
@@ -14,8 +14,8 @@ import (
 )
 
 type databaseCommon struct {
+	*types.ValueStore
 	cch      *cachingChunkHaver
-	vs       *types.ValueStore
 	rt       chunks.RootTracker
 	rootRef  hash.Hash
 	datasets *types.Map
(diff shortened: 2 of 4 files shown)


go/datas/database.go
```diff
@@ -15,8 +15,7 @@ import (
 // Database provides versioned storage for noms values. Each Database instance represents one moment in history. Heads() returns the Commit from each active fork at that moment. The Commit() method returns a new Database, representing a new moment in history.
 type Database interface {
 	// To implement types.ValueWriter, Database implementations provide WriteValue(). WriteValue() writes v to this Database, though v is not guaranteed to be be persistent until after a subsequent Commit(). The return value is the Ref of v.
-	types.ValueWriter
-	types.ValueReader
+	types.ValueReadWriter
```
cmasone-attic: This means that anything on `ValueReadWriter` becomes a part of the Database interface. This includes unexported methods, but they will only be visible in the package where `ValueReadWriter` is defined. Thus, `ValueReadWriter` can declare something unexported (e.g. `opCache() *opCache`), `ValueStore` can implement it, and while it will be a (masked) part of the Database API, the only code that can use it is stuff inside `types` that takes a ValueReadWriter, such as `Map.NewMutator()`.

cmasone-attic:
PTAL

willhite:
This is pretty cool and definitely a way of approaching the problem that I hadn't considered. Thanks for showing me how to do this.

Now that I've looked at it, I have a couple of observations/questions. 

1) The instance variable "vs" becomes "ValueStore" and so becomes exported. I don't think that's a big deal but just wanted to point it out. 
2) Does opCache() belong in the ValueReadWriter interface or should there be a separate interface with the unexported opCache() method?
3) Does it make sense to implement opCache() on ValueStore or should some other object be added to databaseCommon to manage it?

cmasone-attic:
1) Yeah, I agree it's a downside. My thought was that, because stuff outside this package only ever sees the interface Database (as opposed to the concrete types) and that doesn't expose ValueStore, it's pretty much a wash.
2) You could make it a different interface, but then you'd have to add an argument to NewMutator that is of that new interface type.
3) This kind of only works because the thing that implements the unexported method and the code that wants to use it are both in the same package, and because databaseCommon embeds the type that implements the method. You could split it out of ValueStore, but then you'd have to make databaseCommon embed whatever that new type was.

willhite:
Now that I've implemented streaming sets without the singleton, I definitely think it's cleaner to have the database own the opCacheStore. The complexity of the relationship between datas and types is a little unnerving but this doesn't seem to make it much (if any) more complicated than it already is. 

LGTM.

PR #2335, opened 2016-08-10
Port PathSpec implementation to JS

kalman:


kalman:
Not ready for review, just testing out the PR builder.

kalman:
PTAL.

kalman:
This is the last one :-)

kalman:
The comment in the patch is:

```
    This is a breaking change to the JS API. HashSpec and parseObjectSpec
    have been removed in favour of PathSpec, and parse errors now throw
    SyntaxError instead of silently failing (this matches JSON.parse).
```

js/src/specs.js
```diff
@@ -4,155 +4,180 @@
...
-  static parse(spec: string): ?DatabaseSpec {
-    const match = spec.match(/^(.+?)(:.+)?$/);
-    if (!match) {
-      return null;
-    }
-    const [, scheme, p] = match;
-    if (p && p.indexOf('::') > 0) {
-      return null;
+  /** Cache this for testing, or it will reset every time database() is called. */
+  _ms: MemoryStore|null;
```
arv: Whitespace around `|`

js/src/specs.js
```diff
@@ -4,155 +4,180 @@
...
-        return null;
+        throw new SyntaxError(`Unsupported protocol ${protocol}`);
     }
-    return new this(scheme, path || '');
   }
 
   constructor(scheme: string, path: string) {
     this.scheme = scheme;
     this.path = path;
+    this._ms = scheme === 'mem' ? new MemoryStore() : null;
```
arv: I think the comment would be better here.

js/src/specs.js
```diff
@@ -4,155 +4,180 @@
...
+  /**
+   * Returns a tuple of the database backed by this dataset, and value at the HEAD of this
+   * dataset. If this dataset doesn't have a HEAD, the value will be `null`.
+   *
+   * The caller should always call `close()` when done.
+   */
+  value(): Promise<[Database, Value|null]> {
+    const db = this.database.database();
+    return this.dataset().head().then(commit => {
+      if (commit === null || commit === undefined) {
```
arv: if (!commit) {

arv:
Nice. You need to update some of the samples/js though.

js/src/specs.js
```diff
@@ -4,155 +4,180 @@
...
-    // can't call close() on it. Ideally, we'd fix that.
-    return this.dataset().head()
-      .then(commit => commit && commit.value);
+  /**
+   * Returns a tuple of the database backed by this dataset, and value at the HEAD of this
+   * dataset. If this dataset doesn't have a HEAD, the value will be `null`.
+   *
+   * The caller should always call `close()` when done.
+   */
+  value(): Promise<[Database, Value|null]> {
```
arv: This naming is not very idiomatic js but this is old code...

js/src/specs.js
kalman: done

js/src/specs.js
```diff
@@ -4,155 +4,180 @@
...
-  // Returns parsed spec, or null if the spec was invalid.
-  static parse(spec: string): ?DatabaseSpec {
-    const match = spec.match(/^(.+?)(:.+)?$/);
-    if (!match) {
-      return null;
-    }
-    const [, scheme, p] = match;
-    if (p && p.indexOf('::') > 0) {
-      return null;
+  /** Cache this for testing, or it will reset every time database() is called. */
```
kalman: done

js/src/specs.js
kalman: I like it like this :-) but fair enough. done.

js/src/specs.js
kalman: done

js/src/specs.js
kalman: what should it be?

arv: Probably `getValue` since using getters for Promises is too weird.

kalman: It's not a getter, but it does sound like a getter. I actually might as well change it.

kalman: eh...

kalman: It's a breaking change anyway (return type changed) so I *can* change it without any more breakage. Let me know if you want that.

kalman:
PTAL.

js/src/specs.js
arv: Let's leave it as is for now. There are a few other of these already.

js/src/specs.js
```diff
@@ -4,155 +4,175 @@
...
-  value(): Promise<any> {
-    const database = this.database.database();
-    return database.readValue(this.hash).then(v => database.close().then(() => v));
+  /**
+   * Resolves this PathSpec, and returns the database it was resolved in, and the value it
+   * resolved to. If the value wasn't found, it will be `null`.
+   *
+   * The caller should always call `close()` when done.
+   */
+  value(): Promise<[Database, Value|null]> {
```
arv: I still think we should use spaces around `|` since all other binops require it.

arv:
LGTM with nit

js/src/specs.js
kalman: :-(

PR #2337, opened 2016-08-11
Update to Flow 0.30.0

### LINKED ISSUE #2301: Update Flow
We're currently stuck in the past because TOT flow requires all parameterized types to express those parameters and we have alot of js code which doesn't

arv:
This requires all parameterized types to have type params. Fortunately
one can use `T<any>` which has the same behavior as the old `T` syntax.

We should tighten the types further after this but this unblocks us.

Fixes #2301

### DIFF
--- a/js/src/encoding-test.js
+++ b/js/src/encoding-test.js
@@ -20,12 +20,11 @@ import ValueDecoder from './value-decoder.js';
 import ValueEncoder from './value-encoder.js';
 import type Value from './value.js';
 import {ValueBase} from './value.js';
-import type {NomsKind} from './noms-kind.js';
 import {Kind} from './noms-kind.js';
 import {TestDatabase} from './test-util.js';
 import {encodeValue, decodeValue} from './codec.js';
 import {equals} from './compare.js';
-import {invariant} from './assert.js';
+import {invariant, notNull} from './assert.js';
 import {newStruct, newStructWithType} from './struct.js';
 import {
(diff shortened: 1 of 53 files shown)


arv:
Flow now requires the values in a template literal to be `string | number`.

Flow 0.30.0 found a few more issues with null values.

PTAL

js/src/encoding-test.js
```diff
@@ -300,7 +301,7 @@ suite('Encoding', () => {
     assert.deepEqual(encoding, w.toArray());
 
     const r = new TestReader(encoding);
-    const dec = new ValueDecoder(r, null, staticTypeCache);
+    const dec = new ValueDecoder(r, (null: any), staticTypeCache);
```
kalman: weird, but ok

kalman:
wow. lgtm.

It seems like there isn't much extra that flow picked up? Just the buffer stuff?

js/src/encoding-test.js
arv: I'll add a comment. I didn't want to make ValueDecoder accept null but this test never uses the value reader.

arv: I'll just create a `new TestDatabase()`. Doesn't seem worth the hack.

arv:
It picked up a few null cases.

The reason it didn't pick up more is because I ended up using `T<any>` all over. We should tighten these over time.

Issue #2331, opened 2016-08-10
Replace github.com/tsuru/gnuflag with github.com/juju/gnuflag


kalman:
github.com/tsuru/gnuflag describes itself as a mirror of launchpad.net/gnuflag, but the official location of launchpad.net/gnuflag is now https://github.com/juju/gnuflag. We should just use https://github.com/juju/gnuflag directly.

kalman:
@arv

sunglim:
I'm happy to address this issue.

PR #2340, opened 2016-08-11
Replace github.com/tsuru/gnuflag with github.com/juju/gnuflag

sunglim:
Replace vendor folder and where it is used.
Fixes https://github.com/attic-labs/noms/issues/2331

TEST=go install github.com/attic-labs/noms/cmd/noms

@kalman @arv


sunglim:
I already have signed CLA, by the way :)

arv:
test this please

arv:
Thanks @sunglim 

LGTM

I'll merge once the tests pass

(@kalman I still do not see where it says this is the official repo?)

kalman:
@arv tsuru says "Mirror of launchpad.net/gnuflag" implying that launchpad.net/gnuflag is the original. However launchpad.net/gnuflag says "NOTE: The official location of this package has now moved to https://github.com/juju/gnuflag".

P.S. @sunglim we don't need the "TEST=" line in noms, it's only a chrome thing :-)

kalman:
P.P.S. a better test would be `go test ./cmd/... ./samples/go/...` which means "find every test under the `cmd` and `samples/go` directories, compile, and run them".

sunglim:
> P.S. @sunglim we don't need the "TEST=" line in noms, it's only a chrome thing :-)
> P.P.S. a better test would be go test ./cmd/... ./samples/go/... which means "find every test ...

I got it. Thanks for the review! and tips!

PR #2341, opened 2016-08-11
adding basic script to use on jenkins

mikegray:


mikegray:
TBR

arv:
This is cool.

I thought we decided not to use shell scripts and instead use python?

@kalman @cmasone-attic @aboodman

cmasone-attic:
We definitely decided not to use shell scripts back before you joined us, Mike :-/

We have some stuff like this in Go and some other stuff like this in python, mostly because Go has very limited file and directory manipulation stuff. There's tools/run_all_build.go and also some stuff in attic.git. The Right Thing here is maybe Go? I'm not certain. But we should back this out.

kalman:
why TBR?

kalman:
Hm, I'm guessing because this was grabbed from the jenkins config?

kalman:
> because this was grabbed from the jenkins config

in which case lgtm.

mikegray:
"why TBR?" because it is just one step toward having the build script hosted in the repository it is building and I needed to do some more configuration in Jenkins to use the script.  I wanted to keep making progress to get it working, but I can back it out if that is the correct choice here.

kalman:
You can write scripts directly in the Jenkins console if you want to play with it.

If you TBR please explain why. In this case I wouldn't have TBRd.

mikegray:
Yes, correct. Sorry. My goal was to ensure the build script was in the source code repository and that Jenkins was running the script from the repo. I could have done this differently - sorry for the problem.

kalman:
Cool no problem. Incidentally I absolutely agree that the script should eventually be in the repo.

In fact I recently moved a script out of the jenkins console into noms - which is why it was a bash script, not a python script. Unfortunately python scripts are harder to iterate on - just ping me or whoever and I can review ASAP.

Also: much easier to test unit python than sh.

PR #2346, opened 2016-08-11
Add IterRange() function to Map.

willhite:


go/types/map.go
```diff
@@ -227,6 +227,18 @@ func (m Map) IterAll(cb mapIterAllCallback) {
 	})
 }
 
+func (m Map) IterRange(start, end Value, cb mapIterAllCallback) {
```
kalman: Did you see my comments in the previous patch?

It was basically that IMO this should be `IterFrom(start Value, cb mapIterCallback)` so that:
- caller can decide boundary conditions.
- caller can bail early.
it's easy enough for call site to go from
```
map.IterRange(start, end, func(k, v Value) {
  ...
});
```
to
```
map.IterFrom(start, func(k, v Value) (stop bool) {
  ...
  return end.Less(v); // or !v.Less(end)
})
```

willhite: I did see that comment. I had remembered it as just getting rid of the includeStartValue and includeEndValue flags.  But yeah, I think IterFrom() is better. I'll make the change.

willhite:
@kalman PTAL

go/types/map_test.go
```diff
@@ -1155,3 +1155,41 @@ func TestMapRemoveLastWhenNotLoaded(t *testing.T) {
...
+			assert.True(end.Equals(expectedEnd), "end: %s != expectedEnd: %s", EncodedValue(end), EncodedValue(expectedEnd))
+		}
+	}
+
+	mapKVs := ValueSlice{}
+	for i := -98; i <= 98; i += 2 {
+		mapKVs = append(mapKVs, Number(i), Bool(true))
+	}
+
+	m1 := NewMap(mapKVs...)
```
kalman: How about just `NewMap(generateNumbersAsValuesFromToBy(-98, 98, 1)...)`? Your map values will be numbers instead of `Bool` but that should be fine.

go/types/map_test.go
```diff
@@ -1155,3 +1155,41 @@ func TestMapRemoveLastWhenNotLoaded(t *testing.T) {
 		assert.True(tm.toMap().Equals(nm))
 	}
 }
+
+func TestMapIterFrom(t *testing.T) {
+	assert := assert.New(t)
+
+	test := func(m Map, rangeStart, rangeEnd, expectedStart, expectedEnd Value) {
```
kalman: I think this test can be more straightforward given `IterFrom` is dumber than `IterRange`... I'd test:

1. starting before the first element
2. starting on the first element
3. starting in the middle, on an element
4. starting in the middle, not on an element
5. starting on the last element
6. starting past the end

And then instead of the rangeEnd/expectedEnd stuff, just a single test that it terminates if you return `true`.

(You could potentially also test that every value > the previous one when iterating, but that's intrinsic in the map implementation.)

kalman: Oh - one way to do that would be to have a function which flattens a Map into a ValueSlice using IterFrom, then use ValueSlice.Equals(mapKVs[some:range]).

willhite:
@kalman PTAL

kalman:
nice, lgtm

PR #2356, opened 2016-08-12
Let Splore open a noms path

kalman:
Currently Splore can only open a database (db=) or a hash (hash=). This
removes the hash and replaces it with the much more general path (p=).

For example, what used to be:
  ?db=http://demo.noms.io&hash=abcd
is now:
  ?p=http://demo.noms.io::#abcd

and of course more complex paths work, like:
  ?p=http://demo.noms.io::#abcd.field[42].name

I've made 2 changes to the noms API for this:
1. Added support for an authorization token to DatabaseSpec.
2. Exposed getHashOfValue in noms.js.

### DIFF
--- a/js/src/specs.js
+++ b/js/src/specs.js
@@ -22,7 +22,6 @@ import type Value from './value.js';
 export class DatabaseSpec {
   scheme: string;
   path: string;
-  authorization: string;
 
   _ms: MemoryStore | null;
 
@@ -67,7 +66,6 @@ export class DatabaseSpec {
   constructor(scheme: string, path: string) {
     this.scheme = scheme;
     this.path = path;
-    this.authorization = '';
     // Cache the MemoryStore for testing, or it will reset every time database() is called.
     this._ms = scheme === 'mem' ? new MemoryStore() : null;
   }
@@ -80,11 +78,8 @@ export class DatabaseSpec {
       case 'mem':
         return new Database(new BatchStoreAdaptor(notNull(this._ms)));
       case 'http':
-      case 'https': {
-        const auth = this.authorization;
-        const opts = auth !== '' ? {headers: {Authorization: `Bearer ${auth}`}} : undefined;
-        return new Database(new HttpBatchStore(`${this.scheme}:${this.path}`, undefined, opts));
-      }
+      case 'https':
+        return new Database(new HttpBatchStore(`${this.scheme}:${this.path}`));
       default:
         throw new Error('Unreached');
     }
--- a/js/src/specs.js
+++ b/js/src/specs.js
@@ -22,6 +22,7 @@ import type Value from './value.js';
 export class DatabaseSpec {
   scheme: string;
   path: string;
+  authorization: string;
 
   _ms: MemoryStore | null;
 
@@ -66,6 +67,7 @@ export class DatabaseSpec {
   constructor(scheme: string, path: string) {
     this.scheme = scheme;
     this.path = path;
+    this.authorization = '';
     // Cache the MemoryStore for testing, or it will reset every time database() is called.
     this._ms = scheme === 'mem' ? new MemoryStore() : null;
   }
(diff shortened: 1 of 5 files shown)


kalman:
This is a bit... rough, I suppose. I'll have a fresh look tomorrow and see if it can be neater, but I'm not sure without some bigger changes to Splore.

CC @aboodman I've changed some of the messaging

js/src/specs.js
```diff
@@ -22,6 +22,7 @@ import type Value from './value.js';
 export class DatabaseSpec {
   scheme: string;
   path: string;
+  authorization: string;
```
aboodman: Can we just remove this feature from splore -- it is not needed anymore.

samples/js/splore/src/main.js
```diff
@@ -222,30 +248,38 @@ function handleNodeClick(e: MouseEvent, id: string) {
...
+        {this.props.msg}
         <form style={{margin:'0.5em 0'}} onSubmit={e => this._handleOnSubmit(e)}>
           <input type='text' ref='db' autoFocus={true} style={inputStyle}
-            defaultValue={params.db || 'https://demo.noms.io/cli-tour'}
-            placeholder='noms database URL'
+            defaultValue={params.db}
+            placeholder={`database URL (${demoServer})`}
+          />
+          or
+          <input type='text' ref='p' style={inputStyle}
```
aboodman: Why not have this be _only_ the absolute path?

samples/js/splore/src/main.js
```diff
@@ -52,31 +64,45 @@ function load() {
     });
   }
 
-  if (!params.db) {
-    renderPrompt();
+  if (params.p && params.db) {
```
arv: Can we get rid of `params.db`?

js/src/specs.js
kalman: Done

samples/js/splore/src/main.js
kalman: Not if we want Splore to be able to render a database root.

Generally speaking, given a spec, we can't tell if it's for a path or a db. Splore needs to be told what it's looking for.

If you want to drop the feature of inspecting a database... well I find that kind of handy, if awkward in Splore. @aboodman

samples/js/splore/src/main.js
kalman: Not 100% sure what you mean. As in require `db`, then if there is `p` interpret `db+p` as a path spec (otherwise just `db` as a db spec)?

That would work. The nice thing about having it being a full spec is that it's consistent with our samples. We don't say `noms show --db http://demo.noms.io/cli-tour --p sf-crime.value[42]`, we say `noms show http://demo.noms.io::sf-crime.value[42]`. I like that there is a parallel between the two, IMO we should be using paths when we have the opportunity.

On the other hand it does make the code simpler, and the UI a little less ambiguous.

kalman:
Addressed other comments

samples/js/splore/src/main.js
aboodman: I guess that is different because in the case of `show` you can only provide a path. Providing a db is meaningless. In this case, I think I would prefer to have the path field just be the path alone.

samples/js/splore/src/main.js
```diff
@@ -52,31 +64,41 @@ function load() {
...
+      if (value === null) {
+        throw new Error('No value found at ' + params.p);
+      }
+      return [getHashOfValue(value), value];
+    });
   } else {
-    httpStore.getRoot().then(setRootHash);
+    const dbSpec = DatabaseSpec.parse(params.db);
+    database = dbSpec.database();
+    // TODO: Don't access _rt, expose getRoot somewhere.
```
arv: Is there a bug filed?

samples/js/splore/src/main.js
```diff
@@ -222,29 +244,39 @@ function handleNodeClick(e: MouseEvent, id: string) {
   }
 }
 
-class Prompt extends React.Component<void, {}, void> {
+type PromptProps = {
+  msg: string,
+}
```
arv: ;

arv:
LGTM

samples/js/splore/src/main.js
kalman: done

samples/js/splore/src/main.js
kalman: done

kalman:
Ok, `p` is now an absolute path. PTAL.

samples/js/splore/src/main.js
```diff
@@ -301,6 +294,10 @@ function renderPrompt(msg: string) {
 }
 
 function render() {
+  // TODO: Set up better Promise chaining. rootHash is loaded asynchronously elsewhere.
```
arv: Hah, I saw this too!

arv:
LGTM

PR #2357, opened 2016-08-12
Nomdex

willhite:


willhite:
This has been reviewed once in #2208 which was closed because it belonged to a pre-public fork of Noms. The main changes to this version are that it uses StreamingSets, Command and Help modules from noms/cmd/util and it has temporarily relaxed it's type check on indexes pending a bug fix to IsSubtype().

willhite:
@aa This is an updated version of nomdex that should work on your flickr photos index and uses streaming sets.

willhite:
@aa PTAL, see earlier comment for functionality that has changed.

samples/go/nomdex/vrange.go
```diff
@@ -0,0 +1,158 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"sort"
+
+	"github.com/attic-labs/noms/go/types"
+)
+
+type ventry struct {
```
aboodman: I think a better name for this might be: _bound_

samples/go/nomdex/vrange.go
```diff
@@ -0,0 +1,158 @@
...
+	v    types.Value
+	incl bool
+	inf  int8
```
aboodman: What does _inf_ stand for?

samples/go/nomdex/vrange.go
aboodman: Can you please add some commentary to these types? It's a bit dense for a newcomer.

samples/go/nomdex/vrange.go
aboodman: guessing _infinity_... Its fine to use the terse Go style, but in that case, add a comment dereferencing the name for the poor reader :).

samples/go/nomdex/vrange.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+type vrange struct {
```
aboodman: I think that the "v" in this name is superfluous -- it makes me think it is a _virtual_ range or something. Suggest just _range_. Or maybe _resultRange_ if you want to be more descriptive.

samples/go/nomdex/vrange.go
```diff
@@ -0,0 +1,158 @@
...
+	lower ventry
+	upper ventry
+}
+
+type vrangeslice []vrange
```
aboodman: Correct capitalization would be: vRangeSlice

samples/go/nomdex/vrange.go
aboodman: Nope, not infinity.  Hm...

aboodman: OK, after meatspace explanation, I would suggest this:

```
type bound struct {
  v types.Value
  incl bool
  positiveInfinity bool
  negativeInfinity bool
}
```

samples/go/nomdex/vrange.go
```diff
@@ -0,0 +1,158 @@
...
+		if i == 0 {
+			fmt.Fprintf(w, "\n#################\n")
+		}
+		fmt.Fprintf(w, "range %d: %s\n", i, r)
+	}
+	if len(vs) > 0 {
+		fmt.Fprintf(w, "\n")
+	}
+}
+func (r vrange) and(o vrange) (vslice vrangeslice) {
```
aboodman: Keep the method definitions with the structs. So define ventry/bound, then its methods, then vrange, then its methods, then vrangeslice, then it's methods.

samples/go/nomdex/vrange.go
```diff
@@ -0,0 +1,158 @@
...
+		return true
+	}
+	if o.upper.isGreaterThanOrEqual(r.lower) && o.upper.isLessThanOrEqual(r.upper) {
+		return true
+	}
+	return false
+}
+
+func (v ventry) isLessThanOrEqual(o ventry) (res bool) {
+	return v.equals(o) || v.isLessThan(o)
```
aboodman: Because `v.isLessThan()` calls `v.Equals()` you end up calling `v.Equals()` twice here. Is there  a way to avoid that?

samples/go/nomdex/vrange.go
aboodman: But I can see the value in the way you have it. If you want to keep it that way, it is fine with me so long as you add a comment.

samples/go/nomdex/vrange.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+//
+func (v ventry) isLessThan(o ventry) (res bool) {
+	if v.equals(o) {
```
aboodman: I think it would make the code more clear if `Less()` only tested whether the value was less, without delegating to equals().

samples/go/nomdex/vrange.go
aboodman: It would be worth adding a comment saying that this sorts by the lower bound, since it's non-obvious how to sort ranges.

samples/go/nomdex/vrange.go
```diff
@@ -0,0 +1,158 @@
...
+	if !r.intersects(o) {
+		return []vrange{}
+	}
+
+	lower := r.lower.maxValue(o.lower)
+	upper := r.upper.minValue(o.upper)
+	return []vrange{vrange{lower, upper}}
+}
+
+func (v vrange) or(o vrange) (vslice vrangeslice) {
```
aboodman: This is really cool. Never seen an impl of this before...

samples/go/nomdex/expr.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/attic-labs/noms/go/types"
+)
+
+type expr interface {
+	ranges() vrangeslice
+	dbgPrintTree(w io.Writer, level int)
+}
+
+type logExpr struct {
+	op    boolop
```
aboodman: Correct capitalization would be _boolOp_

samples/go/nomdex/nomdex_update.go
```diff
@@ -0,0 +1,142 @@
...
+
+var (
+	inPathArg  = ""
+	outDsArg   = ""
+	relPathArg = ""
+)
+
+var update = &util.Command{
+	Run:       runUpdate,
+	UsageLine: "up -in-path <path> -out-ds <dspath> -by <relativepath>",
```
aboodman: These are all double-dashes.

samples/go/nomdex/nomdex_update.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	outDs := dataset.NewDataset(db, outDsArg)
+	relPath, err := types.ParsePath(relPathArg)
+	if printError(err, "Error parsing -by value\n\t") {
+		return 1
+	}
+
+	typeCacheMutex := sync.Mutex{}
+	typeCache := map[*types.Type]bool{}
```
aboodman: I mentioned this in the other review, but you cannot use the typecache this simplistically. Consider a path like `.foo[42]`. This can resolve on one instance of a type, but not another.

aboodman:
this is really neat - bty

samples/go/nomdex/nomdex_update.go
willhite: Right. An even more compelling case, imo, is something like .foo["geopos"].
I'm still thinking about this one. Smarter solutions aren't trivial. I'm
looking at have a new method on path called MayResolve(v Value) bool, that
examines the type to see if the path can be resolved but it's not trivial.

willhite:
Thanks, I'm hoping to have multiple indexes working by tomorrrow.

willhite:
@aa PTAL, this incorporates all the recommended changes except one. I stuck with infinity being an int rather than two bool values. The name changes you suggested were very helpful and they make the code a lot more understandable.

samples/go/nomdex/range_descriptor.go
```diff
@@ -0,0 +1,159 @@
...
+	"bytes"
+	"fmt"
+	"io"
+	"sort"
+
+	"github.com/attic-labs/noms/go/types"
+)
+
+type bound struct {
+	value        types.Value
```
aboodman: Indent looks wrong

samples/go/nomdex/range_descriptor.go
```diff
@@ -0,0 +1,159 @@
...
+	include  bool
```
aboodman: _inclusive_ would be better

samples/go/nomdex/range_descriptor.go
```diff
@@ -0,0 +1,159 @@
...
+	infinity int8
+}
+
+func (b bound) isLessThanOrEqual(o bound) (res bool) {
+	return b.equals(o) || b.isLessThan(o)
+}
+
+//
+func (b bound) isLessThan(o bound) (res bool) {
+	if b.infinity < o.infinity {
```
aboodman: Why isn't this just:

```
if b.infinity < o.infinity {
  return true
}

if b.value.Less(o.value) {
  return true
}

return false
```

aboodman: Ah, answering own question, because of inclusiveness.

samples/go/nomdex/range_descriptor.go
```diff
@@ -0,0 +1,159 @@
...
+
+	if b.infinity > o.infinity {
+		return false
+	}
+
+	if b.value.Less(o.value) {
+		return true
+	}
+
+    if b.value.Equals(o.value) {
```
aboodman: run `gofmt`

samples/go/nomdex/range_descriptor.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (v bound) maxValue(o bound) (res bound) {
+	if v.isLessThan(o) {
+		return o
+	}
+	return v
+}
+
+type rangeDescriptor struct {
```
aboodman: Why not just `range` ?

aboodman:
lgtm w/ one last naming nit

PR #2362, opened 2016-08-12
Add missing ==nil check to list diff

kalman:
This isn't triggering any problem in particular, I just noticed it. All
it meant was that noms log might take unnecessarily long with list diff
when there are a lot of changes.

### DIFF
--- a/cmd/noms/diff/diff.go
+++ b/cmd/noms/diff/diff.go
@@ -68,7 +68,7 @@ func diffLists(w io.Writer, p types.Path, v1, v2 types.List) (err error) {
 
 		if splice.SpRemoved == splice.SpAdded {
 			// Heuristic: list only has modifications.
-			for i := uint64(0); i < splice.SpRemoved; i++ {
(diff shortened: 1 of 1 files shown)


cmd/noms/diff/diff.go
```diff
@@ -68,7 +68,7 @@ func diffLists(w io.Writer, p types.Path, v1, v2 types.List) (err error) {
 
 		if splice.SpRemoved == splice.SpAdded {
 			// Heuristic: list only has modifications.
-			for i := uint64(0); i < splice.SpRemoved; i++ {
+			for i := uint64(0); i < splice.SpRemoved && err == nil; i++ {
```
willhite: do we want to return if err != nil?

kalman: It'll exit up on line 65

kalman:
PTAL

willhite:
@kalman Did something change since the last time I looked at this?

kalman:
no, but you didn't lgtm.

willhite:
LGTM

PR #2364, opened 2016-08-13
fix error message

jangie:
really silly simple fix for consistency's sake.


arv:
Thanks for the patch.

LGTM

Issue #148, opened 2015-07-29
Implement implicit merge policies


aboodman:
After #147 is done, datastore will enforce that all writes be serialized. Boo :(.

To regain some concurrency, we should implement simple automatic merge policies for our datastructures. What I'm imagining is something where we take a pair of concurrent mutations of the tree, along with the previous version of the tree, an consider the three of them recursively, applying some simple rules like:

- If any of the three nodes are different types, conflict:
- If we are dealing with a set:
  - If the same object is both removed and inserted wrt parent: conflict
- If we are dealing with a map:
  - If the same key is both removed and inserted wrt parent: conflict
  - If the same key is inserted wrt parent, but with different values: conflict
- If we are dealing with a struct:
  - same as map
- If we are dealing with a list:
  - Same as map but substitute "index" for "key"
  - Note that this means that things like "slice" expensive in terms of concurrency

... otherwise, the concurrent modification is allowed.

Another way to look at this (I think) is that the operations we will automatically merge effectively make our datastructures into CRDTs (https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type).

jaseg:
Since you already had a look at CRDTs: Shapiro et al. wrote down a number of CRDTs with their merge strategies in their ["Comprehensive Study of Convergent and Commutative Replicated Data Types"](http://hal.upmc.fr/inria-00555588/document). This might contain some inspiration.

I think generally it would make sense to allow the user to specify a merge strategy for a node when creating said node. In some cases, one might want a conflict to be raised on concurrent modification, in other cases e.g. LWW using the committing device's local wall-clock time would be appropriate. LWW would e.g. be a sensible choice when synchronizing multiple devices that all belong to the same user, since it can be assumed that this user is a) somewhat aware of what modifications she did on each device and b) only using one device at a time.

AFAICT adopting this kind of automated merge policy would make noms data structures into state-based CRDTs, with the difference that noms always stores all history. AFAICT there would be two things that a tighter integration of CRDT semantics could bring:
* Depending on the data type one might be able to use knowledge about the merge procedure to reduce the amount of data transferred  during sync
* The automated merge procedure might be used to compute efficient deltas from full states in order to reduce history storage size (effectively transforming historical data from a state-based "CvRDT" to a operation-based "CmRDT").

aboodman:
jaseg, thanks for chiming in. I agree with both your points.

[...]

PR #2371, opened 2016-08-15
Update FAQ to address conflict resolution

cmasone-attic:
Towards #148

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -14,7 +14,7 @@ Noms can and should be a database, not merely a datastore. We're not quite there
 
 No, decentralized like Git.
 
-Specifically, Noms isn't itself a peer-to-peer network. If you can get two instances to share data, somehow, then they can synchronize. Noms doesn't define how this should happen though. 
+Specifically, Noms isn't itself a peer-to-peer network. If you can get two instances to share data, somehow, then they can synchronize. Noms doesn't define how this should happen though.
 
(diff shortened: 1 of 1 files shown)


cmasone-attic:
PTAL

doc/faq.md
```diff
@@ -28,6 +28,12 @@ That said, it is definitely possible to have write patterns that defeat this. De
 
 Theoretically, definitely. In Git, for example, the concept of "shallow clones" exists, and we could do something similar in Noms. This has not been implemented yet.
 
+### What about conflict resolution?
```
aboodman: Sorry, I'm going to rewrite this a little:

### How does Noms handle conflicts?

Currently, Noms doesn't handle them at all. But in the near future, we plan to add [automatic merge policies](link) for several common cases of conflicts. 

Similar to how Git uses simple heuristics to merge many classes of edits to same same file/directory, we believe that we can use some simple heuristics to merge Noms data. In fact, we think we can even do better than Git, because in general Noms data is more granular than data in Git. We have structs, sets, lists, and maps, where Git just has directories and files.

Whether to use an auto-merge policy will be something that developers choose to use on a case-by-case basis, and can always opt out of for more complex scenarios.

### Why don't you use CRDTs?

<your answer>

cmasone-attic:
NP

cmasone-attic:
PTAL

aboodman:
LGTM

PR #2372, opened 2016-08-15
Use best set/map diff in noms diff

kalman:
In 7c103344 I changed noms log to use left-right diff so that it
completes faster, but it changed noms diff to use left-right as well.
Noms diff is supposed to use best diff.

### DIFF
--- a/cmd/noms/diff/diff.go
+++ b/cmd/noms/diff/diff.go
@@ -17,16 +17,27 @@ type (
 	valueFunc func(k types.Value) types.Value
 )
 
+type diffWriter struct {
+	w         io.Writer
+	leftRight bool
+}
+
+func (w diffWriter) Write(p []byte) (n int, err error) {
+	return w.w.Write(p)
+}
+
 func shouldDescend(v1, v2 types.Value) bool {
 	kind := v1.Type().Kind()
 	return !types.IsPrimitiveKind(kind) && kind == v2.Type().Kind() && kind != types.RefKind
 }
 
-func Diff(w io.Writer, v1, v2 types.Value) error {
-	return diff(w, types.Path{}, nil, v1, v2)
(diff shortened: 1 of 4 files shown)


kalman:
Hold up, I did something dumb.

kalman:
Ok PTAL

cmd/noms/diff/diff.go
```diff
@@ -17,16 +17,27 @@ type (
...
+}
+
 func shouldDescend(v1, v2 types.Value) bool {
 	kind := v1.Type().Kind()
 	return !types.IsPrimitiveKind(kind) && kind == v2.Type().Kind() && kind != types.RefKind
 }
 
-func Diff(w io.Writer, v1, v2 types.Value) error {
-	return diff(w, types.Path{}, nil, v1, v2)
+// Writes the Diff from |v1| to |v2| to |w|.
```
cmasone-attic: If you start this comment with "Diff", godoc can pull the comment out for automatic documentation I believe.

`// Diff writes the diff from....`

cmd/noms/diff/diff.go
```diff
@@ -17,16 +17,27 @@ type (
...
+// If |leftRight| is true then the left-diff diff is used for ordered sequences - see Diff vs DiffLeftRight in Set and Map.
```
cmasone-attic: Do you mean '...then the left-right diff is used...'

cmasone-attic:
LGTM w/comment nits

cmd/noms/diff/diff.go
kalman: done

cmd/noms/diff/diff.go
kalman: I do, thanks

PR #2374, opened 2016-08-15
Go marshal

### LINKED ISSUE #1591: Bring back marshalling mechanism for go
The basic idea is to be able to write go structs and read onto go structs

arv:
Basic support for creating Noms value from Go values as well as reading
Noms values onto Go values.

Typical usage:

```
// Go to Noms
type T struct {
  S string
  N int32
}
nomsVal, err := marshal.Marshal(T{"Hi", 42})

// Noms to Go
var goVal T
err = marshal.Unmarshal(types.NewStruct("T", types.StructData{
  "S": types.String("Bye"),
  "N": types.Number(555),
}, &goVal)
```

Fixes #1591

### DIFF
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -21,7 +21,7 @@ import (
 // fields to the fields used by Marshal (either the struct field name or its tag).
 // Unmarshal will only set exported fields of the struct.
 //
-// If a Noms value is not appropriate for a given target type, or if a Noms number overflows the target type, Unmarshal returns a UnmarshalTypeMismatchError or OverflowError respectively
+// If a Noms value is not appropriate for a given target type, or if a Noms number overflows the target type, Unmarshal returns a UnmarshalTypeMismatchError.
 //
 func Unmarshal(v types.Value, out interface{}) (err error) {
 	defer func() {
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -71,7 +71,7 @@ type UnmarshalTypeMismatchError struct {
 }
 
 func (e *UnmarshalTypeMismatchError) Error() string {
-	return "Cannot unmarshal " + e.Value.Type().Describe() + " into Go value of type " + e.Type.String() + e.details
+	return fmt.Sprintf("Cannot unmarshal %s into Go value of type %s%s", e.Value.Type().Describe(), e.Type.String(), e.details)
 }
 
 // OverflowError describes a Noms value that was not appropriate for a value of a specific Go number type.
@@ -164,8 +164,7 @@ type decField struct {
 }
 
 func structDecoder(t reflect.Type) decoderFunc {
-	// TODO: Is this handling wrong noms type?
-	if t.Implements(nomsValueType) {
+	if t.Implements(nomsValueInterface) {
 		return nomsValueDecoder
 	}
 
@@ -196,6 +195,7 @@ func structDecoder(t reflect.Type) decoderFunc {
 
 	d = func(v types.Value, rv reflect.Value) {
 		s := v.(types.Struct)
+		// If the name is empty then the Go struct has to be anonymous.
 		if s.Type().Desc.(types.StructDesc).Name != name {
 			panic(&UnmarshalTypeMismatchError{v, rv.Type(), ", names do not match"})
 		}
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -1,0 +1,227 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package marshal
+
+import (
+	"fmt"
+	"reflect"
+	"runtime"
+	"sync"
+
+	"github.com/attic-labs/noms/go/types"
+)
+
+// Unmarshal converts a Noms value into a Go value. It decodes v and stores the result
+// in the value pointed to by out.
+//
+// Unmarshal uses the inverse of the encodings that Marshal uses with the following additional rules:
+//
+// To unmarshal a Noms struct into a Go struct, Unmarshal matches incoming object
+// fields to the fields used by Marshal (either the struct field name or its tag).
+// Unmarshal will only set exported fields of the struct.
+//
+// If a Noms value is not appropriate for a given target type, or if a Noms number overflows the target type, Unmarshal returns a UnmarshalTypeMismatchError or OverflowError respectively
+//
+func Unmarshal(v types.Value, out interface{}) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if _, ok := r.(runtime.Error); ok {
+				panic(r)
+			}
+			if s, ok := r.(string); ok {
+				panic(s)
+			}
+			err = r.(error)
+		}
+	}()
+
+	rv := reflect.ValueOf(out)
+	if rv.Kind() != reflect.Ptr || rv.IsNil() {
+		return &InvalidUnmarshalError{reflect.TypeOf(out)}
+	}
+	rv = rv.Elem()
+	d := typeDecoder(rv.Type())
+	d(v, rv)
+	return
+}
+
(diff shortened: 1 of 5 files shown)


arv:
PTAL

go/marshal/decode.go
```diff
@@ -0,0 +1,227 @@
...
+// If a Noms value is not appropriate for a given target type, or if a Noms number overflows the target type, Unmarshal returns a UnmarshalTypeMismatchError or OverflowError respectively
+//
+func Unmarshal(v types.Value, out interface{}) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if _, ok := r.(runtime.Error); ok {
+				panic(r)
+			}
+			if s, ok := r.(string); ok {
+				panic(s)
```
cmasone-attic: Does doing `panic(r)` here preserve the initial stack trace in any way?

go/marshal/decode.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+// OverflowError describes a Noms value that was not appropriate for a value of a specific Go number type.
+type OverflowError struct {
+	Value types.Number
+	Type  reflect.Type // type of Go value it could not be assigned to
+}
+
+func (e *OverflowError) Error() string {
+	return fmt.Sprintf("Cannot unmarshal %g into Go value of type %s", e.Value, e.Type)
```
cmasone-attic: so, very very nitty...but why use Sprintf here and string concatenation above?

go/marshal/decode.go
```diff
@@ -0,0 +1,227 @@
...
+
+type decField struct {
+	name    string
+	decoder decoderFunc
+	index   int
+}
+
+func structDecoder(t reflect.Type) decoderFunc {
+	// TODO: Is this handling wrong noms type?
+	if t.Implements(nomsValueType) {
```
cmasone-attic: I might call `nomsValueType` `nomsValueInterface` or something, because a struct doesn't implement a type, it implements an interface.

go/marshal/decode.go
```diff
@@ -0,0 +1,227 @@
...
+		fields = append(fields, decField{
+			name:    getFieldName(tags, f),
+			decoder: typeDecoder(f.Type),
+			index:   i,
+		})
+	}
+
+	d = func(v types.Value, rv reflect.Value) {
+		s := v.(types.Struct)
+		if s.Type().Desc.(types.StructDesc).Name != name {
```
cmasone-attic: Noms structs can have empty names, right? I think this means that you can't marshal/unmarshal nameless Noms struct types, and I don't know that we want that restriction

go/marshal/decode_test.go
```diff
@@ -0,0 +1,385 @@
...
+	t(types.Bool(true), &b, true)
+	t(types.Bool(false), &b, false)
+
+	for _, s := range []string{"", "s", "hello", "💩"} {
+		var s2 string
+		t(types.String(s), &s2, s)
+	}
+
+	var list types.List
+	list2 := types.NewList()
```
cmasone-attic: In all these tests with Noms collections, I would put something in e.g. `list2`, to make sure that the target isn't just getting an empty collection stuffed in there, but rather getting the appropriate Value assigned.

go/marshal/decode.go
cmasone-attic: ah, I see. This makes unnamed Noms struct types work with anonymous Go structs.

go/marshal/encode.go
```diff
@@ -0,0 +1,291 @@
...
+//
+// Marshal traverses the value v recursively. Marshal uses the following type-dependent encodings:
+//
+// Boolean values are encoded as Noms types.Bool.
+//
+// Floating point and integer values are encoded as Noms types.Number. At the moment this might lead to some loss in precision because types.Number currently takes a float64.
+//
+// String values are encoded as Noms types.String.
+//
+// Struct values are encoded as Noms structs (types.Struct). Each exported struct field becomes a member of the struct unless
```
cmasone-attic: In this paragraph, it'd be good to make clear whether you mean a Noms Struct or a Go struct every time you use "struct"

go/marshal/encode.go
```diff
@@ -0,0 +1,291 @@
...
+// the struct field's tag value is the field name. Examples:
+//
+//   // Field is ignored.
+//   Field int `noms:"-"`
+//
+//   // Field appears in a Noms struct as field "myName".
+//   MyName int
+//
+//   // Field appears in a Noms struct as key "myName".
+//   Field int `json:"myName"`
```
cmasone-attic: `noms:"myName"`

go/marshal/encode.go
```diff
@@ -0,0 +1,291 @@
...
+//
+// Unlike encoding/json Marshal, this does not support "omitempty".
+//
+// Anonymous struct fields are currently not supported.
+//
+// Noms values (values implementing types.Value) are copied over without any change.
+//
+// Go maps, slices, arrays, pointers, complex, interface (non types.Value), function are not supported. Attempting to encode such a value causes Marshal to return an UnsupportedTypeError.
```
cmasone-attic: Are embedded struct supported when Marshalling? I know they're not when Unmarshalling.

go/marshal/encode.go
```diff
@@ -0,0 +1,291 @@
...
+//
+func Marshal(v interface{}) (nomsValue types.Value, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if _, ok := r.(runtime.Error); ok {
+				panic(r)
+			}
+			if s, ok := r.(string); ok {
+				panic(s)
```
cmasone-attic: Same Q about stack traces as above

go/marshal/encode_test.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	t(types.Bool(true), true)
+	t(types.Bool(false), false)
+
+	for _, s := range []string{"", "s", "hello", "💩"} {
+		t(types.String(s), s)
+	}
+
+	t(types.NewList(), types.NewList())
```
cmasone-attic: put something in these collections, methinks

cmasone-attic:
LGTM with comment nits and some test nits

go/marshal/encode.go
cmasone-attic: I see now that they're not, but it wasn't clear to me from this comment. I guess embedding structs leads to anonymous fields, so you do address it here, but I think it'd be good to call it out specifically.

go/marshal/decode.go
arv: I'm not sure. I followed the pattern in encoding/json. I'm going to go through these and see if I can reduce the number of different errors I use and try to figure out the `runtime.Error` case too,

go/marshal/decode.go
arv: I wanted 'g' but I can get that through strconv.FormatFloat. I find the Sprintf cleaner though. Maybe I'll change some of the other ones to use that instead.

go/marshal/decode.go
arv: interfaces are types too but I agree that nomsValueInterface is better (more specific)

go/marshal/decode.go
arv: I'll add a comment

arv: I can also see us relaxing this (and other things) later.

go/marshal/decode_test.go
arv: Agreed, even though `types.List{}` is not equal to `types.NewList()`.

go/marshal/encode.go
arv: Done.

go/marshal/encode.go
arv: Done.

go/marshal/encode.go
arv: Done.

cmasone-attic:
you can land without changing the panic and error stuff if you want, though since errors are part of the public API I understand if you want to take a pass at reducing the variety you use.

go/marshal/decode.go
arv: The stack trace is preserved.

I'm trying to reduce the number of errors. Having 5 different errors seems too many. I could reduce the numbers if we do not care about exposing anything but a the error string.

PR #2375, opened 2016-08-16
Use Go marshal API in the HR sample

arv:


arv:
@aboodman Last commit shows marshal in action for samples/go/hr/

go/marshal/decode.go
```diff
@@ -0,0 +1,227 @@
...
+// To unmarshal a Noms struct into a Go struct, Unmarshal matches incoming object
+// fields to the fields used by Marshal (either the struct field name or its tag).
+// Unmarshal will only set exported fields of the struct.
+//
+// If a Noms value is not appropriate for a given target type, or if a Noms number overflows the target type, Unmarshal returns a UnmarshalTypeMismatchError or OverflowError respectively
+//
+func Unmarshal(v types.Value, out interface{}) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			if _, ok := r.(runtime.Error); ok {
```
aboodman: Interesting - where did you find this pattern?

aboodman: (by which I mean how do you know it must be one of these three types?)

go/marshal/decode.go
```diff
@@ -0,0 +1,227 @@
...
+				panic(r)
+			}
+			if s, ok := r.(string); ok {
+				panic(s)
+			}
+			err = r.(error)
```
aboodman: This seems questionable. If you call some code and it panics with something that implements error, you will return it here. That doesn't seem like what you want. I think you should test for the particular errors you intend to panic with.

aboodman:
seems reasonable to me

go/marshal/decode.go
arv: I didn't intend for this PR to be the review of the marshalling. Only to show the marshalling in action.

I'm going to go through the error handling code and update #2374 instead.

arv:
@aboodman Now ready for review. PTAL.

mikegray:
lgtm

aboodman:
Better PR description needed.

aboodman:
lgtm

PR #2377, opened 2016-08-16
Fix shell script flags

arv:


arv:
@kalman TBR

mikegray:
Hmm, nit - but this is a shell script, right? Shouldn't you remove it?

PR #2378, opened 2016-08-16
JS: Do not shadow struct properties

### LINKED ISSUE #2332: JS: Struct field names can collide with ValueBase interface methods
If you create a struct (e.g. with jsonToNoms) and you happen to have field names of `type`, `hash` or `chunks`, when code goes to use those apis, it'll receive the value of the field, rather than the `ValueBase` method and then bad things happen.

One way to solve this would be that `Struct` doesn't inherit from `ValueBase` and those methods are moved to some reflectively API, e.g. `typeOfValue` gets the `_type` field of `Struct`.

At very least, struct to throw right now if created with field names which collide with API, rather than failing down the road mysteriously.

arv:
When we create JS struct classes we no longer creates properties for
`chunks`, `hash` or `type` (nor for `toString`, `hashOwnProperty` etc)

If you are using structs with field names that clashes with these then
use a `StructMirror`.

Fixes #2332

### DIFF
--- a/js/src/struct.js
+++ b/js/src/struct.js
@@ -43,9 +43,9 @@ export const fieldNameRe = new RegExp(fieldNameComponentRe.source + '$');
  *   setS(value: string): MyStruct;
  * }
  *
- * With one major exception and that is that if the field name conflicts with any of the properties
- * in ValueBase (or Object), such as `chunks`, `hash` or `type` (or `toString`, `hasOwnProperty`
- * etc.), then these are not reflected directly on the struct instance.
+ * With one major exception: if the field name conflicts with any of the properties in ValueBase (or
+ * Object), such as `chunks`, `hash` or `type` (or `toString`, `hasOwnProperty` etc.), then these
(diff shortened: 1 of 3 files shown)


arv:
@kalman PTAL

I also added JSDoc to all exported bindings

js/src/struct.js
```diff
@@ -43,7 +43,12 @@ export const fieldNameRe = new RegExp(fieldNameComponentRe.source + '$');
  *   setS(value: string): MyStruct;
  * }
  *
- * To reflect over structs you can create a new StructMirror.
+ * With one major exception and that is that if the field name conflicts with any of the properties
```
kalman: nit: this would read slightly better as "With one major exception: if the field name conflicts..." rather than "With one major exception and that is that if the field name conflicts..."

kalman:
lgtm

js/src/struct.js
arv: Thanks.

PR #2379, opened 2016-08-16
Render diff struct field paths .foo not ["foo"]

kalman:


cmd/noms/diff/diff.go
```diff
@@ -34,31 +35,33 @@ func shouldDescend(v1, v2 types.Value) bool {
...
+		return line(dw, ADD, nil, v2)
 	}
 
-	line(w, DEL, key, v1)
-	return line(w, ADD, key, v2)
+	return diff(dw, types.Path{}, nil, v1, v2)
+}
+
+func diff(w diffWriter, p types.Path, key, v1, v2 types.Value) error {
+	switch v1.Type().Kind() {
```
kalman: I noticed that `diff` is never called if the values are equal, or if we're not descending. Move checks into Diff.

cmd/noms/diff/diff.go
```diff
@@ -118,44 +121,53 @@ func diffLists(w diffWriter, p types.Path, v1, v2 types.List) (err error) {
...
 
 func diffStructs(w diffWriter, p types.Path, v1, v2 types.Struct) error {
-	return diffOrdered(w, p, field, func(cc chan<- types.ValueChanged, sc <-chan struct{}) {
-		v2.Diff(v1, cc, sc)
-	},
+	str := func(v types.Value) string {
+		return string(v.(types.String))
+	}
+	return diffOrdered(w, p, field,
+		func(v types.Value) types.PathPart { return types.NewFieldPath(str(v)) },
```
kalman: This was the bug

arv:
LGTM

Issue #2376, opened 2016-08-16
Improvements to Go marshal


arv:
We decided to do a pretty bare bone marshalling for Go to Noms (and Noms to Go) to get a better sense for how these are used. There are a bunch of potential features we could add:

- [ ] Slices/Array
- [ ] Map (can only do maps of Number/String/Bool)
- [ ] Embedded structs
- [ ] Blobs - Treat `[]byte` and `io.Reader` as a `Blob`.
- [ ] Support `omitempty`
- [ ] Treat pointers as `Ref`

It is not clear if these are worth it but I wanted to jot them down before I move on.

arv:
Another thing that has come up is how to do unions? We could probably do it using a struct field tag...

arv:
One way to do unions would be to do:

```go
type T struct {
  A int    `noms:"a"`
  B string `noms:"a"`  // same name
}
```

would generate a struct with type:

```
struct T {
  a: Number | String
}
```

but it still has issues with zero values....

rafael-atticlabs:
I feel like there's a (possibly) related issue. Consider a `List<Set<Number> | Number>`. With marshalling, there's no escaping the need to tests which type is `myList.get(0)` before you can do something with it.

Note this is apropos of #1124

sandalwing:
Support for pointers to arbitrary structs would be good.  I don't know that it should automatically encode to `Ref`, though.  I'd just like something like this to be supported, since it works OK with JSON:

```go
package main

import (
	"encoding/json"
	"fmt"
)

type Quux struct {
	Val int
}

type Foo struct {
	Bar *Quux
}

func main() {
	data, _ := json.Marshal(&Foo{&Quux{3}})
	fmt.Println(string(data))
}
```

arv:
Another issue is to allow renaming the struct when we go from Go to Noms:

input:

```go
type X struct {n int}
x := X{n: 42}
v, err := marshal.Marshal(x)
```

wanted output:

```noms
struct Y {n: Number}
```

One way to do this is to introduce interface for MarshalNoms (and UnmarshalNoms) which X struct could implement. Similar to [MarshalJSON](https://golang.org/pkg/encoding/json/#RawMessage.MarshalJSON).

Another way to do this is to have a transformer that can transform the Go value before it is handed to the underlying marshal. This is similar to have the reviver in [JSON.parse](https://tc39.github.io/ecma262/#sec-json.parse) in JS works. A strawman API for this could be:

```go
[...]

PR #2381, opened 2016-08-16
Go marshal slices and arrays

arv:
Slices and arrays are marshaled to Noms List. For arrays we also enforce
that the length is the same.

Towards #2376

### DIFF
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -22,6 +22,7 @@ import (
 // Unmarshal will only set exported fields of the struct.
 //
 // To unmarshal a Noms list into a slice, Unmarshal resets the slice length to zero and then appends each element to the slice.
+// As a special case, to unmarshal an empty Noms list into a slice,  Unmarshal replaces the slice with a new empty slice.
 //
 // To unmarshal a Noms list into a Go array, Unmarshal decodes Noms list elements into corresponding Go array elements.
 //
@@ -155,7 +156,7 @@ func uintDecoder(v types.Value, rv reflect.Value) {
 	}
 }
 
-var structDecoderCache struct {
+var decoderCache struct {
 	sync.RWMutex
 	m map[reflect.Type]decoderFunc
 }
@@ -171,9 +172,9 @@ func structDecoder(t reflect.Type) decoderFunc {
 		return nomsValueDecoder
 	}
 
-	structDecoderCache.RLock()
-	d := structDecoderCache.m[t]
-	structDecoderCache.RUnlock()
+	decoderCache.RLock()
+	d := decoderCache.m[t]
+	decoderCache.RUnlock()
 	if d != nil {
 		return d
 	}
@@ -213,12 +214,12 @@ func structDecoder(t reflect.Type) decoderFunc {
 		}
 	}
 
-	structDecoderCache.Lock()
-	if structDecoderCache.m == nil {
-		structDecoderCache.m = map[reflect.Type]decoderFunc{}
+	decoderCache.Lock()
+	if decoderCache.m == nil {
+		decoderCache.m = map[reflect.Type]decoderFunc{}
 	}
-	structDecoderCache.m[t] = d
-	structDecoderCache.Unlock()
+	decoderCache.m[t] = d
+	decoderCache.Unlock()
 	return d
 }
 
@@ -230,9 +231,22 @@ func nomsValueDecoder(v types.Value, rv reflect.Value) {
 }
 
 func sliceDecoder(t reflect.Type) decoderFunc {
-	decoder := typeDecoder(t.Elem())
-	return func(v types.Value, rv reflect.Value) {
+	decoderCache.RLock()
+	d := decoderCache.m[t]
+	decoderCache.RUnlock()
+	if d != nil {
+		return d
+	}
+
+	var decoder decoderFunc
+
+	d = func(v types.Value, rv reflect.Value) {
 		list := v.(types.List)
+		if list.Len() == 0 {
+			rv.Set(reflect.MakeSlice(t, 0, 0))
+			return
+		}
+
 		slice := rv.Slice(0, 0)
 		list.IterAll(func(v types.Value, i uint64) {
 			elemRv := reflect.New(t.Elem()).Elem()
@@ -241,12 +255,31 @@ func sliceDecoder(t reflect.Type) decoderFunc {
 		})
 		rv.Set(slice)
 	}
+
+	decoderCache.Lock()
+	if decoderCache.m == nil {
+		decoderCache.m = map[reflect.Type]decoderFunc{}
+	}
+	decoderCache.m[t] = d
+	decoderCache.Unlock()
+
+	decoder = typeDecoder(t.Elem())
+
+	return d
 }
 
 func arrayDecoder(t reflect.Type) decoderFunc {
-	size := t.Len()
-	decoder := typeDecoder(t.Elem())
-	return func(v types.Value, rv reflect.Value) {
+	decoderCache.RLock()
+	d := decoderCache.m[t]
+	decoderCache.RUnlock()
+	if d != nil {
+		return d
+	}
+
+	var decoder decoderFunc
+
+	d = func(v types.Value, rv reflect.Value) {
+		size := t.Len()
 		list := v.(types.List)
 		l := int(list.Len())
 		if l != size {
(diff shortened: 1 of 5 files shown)


go/marshal/decode.go
```diff
@@ -217,3 +224,31 @@ func nomsValueDecoder(v types.Value, rv reflect.Value) {
 	}
 	rv.Set(reflect.ValueOf(v))
 }
+
+func sliceDecoder(t reflect.Type) decoderFunc {
+	decoder := typeDecoder(t.Elem())
+	return func(v types.Value, rv reflect.Value) {
+		list := v.(types.List)
+		l := int(list.Len())
+		slice := reflect.MakeSlice(t, l, l)
```
arv: This creates a new slice. I'm not sure this is the desired behavior. The other places we just update the object passed in... I'll check what encoding/json does.

arv:
@kalman PTAL

arv:
I changed how we unmarshal into slices to behave like encoding/json.

PTAL

aboodman:
I still don't think we should do this.

aboodman:
OK, nevermind, go for it.

go/marshal/decode.go
```diff
@@ -217,3 +229,75 @@ func nomsValueDecoder(v types.Value, rv reflect.Value) {
 	}
 	rv.Set(reflect.ValueOf(v))
 }
+
+func sliceDecoder(t reflect.Type) decoderFunc {
+	decoderCache.RLock()
```
arv: Do you have any suggestion how to prevent the code duplication here?

go/marshal/decode_test.go
```diff
@@ -383,3 +377,131 @@ func TestDecodeInvalidNomsType(t *testing.T) {
...
+func TestDecodeRecursive(t *testing.T) {
+	assert := assert.New(t)
+
+	type Node struct {
+		Value    int
+		Children []Node
+	}
+
+	typ := types.MakeStructType("Node", []string{"children", "value"}, []*types.Type{
+		types.MakeListType(types.MakeCycleType(0)),
```
arv: Woot!

go/marshal/decode.go
```diff
@@ -217,3 +229,75 @@ func nomsValueDecoder(v types.Value, rv reflect.Value) {
...
+	if d != nil {
+		return d
+	}
+
+	var decoder decoderFunc
+
+	d = func(v types.Value, rv reflect.Value) {
+		list := v.(types.List)
+		if list.Len() == 0 {
+			rv.Set(reflect.MakeSlice(t, 0, 0))
```
kalman: Doesn't this change the semantics of a an empty list vs a non-empty list?

1. If the list is empty, the struct has a new empty slice
2. If the list isn't empty, the struct reuses the old slice.

I would have thought (1) is better for both (create a new slice each time), then you can use MakeSlice with the exact right size?

Unless I'm misunderstanding.

go/marshal/decode.go
```diff
@@ -217,3 +229,75 @@ func nomsValueDecoder(v types.Value, rv reflect.Value) {
...
+	return d
+}
+
+func arrayDecoder(t reflect.Type) decoderFunc {
+	decoderCache.RLock()
+	d := decoderCache.m[t]
+	decoderCache.RUnlock()
+	if d != nil {
+		return d
+	}
```
kalman: yeah, wonder if you should be able to go
```
d := decoderCache.Get(t)
// and
decoderCache.Set(t, d) // for below
```
save a lot of the lock/unlock repetition.

go/marshal/encode.go
```diff
@@ -131,25 +133,28 @@ func typeEncoder(t reflect.Type) encoderFunc {
...
-func structEncoder(t reflect.Type) encoderFunc {
+func structEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc {
 	if t.Implements(nomsValueInterface) {
 		return nomsValueEncoder
 	}
 
-	structEncoderCache.RLock()
-	e := structEncoderCache.m[t]
-	structEncoderCache.RUnlock()
+	encoderCache.RLock()
```
kalman: ditto

go/marshal/decode.go
arv: I'm following encoding/json here.

One reason to not do 1 is that it would allocate more memory... another reason is that it is not consistent with how structs are filled in. We could change structs to always allocate new memory too if we wanted but it seems useful to marshal some values onto an existing struct.

```go
s := struct {
  X int `noms:"-"`
  Y int
  Z int
} {
  X: 1,
}
Unmarshal(types.NewStruct("", types.StructData{
  "Y": types.Number(2),
  "Z": types.Number(3),
}), &s)
```

arv:
PTAL

go/marshal/decode.go
```diff
@@ -144,11 +155,29 @@ func uintDecoder(v types.Value, rv reflect.Value) {
...
-var structDecoderCache struct {
+type decoderCacheT struct {
 	sync.RWMutex
 	m map[reflect.Type]decoderFunc
 }
 
+var decoderCache = &decoderCacheT{}
+
+func (c *decoderCacheT) get(t reflect.Type) decoderFunc {
+	c.RLock()
```
kalman: You _could_ do this in 3 lines with
```
c.RLock()
defer c.RUnlock()
return c.m[t]
```
and using `defer` is probably a better habit anyway.

go/marshal/decode.go
```diff
@@ -144,11 +155,29 @@ func uintDecoder(v types.Value, rv reflect.Value) {
...
+	return d
+}
+
+func (c *decoderCacheT) set(t reflect.Type, d decoderFunc) {
+	c.Lock()
+	if c.m == nil {
+		c.m = map[reflect.Type]decoderFunc{}
+	}
+	c.m[t] = d
+	c.Unlock()
```
kalman: consider deferring this as well, though it makes no difference in this case

kalman:
lgtm but you might want to get a 2nd opinion on the parentStructTypes stuff

go/marshal/encode.go
```diff
@@ -159,7 +162,7 @@ func structEncoder(t reflect.Type) encoderFunc {
 			return types.NewStructWithType(structType, values)
 		}
 	} else {
-		// Cannot precompute the type since there are noms collections.
+		// Cannot precompute the Noms type since there are noms collections.
```
ahl: consistent capitalization of Noms/noms

go/marshal/decode.go
```diff
@@ -144,11 +155,28 @@ func uintDecoder(v types.Value, rv reflect.Value) {
 	}
 }
 
-var structDecoderCache struct {
+type decoderCacheT struct {
 	sync.RWMutex
```
ahl: fwiw I bet that a mutex is as good as an rwlock here (and below). rwlocks typically make a bigger impact around longer hold times; otherwise the RFO of the cache line is going to dominate

go/marshal/encode.go
```diff
@@ -260,15 +274,24 @@ func nomsType(t reflect.Type) *types.Type {
 	case reflect.String:
 		return types.StringType
 	case reflect.Struct:
-		return structNomsType(t)
-	default:
-		// This will be reported as an error at a different layer.
-		return nil
+		for i, pst := range parentStructTypes {
```
ahl: I would suggest that this makes more sense in structNomsType(), but ... it's a gentle suggestion.

ahl:
LGTM

go/marshal/encode.go
arv: I agree. Done.

go/marshal/decode.go
```diff
@@ -217,3 +238,51 @@ func nomsValueDecoder(v types.Value, rv reflect.Value) {
...
+func sliceDecoder(t reflect.Type) decoderFunc {
+	d := decoderCache.get(t)
+	if d != nil {
+		return d
+	}
+
+	var decoder decoderFunc
+
+	d = func(v types.Value, rv reflect.Value) {
+		slice := rv.Slice(0, 0)
```
cmasone-attic: You should be able to pre-allocate this slice to be the right length, right? To avoid having to grow it as items are appended.

Also, I wonder if it's worth seeing if rv is already a slice that has enough capacity for the list contents?

arv: It uses the slice that was passed in. This is for consistency with encoding/json but also to be consistent with structs (which gets filled in)

PR #2383, opened 2016-08-17
Set, Union, & Intersection iterators.

willhite:


go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+import "github.com/attic-labs/noms/go/d"
+
+type SetIterator interface {
+	// Next returns nil, when no objects remain
+	Next() Value
+
+	// AdvanceTo(v) returns:
+	// * nil if there are no objects >= v.
+	// * the next value in the iteration if  v <= curValue
+	// * the first value > v if v is not in the set, or nil if there is no value > v
```
kalman: The second part of this point "nil if there is no value > v" contradicts the first point "nil if there are no objects >= v".

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+
+package types
+
+import "github.com/attic-labs/noms/go/d"
+
+type SetIterator interface {
+	// Next returns nil, when no objects remain
+	Next() Value
+
+	// AdvanceTo(v) returns:
```
kalman: AdvanceTo ... also advances the cursor right? So this comment could be "AdvanceTo advances the cursor to the next value >= `v`".

kalman: Followed by "The return value is ...?"
And for that matter what state is the cursor in after this point?

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+	// * the first value > v if v is not in the set, or nil if there is no value > v
+	AdvanceTo(v Value) Value
+}
+
+type setIterator struct {
+	s      Set
+	cursor *sequenceCursor
+}
+
+func (s Set) Iterator() SetIterator {
```
kalman: I think it's better to define this in set.go

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+	return &setIterator{s: s, cursor: nil}
+}
+
+func (si *setIterator) Next() (v Value) {
```
kalman: the `v` in `(v Value)` isn't used, likewise the `res` in `(res Value)` below... and almost all of them?

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+	} else {
+		si.cursor.advance()
+	}
+	if si.cursor.valid() {
+		return si.cursor.current().(Value)
+	}
+	return nil
+}
+
+func (si *setIterator) AdvanceTo(v Value) (res Value) {
```
kalman: It seems like a simpler to understand API here would be for AdvanceTo to advance the cursor to `v`, and return the Value it's now on. The caller can then call `Next` if they want. It would also make the code simpler I think.

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	si.cursor, _ = si.s.getCursorAtValue(v)
+	if si.cursor.valid() {
+		return si.cursor.current().(Value)
+	}
+	return nil
+}
+
+// iterState contains iterator and it's current value
```
kalman: You could also just put a Value() method on Iterator? a little bit weird for an iterator to have such a method, but I could see it being useful in other circumstances.

go/types/set_iterator.go
willhite: The problem is that these iterators can be nested, which means shifting the cursor around can mean shifting lots of cursors. I was looking for a semantic that would be universally applicable across all 3 types of iterators and constrain the user to only do sensible things.

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+	return st.v
+}
+
+type UnionIterator struct {
+	aState iterState
+	bState iterState
+}
+
+func NewUnionIterator(iterA, iterB SetIterator) SetIterator {
+	if iterA == nil || iterB == nil {
```
kalman: Why not just `d.Chk.NotNil(iterA); d.Chk.NotNil(iterB)`?

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+		if iterA == nil && iterB == nil {
+			return nil
+		}
+		if iterA == nil {
+			return iterB
+		}
+		return iterA
+	}
+	a := iterState{i: iterA, v: iterA.Next()}
```
kalman: I also think having Value on SetIterator would make this simpler, you wouldn't need to call Next I think

go/types/set_iterator.go
willhite: I did think about this. It would allow AdvanceTo(v) to return the current value when v == current value. But I opted to keep it as streamlined as possible. I'm open to changing that.

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+	b := iterState{i: iterB, v: iterB.Next()}
+	return &UnionIterator{aState: a, bState: b}
+}
+
+func (u *UnionIterator) Next() (res Value) {
+	i := compareValue(u.aState.v, u.bState.v)
```
kalman: `switch (compareValue)`?

go/types/set_iterator.go
willhite: Whatever gets supported on SetIterator would need to be supported by Union and Intersection as well.

go/types/set_iterator_test.go
```diff
@@ -0,0 +1,174 @@
...
+	assert := assert.New(t)
+
+	numbers := append(generateNumbersAsValues(6), Number(10), Number(15))
+	s := NewSet(numbers...)
+	i := s.Iterator()
+
+	vs := ValueSlice{}
+	for v := i.Next(); v != nil; v = i.Next() {
+		vs = append(vs, v)
+	}
```
kalman: Pattern is repeated a lot, how about a method like `iterToSlice(SetIterator) ValueSlice`? then just `vs := iterToSlice(i)` here.

go/types/set_iterator_test.go
```diff
@@ -0,0 +1,174 @@
...
+	assert.True(vs.Equals(numbers), "Expected: %v != actual: %v", numbers, vs)
+
+	i = s.Iterator()
+	assert.Panics(func() { i.AdvanceTo(nil) })
+	assert.Equal(Number(0), i.AdvanceTo(Number(-20)))
```
kalman: It would be good to test what happens when calling Next after AdvanceTo. E.g. presumably `i.AdvanceTo(5); iterToSlice(i) => [6, 10, 15]`?

go/types/set_iterator.go
kalman: Another independent comment here: looking at the tests, the behavior here does seem reasonable, though the method might be better named similar to `NextFrom`. `AdvanceTo` sounds like it does just take the iterator to a position such that the next call to `Next` returns that value.

kalman:
Back to you

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+	if !didAdvance {
+		return u.Next()
+	}
+	if u.aState.v == nil {
+		return u.bState.Next()
+	}
+	if u.bState.v == nil {
+		return u.aState.Next()
+	}
+	if u.aState.v.Less(u.bState.v) {
```
kalman: use `compareValue`?

go/types/set_iterator.go
kalman: also switch here, etc

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil
+	}
+	a := iterState{i: iterA, v: iterA.Next()}
+	b := iterState{i: iterB, v: iterB.Next()}
+	return &IntersectionIterator{aState: a, bState: b}
+}
+
+func (u *IntersectionIterator) Next() (res Value) {
+	cnt := 0
+	for i := compareValue(u.aState.v, u.bState.v); i != 0; i = compareValue(u.aState.v, u.bState.v) {
```
kalman: Perhaps nicer:
```
for {
  switch compareValue(...) {
    case: -1: ...
    case: 1: ...
    case: 0
      break
  }
}
```

go/types/set_iterator.go
kalman: Fair enough, you've convinced me.

willhite:
@kalman PTAL, I didn't change the name of AdvanceTo(). I'm not convinced there's anything better. Other than that, I think I've addressed all your comments.

go/types/set_iterator.go
```diff
@@ -0,0 +1,213 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package types
+
+import "github.com/attic-labs/noms/go/d"
+
+type SetIterator interface {
```
arv: Please document all new public API.

arv: Should this just be `Iterator`? I don't see what is specific to Set here?

kalman: Abstractly yes, but practically no:

A Next() function makes sense to return a Value, but only for Sets and Lists; Maps would return (Value, Value).

And AdvanceTo isn't possible on Lists.

So this Iterator could only be used (in our current data structures) on a Set.

go/types/set_iterator.go
kalman: ^

go/types/set_iterator.go
```diff
@@ -0,0 +1,208 @@
...
+	// Next returns nil, when no objects remain
+	Next() Value
+
+	// AdvanceTo(v) advances the cursor to a new position based on the following rules and returns the value at that position.
+	// * if v <= curValue, advances to the next value in the iteration.
```
kalman: More thoughts:

The behaviour that if the current value < v then it advances is strange. If I have [1,2,3,4,5], current value is 3, and I "advance to" 1, then the iterator is now at 4. That's counterintuitive to me.

I still think this would ideally be either better named, or have a different interface, or different behavior depending on which is most sensible.

E.g. what if the comment were actually "like Next(), but returns a value > `v`", which I think is what it's doing at the moment? That makes a lot more sense to me (and the name NextFrom evokes that more than AdvanceTo, but don't feel particularly strongly about that).

kalman: I guess it should be "returns a value >= v".

go/types/set_iterator.go
```diff
@@ -0,0 +1,208 @@
...
+	if !si.cursor.valid() {
+		return nil
+	}
+
+	curValue := si.cursor.current().(Value)
+	if first {
+		return curValue
+	}
+
+	if v.Less(curValue) {
```
kalman: Should this be <= not <? I.e. !curValue.Less(v)

go/types/set_iterator.go
```diff
@@ -0,0 +1,208 @@
...
+	if st.v == nil {
+		return nil
+	}
+	v := st.v
+	st.v = st.i.Next()
+	return v
+}
+
+func (st *iterState) AdvanceTo(v Value) Value {
+	if st.v == nil || v == nil {
```
kalman: would prefer d.Chk.NotNil(v) and then just `if st.v == nil { return nil }`

go/types/set_iterator.go
```diff
@@ -0,0 +1,208 @@
...
+	case -1:
+		return u.aState.Next()
+	case 0:
+		u.aState.Next()
+		return u.bState.Next()
+	case 1:
+		return u.bState.Next()
+	}
+
+	return nil
```
kalman: panic("unreachable")?

go/types/set_iterator.go
kalman: ^

I.e. would this work
```
...
if !didAdvance {
  return u.Next()
}
switch compareValue(u.aState.v, u.bState.v) {
  case -1:
    return u.bState.Next()
  case 1:
    return u.aState.Next()
  case 0:
    u.aState.Next()
    return u.bState.Next()
}
panic("unreachable")
```

go/types/set_iterator.go
```diff
@@ -0,0 +1,208 @@
...
+	// we only get here if aState and bState are equal
+	res := u.aState.v
+	u.aState.Next()
+	u.bState.Next()
+	return res
+}
+
+func (u *IntersectionIterator) AdvanceTo(v Value) Value {
+	d.Chk.NotNil(v)
+	if u.aState.v != nil && !v.Less(u.aState.v) {
```
kalman: I've seen this pattern a few times now, maybe everything should just be using `compareValue` which both documents the behavior of nil, and does the check properly.

Up to you.

kalman: i.e. `if compareValue(v, u.aState.v) == -1 {`

go/types/set_iterator.go
```diff
@@ -0,0 +1,208 @@
...
+		u.aState.AdvanceTo(v)
+	}
+	if u.bState.v != nil && !v.Less(u.bState.v) {
+		u.bState.AdvanceTo(v)
+	}
+	return u.Next()
+}
+
+// considers nil max value, return -1 if v1 < v2, 0 if v1 == v2, 1 if v1 > v2
+func compareValue(v1, v2 Value) int {
```
kalman: IMO this function would be more explicit like

```
if v1 == nil && v2 == nil {
  return 0
}
if v2 == nil || v1.Less(v2) {
  return -1
}
if v1 == nil || v2.Less(v1) {
  return 1
}
return 0
```

up to you.

kalman:
lgtm with a bunch more comments

go/types/set_iterator.go
```diff
@@ -113,45 +116,45 @@ func (u *UnionIterator) Next() Value {
 	case 1:
 		return u.bState.Next()
 	}
-
+	d.Chk.Fail("Unreachable code")
```
kalman: I would just panic (don't see the point in using our test-assert wrappers here)

go/types/set_iterator_test.go
```diff
@@ -161,3 +163,146 @@ func TestCombinationIterator(t *testing.T) {
 	expectedRes = intsToValueSlice(0, 30, 60)
 	assert.True(vs.Equals(expectedRes), "Expected: %v != actual: %v", expectedRes, vs)
 }
+
+var callCount = 0
```
kalman: can you put this on UnionTestIterator/IntersectionTestIterator instead?

kalman: or a pointer

kalman:
lgtm

go/types/set_iterator.go
```diff
@@ -0,0 +1,207 @@
...
+package types
+
+import (
+	"github.com/attic-labs/noms/go/d"
+)
+
+// SetIterator defines methods that can be used to efficiently iterate through a set in 'Noms-defined'
+// sorted order.
+type SetIterator interface {
+	// Next returns subsequent values from a set. It returns nil, when no objects remain
```
aboodman: Missing period at end of comment.

go/types/set_iterator.go
```diff
@@ -0,0 +1,207 @@
...
+	Next() Value
+
+	// NextFrom(v) advances the cursor to a new position based on the following rules and returns the value at that position.
+	// * if v <= curValue, advances to the next value in the iteration.
```
aboodman: I don't understand this case at all -- first of all 'curValue' is not defined in this interface anywhere, so I don't know what it refers to...

willhite: Would this work?
+	// * if v <= last returned value of the iterator, the next value in the iteration is returned.

PR #2384, opened 2016-08-17
Add perf test suite infrastructure, and a perf test for csv-import

kalman:
You can run these yourself using the -perf flag, e.g.

> noms serve &
> go test -v -perf http://localhost:8000 ./samples/go/csv/csv-import
> noms ds http://localhost:8000

Though you'll need to go-get github.com/attic-labs/testdata.

Note that all of this only records test results, it doesn't have any
concept of failing perf (unless test assertsions themselves fail). It
will be the job of some other Noms client (work in progress) to do that.

I will be setting this up to run continuously momentarily.


kalman:
Ignore ./vendor/... of course.

go/perf/suite/suite.go
```diff
@@ -0,0 +1,322 @@
...
+
+func (suite *PerfSuite) getEnvironment() types.Struct {
+	assert := suite.NewAssert()
+
+	cpuInfos, err := cpu.Info()
+	assert.NoError(err)
+	cpus := make([]types.Value, 2*len(cpuInfos))
+	for i, c := range cpuInfos {
+		var err error
+		cpus[i*2] = types.Number(i)
```
kalman: This can all get even cleaner when @arv lands marshalling for Go maps.

go/perf/suite/suite.go
```diff
@@ -0,0 +1,322 @@
...
+// Package suite implements a performance test suite for Noms, intended for measuring and reporting long running tests.
+//
+// Usage is similar to testify's suite:
+//
+// 1. Define a test suite struct which inherits from `suite.PerfSuite`.
+// 2. Define methods on that struct that start with the word "Test", optionally followed by digits, then followed a non-empty capitalized string.
+// 3. Call `suite.Run` with an instance of that struct.
+// 4. Run `go test` with the `-perf <path to noms db>` flag.
+//
+// Test results are written to Noms, along with a soup of the environment they were recorded in.
```
cmasone-attic: ...soup?

go/perf/suite/suite.go
```diff
@@ -0,0 +1,322 @@
...
+//
+// Test names are derived from that "non-empty capitalized string": `"Test"` is ommitted because it's redundant, and leading digits are omitted to allow for manual test ordering. For example:
```
cmasone-attic: Manual test ordering smells fishy to me. Tests shouldn't have order effects, right?

go/perf/suite/suite.go
kalman: Seemed like a reasonable description. Like a minestrone.

go/perf/suite/suite.go
kalman: Agree in theory, but here I wanted to be able to measure steps.

* "importing blob" is a useful test in itself, and it the whole import-blob phase was in a Pause clause, it's a waste of perfectly good stats.
* or, we could include both importing blob + parsing CSV in the same perf test, but that muddies what the test is trying to do.
* or, I could make it so that individual perf tests can declare multiple clauses (similar to suite.Pause, have suits.Stage), but this sounded like a PITA to me. I liked that there was a 1:1 correlation between the name of the test, and the measurement that it shows up with.

Besides, these are test suites... arguably, maybe, they're related enough to coordinate ordering amongst themselves.

go/perf/suite/suite.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+
+	assert := assert.New(t)
+
+	// Piggy-back off the go test -v flag.
+	verboseFlag := flag.Lookup("test.v")
+	assert.NotNil(verboseFlag)
+	verbose := verboseFlag.Value.(flag.Getter).Get().(bool)
+
+	suite := suiteT.Suite()
```
cmasone-attic: This leaves me a bit curious as to why you don't just require that people's perf suite struct directly embeds PerfSuite. Either way they wouldn't be able to access private members of PerfSuite

kalman: People's perf suite does embed a PerfSuite (and inherits the Suite method). This interface is just a subclass hack so that Run works i.e. it can't take a suite.PerfSuite because that's a struct.

Testify did something similar-ish here: https://github.com/stretchr/testify/blob/master/suite/interfaces.go

go/perf/suite/suite.go
```diff
@@ -0,0 +1,334 @@
...
+	suite.T = t
+	if verbose {
+		suite.W = os.Stdout
+	} else {
+		suite.W = nopWriter{}
+	}
+
+	gopath := os.Getenv("GOPATH")
+	assert.True(gopath != "")
```
cmasone-attic: You could use `assert.NotEmpty()` here

go/perf/suite/suite.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// This is the temporary database for tests to use.
+	server := datas.NewRemoteDatabaseServer(chunks.NewMemoryStore(), 0)
+	portChan := make(chan int)
+	server.Ready = func() { portChan <- server.Port() }
+	go server.Run()
+
+	port := <-portChan
+	suite.DatabaseSpec = fmt.Sprintf("http://localhost:%d", port)
+	suite.Database = datas.NewRemoteDatabase(suite.DatabaseSpec, "")
```
cmasone-attic: Perhaps I'll come to understand this later, but I wonder why this isn't just a LocalDatabase backed by a MemoryStore

go/perf/suite/suite.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// List of test runs, each a map of test name => timing info.
+	testRuns := make([]testRun, *perfRepeatFlag)
+
+	defer func() {
+		for _, f := range suite.tempFiles {
+			os.Remove(f.Name())
```
cmasone-attic: maybe `f.Close()` before remove? I guess it doesn't really matter since these files don't have anything in that we care about...

go/perf/suite/suite.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+
+		runs := make([]types.Value, *perfRepeatFlag)
```
cmasone-attic: There are a bunch of things for which we have a type and also a `Slice` type, like `types.Value` and `types.ValueSlice`. That's so we can make them sortable, I know. I always wonder, when I create a `[]types.Value` (or `[]types.Ref` or whatever) if I should be making a `types.ValueSlice` instead. Is there any downside to using the custom slice type?

go/perf/suite/suite.go
```diff
@@ -0,0 +1,334 @@
...
+					"paused":  types.Number(info.paused.Nanoseconds()),
+					"total":   types.Number(info.total.Nanoseconds()),
+				}))
+			}
+			runs[i] = types.NewMap(timesSlice...)
+		}
+
+		record := types.NewStruct("", map[string]types.Value{
+			"environment":     suite.getEnvironment(),
+			"nomsVersion":     types.String(suite.getGitHead(path.Join(suite.AtticLabs, "noms"))),
```
cmasone-attic: Nit: Should these be "version" or "revision"?

go/perf/suite/suite.go
```diff
@@ -0,0 +1,334 @@
...
+			"testdataVersion": types.String(suite.getGitHead(path.Join(suite.AtticLabs, "testdata"))),
+			"runs":            types.NewList(runs...),
+		})
+
+		ds := dataset.NewDataset(db, datasetID)
+		var err error
+		ds, err = ds.CommitValue(record)
+		assert.NoError(err)
+		assert.NoError(db.Close())
+		server.Stop()
```
cmasone-attic: Maybe `server.Stop()` should be deferred on its own back up where you run the server?

go/perf/suite/suite.go
```diff
@@ -0,0 +1,334 @@
...
+	}()
+
+	for runIdx := 0; runIdx < *perfRepeatFlag; runIdx++ {
+		run := testRun{}
```
cmasone-attic: `run` seems mostly unneeded here, and while this works fine because it's a Go reference type, that fact is obscured by using a type alias (or whatever Go calls them) here. I had to go remind myself that `testRun` is _just_ a Go map to be sure this works OK.

I'd suggest just using `testRuns[runIdx]` in this loop, since you only use it once anyway.

go/perf/suite/suite_test.go
```diff
@@ -0,0 +1,137 @@
...
+	assert.NotNil(err)
+
+	// The results should have been written to the "ds" dataset.
+	ds, err := spec.GetDataset(ldbDir + "::ds")
+	assert.NoError(err)
+	head := ds.HeadValue().(types.Struct)
+
+	// The general structure should be correct.
+	env := head.Get("environment").(types.Struct)
+	env.Get("diskUsages")
```
cmasone-attic:
```
getHelper := func (field string) {
  _, ok := env.Get(field)
  assert.True(ok)
}
getHelper("diskUsages")
getHelper("cpus")
```

samples/go/csv/csv-import/perf_test.go
```diff
@@ -0,0 +1,93 @@
...
+	assert := s.NewAssert()
+
+	var raw []io.Reader
+
+	s.Pause(func() {
+		// The raw data is split into a bunch of files foo.a, foo.b, etc.
+		glob, err := filepath.Glob(path.Join(s.AtticLabs, "testdata", "sf-crime", "2016-07-28.*"))
+		assert.NoError(err)
+
+		raw = make([]io.Reader, len(glob))
```
cmasone-attic: Why not just make these io.ReadClosers? then you don't need to type assert below

samples/go/csv/csv-import/perf_test.go
```diff
@@ -0,0 +1,93 @@
...
+	var importCmd *exec.Cmd
+
+	s.Pause(func() {
+		// Trick the temp file logic into creating a unique path for the csv-import binary.
+		f := s.TempFile()
+		f.Close()
+		out := f.Name()
+		os.Remove(out)
+
+		// Build csv-import into that temporary file.
```
cmasone-attic: This sort of thing makes me wonder if we should add a notion of Suite setup. So that, for example, you could just build this binary once and then use it many times to import many different datasets.

cmasone-attic: Wouldn't need to land in this PR, though

mikegray: The testify/suite does have a SetupTest and TearDownTest - not sure that perfSuite does just noting it in case anyone didn't know about the other.

go/perf/suite/suite.go
kalman: done

go/perf/suite/suite.go
kalman: as discussed -

* it can't be a straight memory store because spec doesn't know how to re-use those...
* and it could be a leveldb store, it just seems more realistic to use a HTTP store, nice to exercise the HTTP stack...
* by which argument perhaps I should be using an HTTP store backed by a leveldb store...
* but perhaps the practical difference here is whether the store ends up in-memory or on-disk...
* on my own laptop I was running out of HDD space but not memory, it'll be interesting to see how jenkins or whatever builder responds...
* come to think of it, memory does seem more reliable for timing?

kalman: I'll distill this into a comment.

kalman: writing the comment and thinking about this a bit more - I'm actually going to do remote db backed by LDB, with a flag to back it by mem (comment explains).

go/perf/suite/suite.go
kalman: can't hurt

go/perf/suite/suite.go
kalman: I think it's good to use the custom slice types, generally. I just forget.

go/perf/suite/suite.go
kalman: hm, looks like revision is what git uses. I'll change.

go/perf/suite/suite.go
kalman: I was avoiding using multiple defer, but changing it SGTM.

go/perf/suite/suite.go
kalman: I guess it's syntactically once, but actually once per method, though I doubt it makes a big difference. I'll inline like you suggest.

go/perf/suite/suite_test.go
kalman: lame

kalman: done, more or less

samples/go/csv/csv-import/perf_test.go
kalman: because go... io.MultiReader takes io.Readers, not io.ReadClosers.

samples/go/csv/csv-import/perf_test.go
kalman: I might as well implement SetupSuite now (and I might as well just use the interfaces that testify provides, in fact).

kalman:
PTAL. There are 3 changes in there.

go/perf/suite/suite.go
```diff
@@ -107,9 +114,20 @@ type PerfSuite struct {
...
+	tempDirs  []string
 	paused    time.Duration
 }
 
+// SetupRepSuite has a SetupRep method, which runs every repetition of the test, i.e. `-perf.repeat` times in total.
+type SetupRepSuite interface {
+	SetupRep()
+}
+
+// TearDownRepSuite has a TearDownRep method, which runs every repetition of hte test, i.e. `-perf.repeat` times in total.
```
cmasone-attic: s/hte/the/

go/perf/suite/suite_test.go
```diff
@@ -63,70 +67,132 @@ func (s *testSuite) Testimate() {
...
+	// These tests mostly assert that the structure of the results is correct. Specific values are hard.
+
+	getOrFail := func(s types.Struct, f string) types.Value {
+		val, ok := s.MaybeGet(f)
+		assert.True(ok)
+		return val
+	}
+
+	envVal, ok := head.MaybeGet("environment")
+	assert.True(ok)
```
cmasone-attic: You could use `getOrFail(head, "environment")` right here :-)

go/perf/suite/suite_test.go
```diff
@@ -63,70 +67,132 @@ func (s *testSuite) Testimate() {
...
+	env := envVal.(types.Struct)
+
+	getOrFail(env, "diskUsages")
+	getOrFail(env, "cpus")
+	getOrFail(env, "mem")
+	getOrFail(env, "host")
+	getOrFail(env, "partitions")
+
+	nomsRevision, ok := head.MaybeGet("nomsRevision")
+	assert.True(ok)
```
cmasone-attic: same here, no?

go/perf/suite/suite_test.go
```diff
@@ -63,70 +67,132 @@ func (s *testSuite) Testimate() {
...
+	assert.True(string(nomsRevision.(types.String)) != "")
+	getOrFail(head, "testdataRevision")
 
-	assert.NotEqual("", head.Get("nomsVersion"))
-	head.Get("testdataVersion") // don't assert it's not empty, it might not be checked out
+	runsVal, ok := head.MaybeGet("runs")
+	assert.True(ok)
```
cmasone-attic: And here

cmasone-attic:
LGTM with nitty-nits

go/perf/suite/suite.go
kalman: done

go/perf/suite/suite_test.go
kalman: done

go/perf/suite/suite_test.go
kalman: done

Issue #1956, opened 2016-07-02
Add support for nested maps to csv-importer and exporter


rafael-atticlabs:
Summary of resolution: extend the current `-dest-type:map,4` thing to support multiple keys and output a nested map

===

for @kylelundstedt , the monthly loan perf data key'd by loan-id + month.

mikegray:
Do we have some sample data with multiple keys?

aboodman:
I was looking at https://nycopendata.socrata.com/Transportation/Medallion-Vehicles-Authorized/rhe8-mgbb/data. I think that some combination of VIN, License Plate, and License Number may be unique but I never figured it out.

aboodman:
This data changes everyday, so if you can figure out what set of columns is unique that'd be cool to track on demo.noms.io.

mikegray:
That data does not have a set of columns that are unique.

aboodman:
good to know

aboodman:
I guess we'll have to track it with `-dest-type:set` then

aboodman:
Or `-dest-type:multiset` (`Map<T, number>`)

aboodman:
@mikegray: @rafael-atticlabs has a dataset that has a compound key that you can use for development. We cannot redistribute that one though.

mikegray:
where can i get a copy?

mikegray:
@rafael-atticlabs I have this working, but I am concerned with the performance ramifications. We moved to using a StreamingMap/mapMutator/opcache to improve performance for csv-import with dest-type=map, but with the work to fix this issue we have potentially several levels of maps within maps. Since the csv file isn't guaranteed to be sorted (and likely isn't), we need to be able to either ask the StreamingMap/mapMutator/opcache if a key already exists to re-use an already created nested map or we need to basically use non-streaming Maps to be able to re-use already created Maps from earlier processing. Or maybe you have another idea?  Interested in your thoughts.

mikegray:
fixed with #2433

PR #2385, opened 2016-08-18
import nested maps for csv-import like --dest-type=map:0,1

mikegray:
support for csv-import portion of #1956 

I still need to finish csv-export of this imported structure, but PTAL at this for landing csv-import.

### DIFF
--- a/samples/go/csv/read.go
+++ b/samples/go/csv/read.go
@@ -100,17 +100,7 @@ func ReadToList(r *csv.Reader, structName string, headers []string, kinds KindSl
 			panic(err)
 		}
 
-		fields := make(types.ValueSlice, len(headers))
-		for i, v := range row {
-			if i < len(headers) {
-				fieldOrigIndex := fieldOrder[i]
-				val, err := StringToValue(v, kindMap[fieldOrigIndex])
-				if err != nil {
-					d.Chk.Fail(fmt.Sprintf("Error parsing value for column '%s': %s", headers[i], err))
-				}
-				fields[fieldOrigIndex] = val
-			}
-		}
+		fields := readFieldsFromRow(row, headers, fieldOrder, kindMap)
 		valueChan <- types.NewStructWithType(t, fields)
 	}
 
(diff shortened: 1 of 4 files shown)


mikegray:
@rafael-atticlabs or @aboodman or someone else? not sure who to assign.

aboodman:
How about Dan?

samples/go/csv/read.go
```diff
@@ -131,22 +161,31 @@ func ReadToMap(r *csv.Reader, structName string, headersRaw []string, pkIdx int,
...
 			}
 		}
-		kvChan <- pk
-		kvChan <- types.NewStructWithType(t, fields)
+
+		// work backward through the pks so that outermost one is the first/primary one
+		// e.g. map<pk1, map<pk2, map<pk3, fields>>>
+		var v types.Value
+		v = types.NewStructWithType(t, fields)
+		for i := len(pkIndices); i > 1; i-- {
```
willhite: I'm not sure if this gives us the results we're expecting. Let's talk offline and see if we're on the same page.

mikegray:
@willhite PTAL again, thanks!

samples/go/csv/read.go
```diff
@@ -99,28 +100,64 @@ func ReadToList(r *csv.Reader, structName string, headers []string, kinds KindSl
...
+
 // ReadToMap takes a CSV reader and reads data into a typed Map of structs. Each row gets read into a struct named structName, described by headers. If the original data contained headers it is expected that the input reader has already read those and are pointing at the first data row.
 // If kinds is non-empty, it will be used to type the fields in the generated structs; otherwise, they will be left as string-fields.
-func ReadToMap(r *csv.Reader, structName string, headersRaw []string, pkIdx int, kinds KindSlice, vrw types.ValueReadWriter) types.Map {
+func ReadToMap(r *csv.Reader, structName string, headersRaw []string, primaryKeys []string, kinds KindSlice, vrw types.ValueReadWriter) types.Map {
 	t, fieldOrder, kindMap := MakeStructTypeFromHeaders(headersRaw, structName, kinds)
 
+	pkIndices := getPkIndices(primaryKeys, headersRaw)
+
+	topMap := types.NewMap()
```
willhite: This variable is now being used to build the entire map in memory. I think doing it this way will break csv importer for large files. I've been thinking about how to build nested maps efficiently using our current streaming approach and, so far, I haven't hit upon any great ideas. Perhaps you can do something similar to what I did in nomdex_build.go. Create the map of keys as a go map of maps (presumably the keys are small enough that this can fit into memory though it's not a guarantee). At the lowest level, instead of adding rows to the map, create a streamingSet for each keys and insert rows into them. When you've finished reading all the rows, create streamingMaps from the bottom up and insert all entries in them.

samples/go/csv/read.go
```diff
@@ -131,24 +168,40 @@ func ReadToMap(r *csv.Reader, structName string, headersRaw []string, pkIdx int,
...
+			return m.Set(key, encloseInMapFunc(subMap, keyLevel+1))
 		}
-		kvChan <- pk
-		kvChan <- types.NewStructWithType(t, fields)
+
+		encloseInMapFunc = encloseInMap
+		topMap = encloseInMap(topMap, 0)
 	}
 
+	topMap.IterAll(func(k types.Value, v types.Value) {
```
willhite: This is just taking things out of the map you've already build in memory, streaming them to the db, and then building another map. It doesn't seem useful.

willhite:
One other suggestion might be to use the old strategy when not building nested maps so that we don't make this slower for people who are currently depending on it.

BTY

samples/go/csv/read.go
mikegray: Yeah, I have continued to think about this problem on and off all weekend too.  I wrote a note in the original bug/issue and am hoping to hear from Raf if he has better ideas. I thought about doing exactly what you suggest, but am wondering if memory will be a problem. I guess the only way to know is to do it. Another idea would be similar to the opcache notion - if we see memory pressure then start backing the temporary work into levelDB. I will continue thinking, hopefully we will hear input from Raf this week too. Thanks for the review!

samples/go/csv/read.go
mikegray: Yeah, I decided not to change the API but I recognized this was redundant work.  I figured this is a temporary state of not using the streaming map in intermediate state, so didn't want to change the API of csv ReadToMap.  I could easily just send the in memory Map to the chan instead. Good point.

mikegray:
Yeah, doing it the old way for when we only have 1 key seems like a good intermediate solution to not cause problems for current users while we get more/better infrastructure in place to do nested-maps quicker. Thanks again for the review.

mikegray:
please test this

mikegray:
closing in favor of #2433 - it is this PR plus the export functionality

PR #2388, opened 2016-08-18
Go: Implement best-effort three-way merge for Maps

cmasone-attic:
Add the Go 'merge' package, which exports one function called ThreeWay().
ThreeWay() attempts a three-way merge between two candidates and a common
ancestor. It considers the three of them recursively, applying some simple rules:

- If any of the three nodes are different types: conflict
- If we are dealing with a map:
  - If the same key is both removed and inserted wrt parent: conflict
  - If the same key is inserted wrt parent, but with different values: conflict
- If we are dealing with a struct:
  - same as map
- If we are dealing with a list:
  - Same as map but substitute "index" for "key"
- If we are dealing with a set:
  - If the same object is both removed and inserted wrt parent: conflict
... otherwise, the concurrent modification is allowed.

Currently, ThreeWay() only works on types.Map.

Towards #148

### DIFF
--- a/go/merge/three_way.go
+++ b/go/merge/three_way.go
@@ -1,0 +1,259 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package merge
+
+import (
+	"fmt"
+
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// ErrMergeConflict indicates that a merge attempt failed and must be resolved manually for the provided reason.
+type ErrMergeConflict struct {
+	msg string
+}
+
+func (e *ErrMergeConflict) Error() string {
+	return e.msg
+}
+
+func newMergeConflict(format string, args ...interface{}) *ErrMergeConflict {
+	return &ErrMergeConflict{fmt.Sprintf(format, args...)}
+}
+
+// ThreeWay attempts a three-way merge between two candidates and a common ancestor. It considers the three of them recursively, applying some simple rules to identify conflicts:
+//  - If any of the three nodes are different NomsKinds: conflict
+//  - If we are dealing with a map:
+//    - If the same key is both removed and inserted wrt parent: conflict
+//    - If the same key is inserted wrt parent, but with different values: conflict
+//  - If we are dealing with a struct:
+//    - If the same field is both removed and inserted wrt parent: conflict
+//    - If the same field is inserted wrt parent, but with different values: conflict
+//  - If we are dealing with a list:
+//    - If the same index is both removed and inserted wrt parent: conflict
+//    - If the same index is inserted wrt parent, but with different values: conflict
+//  - If we are dealing with a set:
+//    - If the same object is both removed and inserted wrt parent: conflict
+//
+// All other modifications are allowed.
+// Currently, ThreeWay() only works on types.Map.
+func ThreeWay(a, b, parent types.Value, vwr types.ValueReadWriter) (merged types.Value, err error) {
+	if a == nil && b == nil {
+		return parent, nil
+	} else if a == nil {
+		return parent, newMergeConflict("Cannot merge nil Value with %s.", b.Type().Describe())
+	} else if b == nil {
+		return parent, newMergeConflict("Cannot merge %s with nil value.", a.Type().Describe())
+	} else if unmergeable(a, b) {
+		return parent, newMergeConflict("Cannot merge %s with %s.", a.Type().Describe(), b.Type().Describe())
+	}
+
+	return threeWayMerge(a, b, parent, vwr)
+}
+
+// a and b cannot be merged if they are of different NomsKind, or if at least one of the two is nil, or if either is a Noms primitive.
+func unmergeable(a, b types.Value) bool {
+	if a != nil && b != nil {
+		aKind, bKind := a.Type().Kind(), b.Type().Kind()
+		return aKind != bKind || types.IsPrimitiveKind(aKind) || types.IsPrimitiveKind(bKind)
+	}
+	return true
+}
+
+func threeWayMerge(a, b, parent types.Value, vwr types.ValueReadWriter) (merged types.Value, err error) {
+	d.PanicIfTrue(a == nil || b == nil, "Merge candidates cannont be nil: a = %v, b = %v", a, b)
+	newTypeConflict := func() *ErrMergeConflict {
+		pDescription := "<nil>"
+		if parent != nil {
+			pDescription = parent.Type().Describe()
+		}
+		return newMergeConflict("Cannot merge %s and %s on top of %s.", a.Type().Describe(), b.Type().Describe(), pDescription)
+	}
+
+	switch a.Type().Kind() {
+	case types.ListKind:
+		// TODO: Come up with a plan for List (BUG 148)
+		return parent, newMergeConflict("Cannot merge %s.", a.Type().Describe())
+
+	case types.MapKind:
+		if aMap, bMap, pMap, ok := mapAssert(a, b, parent); ok {
+			return threeWayMapMerge(aMap, bMap, pMap, vwr)
+		}
+
+	case types.RefKind:
+		if aValue, bValue, pValue, ok := refAssert(a, b, parent, vwr); ok {
+			merged, err := threeWayMerge(aValue, bValue, pValue, vwr)
+			if err != nil {
+				return parent, err
+			}
+			return vwr.WriteValue(merged), nil
+		}
+
+	case types.SetKind:
+		// TODO: Implement plan from BUG148
+		return parent, newMergeConflict("Cannot merge %s.", a.Type().Describe())
+
+	case types.StructKind:
+		if aStruct, bStruct, pStruct, ok := structAssert(a, b, parent); ok {
+			return threeWayStructMerge(aStruct, bStruct, pStruct, vwr)
+		}
+
+	default:
+		return parent, newMergeConflict("Cannot merge %s.", a.Type().Describe())
+
+	}
+	return parent, newTypeConflict()
+}
+
+func threeWayMapMerge(a, b, parent types.Map, vwr types.ValueReadWriter) (merged types.Map, err error) {
+	aChangeChan, bChangeChan := make(chan types.ValueChanged), make(chan types.ValueChanged)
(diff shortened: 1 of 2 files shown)


cmasone-attic:
@kalman PTAL
@aboodman FYI

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+func (e *ErrMergeConflict) Error() string {
+	return e.msg
+}
+
+func newMergeConflict(format string, args ...interface{}) *ErrMergeConflict {
+	return &ErrMergeConflict{fmt.Sprintf(format, args...)}
+}
+
+// ThreeWay() attempts a three-way merge between two candidates and a common ancestor. It considers the three of them recursively, applying some simple rules to identify conflicts:
+//  - If any of the three nodes are different types: conflict
```
kalman: Should this be "different kinds" not "different types"?

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+	newTypeConflict := func() *ErrMergeConflict {
+		pDescription := "<nil>"
+		if parent != nil {
+			pDescription = parent.Type().Describe()
+		}
+		return newMergeConflict("Cannot merge %s and %s on top of %s.", a.Type().Describe(), b.Type().Describe(), pDescription)
+	}
+
+	switch a.Type().Kind() {
+	default:
```
kalman: default false, what is this

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+			return vwr.WriteValue(merged), nil
+		}
+
+	case types.SetKind:
+		// TODO: Implement plan from BUG148
+		return parent, newMergeConflict("Cannot merge %s.", a.Type().Describe())
+
+	case types.StructKind:
+		if aStruct, bStruct, pStruct, ok := structAssert(a, b, parent); ok {
+			return threeWayStructMerge(aStruct, bStruct, pStruct, vwr)
```
kalman: It's odd that list/set not-implemented are handled in this switch, but struct not-implemented is handled in a function. But I suppose it's temporary either way.

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+		a.DiffLeftRight(parent, aChangeChan, aStopChan)
+		close(aChangeChan)
+	}()
+	go func() {
+		b.DiffLeftRight(parent, bChangeChan, bStopChan)
+		close(bChangeChan)
+	}()
+	stopAndDrain := func(stop chan<- struct{}, drain <-chan types.ValueChanged) {
+		close(stop)
+		for _ = range drain {
```
kalman: just `for range drain`

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+		aChange, bChange = types.ValueChanged{}, types.ValueChanged{}
+	}
+	return merged, nil
+}
+
+func apply(target types.Map, change types.ValueChanged, newVal types.Value) types.Map {
+	switch change.ChangeType {
+	case types.DiffChangeAdded:
+		fallthrough
+	case types.DiffChangeModified:
```
kalman: Can't you do like `case types.DiffChangeAdded, types.DiffChangeModifield:`?

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+		return target.Set(change.V, newVal)
+	case types.DiffChangeRemoved:
+		return target.Remove(change.V)
+	default:
+		panic("Not Reached")
+	}
+}
+
+func changeString(change types.ValueChanged) string {
```
kalman: nit: `describeChange`?

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+		if aValue, bValue, pValue, ok := refAssert(a, b, parent, vwr); ok {
+			merged, err := threeWayMerge(aValue, bValue, pValue, vwr)
+			if err != nil {
+				return parent, err
+			}
+			return vwr.WriteValue(merged), nil
+		}
+
+	case types.SetKind:
+		// TODO: Implement plan from BUG148
```
kalman: I expect this will involve (1) renaming map merge to ordered-sequence merge, then (2) calling it from here and map, with some small modifications.

Likewise structs, actually.

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+		if aChange.ChangeType != bChange.ChangeType {
+			return parent, newMergeConflict("Conflict:\n%s\nvs\n%s\n", changeString(aChange), changeString(bChange))
+		}
+
+		aValue, bValue := a.Get(aChange.V), b.Get(bChange.V)
+		if aChange.ChangeType == types.DiffChangeRemoved || aValue.Equals(bValue) {
+			// If both diffs generated a remove, or if the new value is the same in both, merge is fine.
+			merged = apply(merged, aChange, aValue)
+		} else {
+			// There's one case that might still be OK even if aValue and bValue differ: different, but mergeable, compund values of the same type being added/modified at the same key, e.g. a Map being added to both a and b. If either is a primitive, or Values of different Kinds were added, though, we're in conflict.
```
kalman: compund -> compound

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+	var aOk, bOk, pOk bool
+	aMap, aOk = a.(types.Map)
+	bMap, bOk = b.(types.Map)
+	if parent != nil {
+		pMap, pOk = parent.(types.Map)
+	} else {
+		pMap, pOk = types.NewMap(), true
+	}
+	ok = aOk && bOk && pOk
+	return
```
kalman: I guess the only thing with this is that `ok` may end up false, but some of `aMap`/`bMap`/`pMap` might be non-nil, which is arguably a different contract to what I'd expect in Go (i.e. I would expect the return value to have none of the maps be non-nil). Hardly a big deal though.

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func refAssert(a, b, parent types.Value, vwr types.ValueReadWriter) (aValue, bValue, pValue types.Value, ok bool) {
+	var aOk, bOk, pOk bool
+	var aRef, bRef, pRef types.Ref
+	if aRef, aOk = a.(types.Ref); aOk {
+		aValue = aRef.TargetValue(vwr)
```
kalman: although in this case -

do you want to save the effort of reading (from the database) aRef if bRef isn't actually a ref anyway, and vice versa? it seems like it's lost on the callers.

kalman:
Sorry will finish tomorrow

go/merge/three_way.go
cmasone-attic: Done

go/merge/three_way.go
cmasone-attic: I like to do this if `default` is the common case, but that's not really the case here, so I've changed it.

go/merge/three_way.go
cmasone-attic: Yeah. I was considering making Struct work in this patch as well, then stuff started to get too big. I'd like to keep the struct code in, though, since the next patch I do would just reintroduce it anyway :-)

go/merge/three_way.go
cmasone-attic: Done

go/merge/three_way.go
cmasone-attic: Duh

go/merge/three_way.go
cmasone-attic: Done

go/merge/three_way.go
cmasone-attic: Indeed, though I think I need to add a bit of abstraction in a few places, kinda like was done over in noms diff

go/merge/three_way.go
cmasone-attic: Done

go/merge/three_way.go
cmasone-attic: That's fair, but I'd argue that adding the additional declarations makes stuff more verbose without adding clarity

go/merge/three_way.go
cmasone-attic: This is a good point, though

go/merge/three_way.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func apply(target types.Map, change types.ValueChanged, newVal types.Value) types.Map {
+	switch change.ChangeType {
+	case types.DiffChangeAdded:
+		fallthrough
+	case types.DiffChangeModified:
+		return target.Set(change.V, newVal)
+	case types.DiffChangeRemoved:
+		return target.Remove(change.V)
```
kalman: We can figure this out later - but every call to Set and Remove will re-chunk the map, and it'll get slow.

go/merge/three_way_test.go
```diff
@@ -0,0 +1,178 @@
...
+	merged, err := ThreeWay(aa1a, aa1b, aa1, nil)
+	if assert.NoError(err) {
+		assert.Equal(aaMerged, merged)
+	}
+}
+
+func TestThreeWayMergeMap_RecursiveCreate(t *testing.T) {
+	assert := assert.New(t)
+
+	merged, err := ThreeWay(mm1a, mm1b, mm1, nil)
```
kalman: I think it would be valuable for these tests to assert both `ThreeWay(a, b, parent)` and `ThreeWay(b, a, parent)` ... they should give the same results, right?

Seems like you could fairly easily have a function like `testThreeWay` and similar for errors which does both.

go/merge/three_way_test.go
```diff
@@ -0,0 +1,178 @@
...
+	if assert.NoError(err) {
+		assert.Equal(mm1Merged, merged)
```
kalman: I'm surprised that `assert.Equal` works here... these are Noms map, shouldn't you need `assert.True(mm1Merged.Equals(merged))`?

kalman:
lgtm assuming those comments were actually done (you didn't upload a new patch), and my last couple of comments are easy enough to resolve.

go/merge/three_way.go
cmasone-attic: I have hopes that I can use a Mutator once we make that capable of mutating existing maps and sets

go/merge/three_way_test.go
cmasone-attic: Yeah, I guess it just happens that I wind up with an equal Go structure after the merging, but it'd be better to use the Noms Equals

PR #2389, opened 2016-08-18
Go marshal maps

arv:
Go maps are marshaled to Noms Map.

 Towards #2376

### DIFF
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -21,9 +21,11 @@ import (
 // fields to the fields used by Marshal (either the struct field name or its tag).
 // Unmarshal will only set exported fields of the struct.
 //
-// To unmarshal a Noms list into a slice, Unmarshal resets the slice length to zero and then appends each element to the slice.
+// To unmarshal a Noms list into a slice, Unmarshal resets the slice length to zero and then appends each element to the slice. If the Go slice was nil a new slice is created.
 //
 // To unmarshal a Noms list into a Go array, Unmarshal decodes Noms list elements into corresponding Go array elements.
+//
+// To unmarshal a Noms map into a Go map, Unmarshal decodes Noms key and values into corresponding Go array elements.
 //
 // Unmarshal returns an UnmarshalTypeMismatchError if:
 // - a Noms value is not appropriate for a given target type
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -103,6 +103,8 @@ func typeDecoder(t reflect.Type) decoderFunc {
 		return sliceDecoder(t)
 	case reflect.Array:
 		return arrayDecoder(t)
+	case reflect.Map:
+		return mapDecoder(t)
 	default:
 		panic(&UnsupportedTypeError{Type: t})
 	}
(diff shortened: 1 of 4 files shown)


go/marshal/decode.go
```diff
@@ -248,8 +250,17 @@ func sliceDecoder(t reflect.Type) decoderFunc {
 	var decoder decoderFunc
 
 	d = func(v types.Value, rv reflect.Value) {
-		slice := rv.Slice(0, 0)
-		v.(types.List).IterAll(func(v types.Value, i uint64) {
+		list, ok := v.(types.List)
+		if !ok {
+			panic(&UnmarshalTypeMismatchError{v, t, ""})
```
arv: Handle this case instead of having the panic escape

go/marshal/decode.go
```diff
@@ -248,8 +250,17 @@ func sliceDecoder(t reflect.Type) decoderFunc {
...
+		}
+		var slice reflect.Value
+		if rv.IsNil() {
+			slice = reflect.MakeSlice(t, 0, int(list.Len()))
```
arv: For consistency. Now we do not return nil slices nor nil maps

arv:
@kalman PTAL

go/marshal/decode.go
kalman: At some point we should probably have sanity checks that we're not trying to unmarshall huge lists. Obviously if Len doesn't fit in an int that will be a problem.

go/marshal/decode_test.go
```diff
@@ -500,8 +507,51 @@ func TestDecodeRecursive(t *testing.T) {
...
+	err := Unmarshal(types.NewMap(
+		types.String("a"), types.Number(1),
+		types.String("b"), types.Number(2),
+		types.String("c"), types.Number(3)), &m)
+	assert.NoError(err)
+	assert.Equal(map[string]int{"a": 1, "b": 2, "c": 3}, m)
+
+	m = nil
+	err = Unmarshal(types.NewMap(
+		types.String("a"), types.Number(1),
```
kalman: Is this repeated test for the caching logic or something? Seems like you should be resuing a types.Map and map[string]int instead of repeating.

go/marshal/decode_test.go
```diff
@@ -500,8 +507,51 @@ func TestDecodeRecursive(t *testing.T) {
...
+		types.String("b"), types.Number(2),
+		types.String("c"), types.Number(3)), &m)
+	assert.NoError(err)
+	assert.Equal(map[string]int{"a": 1, "b": 2, "c": 3}, m)
+
+	m = map[string]int{"b": 2}
+	err = Unmarshal(types.NewMap(
+		types.String("a"), types.Number(1),
+		types.String("c"), types.Number(3)), &m)
```
kalman: somewhere also test if there is an existing key, like have `m` start with `d`: 4 and unmarshal `d` to 5 or something.

go/marshal/encode.go
```diff
@@ -332,3 +336,27 @@ func listEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc {
...
+	}
+
+	var keyEncoder encoderFunc
+	var valueEncoder encoderFunc
+	e = func(v reflect.Value) types.Value {
+		keys := v.MapKeys()
+		kvs := make([]types.Value, 2*len(keys))
+		for i, k := range keys {
+			kvs[2*i] = keyEncoder(k)
+			kvs[2*i+1] = valueEncoder(v.MapIndex(k))
```
kalman: I found myself writing this pattern like:
```
kvs := make([]types.Value, 0, 2*len(keys))
for _, k := range keys {
  kvs = append(kvs, keyEncoder(k), valueEncoder(v.MapIndex(k)))
}
return types.NewMap(kvs...)
```
either is fine.

kalman:
lgtm with comments

go/marshal/decode_test.go
arv: The test ensures that `m = nil` works. I'll change it to reuse the noms map and the go map though.

go/marshal/decode_test.go
arv: Good idea.

go/marshal/encode.go
arv: I find myself not doing that because append allocates a new object every time (the backing store is kept)

go/marshal/decode_test.go
kalman: doesn't `var m map[string]int` automatically initialise to `nil` anyway/

arv: You are right. I'm not sure what I was thinking. I'll clean it up.

PR #2391, opened 2016-08-19
Marshal and Unmarshal handling of interface{}

arv:
When unmarshalling onto `interface{}` the following rules are used:
 - `types.Bool` -> `bool`
 - `types.List` -> `[]interface{}`
 - `types.Map` -> `map[interface{}]interface{}`
 - `types.Number` -> `float64`
 - `types.String` -> `string`

Towards #2376

### DIFF
--- a/go/marshal/decode_test.go
+++ b/go/marshal/decode_test.go
@@ -549,3 +549,42 @@ func TestDecodeMapWrongNomsType(t *testing.T) {
 	var m map[string]int
 	assertDecodeErrorMessage(t, types.NewList(types.String("a"), types.Number(1)), &m, "Cannot unmarshal List<Number | String> into Go value of type map[string]int")
 }
+
+func TestDecodeOntoInterface(t *testing.T) {
+	assert := assert.New(t)
+
+	var i interface{}
+	err := Unmarshal(types.Number(1), &i)
+	assert.NoError(err)
+	assert.Equal(float64(1), i)
+
+	err = Unmarshal(types.String("abc"), &i)
+	assert.NoError(err)
+	assert.Equal("abc", i)
+
+	err = Unmarshal(types.Bool(true), &i)
+	assert.NoError(err)
+	assert.Equal(true, i)
+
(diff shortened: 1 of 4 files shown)


go/marshal/decode_test.go
```diff
@@ -549,3 +549,42 @@ func TestDecodeMapWrongNomsType(t *testing.T) {
...
+	assert.NoError(err)
+	assert.Equal("abc", i)
+
+	err = Unmarshal(types.Bool(true), &i)
+	assert.NoError(err)
+	assert.Equal(true, i)
+
+	err = Unmarshal(types.NewList(types.String("abc")), &i)
+	assert.NoError(err)
+	assert.Equal([]interface{}{"abc"}, i)
```
arv: We could generate `[]string` in this case. Is it worth it? The static type will always be `interface{}` or `[]interface` but I guess it would lead to less type casts.

arv:
@ahl PTAL

ahl:
LGTM; I'd love to see it on the marshal side as well (Go -> Noms)

arv:
That should ashtray work. The marshalling already uses the dynamic type.

arv:
retest this please

arv:
retest this please

ahl:
@arv responding to this old thread:

I'm working with Firebase which hands me `[]interface{}` but the contents are actually all, in my case, `float64`. Noms doesn't marshal that because it doesn't know that the array doesn't mix and match types, but it seems like it could bang through the array and decide to proceed if the contents is uniform.

You can find an example of the unpacking I needed to do here: https://github.com/ahl/hackernoms/blob/master/sync/sync.go#L467

arv:
@ahl That works for me:

```go
func TestEncodeSliceInterface(t *testing.T) {
	assert := assert.New(t)

	var i interface{}
	i = []interface{}{float64(1), float64(2)}
	v, err := Marshal(i)
	assert.NoError(err)
	assert.True(types.NewList(types.Number(1), types.Number(2)).Equals(v))
}
```

Am I missing something?

PR #2392, opened 2016-08-19
Use new Go-to-Noms map marshalling in perf suite

kalman:


kalman:
Much nicer, thanks @arv 

No functional changes here.

arv:
LGTM

go/perf/suite/suite.go
```diff
@@ -132,6 +132,14 @@ type perfSuiteT interface {
 	Suite() *PerfSuite
 }
 
+type environment struct {
```
aboodman: Will the Noms type name come out as `Environment` or `environment`?

arv: environment... Should we upper case that?

aboodman: I guess that would be consistent.

kalman:
Tension between not wanting to have this be Go public, and having consistent struct names.

If we're going to be modifying struct fields anyway (to start with lowercase) maybe the marshalling code should make struct names start with an uppercase letter?

arv:
Yes. I plan to make that change.

PR #2393, opened 2016-08-19
Add noms perf builder

kalman:


kalman:
Suuuper basic obvs.

Hm, actually, never mind. This isn't great.

kalman:
... because it writes results to a different database each time, and I forgot you can't noms-log between them

kalman:
This is as far as I got on Friday. It uploads results to demo.noms.io/perf/... where ... is the name of the branch. Typically this will be origin/master, and we can track history there, but for PRs it will be like http://demo.noms.io/perf/origin/pr/2393/merge.

You can see the results now:
```
> noms ds http://demo.noms.io/perf/origin/pr/2393/merge
```

I'm going to turn it off for the weekend. There are 2 main problems.

1. it assumes that the testdata directory is checked out. We'll need to set up a shared workspace, and a testdata builder that pulls it in.
2. the testdata directory, even a shallow clone, is 12G and jenkins only has a 20G hard drive. It's filling up.

kalman:
Ok, let's do this.

This script is actually already up and running on the perf bots.

tools/run_perf_builder.py
```diff
@@ -0,0 +1,84 @@
...
+    './samples/go/csv/csv-import',
+]
+
+# 'go test' timeout. Go's default is 10m, which ideally is long enough, but raise it a little.
+TIMEOUT = '15m'
+
+# Number of perf test repetitions. Choose 4 because:
+# - Any less than 3 is not enough sample size.
+# - Any more than 4 will take too long.
+# - Typically the first test will take longer, so choosing 4 gives you 3 non-first samples.
```
cmasone-attic: Does the actual test take longer, or is that because of the setup (compiling csv-import and other things)

cmasone-attic:
LGTM with one Q

tools/run_perf_builder.py
kalman: The compilation of the binaries actually shouldn't count towards the "elapsed" metric, only the total.

But yes, other things. From the point of view of somebody that doesn't know test specifics, the first test might be slower because the OS hasn't cached files (e.g. the CSV off disk). Or the startup time for those binaries would be slower.

However there is actually a test that loads a CSV off disk, into a blob, then writes that blob to the database. Subsequent runs are much faster because the blob is already in the database, a short circuit must be tripping somewhere.

kalman: For the record, and because I can do this, here is what I mean by elapsed vs paused vs total:
http://splore.noms.io/?db=http://demo.noms.io/perf&p=master/csv-import.value.reps[0][%22ImportSfCrimeBlobFromTestdata%22]

Issue #2298, opened 2016-08-06
Need API usage overview doc


rafael-atticlabs:

mikegray:
Can you provide some additional info for me? Do you have an example of a different tool/site that does this well?  What kinds of items do you think should be covered?  Obviously we will have links to the Go and JS docs, but I am guessing this is meant to be a little higher level - e.g. not the methods on Noms maps, but Noms maps exist, etc.? What other types of overview are you thinking about?

aboodman:
I'm just going to close this. I think we have basics in place, and we can fix on an ongoing basis from here on out.

rafael-atticlabs:
As mentioned in the @mikegray patch, I think the current doc does a good job of describing a high-level what noms is and abstractly it's key abstractions, but we got clear feedback from @ahl early on that just basics of how to think about using the API weren't clear.

I've demoted this to P1 and assigned to myself. I'll take a crack at it when I am able to make time.

rafael-atticlabs:
also, IIRC, it wasn't clear to @vinibaggio when he started as well.

Issue #2297, opened 2016-08-06
Make a first pass over the content of go & js doc


rafael-atticlabs:
There isn't hardly any commentary in js. In go, there is some, but we're not doing correctly so it shows up in the godocs.

PR #2394, opened 2016-08-20
adding godoc synopsis for several top level packages

mikegray:
towards #2297 and #2298

### DIFF
--- a/go/chunks/chunk.go
+++ b/go/chunks/chunk.go
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, version 2.0:
 // http://www.apache.org/licenses/LICENSE-2.0
 
-// Package chunks contains the core storage abstractions and implementations used in Noms.
+// Package chunks provides facilities for representing, storing, and fetching content-addressed chunks of Noms data.
 package chunks
 
 import (
--- a/go/chunks/chunk.go
+++ b/go/chunks/chunk.go
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, version 2.0:
 // http://www.apache.org/licenses/LICENSE-2.0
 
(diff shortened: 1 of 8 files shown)


go/chunks/chunk.go
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, version 2.0:
 // http://www.apache.org/licenses/LICENSE-2.0
 
+// Package chunks contains the core storage abstractions and implementations used in Noms.
```
aboodman: Trying to be a little more specific:

Package chunks provides facilities for representing, storing, and fetching content-addressed chunks of Noms data.

go/d/try.go
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, version 2.0:
 // http://www.apache.org/licenses/LICENSE-2.0
 
+// Package d implements several error and assertion functions used throughout Noms.
```
aboodman: Mention "debug", which is what "d" stands for.

mikegray: why don't we just name it `debug` then?

go/chunks/chunk.go
mikegray: sounds good. done.

go/d/try.go
cmasone-attic: `d` is shorter, essentially :-) Also, those of us from the Chromium world were used to using a macro called DCHECK() for stuff similar to this.

cmasone-attic:
This seems fine to me, now, but Aaron had opinions, so I'd rather that he had the final sign off

aboodman:
lgtm, thanks forthis

Issue #2352, opened 2016-08-11
Move noms/samples/js/perf to noms/js/perf


mikegray:
To be more consistent with noms/go/perf, and to move non-samples out of samples.

Issue #2350, opened 2016-08-11
Jenkins configuration and scripts in version control


mikegray:
The Jenkins server configuration and build scripts should be put into version control or at least backed up.

PR #2397, opened 2016-08-22
Move noms/samples/js/perf to noms/js/perf

sunglim:
Fixes: https://github.com/attic-labs/noms/issues/2352

@mikegray, Could you take a look?


mikegray:
@arv is probably a better reviewer for this, but @aboodman might want to have a look too.

arv:
LGTM

Just to be sure; the js/perf/package.json is there because we want to pick up modules from parent directory, just like we did in samples?

ok to test

arv:
This is causing build failures. I think the problem is that this package needs to call npm install on the parent first.

mikegray:
issue #2350 is to get these CI scripts into this repo - we aren't there yet. It doesn't help you in this situation of course, but wanted to reference that issue here.  Take a look at noms/tools/run_pr_builder.sh - it might be the script you are seeking.

arv:
@sunglim This is all my fault. I should have thought of this.

The reason why this works locally and not on jenkins is that jenkins uses a clean check out every time.

I'm going to take over this bug/pr. My plan is to make the perf tests not its own node module.

PR #2399, opened 2016-08-22
Try to fix the build after move of js-perf

arv:


arv:
@mikegray PTAL
@sunglim FYI

mikegray:
lgtm, but it seems like maybe the PR build didn't perform an actual build?  

please build this, please test this, etc. mr. robot

arv:
This doesn't work either. Having something that depends on `@attic/noms` located inside `/js/` is not going to work. Closing this and reverting the other one.

PR #2400, opened 2016-08-22
Revert "Move noms/samples/js/perf to noms/js/perf"

arv:
Reverts attic-labs/noms#2397

Having the perf suite inside `js` and depending in `@attic/noms` is not going to work.


arv:
@mikegray TBR
@sunglim FYI

mikegray:
This is why I wanted you to look at the initial PR. I had worked on a prior version of moving these but had them outside the attic/noms/js directory because they need to depend on the attic/noms npm. Aaron wanted them inside, but I had them at attic/noms/jsperf - so a peer of the JS SDK. Not sure of the right answer, but having them in samples is not great.

mikegray:
revert lgtm

arv:
One option would be to have the perf directory in there but have no package.json in it. We would need to add some more deps (to devDependencies) but that seems like a good compromise.

PR #2408, opened 2016-08-23
Go Marshal: Title case struct names

arv:
Make the name of the Noms struct start with a capital letter.

When unmarshalling a struct we now ignore the case.

Towards #2376


arv:
@kalman PTAL

kalman:
lgtm

PR #2416, opened 2016-08-23
Move perf tests to js/perf/

arv:
This restructures how the perf tests import noms (by doing `require('../../../')` instead of `require('@attic/noms')` which solves the dependency issue.


arv:
@mikegray @sunglim PTAL

This makes me realize that we should probably just change the whole js structure around to:

```
js/
  noms/  (old js dir)
  perf/
  eslintrc/
  webpack-config/
```

mikegray:
do samples/js then move into js/samples too?

js/package.json
```diff
@@ -36,7 +37,12 @@
     "commander": "^2.9.0",
     "flow-bin": "0.30.0",
     "fs-extra": "^0.30.0",
-    "mocha": "^2.5.3"
+    "humanize": "^0.0.9",
+    "mocha": "^2.5.3",
+    "mz": "^2.4.0",
+    "varint": "^4.0.0",
+    "yargs": "^5.0.0"
+
```
mikegray: remove extra blank line?

js/perf/codec-perf-rig/src/main.js
```diff
@@ -43,9 +43,9 @@ const args = argv
   .argv;
 
 main().catch(ex => {
-  console.error('\nError:', ex);
+  console.error('\nError:', ex);  // eslint-disable-line no-console
```
mikegray: do we still need these with the .eslintrc.js change at root of js/perf/ directory?

mikegray:
lgtm with a few minor nits/questions

js/perf/codec-perf-rig/src/main.js
arv: No. At one point I had the `.eslintrc.js` file removed.

arv:
Lame... that worked locally. I'll just do the big move instead.

PR #2417, opened 2016-08-23
Support a -perf.prefix flag for perf tests

kalman:
This prepends a prefix to every dataset ID written by perf tests.
It will be used for namespacing the test results for different PRs.
Currently they're written to different databases, which isn't as good.

### DIFF
--- a/go/perf/suite/suite.go
+++ b/go/perf/suite/suite.go
@@ -261,10 +261,7 @@ func Run(datasetID string, t *testing.T, suiteT perfSuiteT) {
 			"reps":             types.NewList(reps...),
 		})
 
-		if *perfPrefixFlag != "" {
-			datasetID = *perfPrefixFlag + datasetID
-		}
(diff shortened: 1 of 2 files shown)


kalman:
... prepends.

go/perf/suite/suite.go
```diff
@@ -254,6 +261,9 @@ func Run(datasetID string, t *testing.T, suiteT perfSuiteT) {
 			"reps":             types.NewList(reps...),
 		})
 
+		if *perfPrefixFlag != "" {
```
cmasone-attic: There's not really a need for the conditional, is there? "" + "foo == "foo"

cmasone-attic:
LGTM w/nit

go/perf/suite/suite.go
kalman: oh, good point

PR #2426, opened 2016-08-24
Move samples/js/perf to js/perf

arv:


arv:
@mikegray TBR
@sunglim FYI

mikegray:
lgtm, thanks for cleaning all this up!

PR #2432, opened 2016-08-24
Add informative message at base path of demo-server.

willhite:


samples/go/demo-server/web_server.go
```diff
@@ -152,3 +155,10 @@ func notFound(w http.ResponseWriter, r *http.Request) {
 	newParams = append(newParams, params...)
 	hndl(w, r, newParams)
 }
+
+func handleBaseGet(w http.ResponseWriter, req *http.Request, params httprouter.Params) {
+	d.PanicIfTrue(req.Method != "GET", "Expected get method.")
+
+	fmt.Fprintf(w, nomsBaseHtml)
```
mikegray: should these be reversed? set/send header first then send html to responseWriter? I am guessing maybe responseWriter buffers the Fprintf so as to be able to send appropriate headers before sending the output.

samples/go/demo-server/web_server.go
```diff
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	dbParam = "dbName"
+	dbParam      = "dbName"
+	nomsBaseHtml = "<html><head></head><body><p>Hi. This is a Noms HTTP server.</p><p>To learn more, visit <a href=\"https://github.com/attic-labs/noms\">our github project</a>.</p></body></html>"
```
mikegray: Capitalize Github.

mikegray:
lgtm with nits

samples/go/demo-server/web_server.go
willhite: It does buffer them, but I think you're right that, in general, it's better in the other direction. I'll change it here and in the original code submitted by @jangie.

jangie: Ack! Sorry about that.

willhite: No worries @jangie, thanks for following up on this and doing the original fix.

PR #2433, opened 2016-08-24
csv-import/csv-export support for compound keys/nested maps

mikegray:
This is forked from the other PR that has importing to nested-maps.  I needed to have nested-maps in csv-import and/or read.go to properly test csv-export and/or write.go. I can merge the 2 PRs if that makes it easier to land.  This PR actually is the superset, so maybe I should just kill/close the other one?  Let me know how best to enable review/landing.  I am fine with waiting on this one until after the other one lands too.  Whatever works best for you.

### DIFF
--- a/samples/go/csv/csv-import/importer_test.go
+++ b/samples/go/csv/csv-import/importer_test.go
@@ -40,9 +40,9 @@ type testSuite struct {
 func (s *testSuite) SetupTest() {
 	input, err := ioutil.TempFile(s.TempDir, "")
 	d.Chk.NoError(err)
+	defer input.Close()
 	s.tmpFileName = input.Name()
 	writeCSV(input)
-	defer input.Close()
 }
 
 func (s *testSuite) TearDownTest() {
--- a/samples/go/csv/csv-import/importer_test.go
+++ b/samples/go/csv/csv-import/importer_test.go
@@ -22,46 +22,93 @@ import (
 	"github.com/attic-labs/testify/suite"
 )
 
+const (
+	TEST_DATA_SIZE = 100
+	TEST_YEAR      = 2012
+	TEST_FIELDS    = "Number,String,Number,Number"
+)
+
 func TestCSVImporter(t *testing.T) {
 	suite.Run(t, &testSuite{})
 }
 
 type testSuite struct {
 	clienttest.ClientTestSuite
+	tmpFileName string
+}
+
+func (s *testSuite) SetupTest() {
+	input, err := ioutil.TempFile(s.TempDir, "")
+	d.Chk.NoError(err)
+	s.tmpFileName = input.Name()
+	writeCSV(input)
+	defer input.Close()
+}
+
+func (s *testSuite) TearDownTest() {
+	os.Remove(s.tmpFileName)
 }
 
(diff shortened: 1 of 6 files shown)


samples/go/csv/csv-import/importer_test.go
```diff
@@ -22,46 +22,93 @@ import (
...
 	clienttest.ClientTestSuite
+	tmpFileName string
+}
+
+func (s *testSuite) SetupTest() {
+	input, err := ioutil.TempFile(s.TempDir, "")
+	d.Chk.NoError(err)
+	s.tmpFileName = input.Name()
+	writeCSV(input)
+	defer input.Close()
```
willhite: Do you want to move this line up? Seems a little pointless to defer it here.

samples/go/csv/read.go
```diff
@@ -131,24 +169,76 @@ func ReadToMap(r *csv.Reader, structName string, headersRaw []string, pkIdx int,
...
+			panic(err)
+		}
+
+		fields := readFieldsFromRow(row, headersRaw, fieldOrder, kindMap)
+		rowStruct := types.NewStructWithType(t, fields)
+
+		// needed to allow recursive calls to encloseInMap
+		var encloseInMapFunc func(m map[types.Value]mapOrStruct, keyLevel int) map[types.Value]mapOrStruct
+		encloseInMap := func(m map[types.Value]mapOrStruct, keyLevel int) map[types.Value]mapOrStruct {
+			fieldOrigIndex := fieldOrder[pkIndices[keyLevel]]
```
willhite: You should be able to say:
var encloseInMap func(...
encloseInMap := func...

there's no need to declare a var with a different name than the one you want to use.

samples/go/csv/read.go
```diff
@@ -131,24 +169,76 @@ func ReadToMap(r *csv.Reader, structName string, headersRaw []string, pkIdx int,
...
+			if n, ok := m[key]; !ok {
+				subMap = make(map[types.Value]mapOrStruct)
+			} else {
+				subMap = n.goMap
+			}
+			m[key] = mapOrStruct{encloseInMapFunc(subMap, keyLevel+1), types.Struct{}}
+			return m
+		}
+
+		encloseInMapFunc = encloseInMap
```
willhite: this line can go away (see above commit)

willhite:
LGTM with a nit or two. And the caveat that reading with multi-field primary keys all happens in memory. Which will get resolved once GraphBuilder is implemented.

samples/go/csv/csv-import/importer_test.go
mikegray: good point.

samples/go/csv/read.go
mikegray: good point. done.

samples/go/csv/read.go
mikegray: done.

mikegray:
Thanks for the review. Yes, all happening in memory, not good performance, but previous users with one map key get same performance as before and GraphBuilder will enable better performance of nested maps in the future.

mikegray:
fixed #1956

PR #2434, opened 2016-08-25
Add a perf viewer

kalman:


kalman:
@arv PTAL - I should probably add tests for those utility functions (median, gen colours, etc), but in the meantime, this should be reviewable.

This is what it looks like:
screen shot 2016-08-24 at 11 36 23 pm

js/perf/viewer/README.md
```diff
@@ -0,0 +1,15 @@
+# js/perf/viewer
+
+viewer for Noms perf data
+
+## Getting Started
+
+```
+# First time only:
+npm i
```
arv: spell out install for better readability

js/perf/viewer/index.html
```diff
@@ -0,0 +1,9 @@
+<!--
```
arv: `<!doctype html>` please. No quirks mode allowed!

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+
+async function load() {
+  const params = getParams();
+  if (!params.ds) {
+    window.alert('Must provide a ?ds= param');
+    return;
+  }
+
+  if (params.refresh) {
+    window.setTimeout(() => window.location.reload(), params.refresh);
```
arv: or just 

```js
    setTimeout(() => location.reload(), params.refresh);
```

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+window.onpopstate = load;
+window.onresize = render;
+
+// The maximum number of git revisions to show in the perf history.
+// The larger this number, the more screen real estate needed to render the graph - and the slower
+// it will take to render, since the entire parent commit chain must be walked to form the graph.
+// TODO: Implement paging mechanism.
+const MAX_PERF_HISTORY = 15;
+
+let chartDatasets: Map<string /* test name */, (number | null)[] /* elapsed, in seconds */>;
```
arv: You could name these types to remove the comments:

```js
type TestName = string;
...
let chartDatasets: Map<TestName, ...>;
```

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+    window.setTimeout(() => window.location.reload(), params.refresh);
+  }
+
+  const dsSpec = DatasetSpec.parse(params.ds);
+  const [perfData, gitRevs] = await getPerfHistory(dsSpec.dataset());
+
+  chartDatasets = new Map();
+  chartLabels = gitRevs.map(rev => rev.slice(0, 6));
+
+  // Gather all test names up ahead of time, since some commits may be missing results.
```
arv: The sentence is confusing. Care to give it another go?

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+    datasets.push({
+      backgroundColor,
+      borderColor,
+      borderWidth: 1,
+      data: elapsed,
+      label: testName,
+    });
+  }
+
+  // $FlowIssue: Chart is in modules, not imported, so Flow doesn't know about it.
```
arv: You can use `declare` tell Flow about it:

https://flowtype.org/docs/declarations.html

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+    res += nums[lenDiv2 - 1];
+    res /= 2;
+  }
+  return res;
+}
+
+// Generates a light and dark version of some color randomly (but stable) derived from `str`.
+async function genLightAndDarkColors(str: string): Promise<[string, string]> {
+  const strBuf = new window.TextEncoder().encode(str);
+  const hash = await window.crypto.subtle.digest('sha-256', strBuf);
```
arv: This seems like a very heavy weight way to determine two colors.

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+  if (nums.length % 2 === 0) {
+    res += nums[lenDiv2 - 1];
+    res /= 2;
+  }
+  return res;
+}
+
+// Generates a light and dark version of some color randomly (but stable) derived from `str`.
+async function genLightAndDarkColors(str: string): Promise<[string, string]> {
+  const strBuf = new window.TextEncoder().encode(str);
```
arv: TextEncoder does not work in Safari. Why don't you just do `codePointAt % 256` instead?

http://caniuse.com/#search=TextEncoder

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+  const lenDiv2 = Math.floor(nums.length / 2);
+  let res = nums[lenDiv2];
+  if (nums.length % 2 === 0) {
+    res += nums[lenDiv2 - 1];
+    res /= 2;
+  }
+  return res;
+}
+
+// Generates a light and dark version of some color randomly (but stable) derived from `str`.
```
arv: Why do we want this to be random in the first place. It seems better to hard code these so we can use our color palette.

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+  const strBuf = new window.TextEncoder().encode(str);
+  const hash = await window.crypto.subtle.digest('sha-256', strBuf);
+  invariant(hash instanceof ArrayBuffer);
+  const [r, g, b] = new Uint8Array(hash);
+  return [`rgba(${r}, ${g}, ${b}, 0.3)`, `rgb(${r}, ${g}, ${b})`];
+}
+
+// Returns the keys of `map`.
+function keys<K: Value, V: Value>(map: NomsMap<K, V>): Promise<K[]> {
+  return new Promise(res => {
```
arv: This doesn't look right. You should not need to use `new Promise` when you already have a promise to begin with.

```js
function keys<K: Value, V: Value>(map: NomsMap<K, V>): Promise<K[]> {
  const keys = [];
  return map.forEach((_, key) => {
    keys.push(key);
  }).then(() => keys);
}
```

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+    const keys = [];
+    map.forEach((_, key) => {
+      keys.push(key);
+      return;
```
arv: no empty returns

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+  // Gather all test names up ahead of time, since some commits may be missing results.
+  const firstReps = await Promise.all(perfData.map(pd => {
+    invariant(pd.reps instanceof List);
+    return pd.reps.get(0);
+  }));
+  const testNamesSet = new Set();
+  for (const fr of firstReps) {
+    (await keys(fr)).forEach(testName => testNamesSet.add(testName));
+  }
+  // $FlowIssue: thinks Set is incompatible with spread operand... but it is.
```
arv: https://github.com/facebook/flow/issues/1059

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+  const testNames = [...testNamesSet];
+
+  const getElapsed = async (testName: string, pd: Struct) => {
+    invariant(pd.reps instanceof List);
+    const reps = await pd.reps.toJS();
+    const elapsedOrNulls = await Promise.all(reps.map(rep => {
+      invariant(rep instanceof NomsMap);
+      // Note: despite how this code is structured, either all reps should have test data for this
+      // value, or none should. Ideally we'd be able to bail at this point.
+      return rep.get(testName).then(d => d !== null && d !== undefined ? d.elapsed / 1e9 : null);
```
arv: `d !== null && d !== undefined` => `d`

arv:
```js
return rep.get(testName).then(d => d && d.elapsed / 1e9);
```

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  return [perfData, gitRevs];
+}
+
+// Returns a map of URL param key to value.
+function getParams(): {[key: string]: string} {
+  // Note: this way anything after the # will end up in `params`, which is what we want.
+  const params = {};
+  const paramsIdx = location.href.indexOf('?');
```
arv: here you are using location without window. Let's be consistent.

js/perf/viewer/src/main.js
```diff
@@ -0,0 +1,192 @@
...
+}
+
+async function render() {
+  if (!chartDatasets) {
+    return;
+  }
+
+  const datasets = [];
+  for (const [testName, elapsed] of chartDatasets) {
+    const [backgroundColor, borderColor] = await genLightAndDarkColors(testName);
```
arv: I see. Every test gets a unique color. Didn't we have a PRG used in testing? Could we use that instead?

arv:
Back to you

js/perf/viewer/README.md
kalman: done

js/perf/viewer/index.html
kalman: done

js/perf/viewer/src/main.js
kalman: I could, though I don't particularly like that idiom

js/perf/viewer/src/main.js
kalman: Better

js/perf/viewer/src/main.js
kalman: Nice

js/perf/viewer/src/main.js
kalman: done

js/perf/viewer/src/main.js
kalman: done

js/perf/viewer/src/main.js
kalman: As discussed - I'll use noms hash :-)

js/perf/viewer/src/main.js
kalman: neat

js/perf/viewer/src/main.js
kalman: as discussed

js/perf/viewer/src/main.js
kalman: as discussed

js/perf/viewer/src/main.js
kalman: as discussed

js/perf/viewer/src/main.js
kalman: ah. right, thanks.

js/perf/viewer/src/main.js
kalman: I initially did it like this to make both eslint and flow happy:
* If I don't have the empty return, then eslint will insist that I write this like `map.forEach((_, key) => keys.push(key))`
* but then flow complains about that, because `keys.push` returns an int, but `map.forEach` only takes functions that return a `?Promise<void>`, of which this isn't.
* _but_ now eslint seems to not care, so that's weird, but I'll take it.

kalman:
PTAL

arv:
LGTM with nits

PR #2437, opened 2016-08-25
Camel cased fields and logic consolidation

zcstarr:
Changes address logic consolidation concerns and exposing new serialization policy for fields coming from csv.

 Here I also added a perhaps more strict policy for Escaping Struct Fields to verify that the item that's been escaped also meets the escaping policy standards, might be a little overkill .

Finally, I thought, that perhaps the Escape field logic might be a better fit for struct.go vs type.go , but happy to migrate it back if folk feel otherwise :) .

 Any thoughts, comments etc... are highly welcome.


go/types/struct.go
```diff
@@ -169,18 +169,32 @@ func (s1 Struct) Diff(s2 Struct, changes chan<- ValueChanged, closeChan <-chan s
...
+	for _, ch := range input {
+		output += encode(string([]rune{ch}), pattern)
+		pattern = tailFieldNamePattern
+	}
+	return output
+}
 
 // Escapes names for use as noms structs. Disallowed characters are encoded as
 // 'Q<hex-encoded-utf8-bytes>'. Note that Q itself is also escaped since it is
-// the escape character.
```
mikegray: Looks like the comment got cut off.

go/types/struct.go
```diff
@@ -195,13 +209,45 @@ func EscapeStructField(input string) string {
...
+	return output
 
-	output := ""
-	pattern := headPattern
-	for _, ch := range input {
-		output += encode(string([]rune{ch}), pattern)
-		pattern = tailPattern
+}
+
+// IsValidStructFieldName returns whether the name is valid without as a field name in a struct.
```
mikegray: This comment doesn't read clearly - "...name is valid without as..."? "without encoding" maybe, not sure?

go/types/struct.go
```diff
@@ -195,13 +209,45 @@ func EscapeStructField(input string) string {
...
+		return
 	}
 
-	return output
+	last := names[0]
+	verifyFieldName(last)
+
+	for i := 1; i < len(names); i++ {
+		verifyFieldName(names[i])
+		if names[i] <= last {
```
mikegray: Use strings.Compare(names[i], last).

mikegray: You are also using `<=`, but you want it to be `<` because if it is `=` then it wouldn't be unique.

samples/go/csv/read.go
```diff
@@ -44,6 +46,31 @@ func KindsToStrings(kinds KindSlice) []string {
 	return strs
 }
 
+//EscapeStructFieldForCSV removes special characters and replaces spaces with camelCasing (camel' case turns to camelCase)
```
mikegray: Remove extra quote/apostrophe? `(camel' case turns to camelCase)` maybe `("camel case" becomes "camelCase")`?

samples/go/csv/read.go
```diff
@@ -44,6 +46,31 @@ func KindsToStrings(kinds KindSlice) []string {
...
+
+	if len(splitFields) > 1 {
+		output = strings.ToLower(output)
+	}
+
+	for _, field := range splitFields[1:] {
+		output += strings.Title(strings.ToLower(types.EscapeFields(field, encode)))
+	}
+
+	types.IsValidStructFieldName(output)
```
mikegray: Not checking the return value of this. Left over debugging? If not, please check the return value.

samples/go/csv/read.go
mikegray: Golang style would make the function name here EscapeStructFieldForCsv.

samples/go/csv/read_test.go
```diff
@@ -133,6 +133,28 @@ g,h,i,j
 `
 	testTrailingHelper(t, dataString)
 }
+func TestEscapeStructFieldFromCSV(t *testing.T) {
```
mikegray: Are you running `gofmt` in your editor? Should be a blank line between functions.

samples/go/csv/read_test.go
```diff
@@ -133,6 +133,28 @@ g,h,i,j
...
+	assert := assert.New(t)
+	cases := []string{
+		"a", "a",
+		"1a", "a",
+		"AaZz19_", "AaZz19_",
+		"Q", "Q",
+		"AQ", "AQ",
+		"$", "",
```
mikegray: I don't think we can have an empty string for a struct field name. If you end up stripping out all of the characters, what should happen?  It shouldn't be a valid struct field, so maybe it should panic? This and the emoji case below are interesting test cases that should probably either return an error or panic from EscapeStructFieldFromCsv.

mikegray:
ok, back to you. You will probably want to have @aboodman or @rafael-atticlabs look at your first several commits. They will help with useful style, methodology, etc. feedback too.

aboodman:
Something weird happened with this PR where it includes many more commits that are unrelated.

samples/go/csv/read_test.go
zcstarr: most definitely good catch

samples/go/csv/read_test.go
zcstarr: I was using goreturns, but now that I switched to gofmt I didn't see a difference( I also ran it from commandline), do I need any additional linting information ? New to go :)

samples/go/csv/read.go
zcstarr: Typo from an early function I'll clean this up

go/types/struct.go
zcstarr: I think perhaps <= is correct because its a check that  things are ordered least to greatest and not unique so if prev is greater than next  you end up violating the check as well as if it were ==

mikegray: Right you are, sorry I was just looking at it wrong.

zcstarr:
@aboodman yeah I think I might have fat fingered the rebasing in some sort of way, I'm going to attempt to reset things.

zcstarr:
Moving this pr to https://github.com/attic-labs/noms/pull/2441 struggling to put the commit order and back in a good state . I think I ended up rebasing and pulling in other peoples commits, then committing after those commits, which put things in a really funky state.

PR #2441, opened 2016-08-25
Add special encoding to csv imported struct fields

zcstarr:
…n logic, and changes csv importing to strip csv name fields of invalid struct characters and changes spaces in field names to camel case. i.e. camel case is translated to camelCase.

Additional comments and conversation here. I was struggling to undo my fat finger rebase orz.
https://github.com/attic-labs/noms/pull/2437

### DIFF
--- a/go/types/struct.go
+++ b/go/types/struct.go
@@ -7,11 +7,12 @@ package types
 import (
 	"bytes"
 	"fmt"
-	"github.com/attic-labs/noms/go/d"
-	"github.com/attic-labs/noms/go/hash"
 	"regexp"
 	"sort"
 	"strings"
+
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/hash"
 )
 
 var EmptyStructType = MakeStructType("", []string{}, []*Type{})
@@ -178,24 +179,7 @@ var fieldNameRe = regexp.MustCompile(fieldNameComponentRe.String() + "$")
 
 type encodingFunc func(string, *regexp.Regexp) string
 
-func EscapeFields(input string, encode encodingFunc) string {
-	output := ""
-	pattern := headFieldNamePattern
-	for _, ch := range input {
-		output += encode(string([]rune{ch}), pattern)
-		pattern = tailFieldNamePattern
-	}
-	return output
-}
-
-// EscapeStructField escapes names for use as noms structs. Disallowed characters are encoded as
-// 'Q<hex-encoded-utf8-bytes>'. Note that Q itself is also escaped since it is
-// the escape character.
-func EscapeStructField(input string) string {
-
-	if !escapeRegex.MatchString(input) && IsValidStructFieldName(input) {
-		return input
-	}
+func QEncodeFieldName(input string) string {
 	encode := func(s1 string, p *regexp.Regexp) string {
 		if p.MatchString(s1) && s1 != escapeChar {
 			return s1
@@ -210,10 +194,52 @@ func EscapeStructField(input string) string {
 		buf.WriteString(hs)
 		return buf.String()
 	}
-	output := EscapeFields(input, encode)
-	verifyFieldName(output)
-	return output
+	return escapeFields(input, encode)
+}
 
+func CamelCaseFieldName(input string) string {
+	encode := func(s1 string, p *regexp.Regexp) string {
+		if p.MatchString(s1) {
+			return s1
+		}
+		return ""
+	}
+
+	splitFields := strings.Fields(input)
+	output := escapeFields(splitFields[0], encode)
+
+	if len(splitFields) > 1 {
+		output = strings.ToLower(output)
+	}
+
+	for _, field := range splitFields[1:] {
+		output += strings.Title(strings.ToLower(escapeFields(field, encode)))
+	}
+
+	if !IsValidStructFieldName(output) {
+		return ""
+	}
+	return output
+}
+
+func escapeFields(input string, encode encodingFunc) string {
+	output := ""
+	pattern := headFieldNamePattern
+	for _, ch := range input {
+		output += encode(string([]rune{ch}), pattern)
+		pattern = tailFieldNamePattern
+	}
+	return output
+}
+
+// EscapeStructField escapes names for use as noms structs. Disallowed characters are encoded as
+// 'Q<hex-encoded-utf8-bytes>'. Note that Q itself is also escaped since it is
+// the escape character.
+func EscapeStructField(input string) string {
+	if !escapeRegex.MatchString(input) && IsValidStructFieldName(input) {
+		return input
+	}
+	return QEncodeFieldName(input)
 }
 
 // IsValidStructFieldName returns whether the name is valid as a field name in a struct.
--- a/go/types/struct.go
+++ b/go/types/struct.go
@@ -195,7 +195,7 @@ func CamelCaseFieldName(input string) string {
 
 	//Camelcase field
 	if len(splitField) == 1 {
-		output = splitField[0]
+		output = strings.ToLower(splitField[0])
 	} else if len(splitField) > 1 {
 		output = strings.ToLower(splitField[0])
 		for _, field := range splitField[1:] {
--- a/go/types/struct.go
+++ b/go/types/struct.go
@@ -215,7 +215,7 @@ func CamelCaseFieldName(input string) string {
 	for _, field := range splitFields[1:] {
 		output += strings.Title(strings.ToLower(escapeFields(field, encode)))
 	}
-
+	//Because we are removing characters, we may generate an invalid field name
 	if !IsValidStructFieldName(output) {
 		return ""
 	}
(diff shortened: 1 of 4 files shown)


aboodman:
Nit: we try to follow http://chris.beams.io/posts/git-commit/ for the commit messages.

go/types/struct.go
```diff
@@ -169,18 +169,33 @@ func (s1 Struct) Diff(s2 Struct, changes chan<- ValueChanged, closeChan <-chan s
...
 var escapeChar = "Q"
-var headPattern = regexp.MustCompile("[a-zA-PR-Z]")
-var tailPattern = regexp.MustCompile("[a-zA-PR-Z0-9_]")
-var completePattern = regexp.MustCompile("^" + headPattern.String() + tailPattern.String() + "*$")
+var headFieldNamePattern = regexp.MustCompile("[a-zA-Z]")
+var tailFieldNamePattern = regexp.MustCompile("[a-zA-Z0-9_]")
+var escapeRegex = regexp.MustCompile(escapeChar)
+
+var fieldNameComponentRe = regexp.MustCompile("^" + headFieldNamePattern.String() + tailFieldNamePattern.String() + "*")
+var fieldNameRe = regexp.MustCompile(fieldNameComponentRe.String() + "$")
```
aboodman: Are any of the regexs still used other than `fieldNameRe`? If not, we could just put the whole thing there, simplify.

aboodman: Answering, self: yes, by `EscapeFields`

samples/go/csv/read.go
```diff
@@ -44,6 +46,31 @@ func KindsToStrings(kinds KindSlice) []string {
 	return strs
 }
 
+//EscapeStructFieldFromCSV removes special characters and replaces spaces with camelCasing (camel case turns to camelCase)
+func EscapeStructFieldFromCSV(input string) string {
+
```
aboodman: Remove blank line

go/types/struct.go
zcstarr: Yes all of them are in use, headFieldNamePattern and tailFieldNamePattern are used as incremental regex for the encoding logic and fieldNameComponentRe is used by path

go/types/struct.go
```diff
@@ -169,18 +169,33 @@ func (s1 Struct) Diff(s2 Struct, changes chan<- ValueChanged, closeChan <-chan s
...
+
+type encodingFunc func(string, *regexp.Regexp) string
```
aboodman: I think it is confusing to leave testing the regex to the callee. Why would that not be the responsibility of EscapeFields?

aboodman: Sigh. Answering self again: I see.

aboodman: I see why you are doing this, but I find the signature awkward. As a caller of `EscapeFields`, how do I know what to do with the regex that is passed to `encodingFunc`.

Could you factor this like so instead?

`func QEncodeFieldName(name string) string`

and:

`func CamelCaseFieldName(name string) string`

all the rest of the guts should be package-private implementation details, not exposed to caller.

go/types/struct.go
```diff
@@ -169,18 +169,33 @@ func (s1 Struct) Diff(s2 Struct, changes chan<- ValueChanged, closeChan <-chan s
...
+
+func EscapeFields(input string, encode encodingFunc) string {
```
aboodman: Nit: this only escapes one field name, so it should be signular.

go/types/struct.go
```diff
@@ -195,13 +210,45 @@ func EscapeStructField(input string) string {
 		buf.WriteString(hs)
 		return buf.String()
 	}
+	output := EscapeFields(input, encode)
+	verifyFieldName(output)
```
aboodman: Isn't the output guaranteed to be valid at this point, assuming `EscapeFields` works properly? If so there's no need for this check.

zcstarr: Correct, I suppose if we were to break the encoding function, we'd catch something like that in test in general. Wasn't sure if really bad things happened if we introduced a bug in the encode that was missed in test, but that generated bad data or if we'd have enough checks upstream that we'd just fail to write the struct.

aboodman: We should have thorough unit tests for functions, then callers of those functions can assume they are correct without having to double-check the work.

go/types/struct.go
zcstarr: I think this makes sense and is much cleaner 👍 .

zcstarr:
@aboodman mind giving this another once over

aboodman:
Please improve the PR title

aboodman:
It should be one sentence, at most 50 chars max

aboodman:
see http://chris.beams.io/posts/git-commit/

go/types/struct.go
```diff
@@ -195,13 +194,87 @@ func EscapeStructField(input string) string {
...
+
+	if len(splitFields) > 1 {
+		output = strings.ToLower(output)
+	}
+
+	for _, field := range splitFields[1:] {
+		output += strings.Title(strings.ToLower(escapeFields(field, encode)))
+	}
+
+	if !IsValidStructFieldName(output) {
```
aboodman: Can you put a little comment here explaining how this can happen? The code is correct, but it's not immediately obvious from reading it why this would occur.

go/types/struct.go
```diff
@@ -195,13 +194,87 @@ func EscapeStructField(input string) string {
...
 	for _, ch := range input {
 		output += encode(string([]rune{ch}), pattern)
-		pattern = tailPattern
+		pattern = tailFieldNamePattern
 	}
-
 	return output
 }
+
+// EscapeStructField escapes names for use as noms structs. Disallowed characters are encoded as
```
aboodman: Comment out of date.

aboodman: Actually, sorry, comment is correct. But can we just collapse this with QEncodeFieldName? do we need both?

aboodman:
Almost there. One more comment/question. BTY.

go/types/struct.go
```diff
@@ -215,7 +215,7 @@ func CamelCaseFieldName(input string) string {
 	for _, field := range splitFields[1:] {
 		output += strings.Title(strings.ToLower(escapeFields(field, encode)))
 	}
-
+	//Because we are removing characters, we may generate an invalid field name
```
aboodman: Maybe include an example? This seems subtle enough that it warrants one.

zcstarr:
@aboodman PTAL I changed the camelcase encoding logic a bit to support a few more stripped data edge cases and renamed escapedFields to escapedField.

go/types/struct.go
```diff
@@ -180,31 +181,38 @@ var fieldNameRe = regexp.MustCompile(fieldNameComponentRe.String() + "$")
...
-	output := escapeFields(splitFields[0], encode)
+	strippedField := escapeField(input, encode)
+	splitField := strings.Fields(strippedField)
+	output := ""
 
-	if len(splitFields) > 1 {
-		output = strings.ToLower(output)
+	//Camelcase field
+	if len(splitField) == 1 {
+		output = splitField[0]
```
aboodman: Shouldn't `splitField[0]` be `ToLower()`'d?

samples/go/csv/read_test.go
```diff
@@ -148,6 +148,9 @@ func TestEscapeStructFieldFromCSV(t *testing.T) {
 		"https://picasaweb.google.com/data", "httpspicasawebgooglecomdata",
 		"💩", "",
 		"11 1💩", "",
+		"-- A B", "aB",
+		"-- A --", "A",
```
aboodman: Yeah actually I'd expect the result to be "a" here, no?

go/types/struct.go
zcstarr: In this case it assumes that there's only one string and should just preserve the case of the original input. i.e. 

---- Camels ---- should just be Camels vs camels , if that makes sense ?

samples/go/csv/read_test.go
zcstarr: the thought here is that the dashes are bad, so the algorithm would strip the bad characters and apply camel casing as a best effort. so if we strip the dashes we get "A B" then transforming it to camel case it becomes aB.

zcstarr: uncertain if these expectations are the right one , I can change it if it makes more sense going in a different direction

go/types/struct.go
zcstarr: discussed irl, lowercasing this makes sense especially because is a camelcasing function

zcstarr:
@aboodman  PTAL

go/types/struct.go
```diff
@@ -191,18 +191,18 @@ func CamelCaseFieldName(input string) string {
 
 	strippedField := escapeField(input, encode)
 	splitField := strings.Fields(strippedField)
-	output := ""
+
+	if len(splitField) == 0 {
```
aboodman: good catch. i would not have guessed Fields could return empty slice.

aboodman:
lgtm

PR #2448, opened 2016-08-27
Add types.List support merge.ThreeWay()

cmasone-attic:
While the code in threeWayListMerge() starts similarly to that
in threeWayOrderedSequenceMerge(), the types of the channels
are different and trying to shoehorn them together would
remove a lot of clarity just for the sake of sharing maybe
15 lines of code.

Toward #148

### DIFF
--- a/go/merge/three_way.go
+++ b/go/merge/three_way.go
@@ -75,8 +75,9 @@ func threeWayMerge(a, b, parent types.Value, vwr types.ValueReadWriter) (merged 
 
 	switch a.Type().Kind() {
 	case types.ListKind:
-		// TODO: Come up with a plan for List (BUG 148)
-		return parent, newMergeConflict("Cannot merge %s.", a.Type().Describe())
+		if aList, bList, pList, ok := listAssert(a, b, parent); ok {
+			return threeWayListMerge(aList, bList, pList)
+		}
 
 	case types.MapKind:
 		if aMap, bMap, pMap, ok := mapAssert(a, b, parent); ok {
@@ -189,6 +190,19 @@ func threeWayStructMerge(a, b, parent types.Struct, vwr types.ValueReadWriter) (
 		panic(fmt.Errorf("Bad key type in diff: %s", change.V.Type().Describe()))
 	}
 	return threeWayOrderedSequenceMerge(parent, aDiff, bDiff, makeGetFunc(a), makeGetFunc(b), makeGetFunc(parent), apply, vwr)
+}
+
+func listAssert(a, b, parent types.Value) (aList, bList, pList types.List, ok bool) {
+	var aOk, bOk, pOk bool
+	aList, aOk = a.(types.List)
+	bList, bOk = b.(types.List)
+	if parent != nil {
+		pList, pOk = parent.(types.List)
+	} else {
+		pList, pOk = types.NewList(), true
+	}
+	ok = aOk && bOk && pOk
+	return
 }
 
 func mapAssert(a, b, parent types.Value) (aMap, bMap, pMap types.Map, ok bool) {
--- a/go/merge/three_way.go
+++ b/go/merge/three_way.go
@@ -201,8 +201,7 @@ func listAssert(a, b, parent types.Value) (aList, bList, pList types.List, ok bo
 	} else {
 		pList, pOk = types.NewList(), true
 	}
-	ok = aOk && bOk && pOk
-	return
+	return aList, bList, pList, aOk && bOk && pOk
 }
 
 func mapAssert(a, b, parent types.Value) (aMap, bMap, pMap types.Map, ok bool) {
@@ -214,8 +213,7 @@ func mapAssert(a, b, parent types.Value) (aMap, bMap, pMap types.Map, ok bool) {
 	} else {
 		pMap, pOk = types.NewMap(), true
 	}
-	ok = aOk && bOk && pOk
-	return
+	return aMap, bMap, pMap, aOk && bOk && pOk
 }
 
 func refAssert(a, b, parent types.Value, vwr types.ValueReadWriter) (aValue, bValue, pValue types.Value, ok bool) {
@@ -248,8 +246,7 @@ func setAssert(a, b, parent types.Value) (aSet, bSet, pSet types.Set, ok bool) {
 	} else {
 		pSet, pOk = types.NewSet(), true
 	}
-	ok = aOk && bOk && pOk
-	return
+	return aSet, bSet, pSet, aOk && bOk && pOk
 }
 
 func structAssert(a, b, parent types.Value) (aStruct, bStruct, pStruct types.Struct, ok bool) {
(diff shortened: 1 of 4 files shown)


cmasone-attic:
PTAL

go/merge/three_way_list.go
```diff
@@ -0,0 +1,119 @@
...
+		earlier, later = s2, s1
+	}
+	return s1.SpAt == s2.SpAt || earlier.SpAt+earlier.SpRemoved > later.SpAt
+}
+
+func canMerge(a, b types.List, aSplice, bSplice types.Splice) bool {
+	if aSplice != bSplice {
+		return false
+	}
+	// TODO: Add List.IterFrom() and use that to compare added values.
```
rafael-atticlabs: you can use `getCompareFn` here.

rafael-atticlabs:
lgtm w/ nit. @kalman can you back me up here.

go/merge/three_way.go
```diff
@@ -191,6 +192,19 @@ func threeWayStructMerge(a, b, parent types.Struct, vwr types.ValueReadWriter) (
...
+func listAssert(a, b, parent types.Value) (aList, bList, pList types.List, ok bool) {
+	var aOk, bOk, pOk bool
+	aList, aOk = a.(types.List)
+	bList, bOk = b.(types.List)
+	if parent != nil {
+		pList, pOk = parent.(types.List)
+	} else {
+		pList, pOk = types.NewList(), true
+	}
+	ok = aOk && bOk && pOk
```
kalman: Having a named return value seems more verbose

go/merge/three_way_list.go
cmasone-attic: Sadly...no I can't :-/

go/merge/three_way_list.go
```diff
@@ -0,0 +1,119 @@
...
+		close(stop)
+		for range drain {
+		}
+	}
+
+	defer stopAndDrain(aStopChan, aSpliceChan)
+	defer stopAndDrain(bStopChan, bSpliceChan)
+
+	zeroSplice := types.Splice{}
+	zeroToEmpty := func(sp types.Splice) types.Splice {
```
kalman: I read "empty" similar to "zero". Since splice uses "unassigned", how about "unassignedSplice/zeroToUnassigned"?

kalman: It would be nice to have a quick comment here, because I needed to look up the splice code to understand this. Namely, that the zero actually means something to the splice code, so it uses "unassigned" instead.

go/merge/three_way_list.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		return sp
+	}
+
+	merged = parent
+	offset := uint64(0)
+	emptySplice := zeroToEmpty(types.Splice{})
+	aSplice, bSplice := emptySplice, emptySplice
+	for {
+		// Get the next splice from both a and b. If either diff(a, parent) or diff(b, parent) is complete, aSplice or bSplice will get an empty types.Splice. Generally, though, this allows us to proceed through both diffs in (key) order, considering the "current" splice from both diffs at the same time.
```
kalman: "key" => "index"

go/merge/three_way_list.go
kalman: .... and why that even matters to this code, because really, a zero-initialized splice is meaningless. Removing and deleting 0 elements is a no-op.

go/merge/three_way_list.go
```diff
@@ -0,0 +1,119 @@
...
+	for i := uint64(0); i < aSplice.SpAdded; i++ {
+		if !a.Get(aSplice.SpFrom + i).Equals(b.Get(bSplice.SpFrom + i)) {
+			return false
+		}
+	}
+	return true
+}
+
+func merge(s1, s2 types.Splice) types.Splice {
+	return s1
```
kalman: I don't follow. This is because s1 == s2 right? Why even bother to have a method?

go/merge/three_way_list.go
```diff
@@ -0,0 +1,119 @@
...
+
+func overlap(s1, s2 types.Splice) bool {
+	earlier, later := s1, s2
+	if s2.SpAt < s1.SpAt {
+		earlier, later = s2, s1
+	}
+	return s1.SpAt == s2.SpAt || earlier.SpAt+earlier.SpRemoved > later.SpAt
+}
+
+func canMerge(a, b types.List, aSplice, bSplice types.Splice) bool {
```
kalman: (relates to comment below about the `merge` function)

Maybe just call this splicesEqual.

kalman: Or something which implies more strongly that not only the splices are equal, but the lists which they refer to are equal.

go/merge/three_way_list.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return true
+}
+
+func merge(s1, s2 types.Splice) types.Splice {
+	return s1
+}
+
+func apply(source, target types.List, offset uint64, s types.Splice) types.List {
+	// TODO: Add List.IterFrom() and use that to build up toAdd.
```
kalman: How would that help?

I was thinking that a `Slice` method would be handy. Each call to `Get` is O(log n).

kalman:
LGTM with nits.

go/merge/three_way.go
cmasone-attic: sure

go/merge/three_way_list.go
cmasone-attic: Done

go/merge/three_way_list.go
cmasone-attic: done

go/merge/three_way_list.go
cmasone-attic: I named these functions to line up with the pseudocode we wrote in issue #148. It's possible that we'll change how `canMerge()` and `merge()` work soonish

go/merge/three_way_list.go
cmasone-attic: ...you're right. I had in my head that that would mean I could reuse the same cursor. But that's not actually so. I'll ask @rafael-atticlabs about a Splice method tomorrow.

go/merge/three_way_list.go
cmasone-attic: To give us a spot to hook in different logic later :-)

go/merge/three_way_list.go
cmasone-attic: Oh, wait. Using an Iter() method DOES mean I don't have to rebuild the cursor every time. Basically, I'd do

```
toAdd := make(types.ValueSlice, s.SpAdded)
source.IterFrom(s.SpFrom, func(idx uint64, v types.Value) {
  toAdd[int(idx)] = v
})
```

go/merge/three_way_list.go
cmasone-attic: I wouldn't mind suggestions on how to make this more efficient, though...

go/merge/three_way_list.go
kalman: Slice, not Splice. But yes, that would be equivalent to an IterFrom method. We should totally do that.

(and we might want to save the Slice namespace for a `List -> List` not `List -> []Value` method anyway).

PR #2449, opened 2016-08-27
 Introduce poke: a simple tool for modifying Noms data

aboodman:


go/types/path.go
```diff
@@ -261,7 +265,7 @@ func (hip HashIndexPath) String() string {
 	return fmt.Sprintf("[#%s]%s", hip.Hash.String(), ann)
 }
 
-func parsePathIndex(str string) (idx Value, h hash.Hash, rem string, err error) {
+func ParsePathIndex(str string) (idx Value, h hash.Hash, rem string, err error) {
```
kalman: Comment for public method

go/types/path.go
```diff
@@ -272,6 +276,7 @@ Switch:
 		for ; i < len(str); i++ {
 			c := str[i]
 			if c == '"' {
+				i++
```
kalman: It looks like I missed writing a test like
```
test(`.foo["hello"`, "[ is missing closing ]")
```
I.e. a valid index (`"hello"`) but missing a closing `]`. Could you write that test? Seems like it would have been affected by this patch.

samples/go/poke/main.go
```diff
@@ -0,0 +1,137 @@
...
+		fmt.Fprintf(os.Stderr, "  <new-val> : new value for <path>\n\n")
+		fmt.Fprintln(os.Stderr, "Flags:\n")
+		flag.PrintDefaults()
+	}
+
+	flag.Parse(false)
+
+	if flag.NArg() == 0 {
+		flag.Usage()
+		return
```
kalman: It would be good to get posix exit codes here. From what I've seen, the Go idiom is having a different function which returns an int, then calling `os.Exit` on it, like
```
func poke() (exitCode int) {
  if flag.Narg() == 0 {
    flag.Usage()
    return 1
  }
  ...
  return
}

func main() {
  os.Exit(poke())
}
```
E.g. https://github.com/golang/go/blob/master/src/cmd/gofmt/gofmt.go#L161 (that uses a global exitCode I suppose).

samples/go/poke/main.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	oldVal := inPath.Resolve(inRoot)
+	if oldVal == nil {
+		fmt.Fprintf(os.Stderr, "No value at path '%s' - cannot update\n", inPath.String())
+		return
+	}
+
+	val, _, rem, err := types.ParsePathIndex(flag.Arg(2))
```
kalman: It looks surprising to see path parsing code used here, perhaps comment that it's somewhat of a hack, since it just so happens that path code knows how to parse values.

You know, ideally that stuff would be factored out of path because it does look pretty weird.

samples/go/poke/main.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil || rem != "" {
+		fmt.Fprintf(os.Stderr, "Invalid new value: '%s': %s\n", flag.Arg(2), err)
+		return
+	}
+
+	var outDS dataset.Dataset
+	if *outDSStr == "" {
+		outDS = inDS
+	} else {
```
kalman: IMO cleaner to write this like
```
if *outDSStr == "" {
  ...
} else if !dataset.DatasetRe.MatchString(*outDSStr) {
  ...
} else {
  ...
}
```

samples/go/poke/main.go
```diff
@@ -0,0 +1,137 @@
...
+			err = fmt.Errorf("Invalid output dataset name: %s", *outDSStr)
+		} else {
+			outDS = dataset.NewDataset(inDS.Database(), *outDSStr)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Invalid output dataset '%s': %s\n", outDSStr, err)
+				return
+			}
+		}
+	}
+
```
kalman: defer `outDS.database().Close()`?

samples/go/poke/main.go
```diff
@@ -0,0 +1,137 @@
...
+			return subject.Set(uint64(float64(part.Index.(types.Number))), newVal), nil
+		case types.Map:
+			return subject.Set(part.Index, newVal), nil
+		case types.Set:
+			return subject.Remove(part.Index).Insert(newVal), nil
+		default:
+			return nil, fmt.Errorf("Indexing into noms type %s unsupported", subject.Type().Describe())
+		}
+	}
+	return nil, fmt.Errorf("Unsupported path type: %#v", part)
```
kalman: I guess it doesn't make sense to update a hash path? E.g:
```
// foo[#abc] is currently 42, because hash of 42 is #abc.
foo[#abc] = 43
```
but the hash of 43 is <something else>.

My point: it would be helpful to enumerate these rather than just saying "Unsupported path type". E.g.
```
case types.HashIndexPath:
  if part.IntoKey {
    ???
  }
  return nil, fmt.Errorf("Hash indices not supported, it's impossible to assert that the new value has the same hash")
```
Otherwise it just looks like poke is unfinished.

kalman: ... of course this is a broken example because strings aren't sorted by their hashes, which does beg the question, how can we possibly reach this. It's almost like it should be a `panic`.

samples/go/poke/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+		"str": types.String("foobar"),
+		"lst": types.NewList(types.Number(1), types.String("foo")),
+		"map": types.NewMap(types.Number(1), types.String("foo"),
+			types.String("foo"), types.Number(1)),
+	}))
+	ds.Database().Close()
+
+	changes := map[string]string{
+		".num":          "43",
+		".str":          "\"foobaz\"",
```
kalman: I usually see ` `"foobaz"` ` not `"\"foobaz\""`

samples/go/poke/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+		".lst[0]":       "2",
+		".map[1]":       "\"bar\"",
+		".map[\"foo\"]": "2",
+	}
```
kalman: It would be nice to test error conditions as well, but there is only so much one can be bothered to do.

kalman:
Back to you

go/types/path.go
aboodman: done

go/types/path.go
aboodman: done

samples/go/poke/main.go
aboodman: I don't like having all the `return 1` all over the place.

aboodman: did something else, take a look

samples/go/poke/main.go
aboodman: I don't feel that it is a hack. Let's discuss.

samples/go/poke/main.go
aboodman: You could do hash. Poke just isn't finished. I don't need hash to work right now.

samples/go/poke/main_test.go
aboodman: Good point, forgot about that backtick syntax.

aboodman:
all done except where noted. ptal.

samples/go/poke/main.go
```diff
@@ -47,7 +44,7 @@ func main() {
 
 	inRoot, ok := inDS.MaybeHeadValue()
 	if !ok {
-		fmt.Fprintln(os.Stderr, "Input dataset has no data")
+		fmt.Fprintf(os.Stderr, "Input dataset '%s' does not exist\n", flag.Arg(0))
 		return
```
kalman: For the record I still prefer returning the explicit error code.

samples/go/poke/main.go
aboodman: `poke http://localhost:8000::foo '.bar[#asdfasdfas].thing' 42`

would create a new commit at `::foo` that looked like the old one except the value that used to be at that path will now have the value `42`. The fact that it has a different hash now is beside the poitn.

samples/go/poke/main_test.go
```diff
@@ -58,3 +63,36 @@ func (s *testSuite) TestRoundTrip() {
 		s.True(actual.Equals(v), "value at path %s incorrect (expected: %#v, got: %#v)", p.String(), v, actual)
 	}
 }
+
+func (s *testSuite) TestLose() {
+	sp := fmt.Sprintf("ldb:%s::test", s.LdbDir)
+	cases := []interface{}{
```
kalman: I think these can be more structured?

```
type case struct {
  args []string
  err  string
}

cases := []case{
  {[]string{"foo"}, "Incorrect number of arguments"},
  {[]string{"foo", "bar"}, "Incorrect number of arguments"},
  ...
}

...

for _, c := range cases {
  stdout, stderr := s.Run(main, c.args)
  s.Equal(c.err,stderr)
  // etc
}
```

kalman:
LGTM with 1 more comment (not the exit status one)

samples/go/poke/main_test.go
aboodman: Ah yes, thanks.

PR #2461, opened 2016-08-29
Add a perf test for CSV map import

kalman:
Currently we only have a perf test for CSV list import, which uses the
sf-crime dataset. This test uses the 43MB sf-registered-businesses
dataset instead, since sf-crime is too slow. Which is ironic, since we
normally parse sf-crime into a map.

I've also tightened up some of the other perf tests.
- Pause the blob import tests during commit, because commit is much
  faster the first time. I'll be able to reduce the test repetitions.
- Make the pure CSV parsing test use a smaller dataset, it doesn't need
  to use something as large as ny-vehicle-registrations.

### DIFF
--- a/samples/go/csv/csv-import/perf_test.go
+++ b/samples/go/csv/csv-import/perf_test.go
@@ -5,7 +5,6 @@
 package main
 
 import (
-	"bytes"
 	"fmt"
 	"io"
 	"os"
@@ -43,8 +42,10 @@ func (s *perfSuite) SetupSuite() {
 func (s *perfSuite) Test01ImportSfCrimeBlobFromTestdata() {
 	assert := s.NewAssert()
 
-	raw := s.readGlob(s.Testdata, "sf-crime", "2016-07-28.*")
-	blob := types.NewBlob(raw)
+	files := s.openGlob(s.Testdata, "sf-crime", "2016-07-28.*")
+	defer s.closeGlob(files)
+
+	blob := types.NewBlob(io.MultiReader(files...))
 	fmt.Fprintf(s.W, "\tsf-crime is %s\n", humanize.Bytes(blob.Len()))
 
 	ds := dataset.NewDataset(s.Database, "sf-crime/raw")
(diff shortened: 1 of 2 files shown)


kalman:
Perf this

samples/go/csv/csv-import/perf_test.go
```diff
@@ -50,29 +46,57 @@ func (s *perfSuite) Test01ImportSfCrimeBlobFromTestdata() {
 	defer s.closeGlob(raw)
```
kalman: I might rename this test as well, like `Test01ImportSfCrimeBlobOmitCommit`, so that the test results don't appear to change.

samples/go/csv/csv-import/perf_test.go
```diff
@@ -50,29 +46,57 @@ func (s *perfSuite) Test01ImportSfCrimeBlobFromTestdata() {
...
 
 	blob := types.NewBlob(io.MultiReader(raw...))
-	fmt.Fprintf(s.W, "csv/raw is %s\n", humanize.Bytes(blob.Len()))
+	fmt.Fprintf(s.W, "\tsf-crime is %s\n", humanize.Bytes(blob.Len()))
 
-	ds := dataset.NewDataset(s.Database, "csv/raw")
-	_, err := ds.CommitValue(blob)
-	assert.NoError(err)
+	// CommitValue is much faster the first time, so pause the test while it's happening.
+	s.Pause(func() {
```
rafael-atticlabs: per meatspace: this seems a bit strange to me. it might be better to leave this time in and then have a way in the UI to show "hot" vs "cold" perf.

rafael-atticlabs:
lgtm. "hot" vs "cold" suggestion doesn't need to be addressed here neccesarily

kalman:
Sorry @rafael-atticlabs , PTAL at the 2nd patch. It's changed non-trivially. Almost enough to warrant a new patch, but it's all to achieve the same goal anyway, so it makes sense to be here.

kalman:
Perf this

samples/go/csv/csv-import/perf_test.go
```diff
@@ -85,32 +98,26 @@ func (s *perfSuite) TestParseNyVehicleRegistrations() {
...
+// readGlob returns a bytes.Buffer containing the concatenation of all files
+// that match `pattern`. Large CSV files in testdata are broken up into foo.a,
+// foo.b, etc to get around GitHub file size restrictions.
+func (s *perfSuite) readGlob(pattern ...string) *bytes.Buffer {
 	assert := s.NewAssert()
+	res := &bytes.Buffer{}
 
 	s.Pause(func() {
-		glob, err := filepath.Glob(pattern)
+		glob, err := filepath.Glob(path.Join(pattern...))
```
rafael-atticlabs: mild pref for not excluding the input io time.

rafael-atticlabs:
lgtm

kalman:
FYI no longer pausing nor using buffer

PR #2466, opened 2016-08-30
Allow merge.ThreeWay() to operate across DBs

cmasone-attic:
Much like Diff, we want to be able to merge values from different
databases. Unlike Diff, Merge needs to be able to resolve Refs --
which requires a ValueReadWriter that has access to the referenced
Value. So, merge.ThreeWay() now takes a ValueReadWriter for each Value
being considered for the merge -- the parent and both descendants.

Towards #148

### DIFF
--- a/go/merge/three_way_test.go
+++ b/go/merge/three_way_test.go
@@ -19,8 +19,16 @@ type ThreeWayMergeSuite struct {
 	typeStr string
 }
 
-func (s *ThreeWayMergeSuite) tryThreeWayMerge(a, b, p, exp seq, vs types.ValueReadWriter) {
-	merged, err := ThreeWay(s.create(a), s.create(b), s.create(p), vs)
+func (s *ThreeWayMergeSuite) tryThreeWayMerge(a, b, p, exp seq, stores ...types.ValueReadWriter) {
+	var merged types.Value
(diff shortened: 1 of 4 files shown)


cmasone-attic:
PTAL

go/merge/three_way_test.go
```diff
@@ -19,16 +19,24 @@ type ThreeWayMergeSuite struct {
 	typeStr string
 }
 
-func (s *ThreeWayMergeSuite) tryThreeWayMerge(a, b, p, exp seq, vs types.ValueReadWriter) {
-	merged, err := ThreeWay(s.create(a), s.create(b), s.create(p), vs)
+func (s *ThreeWayMergeSuite) tryThreeWayMerge(a, b, p, exp seq, stores ...types.ValueReadWriter) {
```
kalman: You could make this a little bit safer if instead of
`stores ...types.ValueReadWriter)`
you did
`store0 types.ValueReaderWriter, stores ...types.ValueReaderWriter)`
if you like.

kalman:
lgtm with opt nit

cmasone-attic:
turns out this won't work in all cases :-) Going to close this out and keep the branch locally in case I circle back

PR #2469, opened 2016-08-30
Make the Go doc for go/perf/suite look idiomatic

kalman:
I hope.

We should stop using Markdown syntax in our public Go docs. godoc
doesn't recognise it (https://godoc.org/github.com/fluhus/godoc-tricks
is a reasonable summary of what *is* supported).

This patch:
 - Wraps comments at 100 characters, because godoc doesn't do wrapping.
 - Replaces triple-backtick blocks with indented blocks.
 - Removes single-backtick lines with entirely.
 - Lines things up better.
 - An unfortunate side-effect is bullet points are monospaced in
   godoc.org, but there isn't a way around this.

### DIFF
--- a/go/perf/suite/suite.go
+++ b/go/perf/suite/suite.go
@@ -2,68 +2,69 @@
 // Licensed under the Apache License, version 2.0:
 // http://www.apache.org/licenses/LICENSE-2.0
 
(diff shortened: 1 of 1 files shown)


cmasone-attic:
lgtm

PR #2471, opened 2016-08-31
Fix path in splore build script

arv:
The path to the noms sdk in the build script was not updated when the
js sdk was moved.


arv:
I found out that NomsNpmPublish has been failing since this was changed.

mikegray:
lgtm. how can we make sure we know about build failures more promptly?

arv:
Everything on Jenkins except the PR builders should be treated as a build failure.

arv:
#2473

PR #2474, opened 2016-08-31
Add Noms commit command

mikegray:
towards #2100

### DIFF
--- a/cmd/noms/noms_commit.go
+++ b/cmd/noms/noms_commit.go
@@ -1,0 +1,80 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package main
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/attic-labs/noms/cmd/util"
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/dataset"
+	"github.com/attic-labs/noms/go/spec"
+	flag "github.com/juju/gnuflag"
+)
+
+var allowDupe bool
+
+var nomsCommit = &util.Command{
+	Run:       runCommit,
+	UsageLine: "commit [options] [absolute-path] <dataset>",
+	Short:     "Commits a specified value as head of the dataset",
+	Long:      "If absolute-path is not provided, then it is read from stdin. See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the dataset and absolute-path arguments.",
+	Flags:     setupCommitFlags,
+	Nargs:     1, // if absolute-path not present we read it from stdin
+}
+
+func setupCommitFlags() *flag.FlagSet {
+	commitFlagSet := flag.NewFlagSet("commit", flag.ExitOnError)
+	commitFlagSet.BoolVar(&allowDupe, "allow-dupe", false, "creates a new commit, even if it would be identical (modulo metadata and parents) to the existing HEAD.")
+	spec.RegisterCommitMetaFlags(commitFlagSet)
+	return commitFlagSet
+}
+
+func runCommit(args []string) int {
+	ds, err := spec.GetDataset(args[len(args)-1])
+	d.CheckError(err)
+	defer ds.Database().Close()
+
+	var path string
+	if len(args) == 2 {
+		path = args[0]
+	} else {
+		readPath, _, err := bufio.NewReader(os.Stdin).ReadLine()
+		d.CheckError(err)
+		path = string(readPath)
+	}
+	absPath, err := spec.NewAbsolutePath(path)
+	d.CheckError(err)
+
+	value := absPath.Resolve(ds.Database())
+	if value == nil {
+		d.CheckErrorNoUsage(errors.New(fmt.Sprintf("Error resolving value: %s", path)))
+	}
+
+	oldCommitRef, oldCommitExists := ds.MaybeHeadRef()
+	if oldCommitExists {
+		head := ds.HeadValue()
+		if head.Hash() == value.Hash() && !allowDupe {
+			fmt.Fprintf(os.Stdout, "Commit aborted - allow-dupe is set to off and this commit would create a duplicate\n")
+			return 0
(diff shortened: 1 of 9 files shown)


mikegray:
@aboodman please provide feedback as you have time, continuing to add tests and command line arguments

cmd/noms/noms_commit.go
```diff
@@ -0,0 +1,113 @@
...
+		commitFlagSet.StringVar(&metaP, "meta-p", "", "'<key>:<path>' - creates a metadata field called 'key' set to the value at <path>")
+	*/
+	commitFlagSet.StringVar(&message, "message", "", "alias for -meta 'message:<message>'")
+	commitFlagSet.StringVar(&date, "date", "", "alias for -meta 'date:<date>'. '<date>' must be iso8601-formatted. If '<date>' is empty, it defaults to the current date.")
+	return commitFlagSet
+}
+
+func runCommit(args []string) int {
+	ds, err := spec.GetDataset(args[0])
+	d.CheckError(err)
```
aboodman: Make this a more friendly error.

aboodman: Oh nevermind, I this was d.Chk.

cmd/noms/noms_commit.go
```diff
@@ -0,0 +1,113 @@
...
+	if pathResolvedValue == nil {
+		d.CheckErrorNoUsage(fmt.Errorf("Object not found: %s", args[1]))
+	}
+
+	// Note that spec.GetPath() above gave us pathResolvedValue by calling
+	// ReadValue() already which loaded the cache for value into pathDb,
+	// but we are going to commit to ds.Database() so we need the cache for
+	// pathResolvedValue loaded into ds.Database() to perform the commit
+	// successfully
+	value := ds.Database().ReadValue(pathResolvedValue.Hash())
```
aboodman: This won't work in the general case (`pathResolvedValue.Hash()`) is not guaranteed to exist in target database.

I think you want to arrange things such that you load the path and commit to the same database instance. Probably the way to do this is to make the path argument a relative, not absolute. Then load the dataset the normal way, then parse the relative path, then resolve it against the dataset root.

aboodman: Then in the code, you'd use `spec.GetDataset()`, then use `spec.NewAbsolutePath()` on the path, then call `Resolve()` on that against `ds.Database()`. Now everything is using the exact same database instance.

mikegray: ok, thanks - will do it that way

mikegray:
@aboodman PTAL

mikegray:
`./csv-import --no-progress=1 --commit=0 --dest-type=map:0 sf-registered-businesses/2016-05-15.csv http://localhost:8000::temp | noms commit http://localhost:8000::sf-registered-businesses`

mikegray:
test this

mikegray:
Please test this

mikegray:
build this

aboodman:
@mikegray do you know why the tests are failing?

mikegray:
No, I was unable to get it to build again. I apparently do not know the magic or it is broken. I have another commit coming though anyway - bringing url-fetch, json-import, and xml-import along for the commit ride.

cmd/noms/noms_commit.go
```diff
@@ -0,0 +1,146 @@
...
+	allowDupe   bool
+	date        string
+	message     string
+	metaStrings string
+	metaPaths   string
+)
+
+var nomsCommit = &util.Command{
+	Run:       runCommit,
+	UsageLine: "commit [options] <dataset> [absolute-path]",
```
aboodman: Can `[absolute-path]` come before `<dataset>`?

cmd/noms/noms_commit.go
```diff
@@ -0,0 +1,146 @@
...
+	ds, err := spec.GetDataset(args[0])
+	d.CheckError(err)
+	defer ds.Database().Close()
+
+	var path string
+	if len(args) == 2 {
+		path = args[1]
+	} else {
+		readPath := make([]byte, 256)
+		n, err := os.Stdin.Read(readPath)
```
aboodman: It looks like the actual length limit on linux at least is 4k. Maybe use `ReadLine()` instead, or at least make this longer?

aboodman: I don't believe that this API is guaranteed to return you a line. You might get less or more than a line. You should definitely use ReadLine().

cmd/noms/noms_commit.go
```diff
@@ -0,0 +1,146 @@
...
+	absPath, err := spec.NewAbsolutePath(path)
+	d.CheckError(err)
+
+	value := absPath.Resolve(ds.Database())
+	if value == nil {
+		d.CheckErrorNoUsage(errors.New(fmt.Sprintf("Error resolving value: %s", path)))
+	}
+
+	oldCommitRef, oldCommitExists := ds.MaybeHeadRef()
+	head, ok := ds.MaybeHeadValue()
```
aboodman: If the commit exists, there will definitely be a HeadValue. So you don't need both of these.

aboodman:
Really nicely coming along.

aboodman:
BTY

cmd/noms/noms_commit.go
mikegray: good point, done.

cmd/noms/noms_commit.go
mikegray: good point, using ReadLine now - done.

cmd/noms/noms_commit.go
mikegray: sure, makes sense.

mikegray:
@aboodman the build for this PR was broken because master has testExiter removed in favor of the MustRun/Run changes, but this branch was before that. I have rebased to master to pickup those changes which also allows me to turn on some of the disabled tests where stderr was not completely read when failure tests were run. I know this creates more work for you, the reviewer in this case, but having a working build and more fully tested code is better to have for review and landing purposes. PTAL.

mikegray:
@aboodman PTAL - ping

aboodman:
@mikegray I need to run to a meeting. If you have fixed my last comments, LGTM. I will review async.

mikegray:
TBR

PR #2475, opened 2016-08-31
Add run safe method that reads stderr and stdout regardless of panic

zcstarr:
PR here rearranges Run to properly handle a run command panic to read stdout and stderr. Previously the if a Run command panic'd it would not read stderr and stdout causing subsequent tests to read the output.

@kalman thanks for pairing/helping get this sorted!


aboodman:
What was the case that it panicked during?

zcstarr:
The panic case, occurs if a function under test panics. For instance in noms log test, catching that a command exits is done via overriding exit and triggering a panic  state to catch what happened. 
https://github.com/attic-labs/noms/blob/master/cmd/noms/noms_log_test.go#L61 is an example .

aboodman:
I did something similar, but different for poke_test. I didn't know that the noms_log_test.go had already solved this problem.

Let's formalize this a bit more ... what if we add a util to go/util that is just a single function Exit(), which can be overridden in test. Then ClientTestUtils can override this to panic, and we can change ClientTestUtils.Run() to always work as RunSafe does here.

IOW, I don't think there's any reason to ever call Run() instead of RunSafe().

aboodman:
Sorry, on second thought, more detailed proposal:

1. Factor out `go/util/exit`
2. Change ClientTestSuite to override `/go/util/exit::Exit()` and turn it into a panic
3. Rename current `Run()` method to `MustRun()` (this is the naming convention that means: this will panic if something goes wrong)
4. Add the new method in this PR and call it just `Run()` (since typically in Go functions that can fail return errors, so the signature here is "more normal" than the existing one)

aboodman:
@kalman wdyt? ^

kalman:
I agree

zcstarr:
@aboodman  PTAL

samples/go/noms-merge/main_test.go
```diff
@@ -65,7 +65,7 @@ func (s *testSuite) TestWin() {
 	})
 
 	var mainErr error
-	stdout, stderr := s.Run(func() { mainErr = nomsMerge() }, []string{"--quiet=true", "--parent=" + p, s.LdbDir, l, r})
+	stdout, stderr, _ := s.Run(func() { mainErr = nomsMerge() }, []string{"--quiet=true", "--parent=" + p, s.LdbDir, l, r})
```
kalman: Either use MustRun or handle any error. Any reason why not the former?

samples/go/noms-merge/main_test.go
```diff
@@ -110,7 +110,7 @@ func (s *testSuite) TestLose() {
 
 	for _, c := range cases {
 		var mainErr error
-		stdout, _ := s.Run(func() { mainErr = nomsMerge() }, c.args)
+		stdout, _, _ := s.Run(func() { mainErr = nomsMerge() }, c.args)
```
kalman: Same here

samples/go/poke/main_test.go
```diff
@@ -47,10 +47,10 @@ func (s *testSuite) TestWin() {
 	}
 
 	for k, v := range changes {
-		stdout, stderr := s.Run(main, []string{sp, k, v})
+		stdout, stderr, err := s.Run(main, []string{sp, k, v})
 		s.Equal("", stdout)
 		s.Equal("", stderr)
-		s.Equal(0, s.ExitStatus)
+		s.Equal(nil, err)
```
kalman: You can do `s.NoError`

go/util/clienttest/client_test_suite.go
```diff
@@ -43,7 +48,15 @@ func (suite *ClientTestSuite) TearDownSuite() {
 	defer d.Chk.NoError(os.RemoveAll(suite.TempDir))
 }
 
-func (suite *ClientTestSuite) Run(m func(), args []string) (stdout string, stderr string) {
+func (suite *ClientTestSuite) MustRun(m func(), args []string) (stdout string, stderr string) {
```
kalman: Let's document this (`// MustRun runs...`)

go/util/clienttest/client_test_suite.go
```diff
@@ -43,7 +48,15 @@ func (suite *ClientTestSuite) TearDownSuite() {
...
+	var err interface{}
+	if stdout, stderr, err = suite.Run(m, args); err != nil {
+		panic(err)
+	}
+	return
+}
+
+func (suite *ClientTestSuite) Run(m func(), args []string) (stdout string, stderr string, mainErr interface{}) {
```
kalman: and this

go/util/clienttest/client_test_suite.go
```diff
@@ -23,6 +23,10 @@ type ClientTestSuite struct {
 	err        *os.File
 }
 
+type ExitError struct {
+	Code int
+}
```
kalman: If you define an `Error()` method then callers will be able to pass `ExitError` around as an `error` not just an `interface{}`. I.e. currently:
```go
var err interface{} = ExitError{1}
_, ok := err.(error) // ok is false
```
but with an Error method:
```go
var err interface{} = ExitError{1}
_, ok := err.(error) // ok is true
```

kalman:
back to you.

@aboodman what do you mean by "factor out go/util/exit"?
* the code which clients are supposed to call to Exit in `d.Exit`. Being in the `d` package is pretty weird IMO, go/util/exit would make more sense.
* the code (which @zcstarr has just written) that deals specifically with exit code errors seems like it arguably does belong in the client test suite, because exit codes are a client concept. However, coupling this to a test suite does seem bad. Having in go/util/exit seems nice.

aboodman:
This is looking great to me. I agree that the mock exit should ideally be in `go/util/error`.

I disagree that things that are 'client specific' should go in 'clients'. We're a peer-to-peer database. Everything we do is client-specific :). Less snarkily, it gets too confusing to try and figure out for every piece of code whether it will be used by "clients" or not. That is why I think all of this sort of thing should just go in `go/util` and let the linker figure it out.

Only thing IMO that should be in `samples/go` are actual sample binaries. I'd put everything else in `go/util`.

samples/go/poke/main_test.go
zcstarr: Not sure I can here, if err is nil, I can't type assert it

samples/go/noms-merge/main_test.go
zcstarr: Simply didn't want to muck around with the existing test logic too much in this commit. I think @cmasone-attic  has an issue for me to refactor test a little bit to mesh better with noms-merge

samples/go/noms-merge/main_test.go
zcstarr: ditto

go/util/clienttest/client_test_suite.go
zcstarr: good idea done.

zcstarr:
@aboodman @kalman PTAL didn't add too much, just addressed the issues in the comments

go/util/clienttest/client_test_suite.go
```diff
@@ -27,6 +27,11 @@ type ExitError struct {
 	Code int
 }
 
+func (e ExitError) Error() string {
+
```
aboodman: Remove extraneous line.

go/util/clienttest/client_test_suite.go
```diff
@@ -48,6 +53,7 @@ func (suite *ClientTestSuite) TearDownSuite() {
 	defer d.Chk.NoError(os.RemoveAll(suite.TempDir))
 }
 
+//MustRun is a wrapper around Run that will panic on Exit or Panic
```
aboodman: Add a space after `//`

aboodman: Add a period at end of sentence.

go/util/clienttest/client_test_suite.go
```diff
@@ -56,6 +62,8 @@ func (suite *ClientTestSuite) MustRun(m func(), args []string) (stdout string, s
 	return
 }
 
+//Run will execute a function passing args as commandline args, and recovers the function
```
aboodman: Spaces after '//'

go/util/clienttest/client_test_suite.go
```diff
@@ -43,7 +53,18 @@ func (suite *ClientTestSuite) TearDownSuite() {
...
+	var err interface{}
+	if stdout, stderr, err = suite.Run(m, args); err != nil {
+		panic(err)
+	}
+	return
+}
+
+//Run will execute a function passing args as commandline args, and recovers the function
+//on Panic or os.Exit()
+func (suite *ClientTestSuite) Run(m func(), args []string) (stdout string, stderr string, mainErr interface{}) {
```
aboodman: I believe you can change `mainErr` to `error` type now.

aboodman:
BTY, please try to make the return type `error`,.

aboodman:
lgtm

aboodman:
thanks!

PR #2478, opened 2016-08-31
Add stage.py for perf viewer

kalman:


kalman:
Compare to https://github.com/attic-labs/noms/blob/master/samples/js/splore/stage.py

cmasone-attic:
LGTM

PR #2479, opened 2016-08-31
Build script for building JS SDK Docs

arv:
This script npm installs documentationjs and runs it on noms.js and
puts the output in $WORKSPACE/build

It is intended to be used with Jenkins.

### DIFF
--- a/tools/build-js-documentation.py
+++ b/tools/build-js-documentation.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+
+# Copyright 2016 Attic Labs, Inc. All rights reserved.
+# Licensed under the Apache License, version 2.0:
+# http://www.apache.org/licenses/LICENSE-2.0
+
+'''
+This script builds the JS SDK documentation and puts the generated files
+in $WORKSPACE/build.
+'''
+
+import os, subprocess, json
+
+def main():
+    # Workspace is the root of the builder, e.g. "/var/lib/jenkins/workspace/builder".
+    workspace = os.getenv('WORKSPACE')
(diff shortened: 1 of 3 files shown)


tools/build-js-documentation.py
```diff
@@ -0,0 +1,43 @@
...
+    assert workspace
+
+    noms_dir = os.path.join(workspace, 'src/github.com/attic-labs/noms')
+    noms_js_dir = os.path.join(noms_dir, 'js/noms')
+    with open(os.path.join(noms_js_dir, 'package.json')) as pkg:
+        data = json.load(pkg)
+
+    version = data['version']
+
+    os.chdir(workspace)
```
kalman: noms_js_dir?

arv: Why?

kalman:
lgtm modulo 1 question

I also suggest that you install pylint and run it against all python scripts, fix any sensible errors.

tools/build-js-documentation.py
kalman: You don't want to run it relative to the js directory?

arv: To be honest it doesn't really matter. Maybe it is better to be in a dir that already have a `node_modules` in it but then there might be some conflict on the version of `documentation` we get.

kalman: Hm, I assumed given src/js already had node_modules and so forth you might want to keep everything in there. But I see it doesn't especially matter.

kalman: Conflicting comments :-) SGTM.

Issue #2358, opened 2016-08-12
`DatabaseCommon.descendsFrom` should use `Height` to exit early


rafael-atticlabs:
It was written before we had ref `Height`. Now that we do, it can exit any ancestry path search which has reached height <= `commit.Height()`.

rafael-atticlabs:
Also, now that I think about this, we should probably just move this to be a pubic static method inside of `commit.go`, like `CommitDescendsFrom(ref, ancestor types.Ref) bool`

PR #2480, opened 2016-08-31
Move database_commom.descendsFrom and prune search using Ref.Height

ehalpern:
- Move to commit.CommitDescendsFrom
- When searching for an ancestor in the commit history, prune
  whenever the commit.Height() < ancestor.Height()
- Add new test TestCommitDescendsFrom to verify correctness.
  Note this does not verify that pruning is occurring.

fixes:#2358

### DIFF
--- a/go/datas/commit.go
+++ b/go/datas/commit.go
@@ -53,6 +53,42 @@ func NewCommit(value types.Value, parents types.Set, meta types.Struct) types.St
 	return types.NewStructWithType(t, types.ValueSlice{meta, parents, value})
 }
 
+// CommitDescendsFrom returns true if commit descends from ancestor
+func CommitDescendsFrom(commit types.Struct, ancestor types.Ref, vr types.ValueReader) bool {
+	// BFS because the common case is that the ancestor is only a step or two away
+	ancestors := commit.Get(ParentsField).(types.Set)
+	for !ancestors.Has(ancestor) {
+		if ancestors.Empty() {
+			return false
+		}
+		ancestors = getAncestors(ancestors, ancestor.Height(), vr)
+	}
+	return true
+}
+
+// getAncestors returns set of direct ancestors with height >= minHeight
+func getAncestors(commits types.Set, minHeight uint64, vr types.ValueReader) types.Set {
+	ancestors := types.NewSet()
+	commits.IterAll(func(v types.Value) {
+		r := v.(types.Ref)
+		c := r.TargetValue(vr).(types.Struct)
+		// only consider commit-refs greater than minHeight; commit-refs at same height
+		// can be ignored since their parent heights will be < minHeight
+		if r.Height() > minHeight {
+			next := []types.Value{}
+			c.Get(ParentsField).(types.Set).IterAll(func(v types.Value) {
+				r := v.(types.Ref)
+				// only consider parent commit-refs >= minHeight
+				if r.Height() >= minHeight {
+					next = append(next, v)
+				}
+			})
+			ancestors = ancestors.Insert(next...)
+		}
+	})
+	return ancestors
+}
+
 func makeCommitType(valueType *types.Type, parentsValueTypes []*types.Type, metaType *types.Type, parentsMetaTypes []*types.Type) *types.Type {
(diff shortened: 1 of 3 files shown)


go/datas/commit_test.go
```diff
@@ -82,3 +84,57 @@ func TestCommitWithoutMetaField(t *testing.T) {
 	})
 	assert.False(IsCommitType(noMetaCommit.Type()))
 }
+
+func TestCommitDescendsFrom(t *testing.T) {
+	assert := assert.New(t)
+	cs := chunks.NewTestStore()
+	db := NewDatabase(cs)
+	defer cs.Close()
```
aboodman: Drive-by: do you need to close `cs` manually? I don't usually see this.

go/datas/commit_test.go
```diff
@@ -82,3 +84,57 @@ func TestCommitWithoutMetaField(t *testing.T) {
...
+		commit types.Struct
+		ref    types.Ref
+	}
+	// Add a commit and return commit and ref
+	addCommit := func(ds string, val string, parents types.Set) commitAndRef {
+		commit := NewCommit(types.String(val), parents, types.EmptyStruct)
+		db, err := db.Commit(ds, commit)
+		assert.NoError(err)
+		ref := db.HeadRef(ds)
+		return commitAndRef{commit, ref}
```
aboodman: I believe that `commit.Ref() == ref`. So there's no need to store both.

go/datas/commit.go
```diff
@@ -53,6 +53,36 @@ func NewCommit(value types.Value, parents types.Set, meta types.Struct) types.St
...
+	return true
+}
+
+// getAncestors returns set of direct ancestors with height >= minHeight
+func getAncestors(commits types.Set, minHeight uint64, vr types.ValueReader) types.Set {
+	ancestors := types.NewSet()
+	commits.IterAll(func(v types.Value) {
+		r := v.(types.Ref)
+		c := r.TargetValue(vr).(types.Struct)
+		if r.Height() >= minHeight {
```
rafael-atticlabs: i think this is >

go/datas/commit.go
```diff
@@ -53,6 +53,36 @@ func NewCommit(value types.Value, parents types.Set, meta types.Struct) types.St
...
+			next := []types.Value{}
+			c.Get(ParentsField).(types.Set).IterAll(func(v types.Value) {
+				next = append(next, v)
```
rafael-atticlabs: this also needs to check >=

go/datas/commit_test.go
ehalpern: I copied this pattern from database_test.TearDownTest. If not required, I'll remove from there as well.

go/datas/commit_test.go
```diff
@@ -82,3 +84,57 @@ func TestCommitWithoutMetaField(t *testing.T) {
 	})
 	assert.False(IsCommitType(noMetaCommit.Type()))
 }
+
+func TestCommitDescendsFrom(t *testing.T) {
+	assert := assert.New(t)
```
rafael-atticlabs: high-level issue: I think this test would be clearer if it first setup a DAG of commits which collected local variables named `a`, `b`, `c`, etc... (each of these would be a `Ref<Commit>`) and some ascii-art in a comment which depicts the graph, something like: 

https://github.com/attic-labs/noms/blob/master/go/datas/pull_test.go#L168

Then at the bottom a bunch of assertions for the set of cases you need to test, e.g.

`assert.True(descendsFrom(a, b))`
`assert.False(descendsFrom(c, e))`

etc...

willhite:
test this please

go/datas/commit_test.go
ehalpern: PTAL @rafael-atticlabs. I've restructured the tests in the spirit of your comment.

ehalpern:
@rafael-atticlabs: PYAL to confirm your concerns are addressed. The only change since you reviewed over my shoulder are improvements to the tests.

rafael-atticlabs:
awesome. lgtm. the test is really beautiful. sorry for the delay.

ehalpern:
Thanks for your very helpful feedback.

PR #2481, opened 2016-08-31
merge commit parent type issue

cmasone-attic:
NOT FOR LANDING

### DIFF
--- a/go/datas/commit_test.go
+++ b/go/datas/commit_test.go
@@ -65,6 +65,20 @@ func TestNewCommit(t *testing.T) {
 		types.StringType,
 	})
 	assertTypeEquals(et4, at4)
+
+	// Merge-commit with different parent types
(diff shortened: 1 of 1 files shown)


cmasone-attic:
Here's what I get when I try to run this. It seems like something surprising to me is happening when NewCommit() tries to calculate the appropriate type for the Parent field.

panic:                         	Error Trace:	assert_type.go:11
				struct.go:53
				commit.go:53
				commit_test.go:70
		Error:		Invalid type
		Messages:	Set<Ref<struct Commit {
	  meta: struct {},
	  parents: Set<Ref<Cycle<0>>>,
	  value: Number,
	}> | Ref<struct Commit {
	  meta: struct {},
	  parents: Set<Ref<struct Commit {
	    meta: struct {},
	    parents: Set<Ref<Cycle<0>>>,
	    value: Number | String,
	  }>>,
	  value: String,
	}>> is not a subtype of Set<Ref<struct Commit {
	  meta: struct {},
	  parents: Set<Ref<Cycle<0>>>,
	  value: Number | String,
	}>>

PR #2482, opened 2016-08-31
Add node to the PATH for documentation build

arv:


tools/build-js-documentation.py
```diff
@@ -37,7 +53,9 @@ def main():
         '--output', os.path.join(workspace, 'build', version),
         '--project-version', version
     ]
-    subprocess.check_call(cmd)
+    proc = subprocess.Popen(cmd, env=env)
```
kalman: you could factor this out into a method like `call_with_env(cmd)`

tools/build-js-documentation.py
```diff
@@ -9,22 +9,38 @@
...
+import os
+import subprocess
 
 def main():
     # Workspace is the root of the builder, e.g. "/var/lib/jenkins/workspace/builder".
     workspace = os.getenv('WORKSPACE')
     assert workspace
 
+    # Directory where node/npm binaries have been installed.
+    node_bin = '/var/lib/jenkins/node-v5.11.1-linux-x64/bin'
```
kalman: Every time I see this hardcoded I wonder if we're just creating work for ourselves when we update to node v6. Which, we should do :-\

kalman:
lgtm

tools/build-js-documentation.py
arv: I don't understand why we are not adding a symlink in `/usr/local/node/bin` like we have done for go?

kalman: That sounds like a good idea to me...

Issue #1471, opened 2016-05-10
Auto-generate JS doc


aboodman:
In Go, we get Godoc for free. In JavaScript/Node I'm sure there are tools for this -- probably tons of them, all slightly broken and different.

mikegray:
Once repo became public godoc works - https://godoc.org/github.com/attic-labs/noms. We still have work to do in order to make the content better - and of course we need a JS equivalent.

aboodman:
I don't think we need a bug for "make content better" . That can just be
ongoing task. I'm going to retarget this bug to just JS.

arv:
Some update on the JS front. I've been contributing to [documentation.js](http://documentation.js.org/) and there is one important feature missing. I have a PR for that: https://github.com/documentationjs/documentation/pull/490. With that done we have "a complete" documentation.

We then need to setup a build that generates the HTML and deploy that somewhere.

After that I think we need to tweak the generated HTML quite a bit.

arv:
The documentation.js tool is "good enough" to move on.

For deployment I could set up another ec2 instance or maybe we should just store these in s3 (since these are just static files).

@cmasone-attic @zcstarr @mikegray What do you think?

mikegray:
Probably makes sense to take advantage of AWS CloudFront in front of S3 to serve these if they are all static.

cmasone-attic:
agreed

zcstarr:
agreed as well , I think @mike is right s3 + aws cloudfront sounds smooth

arv:
We now automatically deploy to https://docs.noms.io/js/

I'm closing this. Lets open more specific bugs for things that needs to be improved.

rafael-atticlabs:
So much awesome!

PR #2486, opened 2016-08-31
Drive js docs from npm script

arv:
Turns out that documentationjs reads the babelrc file from js/noms
which leads to us having to do npm install in js/noms. Therefore
putting the dev deps in there and running the script from there
makes things cleaner.

It also allows doing:

  npm run build-docs

Towards #1471

### DIFF
--- a/tools/build-js-documentation.py
+++ b/tools/build-js-documentation.py
@@ -30,33 +30,16 @@ def main():
 
     noms_dir = os.path.join(workspace, 'src/github.com/attic-labs/noms')
     noms_js_dir = os.path.join(noms_dir, 'js/noms')
-    with open(os.path.join(noms_js_dir, 'package.json')) as pkg:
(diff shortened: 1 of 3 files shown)


arv:
@kalman

tools/build-js-documentation.py
```diff
@@ -30,33 +30,16 @@ def main():
...
-        '--name', 'Noms',
-        '--document-exported',
-        '--infer-private', '^_',
-        '--github',
-        '--format', 'html',
-        '--output', os.path.join(workspace, 'build', version),
-        '--project-version', version
-    ]
-    call_with_env(cmd, env)
+    os.chdir(noms_js_dir)
```
kalman: You could consider using `pushd` when my PR lands

kalman:
nice, lgtm.

PR #2487, opened 2016-08-31
Add perf build.py and util

kalman:


kalman:
Hopefully this is the last one.

js/perf/viewer/build.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/python
```
arv: Who is calling this file?

arv:
BTY

js/perf/viewer/build.py
kalman: tools/run_all_build.go

arv:
LGTM

PR #2488, opened 2016-08-31
Add a super quick indeterminite progress meter to noms-merge

aboodman:


cmasone-attic:
This breaks tests because it puts stuff on stdout when the test isn't expecting :-/

Since this isn't really a permanent thing, I'm fine with any number of kinda-hacky ways to silence the output during testing. You could make the updateProgress closure a thing that's passed in, or have merge.ThreeWay() take some kind of progress channel...whatever you want.

aboodman:
Shoot, forgot to run tests. I'll add a --silent flag.

aboodman:
ptal

go/merge/three_way.go
```diff
@@ -63,7 +63,15 @@ func unmergeable(a, b types.Value) bool {
...
-func threeWayMerge(a, b, parent types.Value, vwr types.ValueReadWriter) (merged types.Value, err error) {
+func updateProgress(progress chan struct{}) {
+	// TODO: Eventually we'll want more information than a single bit :).
+	if progress != nil {
+		progress <- struct{}{}
+	}
+}
+
+func threeWayMerge(a, b, parent types.Value, vwr types.ValueReadWriter, progress chan struct{}) (merged types.Value, err error) {
+	defer updateProgress(progress)
```
cmasone-attic: Since you're calling this down in all the `apply()` closures, I don't think it's accurate to also call it here. To a first approximation, you want to do this every time the code calls Set or Remove or similar on `merged`, right? That doesn't happen here in this function

cmasone-attic:
lgtm with nit

PR #2489, opened 2016-09-01
npm run did not work

arv:


arv:
For some reason, the second time call_with_env, the call is not passing the whole array. Do you have any ideas?

```sh
npm install
npm  # oops
```

arv:
TBR

PR #2493, opened 2016-09-01
JS Doc builder

arv:


arv:
I have no idea why `npm install` is called but `npm run build-docs` is not. I'm adding some printing and whatnot.

arv:
TBR

PR #2496, opened 2016-09-01
Add prepublish step to perf/viewer so that out.js gets built

kalman:


kalman:
This is exactly how Splore builds its artifacts, so tbr.

arv:
LGTM

PR #2497, opened 2016-09-01
Print skipped perf tests in verbose mode

kalman:


kalman:
Chatting to you made me decide this is a good idea, though it's not relatedd.

willhite:
LGTM

PR #2498, opened 2016-09-01
Fix test assertion in noms-merge::TestLose()

cmasone-attic:
This was caused by me changing an error string and failing
to update the test.


cmasone-attic:
TBR

PR #2504, opened 2016-09-02
Make poke hang on to parent's Commit metadata

cmasone-attic:
Before this, poke would drop any commit metadata from the dataset being modified. Now, it just pulls it forward.

### DIFF
--- a/samples/go/poke/main.go
+++ b/samples/go/poke/main.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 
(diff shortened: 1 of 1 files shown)


cmasone-attic:
PTAL

aboodman:
LGTM - let's check it in

cmasone-attic:
@arv fixed the type issue that was preventing some merge commits from working right yesterday, so we can revert this if you'd like. Up to you :-)

Issue #2445, opened 2016-08-26
idea: `merge` command


aboodman:
It would be cool to be able to interact with merge a bit to be able to figure out how to work it into our command set. Also, it would make a great demo. To that end, I propose:

```
noms-merge <left> <right> <ds-out>
```

This wouldn't be part of the `noms` binary (yet), just a standalone sample program. It merges the values at `left` and `right` using our currently in-progress merge policies and writes the result to `ds-out`.

In the case of conflict, noms-merge enters an interactive mode that looks like:

```
!? noms-merge http://localhost:8000::foo.value http://localhost:8000::bar.value http://localhost:8000::baz
merging....

Conflict at: .value[42].firstName
Left: "Samuel"
Right: "Simon"
Enter 'l' to accept the left value, 'r' to accept the right value, or 'c <val>' to enter a custom resolution.
```

rafael-atticlabs:
closing in favor of #2535 which more accurately describes the work to be done.

PR #2505, opened 2016-09-02
Update merge.ThreeWay() to allow very basic custom conflict resolution

cmasone-attic:
This patch modifies merge.ThreeWay() to take a callback that allows
for custom conflict resolution. The noms-merge command-line tool uses
this to inject a callback that accepts input from the console
dictating whether to accept the value from the 'left' or 'right' merge
candidates.

Toward #2445

### DIFF
--- a/go/merge/candidate.go
+++ b/go/merge/candidate.go
@@ -1,0 +1,99 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package merge
+
+import (
+	"fmt"
+
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// candidate represents a collection that is a candidate to be merged. This interface exists to wrap Maps, Sets and Structs with a common API so that threeWayOrderedSequenceMerge() can remain agnostic to which kind of collections it's actually working with.
+type candidate interface {
+	diff(parent candidate, change chan<- types.ValueChanged, stop <-chan struct{})
+	get(k types.Value) types.Value
+	pathConcat(change types.ValueChanged, path types.Path) (out types.Path)
+	toValue() types.Value
+}
+
+type mapCandidate struct {
+	m types.Map
+}
+
+func (mc mapCandidate) diff(p candidate, change chan<- types.ValueChanged, stop <-chan struct{}) {
+	mc.m.Diff(p.(mapCandidate).m, change, stop)
+}
+
+func (mc mapCandidate) get(k types.Value) types.Value {
+	return mc.m.Get(k)
+}
+
+func (mc mapCandidate) pathConcat(change types.ValueChanged, path types.Path) (out types.Path) {
+	out = append(out, path...)
+	if kind := change.V.Type().Kind(); kind == types.BoolKind || kind == types.StringKind || kind == types.NumberKind {
+		out = append(out, types.NewIndexPath(change.V))
+	} else {
+		out = append(out, types.NewHashIndexPath(change.V.Hash()))
+	}
+	return
+}
+
+func (mc mapCandidate) toValue() types.Value {
+	return mc.m
+}
+
+type setCandidate struct {
+	s types.Set
+}
+
+func (sc setCandidate) diff(p candidate, change chan<- types.ValueChanged, stop <-chan struct{}) {
+	sc.s.Diff(p.(setCandidate).s, change, stop)
+}
+
+func (sc setCandidate) get(k types.Value) types.Value {
+	return k
+}
+
+func (sc setCandidate) pathConcat(change types.ValueChanged, path types.Path) (out types.Path) {
+	out = append(out, path...)
+	if kind := change.V.Type().Kind(); kind == types.BoolKind || kind == types.StringKind || kind == types.NumberKind {
+		out = append(out, types.NewIndexPath(change.V))
+	} else {
+		out = append(out, types.NewHashIndexPath(change.V.Hash()))
+	}
+	return
+}
+
+func (sc setCandidate) toValue() types.Value {
+	return sc.s
+}
+
+type structCandidate struct {
+	s types.Struct
+}
+
+func (sc structCandidate) diff(p candidate, change chan<- types.ValueChanged, stop <-chan struct{}) {
+	sc.s.Diff(p.(structCandidate).s, change, stop)
+}
+
+func (sc structCandidate) get(key types.Value) types.Value {
+	if field, ok := key.(types.String); ok {
+		val, _ := sc.s.MaybeGet(string(field))
+		return val
+	}
+	panic(fmt.Errorf("Bad key type in diff: %s", key.Type().Describe()))
+}
+
+func (sc structCandidate) pathConcat(change types.ValueChanged, path types.Path) (out types.Path) {
+	out = append(out, path...)
+	str, ok := change.V.(types.String)
+	d.PanicIfTrue(!ok, "Field names must be strings, not %s", change.V.Type().Describe())
+	return append(out, types.NewFieldPath(string(str)))
+}
+
+func (sc structCandidate) toValue() types.Value {
+	return sc.s
+}
--- a/go/merge/candidate.go
+++ b/go/merge/candidate.go
@@ -11,12 +11,15 @@ import (
 	"github.com/attic-labs/noms/go/types"
 )
 
-// candidate represents a collection that is a candidate to be merged. This interface exists to wrap Maps, Sets and Structs with a common API so that threeWayOrderedSequenceMerge() can remain agnostic to which kind of collections it's actually working with.
+// candidate represents a collection that is a candidate to be merged. This
+// interface exists to wrap Maps, Sets and Structs with a common API so that
+// threeWayOrderedSequenceMerge() can remain agnostic to which kind of
+// collections it's actually working with.
 type candidate interface {
 	diff(parent candidate, change chan<- types.ValueChanged, stop <-chan struct{})
 	get(k types.Value) types.Value
 	pathConcat(change types.ValueChanged, path types.Path) (out types.Path)
-	toValue() types.Value
+	getValue() types.Value
 }
 
 type mapCandidate struct {
@@ -41,7 +44,7 @@ func (mc mapCandidate) pathConcat(change types.ValueChanged, path types.Path) (o
 	return
 }
 
-func (mc mapCandidate) toValue() types.Value {
+func (mc mapCandidate) getValue() types.Value {
 	return mc.m
 }
 
@@ -67,7 +70,7 @@ func (sc setCandidate) pathConcat(change types.ValueChanged, path types.Path) (o
 	return
 }
 
-func (sc setCandidate) toValue() types.Value {
+func (sc setCandidate) getValue() types.Value {
 	return sc.s
 }
 
@@ -94,6 +97,6 @@ func (sc structCandidate) pathConcat(change types.ValueChanged, path types.Path)
 	return append(out, types.NewFieldPath(string(str)))
 }
 
-func (sc structCandidate) toValue() types.Value {
+func (sc structCandidate) getValue() types.Value {
 	return sc.s
 }
--- a/go/merge/three_way.go
+++ b/go/merge/three_way.go
@@ -11,9 +11,19 @@ import (
 	"github.com/attic-labs/noms/go/types"
 )
 
-type ResolveFunc func(aChange, bChange types.ValueChanged, a, b types.Value, path types.Path) (change types.ValueChanged, merged types.Value, ok bool)
+// ResolveFunc is the type for custom merge-conflict resolution callbacks.
+// When the merge algorithm encounters two non-mergeable changes (aChange and
+// bChange) at the same path, it calls the ResolveFunc passed into ThreeWay().
+// The callback gets the types of the two incompatible changes (added, changed
+// or removed) and the two Values that could not be merged (if any). If the
+// ResolveFunc cannot device a resolution, ok should be false upon return and
+// the other return values are undefined. If the conflict can be resolved, the
+// function should return the appropriate type of change to apply, the new value
+// to be used (if any), and true.
+type ResolveFunc func(aChange, bChange types.DiffChangeType, a, b types.Value, path types.Path) (change types.DiffChangeType, merged types.Value, ok bool)
 
-// ErrMergeConflict indicates that a merge attempt failed and must be resolved manually for the provided reason.
+// ErrMergeConflict indicates that a merge attempt failed and must be resolved
+// manually for the provided reason.
 type ErrMergeConflict struct {
 	msg string
 }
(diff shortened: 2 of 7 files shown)


cmasone-attic:
Off I go to work on smush-merge :-)

You can review this now, if you want

samples/go/noms-merge/main.go
```diff
@@ -91,6 +95,37 @@ func nomsMerge() error {
...
+	// TODO: Handle removes as well.
+	fmt.Printf("\nConflict at: %s\n", path.String())
+	fmt.Printf("Left:  %s\nRight: %s\n\n", left, right)
+	fmt.Println("Enter 'l' to accept the left value, 'r' to accept the right value")
+	var choice rune
+	for {
+		_, err := fmt.Scanf("%c", &choice)
+		d.PanicIfError(err)
+		switch choice {
+		case 'l', 'L':
```
aboodman: Why don't I see this in red anywhere? The old version had it already.

cmasone-attic: The old version definitely didn't have this code

cmasone-attic:
alright! We now have 'mash'. It concatenates strings and then randomizes them. PTAL

cmasone-attic:
@aboodman Do we want to review and land this? Should I ask @kalman to review since he's more familiar with this code?

aboodman:
oh sorry, yeah. Seems reasonable to land.

cmasone-attic:
You want to review it, or redirect to ben?

aboodman:
ben please

samples/go/noms-merge/main.go
```diff
@@ -91,6 +98,60 @@ func nomsMerge() error {
...
+			case types.String:
+				mashed := mash(a, b.(types.String))
+				fmt.Println("Replacing with", mashed)
+				return aChange, mashed, true
+			}
+		}
+	}
+}
+
+func mash(a, b types.String) types.String {
```
kalman: wat?

samples/go/noms-merge/main.go
```diff
@@ -91,6 +98,60 @@ func nomsMerge() error {
 	}))
 }
 
+func resolve(aChange, bChange types.ValueChanged, a, b types.Value, path types.Path) (change types.ValueChanged, merged types.Value, ok bool) {
```
kalman: This needs to be tested. You may want to drive the majority of the merge logic outside of the stdio loop, like pass the choices over a channel.

go/merge/three_way.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/attic-labs/noms/go/types"
 )
 
+type ResolveFunc func(aChange, bChange types.ValueChanged, a, b types.Value, path types.Path) (change types.ValueChanged, merged types.Value, ok bool)
```
kalman: needs documentation, especially for me

cmasone-attic:
ok, @kalman if you could look at this, I'd appreciate. The biggest change is that now there's a `merger` struct (which is a crap name) that represents a given merge attempt...sorta. There really isn't a lot of state to keep, other than the ValueReadWriter and the new conflict resolution callback. There will probably be more, though, so it seems like having a struct is The Right Thing.

The other big thing is that, rather than passing around `get`, `diff`, and (now) Path-concatenating closures for `a`, `b`, and `parent`, there's now a little interface that I can wrap Maps, Structs and Sets in.

go/merge/three_way.go
```diff
@@ -200,7 +190,81 @@ func threeWayStructMerge(a, b, parent types.Struct, vwr types.ValueReadWriter, p
 		}
 		panic(fmt.Errorf("Bad key type in diff: %s", change.V.Type().Describe()))
 	}
-	return threeWayOrderedSequenceMerge(parent, aDiff, bDiff, makeGetFunc(a), makeGetFunc(b), makeGetFunc(parent), apply, vwr, progress)
+	return m.threeWayOrderedSequenceMerge(structCandidate{a}, structCandidate{b}, structCandidate{parent}, apply, path)
+}
+
+type candidate interface {
```
kalman: I've spent some time trying to understand this change (with the mapCandidate etc). why did you need it?

kalman:
let's start with these comments to help me get started.

samples/go/noms-merge/main.go
cmasone-attic: @aboodman, @rafael-atticlabs 
Are we really planning to keep this UX around? My understanding was that it was mostly for a demo, so I'm kinda loathe to refactor and add tests for it if we really don't care.

go/merge/three_way.go
cmasone-attic: I was explaining while you were reviewing!

The 'candidate' stuff was because the alternative was passing something like 7 closures to `threeWayOrderedSequenceMerge()`: one getFunc for each of a, b, and parent; one diffFunc for each of a and b; and one path-concatenation function for each of a and b.

That seemed absurd, so instead I made a little interface that Maps, Sets and Structs could all be adapted to and made the function take those instead. mapCandidate, setCandidate, and structCandidate are the adaptor implementations.

go/merge/three_way.go
cmasone-attic: // ResolveFunc is the type of a function which takes two conflicting changes and tries to resolve them. The two changes occured at `path`, one resulting in the Value `a` and the other in the Value `b`. If a resolution cannot be devised, `ok` should be false upon return and the other return values are undefined. If the conflict can be resolved, the function should return the appropriate types.ValueChanged to apply, the new value to be used when applying the change, and `true`.

(I will put this in the source too, obviously, but writing it here to hopefully explain)

go/merge/three_way.go
kalman: I see. Yes that would get pretty noisy, though I also find the adaptors quite distracting. What about another entirely different approach: if merger had a map of Kind --> getter/diff functions instead? (in fact it could be global)

samples/go/noms-merge/main.go
cmasone-attic: Aaron specifically asked for this to show that a conflict could be resolved using something other than the previously-existing value in one descendant or the other

samples/go/noms-merge/main.go
```diff
@@ -91,6 +98,60 @@ func nomsMerge() error {
...
+			return aChange, a, true
+		case 'r', 'R':
+			return bChange, b, true
+		case 'm', 'M':
+			if !a.Type().Equals(b.Type()) {
+				fmt.Printf("Sorry, can't smush a %s with a %s\n", a.Type().Describe(), b.Type().Describe())
+			}
+			switch a := a.(type) {
+			case types.Bool:
+				return aChange, types.Bool(bool(a) || bool(b.(types.Bool))), true
```
kalman: If I'm understanding correctly: is the use of aChange here even correct? If it's the change to apply, and we have a=false and b=true, shouldn't it return bChange?

Whether or not this code becomes relevant, i.e. if this stays around, it does raise the question whether the merge function should only return the change to apply, not the final value, because you can derive the final value by the change to apply. And, what if they disagree, as I mentioned with aChange?

Or alternatively don't return the change, only the final value, which would be simpler... but perhaps not enough information?

samples/go/noms-merge/main.go
kalman: Ok let's see what the decision is re keeping this code.

samples/go/noms-merge/main.go
cmasone-attic: The problem is that a `types.ValueChanged` doesn't contain the "changed value" for anything except a Set :-/ For maps, it's the key at which the described change happened, and for structs it's the field name. And, in those cases, the Value in the two ValueChanged structs is equal -- it has to be or we wouldn't be in conflict.

So, the value in aChange and bChange is either the same or it _should be_ ignored. I see now that it's not in the version of `apply()` used for Sets :-)

kalman: aha, I see.

I think the proposed comment on the ResolveFunc documentation needs to be clearer then. And in fact since the value is always going to be the same in the ValueChanged struct, perhaps return a ChangeType instead?

go/merge/three_way.go
cmasone-attic: My gut reaction is that that will result is lots of type-switching, because `a`, `b`, and `parent` will come in as a types.Value. I feel like that's lame, because type-appropriate implementations of the same method is like...literally the thing you're supposed to use interfaces for.

What is it that's distracting? Is it that the definitions are all kind of in-line here? Is it the apply functions having to sorta un-candidate them? In my mind, the stuff that's complicated is all over in `threeWayOrderedSequenceMerge()`, and this use of adaptors doesn't change anything over there. This is really just shoehorning three different things into a similar enough shape that we can run them all through a single, common, complex code path.

samples/go/noms-merge/main.go
cmasone-attic: Can you propose a comment for ResolveFunc? The comment above _was_ my stab at explaining clearly and pithily and it apparently did not succeed

go/merge/three_way.go
kalman: > What is it that's distracting?

A few things, not necessarily in any order:
* PR diff is complicated
* figuring out whether these new collection types introduce any behavioral changes
* a fair amount of non-dense extra code
* boxing/unboxing is more code
* this entire algorithm is merging types.Collections and that's less obvious now
* I just find it less direct

I hear you on the type switching but I was more thinking like:

```go
differs := map[Kind]diffFunc{
  SetKind: func(v1, v2 types.Value, change chan<- types.ValueChanged, stop <-chan struct{}) types.Value {
    return v1.(Set).Diff(v2.(Set), change, stop)
  },
  MapKind: func(v1, v2 types.Value, change chan<- types.ValueChanged, stop <-chan struct{}) types.Value {
    return v1.(Map).Diff(v2.(Map), change, stop)
  },
  StructKind: func(v1, v2 types.Value, change chan<- types.ValueChanged, stop <-chan struct{}) types.Value {
    return v1.(Struct).Diff(v2.(Struct), change, stop)
  },
}
getters := map[Kind]getFunc{
  SetKind: func(v, k types.Value) types.Value {
    return k
  },
  MapKind: func(v, k types.Value) types.Value {
    return v.(Map).Get(k)
  },
  StructKind: func(v, k types.Value) types.Value {
    return v.(Struct).Get(string(k.(String)))
  },
}
pathConcat := map[Kind]pathConcatFunc { /* etc */ }
// then instead of aValue, bValue := a.get(aChange.V), b.get(bChange.V) have:
get := getters[a.Kind]
aValue, bValue := get(aChange.V), get(bChange.V)
```
now that I've typed that out I'm not sure it's any better. Is there a way to make the old callback approach more manageable? I can think of a few other ways to cut it.

samples/go/noms-merge/main.go
kalman: Like, make it clear that value is relative to the key that was changed. I read it as an entirely new value.

Also see change type? does that make sense? Because the key is going to be constant, seems moot.

cmasone-attic: I've got a hopefully-better comment in my latest patch

samples/go/noms-merge/main.go
cmasone-attic: I've refactored this a little, passing in an io.Reader and io.Writer for i/o. That makes it testable without having to involve multiple goroutines and channels just for the sake of testing

cmasone-attic:
Ok, PTAL. Lots of changes here. ResolveFunc is a bit different, the CLI resolve stuff has been tested, and I moved around the 'candidate' stuff to hopefully make the diff and the code more clear

go/merge/candidate.go
```diff
@@ -0,0 +1,99 @@
...
+package merge
+
+import (
+	"fmt"
+
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// candidate represents a collection that is a candidate to be merged. This interface exists to wrap Maps, Sets and Structs with a common API so that threeWayOrderedSequenceMerge() can remain agnostic to which kind of collections it's actually working with.
```
kalman: Might as well wrap, even though this isn't public.

go/merge/candidate.go
```diff
@@ -0,0 +1,99 @@
...
+type candidate interface {
```
kalman: Thanks, this is much nicer.

go/merge/candidate.go
```diff
@@ -0,0 +1,99 @@
...
+	diff(parent candidate, change chan<- types.ValueChanged, stop <-chan struct{})
+	get(k types.Value) types.Value
+	pathConcat(change types.ValueChanged, path types.Path) (out types.Path)
+	toValue() types.Value
```
kalman: nit: getValue()?

go/merge/three_way.go
```diff
@@ -11,7 +11,19 @@ import (
 	"github.com/attic-labs/noms/go/types"
 )
 
-// ErrMergeConflict indicates that a merge attempt failed and must be resolved manually for the provided reason.
+// ResolveFunc is the type for custom merge-conflict resolution callbacks.
+// When the merge algorithm encounters two non-mergeable changes (aChange and
+// bChange) at the same path, it calls the ResolveFunc passed into ThreeWay().
+// The callback gets the types of the two incompatible changes (added, changed
+// or removed) and the two Values that could not be merged (if any). If the
+// ResolveFunc cannot device a resolution, ok should be false upon return and
```
kalman: cannot... _devise_?

go/merge/three_way.go
```diff
@@ -11,7 +11,19 @@ import (
 	"github.com/attic-labs/noms/go/types"
 )
 
-// ErrMergeConflict indicates that a merge attempt failed and must be resolved manually for the provided reason.
+// ResolveFunc is the type for custom merge-conflict resolution callbacks.
+// When the merge algorithm encounters two non-mergeable changes (aChange and
+// bChange) at the same path, it calls the ResolveFunc passed into ThreeWay().
+// The callback gets the types of the two incompatible changes (added, changed
+// or removed) and the two Values that could not be merged (if any). If the
```
kalman: Would it be correct to say "two keys" rather than "two Values"? Because this Value is different to the one referenced below "the new value to be used (if any)" and it would be helpful to disambiguate.

And if so - perhaps instead of saying "the new value to be used (if any)" say "the new value for that key (if any)"?

go/merge/three_way.go
```diff
@@ -63,23 +82,26 @@ func unmergeable(a, b types.Value) bool {
 	return true
 }
 
-func updateProgress(progress chan struct{}) {
+type merger struct {
+	vwr      types.ValueReadWriter
```
kalman: nit: vrw not vwr?

samples/go/noms-merge/main.go
```diff
@@ -91,6 +99,52 @@ func nomsMerge() error {
...
+				return change, merged, false
+			}
+			switch a := a.(type) {
+			case types.Bool:
+				return aType, types.Bool(bool(a) || bool(b.(types.Bool))), true
+			case types.Number:
+				return aType, types.Number(float64(a) + float64(b.(types.Number))), true
+			case types.String:
+				concatenated := string(a) + string(b.(types.String))
+				fmt.Fprintln(out, "Replacing with", concatenated)
```
kalman: Should this "replacing with" output be shown for bool/number as well?

kalman:
lgtm

go/merge/candidate.go
cmasone-attic: done

go/merge/candidate.go
cmasone-attic: I'm glad it worked out :-)

go/merge/three_way.go
cmasone-attic: whoops!

go/merge/candidate.go
cmasone-attic: Yep, that's more accurate.

go/merge/three_way.go
cmasone-attic: But it's actually the two Values that couldn't be merged; if the code is trying to merge two maps {foo: bar} and {foo: baz}, `a` will be "bar" and `b` wil be "baz". And the thing that's being returned is the new value to use for 'foo', in the above example.

go/merge/three_way.go
cmasone-attic: ...I may have made this mistake in many, many places. Accursèd mild dyslexia!

samples/go/noms-merge/main.go
cmasone-attic: Done

go/merge/three_way.go
kalman: Ah, got it. Never mind then.

PR #2506, opened 2016-09-02
Fix subtype issue with unions as elem types

arv:
For compound types (List, Set, Map, Ref) the concrete type may be a
union. If that is the case all the types in the union must be a
subtype of the concrete type's element type.

`C<T | V>` is a subtype of `C<S>` iff T is a subtype of S and V is a
subtype of S.


cmasone-attic:
LGTM

rafael-atticlabs:
js?

arv:
Never heard of it... will fix.

PR #2508, opened 2016-09-02
Disable perf suite check for nomsRevision

willhite:
This needs to be disabled so that test cases succeed when
building demo-server. Currently this fails because we build
with CodePipeline so the source code is not in a git repo
when getting built.


kalman:
lgtm

PR #2510, opened 2016-09-02
Fix import of noms.staging in stage.py

arv:
Also fix unit test that was not updated when the functions were
renamed.


kalman:
oops. LGTM.

PR #2512, opened 2016-09-02
Go: Add List.IteratorAt() and use in list merging

cmasone-attic:
When merge.ThreeWay() merges list splices, it frequently needs to
extract a slice of a List or check whether a given range of values is
exactly equal in two different Lists. Repeatedly Getting elements from
a List is expensive, because it creates a new cursor internally every
time. Adding IteratorAt(idx uint64) allows code to iterate over a
range of a List while only creating a cursor once. This allows a slice
of Values to be extracted or used in comparisons efficiently.

Toward #2445

### DIFF
--- a/go/merge/three_way_list.go
+++ b/go/merge/three_way_list.go
@@ -7,6 +7,7 @@ package merge
 import (
 	"fmt"
 
+	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/types"
 )
 
@@ -94,9 +95,10 @@ func canMerge(a, b types.List, aSplice, bSplice types.Splice) bool {
 	if aSplice != bSplice {
 		return false
 	}
-	// TODO: Add List.IterFrom() and use that to compare added values.
-	for i := uint64(0); i < aSplice.SpAdded; i++ {
-		if !a.Get(aSplice.SpFrom + i).Equals(b.Get(bSplice.SpFrom + i)) {
+	aIter, bIter := a.IteratorAt(aSplice.SpFrom), b.IteratorAt(bSplice.SpFrom)
+	for count := uint64(0); count < aSplice.SpAdded; count++ {
+		aVal, bVal := aIter.Next(), bIter.Next()
+		if aVal == nil || bVal == nil || !aVal.Equals(bVal) {
 			return false
 		}
 	}
@@ -109,10 +111,12 @@ func merge(s1, s2 types.Splice) types.Splice {
 }
 
 func apply(source, target types.List, offset uint64, s types.Splice) types.List {
-	// TODO: Add List.IterFrom() and use that to build up toAdd.
 	toAdd := make(types.ValueSlice, s.SpAdded)
-	for i := range toAdd {
-		toAdd[i] = source.Get(s.SpFrom + uint64(i))
+	iter := source.IteratorAt(s.SpFrom)
+	for i := 0; uint64(i) < s.SpAdded; i++ {
+		v := iter.Next()
+		d.PanicIfTrue(v == nil, "List diff returned a splice that inserts a nonexistent element.")
+		toAdd[i] = v
 	}
 	return target.Splice(s.SpAt+offset, s.SpRemoved, toAdd...)
 }
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -185,6 +185,14 @@ func (l List) IterAll(f listIterAllFunc) {
 	})
 }
 
+func (l List) Iterator() ListIterator {
+	return l.IteratorAt(0)
+}
+
+func (l List) IteratorAt(index uint64) ListIterator {
+	return ListIterator{newCursorAtIndex(l.seq, index)}
+}
+
 func (l List) Diff(last List, changes chan<- Splice, closeChan <-chan struct{}) {
 	l.DiffWithLimit(last, changes, closeChan, DEFAULT_MAX_SPLICE_MATRIX_SIZE)
 }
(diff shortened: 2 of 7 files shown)


cmasone-attic:
PTAL

go/merge/three_way_list.go
```diff
@@ -94,13 +94,26 @@ func canMerge(a, b types.List, aSplice, bSplice types.Splice) bool {
...
 	}
-	// TODO: Add List.IterFrom() and use that to compare added values.
-	for i := uint64(0); i < aSplice.SpAdded; i++ {
-		if !a.Get(aSplice.SpFrom + i).Equals(b.Get(bSplice.SpFrom + i)) {
-			return false
-		}
+	aChan, bChan := make(chan types.Value), make(chan types.Value)
+	iterFrom := func(l types.List, s types.Splice, c chan<- types.Value) {
+		l.IterFrom(s.SpFrom, func(v types.Value, idx uint64) (stop bool) {
+			if i := idx - s.SpFrom; i < s.SpAdded {
```
kalman: just `idx - s.spFrom < s.SpAdded`?

go/merge/three_way_list.go
```diff
@@ -94,13 +94,26 @@ func canMerge(a, b types.List, aSplice, bSplice types.Splice) bool {
...
+		})
+		close(c)
+	}
+	go iterFrom(a, aSplice, aChan)
+	go iterFrom(b, bSplice, bChan)
+
+	mergeable := true
+	for aVal := range aChan {
+		bVal := <-bChan
+		mergeable = mergeable && aVal.Equals(bVal)
```
kalman: Should you break if this is false?

go/types/list.go
```diff
@@ -162,24 +162,30 @@ func (l List) RemoveAt(idx uint64) List {
...
 		return false
 	})
 }
 
-type listIterAllFunc func(v Value, index uint64)
-
-func (l List) IterAll(f listIterAllFunc) {
-	idx := uint64(0)
+func (l List) IterFrom(idx uint64, f listIterFunc) {
+	if idx >= l.Len() { // Is this cheap?
```
kalman: yes this is cheap

go/types/list.go
```diff
@@ -162,24 +162,30 @@ func (l List) RemoveAt(idx uint64) List {
...
 		idx++
 		return false
 	})
 }
 
-type listIterAllFunc func(v Value, index uint64)
-
-func (l List) IterAll(f listIterAllFunc) {
-	idx := uint64(0)
+func (l List) IterFrom(idx uint64, f listIterFunc) {
```
kalman: I'm wondering if we should break with tradition and make IterFrom a channel based API? It's more flexible, and you've gone through some contortions to adapt the callback API into a channel one.

(using the same idiom as diff where it takes the result and close channels, and sends back on the same goroutine)

kalman:
WDYT? You could even call the new API `RangeFrom` or something, to distinguish it.

also should add a quick test either way.

go/merge/three_way_list.go
cmasone-attic: yis

go/merge/three_way_list.go
cmasone-attic: Ideally, but to do so I'd need to also have code to drain the channels and/or stop the iterations and THEN drain the channels.

It seems like a significant reduction in code clarity and brevity for an optimization that only helps in the case when we're about to bail on the merge anyway

go/types/list.go
cmasone-attic: good!

go/types/list.go
cmasone-attic: In one case I went through those contortions, definitely. It was really only because I wanted to co-iterate across two lists at the same time. In the other case, the callback model was just fine, so I don't feel like this code gives a clear mandate one way or the other :-/

Do you have other potential use cases that might break the tie?

kalman: Generally speaking I needed to add an `iterator/iteratorAt` method to JS: https://github.com/attic-labs/noms/blob/master/js/noms/src/list.js#L87 - it was for some noms client I wrote. It was related to needing to compose iterations over two lists :-)

If would at least bring the SDKs closer to parity (plus I think in the case of this CL, easier to reason about the merge code).

go/merge/three_way_list.go
kalman: Good point.

The channel drain is especially lame. At this moment I can't remember why just sending on the closer channel doesn't terminate the goroutine (in the case of diff, which would apply here).

(Note that if you had the channel based iteration API this would fall out, I think.)

cmasone-attic:
Ok. I've changed to List.Iterator() and List.IteratorAt(). PTAL

go/types/list_iterator.go
```diff
@@ -0,0 +1,24 @@
...
+
+package types
+
+import (
+	"github.com/attic-labs/noms/go/d"
+)
+
+// ListIterator can be used to efficiently iterate through a Noms List starting at an index of your choice.
+type ListIterator struct {
+	*sequenceCursor
```
kalman: can this compose instead of embed?

kalman:
lgtm but see comment about composition. I don't want methods from the cursor being available on the iterator.

go/types/list_iterator.go
cmasone-attic: yeah. I tried doing it as a type alias so I didn't have to have the extra struct, but there was a lot of `(*sequenceCursor)(li)` going on and that was ugly.

kalman:
Another trickle of comments. Feel free to ping me directly if my turnaround is too slow.

PR #2515, opened 2016-09-02
Rename files to their hashes when staging

arv:


arv:
What do you think?

tools/noms/staging.py
```diff
@@ -51,13 +53,54 @@ def stage(stagingDir):
...
+
+                digest = ''
+                with open(f) as fh:
+                    sha = hashlib.sha256()
+                    sha.update(fh.read())
+                    digest = sha.hexdigest()
+                    # print digest[:20]
+                basename = os.path.basename(f)
+                name, ext = os.path.splitext(basename)
+                new_name = name + '.' + digest[:20] + ext
```
kalman: use % syntax

tools/noms/staging.py
```diff
@@ -51,13 +53,54 @@ def stage(stagingDir):
...
-                shutil.copy2(f, toDir)
+                to_dir = os.path.join(stagingDir, from_dir)
+                if not os.path.exists(to_dir):
+                    os.makedirs(to_dir)
+                shutil.copy2(f, to_dir)
+    return stage
+
+def HashGlobCopier(index_file, *globs):
+    exclude = ('webpack.config.js',)
+    def stage(stagingDir):
```
kalman: staging_dir?

tools/noms/staging.py
```diff
@@ -51,13 +53,54 @@ def stage(stagingDir):
...
+                shutil.copy2(f, to_dir)
+                shutil.move(os.path.join(to_dir, basename), os.path.join(to_dir, new_name))
+
+        from_dir, name = os.path.split(index_file)
+        to_dir = os.path.join(stagingDir, from_dir)
+        data = None
+        with open(index_file, 'r') as f:
+            data = f.read()
+        for old_name, new_name in rename_dict.iteritems():
+            r = re.compile(r'(\W)' + re.escape(old_name) + r'(\W)')
```
kalman: use % syntax

tools/noms/staging.py
```diff
@@ -51,13 +53,54 @@ def stage(stagingDir):
...
+                basename = os.path.basename(f)
+                name, ext = os.path.splitext(basename)
+                new_name = name + '.' + digest[:20] + ext
+                rename_dict[basename] = new_name
+                shutil.copy2(f, to_dir)
+                shutil.move(os.path.join(to_dir, basename), os.path.join(to_dir, new_name))
+
+        from_dir, name = os.path.split(index_file)
+        to_dir = os.path.join(stagingDir, from_dir)
+        data = None
```
kalman: You don't really need to initialise this

kalman: what is the \W for? detecting word boundaries? if so you can use \b for that

tools/noms/staging.py
```diff
@@ -51,13 +53,54 @@ def stage(stagingDir):
...
+            for f in glob.glob(pattern):
+                if os.path.isdir(f):
+                    continue
+                from_dir, name = os.path.split(f)
+                if name in exclude:
+                    continue
+                to_dir = os.path.join(stagingDir, from_dir)
+                if not os.path.exists(to_dir):
+                    os.makedirs(to_dir)
+
```
kalman: How about you write a method which yields all paths to consider, then use it here and in GlobCopier? like:

```python
def run_globs(globs, exclude):
  for pattern in globs:
    for f in glob.glob(pattern):
      if os.path.isdir(f):
        continue
      from_dir, name = os.path.split(f)
      if name in exclude:
        continue
      to_dir = os.path.join(staging_dir, from_dir)
        if not os.path.exists(to_dir):
          os.makedirs(to_dir)
      yield (f, to_dir)

def GlobCopier(*globs):
  exclude = ...
  def stage(staging_dir):
    for f, to_dir in run_globs(globs, exclude):
      shutil.copy2(f, to_dir)
  return stage

def HashGlobCopier(*globs):
  exclude = ...
  def stage(staging_dir):
    for f, to_dir in run_globs(globs, exclude):
      ...
  return stage
```

kalman:
The approach seems fine to me. maybe "HashGlobCopier" isn't the best name but the abstraction is fine.

tools/noms/staging.py
arv: I like the generator idea.

arv:
I ended up folding HashGlobCopier into GlobCopier. It now takes kwargs which allows you to pass in rename (bool) and index_file (str)

arv:
PTAL

tools/noms/staging.py
```diff
@@ -28,36 +30,76 @@ def Main(projectName, stagingFunction):
         staging.Main('nerdosphere', staging.GlobCopier('index.html', 'styles.css', '*.js'))
     """
     parser = argparse.ArgumentParser(description='Stage build products from this directory.')
-    parser.add_argument('stagingDir',
+    parser.add_argument('staging_dir',
```
kalman: Make sure every call site (script) has been updated. My comment was about the little closures that use `stagingDir` which doesn't affect the variable name. You might want to leave the flag the same.

tools/noms/staging.py
```diff
@@ -28,36 +30,76 @@ def Main(projectName, stagingFunction):
...
+            if name in exclude:
+                continue
+            to_dir = os.path.join(staging_dir, from_dir)
+            if not os.path.exists(to_dir):
+                os.makedirs(to_dir)
+            yield (f, to_dir)
+
+
+def rename_with_hash(f, to_dir, rename_dict):
+    digest = ''
```
kalman: Is this a pylint thing? Because python scoping rules make the assignment on line 68 perfectly happy.

tools/noms/staging.py
```diff
@@ -28,36 +30,76 @@ def Main(projectName, stagingFunction):
...
+        digest = sha.hexdigest()
+        # print digest[:20]
+    basename = os.path.basename(f)
+    name, ext = os.path.splitext(basename)
+    new_name = '%s.%s%s' % (name, digest[:20], ext)
+    rename_dict[basename] = new_name
+    shutil.move(os.path.join(to_dir, basename), os.path.join(to_dir, new_name))
+
+
+def GlobCopier(*globs, **kwargs):
```
kalman: Use explicit kwargs? `def GlobCopier(*globs, rename=False, index_file=None)`

tools/noms/staging.py
arv: This is`add_argument` is not a named argument

tools/noms/staging.py
```diff
@@ -28,36 +30,76 @@ def Main(projectName, stagingFunction):
...
+        sha.update(fh.read())
+        digest = sha.hexdigest()
+        # print digest[:20]
+    basename = os.path.basename(f)
+    name, ext = os.path.splitext(basename)
+    new_name = '%s.%s%s' % (name, digest[:20], ext)
+    rename_dict[basename] = new_name
+    shutil.move(os.path.join(to_dir, basename), os.path.join(to_dir, new_name))
+
+
```
kalman: pydoc would be nice

kalman:
lgtm with more comments

tools/noms/staging.py
kalman: Ah, got it.

tools/noms/staging.py
arv: python scopes are so strange

tools/noms/staging.py
arv: That does not work with `*glob`. I tried that first.

kalman: Bummer.

Follow up idiom: use get() with a default argument. E.g.:
```python
// instead of:
rename = 'rename' in kwargs and kwargs['rename']
// do this:
rename = kwargs.get('rename', False)
```

PR #2518, opened 2016-09-06
JS: Fix subtype issue with unions as elem types

arv:
For compound types (List, Set, Map, Ref) the concrete type may be a
union. If that is the case all the types in the union must be a
subtype of the concrete type's element type.

`C<T | V>` is a subtype of `C<S>` iff T is a subtype of S and V is a
subtype of S.

### DIFF
--- a/js/noms/src/assert-type.js
+++ b/js/noms/src/assert-type.js
@@ -115,11 +115,12 @@ function isSubtypeInternal(requiredType: Type<any>, concreteType: Type<any>,
   invariant(false);
 }
 
-// compoundSubtype is called when comparing the element types of two compound types. This is the
-// only case where a concrete type may have be a union type.
+/**
+ * compoundSubtype is called when comparing the element types of two compound types.
+ * This is the only case where a concrete type may have be a union type.
+ */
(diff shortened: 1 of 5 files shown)


arv:
JS part of #2506

js/noms/src/assert-type.js
```diff
@@ -111,10 +115,17 @@ export function isSubtypeInternal(requiredType: Type<any>, concreteType: Type<an
   invariant(false);
 }
 
+// compoundSubtype is called when comparing the element types of two compound types. This is the
```
rafael-atticlabs: are we only to use `/*` style comments for public API doc?

rafael-atticlabs: word wrapping.

rafael-atticlabs:
lgtm

js/noms/src/assert-type.js
arv: For now the API doc picks up all `/** */`. I'm working on a system that follows the exported bindings (+methods) and then we should probably use JSDoc comments here.

PR #2520, opened 2016-09-06
Fix cmdline help message color argument

zcstarr:
Quick little noms-log patch that fixes the help message for noms-log to turn off correctly


zcstarr:
@willhite  PTAL

willhite:
LGTM

PR #2521, opened 2016-09-06
Improve code based on tool analysis feeback

mikegray:
Fixes are based on Go report card output:
- `gofmt -s` eliminates some duplication in struct/slice initialization
- `golint` found some issues like: `warning: should drop = nil from declaration of var XXX; it is the zero value`
- `golint` found some issues like: `warning: receiver name XXX should be consistent with previous receiver name YYY for ZZZ`
- `golint` says not to use underscores for function/variable names
- `golint` found several issues like: `warning: if block ends with a return statement, so drop this else and outdent its block`

No functional changes are included - just source code quality improvements.

### DIFF
--- a/go/types/struct.go
+++ b/go/types/struct.go
@@ -126,11 +126,11 @@ func (s Struct) Set(n string, v Value) Struct {
 	return Struct{values, s.t, &hash.Hash{}}
 }
 
-func (s1 Struct) Diff(s2 Struct, changes chan<- ValueChanged, closeChan <-chan struct{}) {
-	if s1.Equals(s2) {
+func (s Struct) Diff(last Struct, changes chan<- ValueChanged, closeChan <-chan struct{}) {
+	if s.Equals(last) {
 		return
 	}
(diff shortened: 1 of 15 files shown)


mikegray:
@willhite PTAL

go/types/struct.go
```diff
@@ -126,19 +126,19 @@ func (s Struct) Set(n string, v Value) Struct {
 	return Struct{values, s.t, &hash.Hash{}}
 }
 
-func (s1 Struct) Diff(s2 Struct, changes chan<- ValueChanged, closeChan <-chan struct{}) {
-	if s1.Equals(s2) {
+func (s Struct) Diff(last Struct, changes chan<- ValueChanged, closeChan <-chan struct{}) {
+	if s.Equals(last) {
```
willhite: Just a suggestion, would "other" be a better variable name than "last"?

mikegray: map, list, set all use last.

willhite:
Nice. LGTM

PR #2525, opened 2016-09-06
Adds checks for bad column-type and header csv-import flag values

zcstarr:
Adds checks to ensure column types correspond to the number of header types, makes partial specification invalid and ensures header values are unique.


aboodman:
lgtm!

PR #2526, opened 2016-09-07
noms diff should exit after the pager exits

ahl:
Try `noms log http://104.198.2.60::hn`; exiting the pager (in my case `less`) doesn't necessarily exit the `noms` command in a timely fashion (though it depends a bit on when you quit less).

### DIFF
--- a/cmd/noms/noms_log.go
+++ b/cmd/noms/noms_log.go
@@ -87,8 +87,10 @@ func runLog(args []string) int {
 		return buff.Bytes()
 	}, parallelism)
 
+	var done = false
+
 	go func() {
-		for ln, ok := iter.Next(); ok && displayed < maxCommits; ln, ok = iter.Next() {
+		for ln, ok := iter.Next(); !done && ok && displayed < maxCommits; ln, ok = iter.Next() {
 			inChan <- ln
 			displayed++
 		}
@@ -99,8 +101,15 @@ func runLog(args []string) int {
 	defer pgr.Stop()
 
 	for commitBuff := range outChan {
-		io.Copy(pgr.Writer, bytes.NewReader(commitBuff.([]byte)))
+		_, err := io.Copy(pgr.Writer, bytes.NewReader(commitBuff.([]byte)))
+		if err != nil {
+			done = true
+			for range outChan {
+				// drain the output
+			}
+		}
 	}
+
 	return 0
 }
 


arv:
@kalman and @willhite has dealt with lots of edge cases in this area.

cmd/noms/noms_log.go
```diff
@@ -99,8 +100,12 @@ func runLog(args []string) int {
 	defer pgr.Stop()
 
 	for commitBuff := range outChan {
-		io.Copy(pgr.Writer, bytes.NewReader(commitBuff.([]byte)))
+		_, err := io.Copy(pgr.Writer, bytes.NewReader(commitBuff.([]byte)))
+		if err != nil {
+			os.Exit(0)
```
kalman: I've had code like this in the past and it caused other problems, like the terminal not resetting correctly. Any idea what is still running? Does just `break` here fix it (in which case I would wonder why)?

ahl: There are a shit ton of commits so even when I exit my pager (less) it keep cranking along (I can see it w/ DTrace).

If I just do `break` it will occasionally hang forever. The short of it is that if you do a `datas.httpBatchStore.Get()` after the `Close()` method of that class has been called you'll hang forever because the background task (spawned from `batchReadRequests()`) that processes those requests has exited.

I've chased this down and found a better fix, but I'd suggest that there are still some gaps here. In particular invoking `Close()` should cause outstanding `Get()` calls to fail and return immediately, but there doesn't appear to be a mechanism to achieve this (e.g. outstanding and in-flight requests are not tracked centrally).

I'd strongly suggesting asserting that untoward stuff doesn't happen after `Close()` as it will certainly lead to problems.

New fix coming.

ahl: Here are the stacks when I use `break` in case you're interested:

https://gist.github.com/ahl/35937d020e585d6d3aa28f638dc46273

cmd/noms/noms_log.go
```diff
@@ -99,8 +101,16 @@ func runLog(args []string) int {
...
 
 	for commitBuff := range outChan {
-		io.Copy(pgr.Writer, bytes.NewReader(commitBuff.([]byte)))
+		_, err := io.Copy(pgr.Writer, bytes.NewReader(commitBuff.([]byte)))
+		if err != nil {
+			done = true
+			for range outChan {
+				// drain the output
+			}
+			break
```
kalman: Presumably don't need this break?

cmd/noms/noms_log.go
kalman: I've totally run into this problem before, or something very similar to it. I don't remember when, but it probably was either noms show, diff, or log. @cmasone-attic might be interested.

kalman:
lgtm, thanks so much.

cmd/noms/noms_log.go
ahl: true that

PR #2533, opened 2016-09-08
Allow changes to different Datasets to proceed concurrently

### LINKED ISSUE #2524: (JS) Auto-merge concurrent commits to noms datasets
Currently a noms database has zero concurrency. Even commits to separate datasets must be serialized.

Now that we have auto-merge, we should take the first step of adding back some concurrency by allowing concurrent commits on separate datasets to auto-merge.

cmasone-attic:
Prior to this change, concurrent modification of two different
Datasets in a given Database would result in an
ErrOptimisticLockFailed for whichever process lost the race. As of
this patch, concurrent changes to the _same_ Dataset will result in an
ErrMergeNeeded for the loser, but concurrent changes to different
Datasets will succeed.

Towards #2524

### DIFF
--- a/go/datas/database_common.go
+++ b/go/datas/database_common.go
@@ -96,34 +96,60 @@ func (ds *databaseCommon) doSetHead(datasetID string, commit types.Struct) error
 func (ds *databaseCommon) doCommit(datasetID string, commit types.Struct) error {
 	d.PanicIfTrue(!IsCommitType(commit.Type()), "Can't commit a non-Commit struct to dataset %s", datasetID)
 
-	currentRootRef, currentDatasets := ds.getRootAndDatasets()
-	commitRef := ds.WriteValue(commit) // will be orphaned if the tryUpdateRoot() below fails
+	var err error
+	for err = ErrOptimisticLockFailed; err == ErrOptimisticLockFailed; {
+		currentRootRef, currentDatasets := ds.getRootAndDatasets()
+		commitRef := ds.WriteValue(commit) // will be orphaned if the tryUpdateRoot() below fails
 
-	// First commit in store is always fast-forward.
-	if !currentRootRef.IsEmpty() {
-		r, hasHead := currentDatasets.MaybeGet(types.String(datasetID))
+		// Allow only fast-forward commits.
+		// If there's nothing in the DB yet, skip all this logic.
+		if !currentRootRef.IsEmpty() {
+			r, hasHead := currentDatasets.MaybeGet(types.String(datasetID))
 
-		// First commit in dataset is always fast-forward.
-		if hasHead {
-			currentHeadRef := r.(types.Ref)
-			// Allow only fast-forward commits.
-			if commitRef.Equals(currentHeadRef) {
-				return nil
-			}
-			if !CommitDescendsFrom(commit, currentHeadRef, ds) {
-				return ErrMergeNeeded
+			// First commit in dataset is always fast-forward, so go through all this iff there's already a Head for datasetID.
+			if hasHead {
+				currentHeadRef := r.(types.Ref)
+				if commitRef.Equals(currentHeadRef) {
+					return nil
+				}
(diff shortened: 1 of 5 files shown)


cmasone-attic:
PTAL

aboodman:
Sweet

go/datas/database_common.go
```diff
@@ -96,34 +96,60 @@ func (ds *databaseCommon) doSetHead(datasetID string, commit types.Struct) error
...
-		if hasHead {
-			currentHeadRef := r.(types.Ref)
-			// Allow only fast-forward commits.
-			if commitRef.Equals(currentHeadRef) {
-				return nil
-			}
-			if !CommitDescendsFrom(commit, currentHeadRef, ds) {
-				return ErrMergeNeeded
+	var err error
+	for err = ErrOptimisticLockFailed; err == ErrOptimisticLockFailed; {
```
rafael-atticlabs: can we put a comment here to a bug that we open that there's no guarentee about termination here? i.e. if a client gets unlucky and the database is under heavy load, it could theoretically stay in this loop forever. Eventually we probably need some kind of consensus mechanism which orders commits or something.

arv:
jenkins: test this please

go/datas/database_common.go
rafael-atticlabs: I feel like at one point we had calling code in places that did this looping because it wasn't done here. Can you just double check that if we do, those places don't loop anymore over this specific condition?

go/datas/database_common.go
```diff
@@ -96,34 +96,60 @@ func (ds *databaseCommon) doSetHead(datasetID string, commit types.Struct) error
...
 
 // doDelete manages concurrent access the single logical piece of mutable state: the current Root. doDelete is optimistic in that it is attempting to update head making the assumption that currentRootRef is the hash of the current head. The call to UpdateRoot below will return an 'ErrOptimisticLockFailed' error if that assumption fails (e.g. because of a race with another writer) and the entire algorithm must be tried again.
-func (ds *databaseCommon) doDelete(datasetID string) error {
+func (ds *databaseCommon) doDelete(datasetIDstr string) error {
+	datasetID := types.String(datasetIDstr)
 	currentRootRef, currentDatasets := ds.getRootAndDatasets()
-	currentDatasets = currentDatasets.Remove(types.String(datasetID))
-	return ds.tryUpdateRoot(currentDatasets, currentRootRef)
+	var initialHead types.Ref
+	if r, hasHead := currentDatasets.MaybeGet(datasetID); !hasHead {
```
rafael-atticlabs: heh. nice catch

rafael-atticlabs: actually, shouldn't this be in the loop below? it's possible someone will delete this ds behind your back.

go/datas/database_common.go
```diff
@@ -96,34 +96,60 @@ func (ds *databaseCommon) doSetHead(datasetID string, commit types.Struct) error
...
+	var err error
+	for {
+		currentDatasets = currentDatasets.Remove(datasetID)
+		err = ds.tryUpdateRoot(currentDatasets, currentRootRef)
+		if err != ErrOptimisticLockFailed {
+			break
+		}
+		// If the optimistic lock failed because someone changed the Head of datasetID, then return ErrMergeNeeded. If it failed because someone changed a different Dataset, we should try again.
+		currentRootRef, currentDatasets = ds.getRootAndDatasets()
+		if r, hasHead := currentDatasets.MaybeGet(datasetID); hasHead && !initialHead.Equals(r) {
```
rafael-atticlabs: this is the part that was kind of confusing. because the initialHead is just the commit which happened to be head at the time you called Delete() -- not as of the time you created the `Database`. Can you just put a todo here that presumably will be resolved when `Delete` takes a `Dataset` as an argument?

rafael-atticlabs:
lgtm

rafael-atticlabs:
Note that this needs impl in JS.

go/datas/database_common.go
cmasone-attic: Yes and yes

PR #2539, opened 2016-09-09
Fix JSDoc

arv:


arv:
Waiting for tests to pass this time...

PR #2542, opened 2016-09-12
Allow struct set to cause a type change

### LINKED ISSUE #2181: type accretion does not work with Go structs
```
s := NewStruct("", nil)
// error, because s does not have field bar
s = s.Set("foo", String("bar"))
```

Similarly:

```
s := NewStruct("", map[string]Value{"foo", String("bar")})
// error, because number not same type as string
s = s.Set("foo", Number(42))
```

I sort of expected both these to work and the type to just change, similar to what happens with List.Append() or Set.Insert().

arv:
This allows setting a field in a struct to a new type or to set a
non-existig field in a struct.

In JS this is done through the StructMirror.p.set and in Go this is
done through Struct Set.

Fixes #2181

### DIFF
--- a/js/noms/package.json
+++ b/js/noms/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@attic/noms",
   "license": "Apache-2.0",
-  "version": "63.0.1",
+  "version": "63.1.0",
   "description": "Noms JS SDK",
   "repository": "https://github.com/attic-labs/noms",
   "main": "dist/commonjs/noms.js",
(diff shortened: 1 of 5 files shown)


arv:
Ping

arv:
Ping

arv:
@willhite PTAL

js/noms/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@attic/noms",
   "license": "Apache-2.0",
-  "version": "59.0.2",
+  "version": "59.1.0",
```
willhite: I'm sure this is out of date.

arv:
Type errors. Will have to look closer.

willhite:
Let me know if you're waiting on me for anything.

Issue #2532, opened 2016-09-08
Create landing page for docs.noms.io 


mikegray:
Currently docs.noms.io displays only the documentation generated from the Noms JavaScript code. It would be great if it had a landing page to allow you select JS or Go. For Go, it could just link to https://godoc.org/github.com/attic-labs/noms, and for JS it could go to the most recent version like it redirects to now - https://docs.noms.io/59.0.2/.  This would enable us to point developers to https://docs.noms.io/ for all generated documentation regardless of SDK/platform.

mikegray:
Might be a separate bug/issue, but as I am writing API overview I am finding myself wanting to link to specific functions/methods in the JavaScript documentation, but I don't think it would be good to include the `/59.0.2/` portion. If possible, could the index page that redirects to the current version also carry forward in the redirect the query string portion of the URL?

e.g. https://docs.noms.io/#databasespec would redirect to https://docs.noms.io/59.0.2/#databasespec

arv:
@mikegray Those two comments are kind of contradicting each other.

We could do `/js/1.2.3` and `/js/` which would forward dynamically.

For `/` I think we need some proposal what to put there, but I do agree that squatting on `/` for the JS SDK API docs does not seem right.

mikegray:
Sorry, yes was thinking https://docs.noms.io/js/ would be the location of JS documentation. If I link to https://docs.noms.io/js/#databasespec then it would be great if the redirect within JS documentation from root (/js/) to current version (/js/1.2.3/) would carry forward the anchor ref (#databasespec).

arv:
Now the question is where we should checkin the /index.html file?

mikegray:
Not sure - part of the build of the js documentation so next to that build script?

arv:
@mikegray I'll put it in the tools/build-js-documentation.py which is what jenkins invokes

PR #2543, opened 2016-09-12
Generate js docs in /js/

arv:
Also keep the hash when redirecting

Towards #2532

### DIFF
--- a/js/noms/build/make-index-html.js
+++ b/js/noms/build/make-index-html.js
@@ -4,18 +4,14 @@
 
 'use strict';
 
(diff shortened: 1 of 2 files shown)


js/noms/build/make-index-html.js
```diff
@@ -4,18 +4,14 @@
 
 'use strict';
 
-const fs = require('fs');
+const fs = require('fs-extra');
```
arv: We already had a dev dep on fs-extra

mikegray:
lgtm, thanks!

PR #2544, opened 2016-09-12
flickr/slurp: Treat fail as an error

arv:
Reject the promise when we get a fail status.


arv:
This makes it clear why the cron job is failing... "Invalid auth token"!

kalman:
LGTM

kalman:
@aboodman as suspected

PR #2545, opened 2016-09-12
API overview documentation

mikegray:
towards #2298

### DIFF
--- a/doc/api-overview.md
+++ b/doc/api-overview.md
@@ -1,0 +1,91 @@
+# Noms API Overview
+
+Noms provides SDKs for Go and JavaScript development. This overview introduces the basic types and APIs available to Go and Javascript developers. The SDKs are equivalent in terms of feature parity with the differences being to enable more natural development by the platform specific developer.
+
+## Generated API Documentation
+
+The API documentation that is derived from the source code automatically can be found by following these links: 
+
+* [Go Docs](https://godoc.org/github.com/attic-labs/noms)
+* [JavaScript Docs](https://docs.noms.io)
+
+## Noms Setup
+
+If you haven't already done so, please read the [Intro to Noms](intro.md). You should also review the [Go Tour](go-tour.md) or the [JavaScript Tour](js-tour.md) for help getting the appropriate SDK installed.
+
+## Specifying Data
+
+In order to store or retrieve data you will need to connect to a database and specify a dataset. Noms has `paths` for specifying resources like databases, datasets and values. The [Spelling in Noms](spelling.md) documentation explains the finer details on how to construct these `paths`. The Noms [go/spec package](https://godoc.org/github.com/attic-labs/noms/go/spec) and [JavaScript specs.js](https://github.com/attic-labs/noms/blob/master/js/noms/src/specs.js) are the source code locations to review for parsing and building `paths`. Several of the [samples](https://github.com/attic-labs/noms/tree/master/samples) use `paths` specified as command-line arguments and can be used as examples for properly parsing, building and checking for the errors that are returned from these APIs.
+
+## Data Versioning
+
+Once you have a database/dataset specified, you can store and retrieve data. A dataset, like a `git` repository has a `HEAD` revision - or a marker for the current version of the data. You can access a Noms dataset `HEAD` using the dataset API - [Head](https://godoc.org/github.com/attic-labs/noms/go/dataset#Dataset.Head). If you want to set a different version of your data to be `HEAD` then you would use the dataset [`Commit`](https://godoc.org/github.com/attic-labs/noms/go/dataset#Dataset.Commit) method.
+
+## Noms Types
+Noms supports [several data types](intro.md#types). Most of these types will be familiar and behave as you expect. Most Noms types are [immutable](https://en.wikipedia.org/wiki/Immutable_object). For example, at the end of the last section when calling `Commit` on dataset, the dataset you were holding was not updated. Instead `Commit` returned a new `dataset` for the new `HEAD` you created.
+
+### Noms Values
+
+All Noms values implement the [Value](https://godoc.org/github.com/attic-labs/noms/go/types#Value) interface. Noms lists, sets and maps are all collections of Noms values - but they are also Noms values themselves and can therefore be nested to create complex data structures. Noms includes value implementations for booleans, strings, structs and numbers too.
+
+### Noms Lists
+
+Lists are one of the value collection types supported in Noms - [Go](https://godoc.org/github.com/attic-labs/noms/go/types#List) and [JavaScript](https://docs.noms.io/59.0.2/#list). Lists are similar to arrays or vectors in most languages. 
+
+- `Empty` returns a boolean and enables you to test if a list is empty
+- `Len` returns the number of values in the list 
+- `Get` and `Set` enables retrieving and storing, respectively, Noms values at specific indices.
+- `Remove` and `RemoveAt` allow for deleting values from a Noms list by index.
+- `Append` allows for adding additional Noms values to the end of the list.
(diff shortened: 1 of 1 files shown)


mikegray:
@rafael-atticlabs PTAL, not sure what all you want covered but I took a stab at an initial version

doc/api-overview.md
```diff
@@ -0,0 +1,91 @@
+# Noms API Overview
+
+Noms provides SDKs for Go and JavaScript development. This overview introduces the basic types and APIs available to Go and Javascript developers. The SDKs are equivalent in terms of feature parity with the differences being to enable more natural development by the platform specific developer.
```
arv: Javascript -> JavaScript

doc/api-overview.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Data Versioning
+
+Once you have a database/dataset specified, you can store and retrieve data. A dataset, like a `git` repository has a `HEAD` revision - or a marker for the current version of the data. You can access a Noms dataset `HEAD` using the dataset API - [Head](https://godoc.org/github.com/attic-labs/noms/go/dataset#Dataset.Head). If you want to set a different version of your data to be `HEAD` then you would use the dataset [`Commit`](https://godoc.org/github.com/attic-labs/noms/go/dataset#Dataset.Commit) method.
+
+## Noms Types
+Noms supports [several data types](intro.md#types). Most of these types will be familiar and behave as you expect. Most Noms types are [immutable](https://en.wikipedia.org/wiki/Immutable_object). For example, at the end of the last section when calling `Commit` on dataset, the dataset you were holding was not updated. Instead `Commit` returned a new `dataset` for the new `HEAD` you created.
```
arv: > Most Noms types are [immutable]

actually, all of them are immutable.

doc/api-overview.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Noms Values
+
+All Noms values implement the [Value](https://godoc.org/github.com/attic-labs/noms/go/types#Value) interface. Noms lists, sets and maps are all collections of Noms values - but they are also Noms values themselves and can therefore be nested to create complex data structures. Noms includes value implementations for booleans, strings, structs and numbers too.
+
+### Noms Lists
+
+Lists are one of the value collection types supported in Noms - [Go](https://godoc.org/github.com/attic-labs/noms/go/types#List) and [JavaScript](https://docs.noms.io/59.0.2/#list). Lists are similar to arrays or vectors in most languages. 
+
```
arv: You can use https://docs.noms.io/js/#list now

doc/api-overview.md
```diff
@@ -0,0 +1,91 @@
...
+- `Empty` returns a boolean and enables you to test if a list is empty
+- `Len` returns the number of values in the list 
+- `Get` and `Set` enables retrieving and storing, respectively, Noms values at specific indices.
+- `Remove` and `RemoveAt` allow for deleting values from a Noms list by index.
+- `Append` allows for adding additional Noms values to the end of the list.
+- `Insert` allows for adding Noms values at a specific index.
+- Similarly `Splice` supports removing items at an index and inserting Noms values at the same index.
+- Iterating over Noms values in the list can be done by calling `Iterator` or `IteratorAt` and using the [ListIterator](https://godoc.org/github.com/attic-labs/noms/go/types#ListIterator) returned to call `Next` to retrieve the values in order by their index.
```
arv: Maybe add a comment that using  a c-style for loop is highly ineffective?

doc/api-overview.md
```diff
@@ -0,0 +1,91 @@
...
+- `Set`
+- `Diff`
+
+*TODO: write about Noms structs here*
+
+## JavaScript Specifics
+
+### Flow, Promises, Unicorns, Oh My!
+
+Noms uses several JavaScript packages to improve source code quality, readability and performance.  You can see the dependencies by reviewing [package.json](https://github.com/attic-labs/noms/blob/master/js/noms/package.json). Many of these will not affect you, but a few definitely will. The usage of [Promises](https://www.promisejs.org/) and [Flow](https://flowtype.org/) type annotations will be noticeable in every JavaScript file and most APIs. If you are not already familiar with these projects or concepts it is worth your time to go learn the basics as it will make it easier for you to develop using the Noms JavaScript SDK.
```
arv: Better docs for Promise:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

arv: I think it is worth mentioning that there is no need to use Flow in js clients that depend on noms.

mikegray:
thanks @arv for the review thus far! All of your comments have now been incorporated.

doc/api-overview.md
```diff
@@ -0,0 +1,91 @@
+# Noms API Overview
+
+Noms provides SDKs for Go and JavaScript development. This overview introduces the basic types and APIs available to Go and JavaScript developers. The SDKs are equivalent in terms of feature parity with the differences being to enable more natural development by the platform specific developer.
```
rafael-atticlabs: feature parity isn't totally true. maybe worth making a slightly weaker statement or leaving out altogether?

rafael-atticlabs:
I think this a good start. What about this as an outline:

- How Noms models data (progression of commits, all commits have a value, values form a graph, examples of modeling data in graph form)
- Working with immutable APIs (mutating a thing,\ returns a new thing which reflects the logical change. if you want to mutate a thing down deep in a graph, you have to mutate all the things above it to include the change, code same of making that look clean)
- Quick Review of Noms data-types (scalar types w/ possible discussion of marshalling in go, struct type (including recursive structs), collection types (iterating over data, discussion of lazy loading, in-memory mutation, graph builder), dataset / database)
- Noms typing (show how types roll up to the commit)
- Refs, ReadValue, WriteValue
- Managing concurrency (optimistic locking, ErrMergeNeeded, diffing, explicit merging)

Obviously, this is alot, but you don't need to take it all on at once. @aboodman Thoughts?

mikegray:
I see the value is all of the parts you listed above. I believe many of them fall outside the boundary of what I would consider `API overview`.  I only mention it because that was the original issue/bug here. I agree all of these are worthwhile and should exist in the future world we shall continue to inhabit. It might be worthwhile to create a meta-bug that has links to bugs for all of the parts you suggested (and others to come).  I doubt I am the best person to write all of the different documents but will continue with it and probably ask others to contribute to sections for which they have developed the code. Does that make sense to you @rafael-atticlabs and is it an okay path forward?

rafael-atticlabs:
sgtm

doc/api-overview.md
mikegray: ok, done

aboodman:
Much of the content of this seems to overlap with this: https://github.com/attic-labs/noms/blob/master/doc/intro.md#types -- is there a plan for deduping?

mikegray:
Which do you like more? Yes, we should reduce duplication.

My understanding was that the request to have an API overview doc meant this API doc was less of an introduction and was more specifically targeted toward an application developer audience. The intro doc could be targeted toward anyone that is technical and interested in knowing about Noms in general. 

I might be thinking about it incorrectly though so please let me know.  I don't have heartburn about having duplicate content if it is targeting different audiences with more specific or more depth for those audiences.

rafael-atticlabs:
@mikegray, really sorry. I think this is a case of poor communication between @aboodman & myself and the result is that did some good work without clear goals. I'll take the blame, and I'll also take over this bug.

I do think the existing documentation doesn't sufficiently describe how to use the noms api, but I'll take it as an exercise for myself to work with @aboodman to get it to that point.

I'm going to close this PR but keep it as raw material for a second attempt that I'll take at it.

PR #2548, opened 2016-09-12
Implement GraphBuilder

willhite:


go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+	oc      opCache
+	vrw     ValueReadWriter
+	stack   graphStack
+	verbose bool
+}
+
+// NewGraphBuilder() returns an object that is used in building a graph of
+// objects that are stored on disk and then assembled into a noms collection.
+// Graphs can exist of nested maps, with maps, sets and lists at the outermost
+// levels of the graph. Any of the collections (nested or leaf) can contain onto
```
aboodman: extra word: 'onto' ?

rafael-atticlabs:
Sorry, @willhite. I know you were just waiting on me to get to this (and I said send it to me), but looking at it now, I realize that @cmasone-attic needs to be the primary reviewer here since this is really an evolution of the initial work he. did.

@cmasone-attic, I'll do a high-level review as well, but please do the detail review.

go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+func newGraphBuilder(vrw ValueReadWriter, opc opCache, rootKind NomsKind, verbose bool) *GraphBuilder {
+	b := &GraphBuilder{oc: opc, vrw: vrw, verbose: verbose}
+	b.pushNewKeyOnStack(String("ROOT"), rootKind, nil)
+	return b
+}
+
+// MapSet() will add the key/value pair |k, v| to the map found by traversing
+// the graph using the |keys| slice. Intermediate maps referenced by |keys| are
+// created as necessary
+func (b *GraphBuilder) MapSet(keys []Value, k Value, v Value) {
```
rafael-atticlabs: can we call `keys`, `path`, since that's what it is: the path to the place you wish to add values.

Also, this makes sense, but isn't entirely explained in the comment that since right now graphs can only have maps as their internal nodes, that a path is just a sequence of map keys.

go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+// NewGraphBuilder() returns an object that is used in building a graph of
+// objects that are stored on disk and then assembled into a noms collection.
+// Graphs can exist of nested maps, with maps, sets and lists at the outermost
+// levels of the graph. Any of the collections (nested or leaf) can contain onto
+// any type of Value. |rootKind| indicates what the topLevel collection should be.
+// It must be one of: MapKind, SetKind, or ListKind. If this GraphBuilder is
+// intended to build nested maps, |rootKind| must be MapKind.
+func NewGraphBuilder(vrw ValueReadWriter, rootKind NomsKind, verbose bool) *GraphBuilder {
```
rafael-atticlabs: Is this abstraction intended to be thread safe (usable from multiple goroutines)?

go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+
+type GraphOpContainer struct {
+	keys []Value
+	kind NomsKind
+	item sequenceItem
+}
+
+// Builds and returns the graph.
+func (b *GraphBuilder) Build() Value {
+	d.Chk.True(b.oc != nil, "Can only call Finish() once")
```
rafael-atticlabs: if this thing is intended to be threadsafe, then this probably needs to be different.

go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+		close(graphOpChan)
+	}()
+
+	// iterator returns keys, in sort order by array
+	for goc := range graphOpChan {
+		keys, kind, item := goc.keys, goc.kind, goc.item
+
+		// Get index of first key that is different than what is on the stack
+		idx := findDiff(b.stack, keys)
```
rafael-atticlabs: `findDiff` isn't a great name for this. how about `commonPrefixCount`?

go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+		status.Done()
+	}
+	return res
+
+}
+
+// pushNewKeyOnStack() creates a new graphStackElem node and pushes it on the
+// stack. The new element contains the |key| and a new sequenceChunker that will
+// be appended to to build this node in the graph.
+func (b *GraphBuilder) pushNewKeyOnStack(key Value, kind NomsKind, item sequenceItem) {
```
rafael-atticlabs: why have `item` here, as opposed to just having caller immediately call `appendItemToCurrentTopOfStack`?

go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+	var seq *sequenceChunker
+	switch kind {
+	case MapKind:
+		seq = newEmptySequenceChunker(b.vrw, b.vrw, makeMapLeafChunkFn(b.vrw), newOrderedMetaSequenceChunkFn(MapKind, b.vrw), mapHashValueBytes)
```
rafael-atticlabs: can we abstract these, e.g. `newListChunker`, `newSetChunker`, `newMapChunker` to avoid code duplication with other places that are constructing appropriate chunkers?

go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	return res
+
+}
+
+// pushNewKeyOnStack() creates a new graphStackElem node and pushes it on the
+// stack. The new element contains the |key| and a new sequenceChunker that will
+// be appended to to build this node in the graph.
+func (b *GraphBuilder) pushNewKeyOnStack(key Value, kind NomsKind, item sequenceItem) {
+	var seq *sequenceChunker
```
rafael-atticlabs: nit: it's common in other places to use `seq` to review to a sequence. please use `sc` or `chunker` to refer to a sequence chunker (here and elsewhere).

go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (s graphStack) String() string {
+	buf := bytes.Buffer{}
+	for i := len(s.elems) - 1; i >= 0; i-- {
+		fmt.Fprintln(&buf, "#:", i, s.elems[i])
+	}
+	return buf.String()
+}
+
+type graphStackElem struct {
```
rafael-atticlabs: put the type def above the member functions.

rafael-atticlabs: i.e. with the type def for graphStack

go/types/graph_builder.go
```diff
@@ -0,0 +1,263 @@
...
+	key  Value
+	kind NomsKind
+	seq  *sequenceChunker
+}
+
+// done() creates the appropriate collection for this element and returns it
+func (e *graphStackElem) done() Value {
```
rafael-atticlabs: Return value can be `Collection`

go/types/opcache.go
```diff
@@ -28,12 +28,19 @@ type opCacheStore interface {
 type opCache interface {
 	MapSet(mapKey Value, mapVal Value)
```
rafael-atticlabs: are MapSet & SetInsert still needed?

rafael-atticlabs: also ListAppend?

go/types/graph_builder.go
willhite: Can you explain why this would be different? It is intended to be threadsafe but this method should only be called once and only after any other "mutating" calls are completed.

go/types/graph_builder.go
willhite: Great suggestion, I hated that name, too.

go/types/graph_builder.go
willhite: Seemed convenient at the time but now that I look at the code, I'm willing to have the caller call appendItem... when necessary.

go/types/opcache.go
willhite: I debated whether or not to remove these. Each one can be replaced with a to it's corresponding Graph method like so:
MapSet(key, val)   => GraphMapSet(nil, key, val)

I'll take the approach that less api calls wins in this case against slightly more complex function signatures. I also deleted the corresponding, ListOp(), SetOp(), and Map() functions on opCacheIterator().

go/types/graph_builder.go
willhite: Yes. I created:
 newEmptyListSequenceChunker(vr ValueReader, vw vaueWriter) *sequenceChunker
 newEmptyMapSequenceChunker(vr ValueReader, vw vaueWriter) *sequenceChunker
 newEmptySetSequenceChunker(vr ValueReader, vw vaueWriter) *sequenceChunker

go/types/graph_builder.go
willhite: I'll leave this as keys for now and try to explain it better than I have in the comment for NewGraphBuilder() above. I would rather use path for actual Path objects.

go/types/graph_builder.go
willhite: All the methods are thread safe (i.e. MapSet(), SetInsert(), and ListAppend()). Build() is only meant to be called once so it really doesn't make sense to call it from multiple threads.

go/types/graph_builder.go
willhite: done

go/types/graph_builder.go
cmasone-attic: I think the issue isn't that this might be called on multiple goroutines, but rather that it should probably provide some kind of synchronization so that it's guaranteed to run only after all "mutating" calls have completed, which themselves may be running on any number of goroutines. Otherwise the caller has to do that manually, which is kinda weird as they don't need to do any synchronization among all the mutating calls themselves.

willhite: I talked to Raf about this yesterday and his issue I think was that this could be called once by two different threads and possibly not have this check fail since oc didn't get set to nil until the deferred function was called. I can fix that pretty easily by coping the oc to a local variable and setting oc to nil immediately.

I did think of another issue about calling the mutation functions from multiple threads though. The order that values get appended to lists depends on the order that ListAppend() is called. So any one who cared about that would have to manage when that method got called.

rafael-atticlabs: I'm not really asserting an opinion about whether this thing should be goroutine safe right now, just that you should make a decision, document it and have it the provided semantics be correct.

in the abstract, having this threadsafe seems highly useful because the basic goal here is to be able to efficient operate over data which won't fit in memory. it seems likely we're going to want to do things like that in parallel.

Abstractly, it would make more sense to me that `ListAppend` because `ListSet(index uint64, v Value)`. If you do that, then you get a couple of things:

 - It's more straightforward how to make this threadsafe
 - You have an abstraction which allows you specify deterministically an location in graph.
 - Lists could then be internal nodes in trees, which seems somewhat likely to be useful (i.e. list of structs).
 - The mental model of how GraphBuilder works is self consistent (I add items to a graph based on path to their location, in any order => I can create a graph of any size).

I'm fine with proceeding in multiple directions. For instance, documenting this as non-threadsafe and landing as is, is an acceptable answer -- leaving the more general solution for later.

cmasone-attic:
@willhite if you're waiting for me to review before posting new revisions, I'm waiting for you to post revisions addressing the comments from @rafael-atticlabs before reviewing -- particularly the "big theory" stuff so I can orient myself before trying to understand what's going on :-)

willhite:
@cmasone-attic @rafael-atticlabs PTAL. I believe all review comments have been addressed. I left ListAppend() like it was with a comment explaining the behavior in a multi-threaded environment might not be what the user expects.

go/types/graph_builder.go
```diff
@@ -0,0 +1,301 @@
...
+	for goc := range graphOpChan {
+		keys, kind, item := goc.keys, goc.kind, goc.item
+
+		// Get index of first key that is different than what is on the stack
+		idx := commonPrefixCount(b.stack, keys)
+
+		if idx == -1 {
+			// no keys have changed we're working on same coll as previous
+			// iteration, just append to sequenceChunker at top of stack
+			b.appendItemToCurrentTopOfStack(kind, item)
```
cmasone-attic: I'd put a `continue` after this, and then not have the `else` clause at all, so you can get rid of a level of nesting :-)

go/types/graph_builder.go
```diff
@@ -0,0 +1,301 @@
...
+		} else {
+			// Some keys that were in the last graphOp are no longer present
+			// which indicates that we are finished adding to those cols. Pop
+			// those keys from the stack. This will cause any popped cols to be
+			// closed and added to their parents.
+			for idx < b.stack.len()-1 {
```
cmasone-attic: you do `b.stack.len()-1` so much that I wonder if you should replace graphStack.len() with some other method that returns that value. `lastIdx()` or something?

go/types/graph_builder.go
```diff
@@ -0,0 +1,301 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// GraphBuilder allows non-RAM-bound construction of a graph of nested Maps whose
+// leaf collections can be Lists, Sets, or Maps that contain any type of NOMS
```
willhite: good call.

go/types/graph_builder.go
```diff
@@ -0,0 +1,301 @@
...
+// the nodes at each level of the graph. (See opcache.go for more info on how
+// that is done)
+//
+// GraphBuilder.Build() does the work of assembling the graph. Build() gets an
+// iterator for this graph from the opCache and uses it to iterate over all the
+// operations that have been stored for this graph. opCache ensures that the
+// operations are returned in optimal sorted order so that sequenceChunker can
+// most efficiently assemble the graph. Build() will ensure that there is a Map
+// object for each key in |graphKeys|. Any node that falls in the middle of the
+// graph must be a Map, although, intermediate apps may have any element as keys
```
willhite: I meant 'intermediate nodes'

go/types/graph_builder.go
```diff
@@ -0,0 +1,301 @@
...
+// as long as the path formed by the graphKeys doesn't conflict.
+//
+// MapSet(), SetInsert(), and ListAppend() are threadsafe meaning they can safely
+// be called from different go routines. However, the semantics of ListAppend()
+// are such that the order of the list will be determined by which thread() calls
+// ListAppend first (this function call may be modified later to allow specification
+// of index or order).
+//
+// Build() should only be called once, after all the operations for the graph
```
willhite: Done.

go/types/graph_builder.go
```diff
@@ -0,0 +1,301 @@
...
+// have been stored.
+//
+
+package types
+
+import (
+	"bytes"
+	"fmt"
+
```
willhite: Fixed.

go/types/graph_builder.go
```diff
@@ -0,0 +1,301 @@
...
+// maps referenced by |keys| are created as necessary. This is threadsafe, may
+// call from multiple go routines, however append semantics are such that the
+// elements will be appended in order that functions are called, so order has
+// to be managed by caller.
+func (b *GraphBuilder) ListAppend(keys []Value, v Value) {
+	d.Chk.True(b.oc != nil, "Can't call ListAppend() again after Build()")
+	b.oc.GraphListAppend(keys, v)
+}
+
+type GraphOpContainer struct {
```
willhite: Done.

go/types/graph_builder.go
```diff
@@ -0,0 +1,301 @@
...
+	keys []Value
+	kind NomsKind
+	item sequenceItem
+}
+
+// Builds and returns the graph.
+func (b *GraphBuilder) Build() Value {
+	var opc opCache
+
+	checkFirstCall := func() {
```
willhite: I'm still thinking about this one.

go/types/graph_builder.go
willhite: Done.

go/types/graph_builder.go
willhite: Good idea. Done.

go/types/opcache.go
```diff
@@ -26,14 +83,21 @@ type opCacheStore interface {
 }
 
 type opCache interface {
-	MapSet(mapKey Value, mapVal Value)
-	SetInsert(val Value)
+	// This method is threadsafe.
```
willhite: Changed to: This method can be called from multiple go routines

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
-		d.Chk.NoError(err)
-
-	case BoolKind, NumberKind, StringKind:
-		ldbKey := ldbKeyFromValue(mapKey, opc.colId, opc.vrw)
-		encodedMapVal := encToSlice(mapVal, mapValArray[:], opc.vrw)
-		err := opc.ldb.Put(ldbKey, encodedMapVal, nil)
-		d.Chk.NoError(err)
+// insertLdbOp encodes allKeys into the ldb key. Bool, Number, and String values
+// are encoded directly into the ldb key bytes. All other types are encoded as
+// there Hash() digest. Their actual value is then stored in ldb value.
```
willhite: Done.

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
+func (opc *ldbOpCache) insertLdbOp(allKeys ValueSlice, opKind NomsKind, val Value) {
+	d.Chk.True(len(allKeys) < 0x00FF, "Number of keys in GraphMapSet exceeds max of 256")
```
willhite: Right. I merged that change into this branch a while back and must have missed this one. This line also had another bug. len(allKeys) can be <= 256. This changed to:
```    d.PanicIfTrue(len(allKeys) > 0x00FF```

File: go/types/opcache.go [hunk 23]
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
-		entry.key = DecodeFromBytes(ldbValue[uint32Size:uint32Size+keyBytesLen], i.vr, staticTypeCache)
-		entry.value = DecodeFromBytes(ldbValue[uint32Size+keyBytesLen:], i.vr, staticTypeCache)
+	ldbVal := i.iter.Value()
+
+	// skip over 4 bytes of colId and get opKind, and numKeys from bytes 4 & 5
+	opKind := NomsKind(ldbKey[4])
+	numKeys := uint8(ldbKey[5])
+	ldbKey = ldbKey[6:]
+
+	// Call decodeValue for each encoded graphKey
```
willhite: Done.

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
-func (opc *ldbOpCache) SetInsert(val Value) {
-	switch val.Type().Kind() {
-	default:
-		ldbKey := ldbKeyFromValueHash(val, opc.colId)
-		valArray := [initialBufferSize]byte{}
-		encodedVal := encToSlice(val, valArray[:], opc.vrw)
-		err := opc.ldb.Put(ldbKey, encodedVal, nil)
-		d.Chk.NoError(err)
+	// Fold in any non-primitive key values that were stored in ldbVal
+	keys := ValueSlice{}
```
willhite: Wow. Mutability can be a wonderful thing. :-) Done

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
-// hash doesn't contain it. In cases were we are encoding a primitive value into the key, the first byte
-// of the value is the nomsKind so there is no reason to write it again.
-func writeLdbKeyHeaderBytes(ldbKey []byte, colId uint32, v Value) []byte {
-	binary.LittleEndian.PutUint32(ldbKey, colId)
-	length := uint32Size
-	if v != nil {
-		ldbKey[length] = byte(v.Type().Kind())
-		length++
+func encodeKeys(bs []byte, opKind NomsKind, keys []Value, vrw ValueReadWriter) ([]byte, []Value) {
+	// All ldb keys start with a 4-byte collection id that serves as a namespace
```
willhite: Decided to move the insertion of the first colId into the function to keep it all together.

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
+	// that keeps them separate from other collections.
+
+	// ldbKey[4] is a NomsKind value which represents the type of leaf
```
willhite: Fixed.

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
+	//   collection being operated on (i.e. MapKind, SetKind, or ListKind)*[]:
```
willhite: Crazy code auto-completion functionality? Deleted.

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
+	// ldbKey[5] is a single uint8 value representing the number of keys
+	//   encoded in the ldb key.
+	bs = append(bs, byte(opKind), uint8(len(keys)))
```
willhite: Went with byte...

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
+
+	valuesToEncode := []Value{}
```
willhite: Done.

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
 	}
-	return ldbKey[0:length]
+	return bs
 }
 
-// ldbKeys for non-primitive Nom Values (e.g. blobs, structs, lists, maps, etc) use a serialization of the values hash:
-// colId(4 bytes) + nomsKind(val)(1 byte) + val.Hash()(20 bytes).
+func encodeForGraph(bs []byte, v Value, asValue bool, vrw ValueReadWriter) []byte {
+	if asValue || isKindOrderedByValue(v.Type().Kind()) {
+		// if we're encoding value, the put:
```
willhite: Done

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
+		// noms-kind(1-byte), serialization-len(4-bytes), serialization(n-bytes)
+		buf := [initialBufferSize]byte{}
+		uint32buf := [4]byte{}
+		encodedVal := encToSlice(v, buf[:], vrw)
+		binary.BigEndian.PutUint32(uint32buf[:], uint32(len(encodedVal)))
+		bs = append(bs, uint8(v.Type().Kind()))
+		bs = append(bs, uint32buf[:]...)
+		bs = append(bs, encodedVal...)
```
willhite: Done.

go/types/opcache.go
```diff
@@ -80,144 +147,185 @@ func (store *ldbOpCacheStore) opCache() opCache {
...
+		bs = append(bs, v.Hash().DigestSlice()...)
+	}
+	return bs
+}
 
-func ldbKeyFromValueHash(val Value, colId uint32) []byte {
-	ldbKeyArray := [uint32Size + 1 + hash.ByteLen]byte{}
-	ldbKey := writeLdbKeyHeaderBytes(ldbKeyArray[:], colId, val)
-	return append(ldbKey, val.Hash().DigestSlice()...)
+func encodeGraphKey(bs []byte, v Value, vrw ValueReadWriter) []byte {
```
willhite: Done.

go/types/opcache_compare.go
```diff
@@ -6,39 +6,124 @@ package types
...
+	"encoding/binary"
 
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
 )
 
 type opCacheComparer struct{}
 
-func (opCacheComparer) Compare(a, b []byte) int {
+func (opCacheComparer) Compare(a, b []byte) (res int) {
```
willhite: Done.

go/types/opcache_compare.go
```diff
@@ -6,39 +6,124 @@ package types
...
 		}
-		a, b = a[1:], b[1:]
+	}
+	return cres
+}
+
+// compareKey accepts two bytestreams that each contain a number of encoded keys.
+// It compares the first key in the encoding and returns the comparison result for
+// that key along with two slices that hold the bytes following the current key
+func compareEncodedKey(a, b []byte) (res int) {
```
willhite: Fixed.

go/types/opcache_compare.go
```diff
@@ -6,39 +6,124 @@ package types
...
+	// each key is encoded as NomsKind(1-byte) + length(4-bytes) + encoding(n-bytes)
```
willhite: I think the comment is clearer now.

go/types/opcache_compare.go
```diff
@@ -6,39 +6,124 @@ package types
...
+	aKind, bKind := NomsKind(a[0]), NomsKind(b[0])
+	if isKindOrderedByValue(aKind) || isKindOrderedByValue(bKind) {
+		if res := compareKinds(aKind, bKind); res != 0 {
+			return res
+		}
+	}
+
+	// we know the noms kinds are identical. If they are both hashes just compare
+	// the hashes
+	if !isKindOrderedByValue(aKind) && !isKindOrderedByValue(bKind) {
```
willhite: I don't think I can easily switch positions here. The second "simpler" case relies on knowing that the noms kinds are identical.

go/types/opcache_compare.go
```diff
@@ -6,39 +6,124 @@ package types
...
+		a, b := a[1:], b[1:]
 		d.PanicIfFalse(len(a) == hash.ByteLen && len(b) == hash.ByteLen)
 		res := bytes.Compare(a, b)
-		d.PanicIfFalse(res != 0 || aKind == bKind)
+		d.PanicIfFalse(res != 0 || a[0] == b[0])
```
willhite: Yikes. This line looks completely bogus. I've deleted this check as I don't think it's really necessary.

go/types/opcache_compare.go
```diff
@@ -6,39 +6,124 @@ package types
...
+	lenA := binary.BigEndian.Uint32(a[1:5])
+	lenB := binary.BigEndian.Uint32(b[1:5])
+
+	// create a1, b1 slices that just contain encoding
+	a1, b1 := a[1+uint32Size:1+uint32Size+lenA], b[1+uint32Size:1+uint32Size+lenB]
+
+	return compareEncodedValue(a1, b1)
+}
+
+// compareEncodedKey compares two slices. Each slice contains a first byte that
```
willhite: Fixed comment.

go/types/opcache.go
```diff
@@ -193,7 +193,8 @@ func (i *ldbOpCacheIterator) GraphOp() (ValueSlice, NomsKind, sequenceItem) {
 	numKeys := uint8(ldbKey[5])
 	ldbKey = ldbKey[6:]
 
-	// Call decodeValue for each encoded graphKey
+	// Call decodeValue for each encoded graphKey. nil will be appended to
+	// encoded device for any keys that were encoded as hash digests.
```
willhite: Yes. This inspired some minor name changes and an additional comment or two.

willhite:
For some reason, I can't reply to your comment about simpler code but you were right, it is nicer the this way. I made the change and few more tweaks and comments to make it more readable.
@cmasone-attic  PTAL

cmasone-attic:
lgtm

willhite:
@rafael-atticlabs Raf, are you good with this now, too?

rafael-atticlabs:
i defer to chris at this point.

PR #2549, opened 2016-09-12
Refactor Chunks and ChildValues API to work iteratively

zcstarr:
First pass at an iterative api for Chunks, additionally changes include a refactor of the parallel walk apis


zcstarr:
@rafael-atticlabs @aboodman  PTAL. I'm holding off on merging upstream changes, simply because the changes might be a little easier to digest. Looking for feedback on naming and if I should refactor pull perhaps. So for right now I use a helper function to essentially do the same thing pull was doing before.

go/dataset/dataset.go
```diff
@@ -92,9 +92,18 @@ func (ds *Dataset) Commit(v types.Value, opts CommitOptions) (Dataset, error) {
 	}
 
 	meta := opts.Meta
+
+	numChildren := func(val types.Value) uint64 {
+		var childrenCount uint64
+		count := func(val types.Value) {
```
rafael-atticlabs: nit: for these little closures, just put the function inline. e.g.

```
meta.WalkValues(func(val types.Value) uint64 {
...
})
```

rafael-atticlabs:
In general, I think this is good. Couple of things:

You didn't remove `ChildValues()` & `Chunks()` implementation methods from most types, even though you removed them from the `Value.go` interface. Can you do that (and try to line up the edits so that it's clear that `ChildValues` -> `WalkValues`)?

I realize I said in my proposal specifically to move the static `WalkValues` & `WalkRefs` to types, but it's making it hard to see the diffs in the implementation of those functions, and since the instance methods are public, I think it'd prefer to leave those in the `walk` package, at least for this patch so that the diffs make sense.

zcstarr:
pull request is  a duplicate and no longer valid. The outstanding PR has been closed https://github.com/attic-labs/noms/pull/2599

PR #2550, opened 2016-09-12
Introduce List.Concat to the Go API

kalman:
It exploits the chunked structure of Lists to allow concatenating
arbitrarily large Lists.

I've added both functional and perf tests for Concat, and the perf tests
only made sense with tests for building and reading Lists, so now we
have those too.

### DIFF
--- a/go/perf/suite/suite.go
+++ b/go/perf/suite/suite.go
@@ -83,6 +83,7 @@ import (
 	"time"
 
 	"github.com/attic-labs/noms/go/chunks"
+	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/datas"
 	"github.com/attic-labs/noms/go/dataset"
 	"github.com/attic-labs/noms/go/marshal"
@@ -339,16 +340,14 @@ func (suite *PerfSuite) Pause(fn func()) {
 	suite.paused += time.Since(start)
 }
 
-func callSafe(name string, fun reflect.Value, args ...interface{}) (err interface{}) {
-	defer func() {
-		err = recover()
-	}()
+func callSafe(name string, fun reflect.Value, args ...interface{}) error {
 	funArgs := make([]reflect.Value, len(args))
 	for i, arg := range args {
 		funArgs[i] = reflect.ValueOf(arg)
 	}
-	fun.Call(funArgs)
-	return
+	return d.Try(func() {
+		fun.Call(funArgs)
+	})
 }
 
 func (suite *PerfSuite) getEnvironment() types.Value {
--- a/go/types/indexed_sequences.go
+++ b/go/types/indexed_sequences.go
@@ -59,22 +59,6 @@ func (ims indexedMetaSequence) getCompareFn(other sequence) compareFn {
 	}
 }
 
-func newCursorAtIndex(seq indexedSequence, idx uint64) *sequenceCursor {
-	var cur *sequenceCursor
-	for {
-		cur = newSequenceCursor(cur, seq, 0)
-		idx = idx - advanceCursorToOffset(cur, idx)
-		cs := cur.getChildSequence()
-		if cs == nil {
-			break
-		}
-		seq = cs.(indexedSequence)
-	}
-
-	d.Chk.True(cur != nil)
-	return cur
-}
-
 func advanceCursorToOffset(cur *sequenceCursor, idx uint64) uint64 {
 	seq := cur.seq.(indexedSequence)
 	cur.idx = sort.Search(seq.seqLen(), func(i int) bool {
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -155,14 +155,6 @@ func (l List) Insert(idx uint64, vs ...Value) List {
 // visit the rightmost prolly tree chunks of this list, and the leftmost prolly
 // tree chunks of other.
 func (l List) Concat(other List) List {
-	if l.Empty() {
-		return other
-	}
-	if other.Empty() {
-		return l
-	}
-	d.Chk.True(l.seq.valueReader() == other.seq.valueReader())
-
 	seq := concat(l.seq, other.seq, func(cur *sequenceCursor, vr ValueReader) *sequenceChunker {
 		return l.newChunker(cur, vr)
 	})
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -135,7 +135,7 @@ func (l List) Splice(idx uint64, deleteCount uint64, vs ...Value) List {
 	d.Chk.True(idx+deleteCount <= l.Len())
 
 	cur := newCursorAtIndex(l.seq, idx)
-	ch := l.newChunker(cur)
+	ch := l.newChunker(cur, l.seq.valueReader())
 	for deleteCount > 0 {
 		ch.Skip()
 		deleteCount--
@@ -163,8 +163,8 @@ func (l List) Concat(other List) List {
 	}
 	d.Chk.True(l.seq.valueReader() == other.seq.valueReader())
 
-	seq := concat(l.seq, other.seq, func(cur *sequenceCursor) *sequenceChunker {
-		return l.newChunker(cur)
+	seq := concat(l.seq, other.seq, func(cur *sequenceCursor, vr ValueReader) *sequenceChunker {
+		return l.newChunker(cur, vr)
 	})
 	return newList(seq.(indexedSequence))
 }
@@ -235,8 +235,8 @@ func (l List) DiffWithLimit(last List, changes chan<- Splice, closeChan <-chan s
 	indexedSequenceDiff(last.seq, lastCur.depth(), 0, l.seq, lCur.depth(), 0, changes, closeChan, maxSpliceMatrixSize)
 }
 
-func (l List) newChunker(cur *sequenceCursor) *sequenceChunker {
-	return newSequenceChunker(cur, l.seq.valueReader(), nil, makeListLeafChunkFn(l.seq.valueReader()), newIndexedMetaSequenceChunkFn(ListKind, l.seq.valueReader()), hashValueBytes)
+func (l List) newChunker(cur *sequenceCursor, vr ValueReader) *sequenceChunker {
+	return newSequenceChunker(cur, vr, nil, makeListLeafChunkFn(vr), newIndexedMetaSequenceChunkFn(ListKind, vr), hashValueBytes)
 }
 
 // If |sink| is not nil, chunks will be eagerly written as they're created. Otherwise they are
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -27,7 +27,7 @@ func NewList(values ...Value) List {
 	return newList(seq.Done().(indexedSequence))
 }
 
-// NewStreamingList creates a new List, populated with values, chunking if and when needed. As chunks are created, they're written to vrw -- including the root chunk of the list. Once the caller has closed values, she can read the completed List from the returned channel.
+// NewStreamingList creates a new List, populated with values, chunking if and when needed. As chunks are created, they're written to vrw -- including the root chunk of the list. Caller should close the values channel to read completed list.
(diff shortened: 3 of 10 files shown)


go/perf/suite/suite.go
```diff
@@ -339,16 +340,14 @@ func (suite *PerfSuite) Pause(fn func()) {
...
-		err = recover()
-	}()
+func callSafe(name string, fun reflect.Value, args ...interface{}) error {
 	funArgs := make([]reflect.Value, len(args))
 	for i, arg := range args {
 		funArgs[i] = reflect.ValueOf(arg)
 	}
-	fun.Call(funArgs)
-	return
+	return d.Try(func() {
```
kalman: It was losing the stack trace.

go/types/indexed_sequences.go
```diff
@@ -59,22 +59,6 @@ func (ims indexedMetaSequence) getCompareFn(other sequence) compareFn {
 	}
 }
 
-func newCursorAtIndex(seq indexedSequence, idx uint64) *sequenceCursor {
```
kalman: It now works with all sequences, not just indexed sequences, so I moved it to sequence_cursor.go.

go/types/list.go
```diff
@@ -27,7 +27,7 @@ func NewList(values ...Value) List {
 	return newList(seq.Done().(indexedSequence))
 }
 
-// NewStreamingList creates a new List, populated with values, chunking if and when needed. As chunks are created, they're written to vrw -- including the root chunk of the list. Once the caller has closed values, she can read the completed List from the returned channel.
+// NewStreamingList creates a new List, populated with values, chunking if and when needed. As chunks are created, they're written to vrw -- including the root chunk of the list. Caller should close the values channel to read completed list.
```
kalman: Gendered comment seemed unnecessary.

go/types/list.go
```diff
@@ -75,11 +75,12 @@ func (l List) Hash() hash.Hash {
 	return *l.h
 }
 
-func (l List) ChildValues() (values []Value) {
+func (l List) ChildValues() []Value {
+	values := make([]Value, l.Len())
 	l.IterAll(func(v Value, idx uint64) {
-		values = append(values, v)
+		values[idx] = v
```
kalman: Unrelated, just more efficient.

go/types/perf/perf_test.go
```diff
@@ -0,0 +1,99 @@
...
+	assert.NoError(err)
+	s.Database = ds.Database()
+}
+
+func (s *perfSuite) headList(dsName string) types.List {
+	ds := dataset.NewDataset(s.Database, dsName)
+	return ds.HeadValue().(types.List)
+}
+
+func TestPerf(t *testing.T) {
```
kalman: On my laptop:
```
> go test ./go/types/perf -run Perf -v -perf mem
=== RUN   TestPerf
Listening on port 56012...
(perf) RUN(1/1) Test01BuildList10mNumbers (as "BuildList10mNumbers")
(perf) PASS:    Test01BuildList10mNumbers (19.269953152s, paused for 0, total 19.269953152s)
(perf) RUN(1/1) Test02BuildList10mStructs (as "BuildList10mStructs")
(perf) PASS:    Test02BuildList10mStructs (45.171747431s, paused for 0, total 45.171747431s)
(perf) RUN(1/1) Test03Read10mNumbers (as "Read10mNumbers")
(perf) PASS:    Test03Read10mNumbers (18.05143979s, paused for 0, total 18.05143979s)
(perf) RUN(1/1) Test04Read10mStructs (as "Read10mStructs")
(perf) PASS:    Test04Read10mStructs (26.579869908s, paused for 0, total 26.579869908s)
(perf) RUN(1/1) Test05Concat10mValues2kTimes (as "Concat10mValues2kTimes")
(perf) PASS:    Test05Concat10mValues2kTimes (10.425549004s, paused for 0, total 10.425549004s)
--- PASS: TestPerf (119.69s)
PASS
ok  	github.com/attic-labs/noms/go/types/perf	119.817s
```

go/types/list.go
```diff
@@ -150,6 +151,24 @@ func (l List) Insert(idx uint64, vs ...Value) List {
...
+// visit the rightmost prolly tree chunks of this list, and the leftmost prolly
+// tree chunks of other.
+func (l List) Concat(other List) List {
+	if l.Empty() {
+		return other
+	}
+	if other.Empty() {
+		return l
+	}
+	d.Chk.Equal(l.seq.valueReader(), other.seq.valueReader())
```
arv: Don't use d.Chk.Equal. It is slow

go/types/perf/perf_test.go
```diff
@@ -0,0 +1,99 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package perf
```
kalman: Note this needs to be in a separate package because the perf suite relies on the types package, so types can't have a perf suite.

The same will be true for datas, dataset, etc as well. We'll need to use this idiom a bunch.

go/types/sequence_concat.go
```diff
@@ -0,0 +1,80 @@
...
+
+func appendCursorToChunker(chunker *sequenceChunker, cur *sequenceCursor) {
+	// Append beyond the chunk window, then through to the next definite chunk
+	// boundary.
+	for i := uint32(0); cur.valid() && (i < chunkWindow || cur.indexInChunk() != 0); i++ {
+		chunker.Append(cur.current())
+		cur.advance()
+	}
+
+	// There may not be a chunk bounary beyond the chunk window - terminate.
```
arv: typo

go/types/sequence_concat.go
```diff
@@ -0,0 +1,80 @@
...
+	if !cur.valid() {
+		return
+	}
+
+	d.Chk.True(len(chunker.current) == 0)
```
arv: You know that all of these gets boxed right?

go/types/list.go
rafael-atticlabs: ha. nice.

go/types/list.go
kalman: Done

go/types/sequence_concat.go
kalman: Done

go/types/sequence_concat.go
kalman: I don't know what you mean

go/types/sequence_concat.go
```diff
@@ -0,0 +1,80 @@
...
+	// resulting concatenated sequence.
+	chunker := newSequenceChunker(newCursorAtIndex(fst, fst.numLeaves()))
+	appendCursorToChunker(chunker, newCursorAtIndex(snd, 0))
+	return chunker.Done()
+}
+
+func appendCursorToChunker(chunker *sequenceChunker, cur *sequenceCursor) {
+	// Append beyond the chunk window, then through to the next definite chunk
+	// boundary.
+	for i := uint32(0); cur.valid() && (i < chunkWindow || cur.indexInChunk() != 0); i++ {
```
rafael-atticlabs: i don't think this is right. `chunkWindow` is now in bytes. You'll need to compute how many bytes each cursor item represents. SequenceChunker has a similar issue for the case when it re-chunks an existing ptree.

rafael-atticlabs: I think this loop should look pretty similar to the one in `finalize` since they are kind of doing the same thing. not sure if it's possible to abstract and use common code.

go/types/sequence_concat.go
arv: Most of the testify functions take `interface{}` so a boxed value is created in every case (actually True takes a bool) which is why we should not use d.Chk.<TestifyFunction>.

go/types/sequence_concat.go
```diff
@@ -0,0 +1,80 @@
...
+		chunker.Append(cur.current())
+		cur.advance()
+	}
+
+	// There may not be a chunk bounary beyond the chunk window - terminate.
+	if !cur.valid() {
```
rafael-atticlabs: this worries me a bit. what if the final item in a sequence is also an explicit chunk boundary?

go/types/sequence_concat.go
```diff
@@ -0,0 +1,80 @@
...
+	// beyond the chunk window, to solve Incorrect Reason 2, then (b) through to
+	// the next definite chunk boundary of snd, to solve Reason 1.
+	//
+	// Implementation-wise wise we can simply reuse sequenceChunker re-chunking
+	// logic, which automatically steps back to the start of the previous chunk
+	// (and can correctly prime the chunk window, etc), plus how to construct that
+	// resulting concatenated sequence.
+	chunker := newSequenceChunker(newCursorAtIndex(fst, fst.numLeaves()))
+	appendCursorToChunker(chunker, newCursorAtIndex(snd, 0))
+	return chunker.Done()
```
rafael-atticlabs: hmm.. this doesn't seem right to me. if you call Done() on the root chunker (L0 of the tree), it's going to call finalize, but in some sense this level has already been finalized.

rafael-atticlabs:
you've thought alot about this, so my comments may be wrong. think about it. if i'm right, then I have an idea how to approach.

go/types/sequence_concat.go
kalman: Ok, well your original comment wasn't possible to understand.

In any case what should I do instead?

go/types/sequence_concat.go
kalman: Yes. Egregious bug, thank you.

kalman: I can't imagine trying to have common code would help readability, though :-\

go/types/sequence_concat.go
kalman: ITYM "leaf" not "root".

The finalise thing doesn't matter on the leaf because the fst cursor points to the end of the sequence, so there is nothing to finalise.

But I tested, and Done does need to be called on the leaf chunker (I tried calling it on the root, and the test fails). This is what made most sense to me at the time, but come to think of it, I can see it might seem weird: we've been circumventing the assumptions that sequenceChunker makes in its Append flow. The parent is the only sequenceChunker that hasn't been circumvented.

As to why it fails on the parent: I'll give this a bit of thought tomorrow :-)

go/types/sequence_concat.go
kalman: Not sure how that affects anything, I would assume sequenceChunker knows how to deal with it. Perhaps I'm misunderstanding/missing something.

kalman:
Done, we can discuss other comments tomorrow.

kalman:
jenkins: perf this

kalman:
PTAL, I incorporated raf's idea from this morning. You will probably just want to review the whole PR again, not the individual patches (it's a pretty small PR).

kalman:
jenkins: perf this

go/types/list.go
kalman: (FYI arv is about to fix this in a separate patch)

kalman:
PTAL

go/types/sequence_concat.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	if snd.numLeaves() == 0 {
+		return fst
+	}
+
+	// concat works by tricking the sequenceChunker into resuming chunking at a
+	// cursor to the end of fst, then finalizing chunking to the start of snd - by
+	// swapping fst cursors for snd cursors in the middle of chunking.
+	vr := fst.valueReader()
+	d.PanicIfTrue(vr != snd.valueReader(), "cannot concat sequences from different databases")
```
rafael-atticlabs: Excellent check.

rafael-atticlabs:
LGTM. I kind of want to frame this patch and put it on the wall.

ghost (approved):
lgtm

PR #2553, opened 2016-09-14
Add jsdoc/godoc for Hash and Value

arv:
Towards #2297

### DIFF
--- a/go/hash/hash.go
+++ b/go/hash/hash.go
@@ -39,7 +39,10 @@ import (
 )
 
 const (
-	ByteLen   = 20
+	// ByteLen is the number of bytes used to represent the Hash.
+	ByteLen = 20
+
+	// StringLen is the number of characters need to represent the Hash using Base32.
 	StringLen = 32 // 20 * 8 / log2(32)
 )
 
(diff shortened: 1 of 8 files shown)


go/hash/hash.go
```diff
@@ -107,25 +121,32 @@ func Parse(s string) Hash {
...
 }
 
+// Less compares two hashes returning whether this Hash is less than other.
 func (r Hash) Less(other Hash) bool {
 	return bytes.Compare(r.digest[:], other.digest[:]) < 0
 }
 
+// Greater compares two hashes returning whether this Hash is greater than other.
 func (r Hash) Greater(other Hash) bool {
+	// TODO: Remove this
```
mikegray: Hmm, did you mean to add this comment here?

go/types/value.go
```diff
@@ -8,15 +8,31 @@ import (
...
 )
 
-// Value is implemented by every noms value
+// Value is the interface all Noms values implement.
 type Value interface {
+	// Equals determines if two different Noms values represents the same underlying value.
 	Equals(other Value) bool
+
+	// Less determines if this Noms value is less than another Noms value.
+	// When comparing two Noms values and both are comparable and the same type (Bool, Number or
```
mikegray: Should we document the details of the implementation here or maybe just leave that to the documentation of the implementations of Less?

mikegray:
lgtm with few open questions.

go/hash/hash.go
arv: Yes. I didn't want to change the API in this PR.

Having Less and Greater is pretty silly. I'll open an issue.

go/types/value.go
arv: I wanted this somewhere. It is not clear where it belongs though.

PR #2554, opened 2016-09-14
Add jsdoc/godoc for List

arv:
Towards #2297

### DIFF
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -116,8 +116,10 @@ func (l List) Get(idx uint64) Value {
 
 type MapFunc func(v Value, index uint64) interface{}
 
+// Deprecated: This API may change in the future. Use IterAll or Iterator instead.
 func (l List) Map(mf MapFunc) []interface{} {
 	// TODO: This is bad API. It should have returned another List.
+	// https://github.com/attic-labs/noms/issues/2557
 	idx := uint64(0)
 	cur := newCursorAtIndex(l.seq, idx)
 
@@ -142,7 +144,7 @@ func (l List) Set(idx uint64, v Value) List {
 	return l.Splice(idx, 1, v)
 }
 
-// Append returns a new list where vs have bben appended to the resulting list.
+// Append returns a new list where vs have been appended to the resulting list.
 func (l List) Append(vs ...Value) List {
 	return l.Splice(l.Len(), 0, vs...)
 }
@@ -177,7 +179,7 @@ func (l List) Insert(idx uint64, vs ...Value) List {
 }
 
 // Remove returns a new list where the items at index start (inclusive) through end (exclusive) have
-// been removed.
+// been removed. This panics if end is smaller than start.
 func (l List) Remove(start uint64, end uint64) List {
 	d.Chk.True(start <= end)
 	return l.Splice(start, end-start)
@@ -207,7 +209,7 @@ func (l List) Iter(f listIterFunc) {
 type listIterAllFunc func(v Value, index uint64)
 
 // IterAll iterates over the list and calls f for every element in the list. Unlike Iter there is no
-// way to stop the iteration and all elements are visited no matter what.
+// way to stop the iteration and all elements are visited.
 func (l List) IterAll(f listIterAllFunc) {
 	// TODO: Consider removing this and have Iter behave like IterAll.
 	idx := uint64(0)
@@ -224,17 +226,20 @@ func (l List) Iterator() ListIterator {
 	return l.IteratorAt(0)
 }
 
-// IteratorAt returns a ListIterator starting at index.
+// IteratorAt returns a ListIterator starting at index. If index is out of bound the iterator will
+// have reached its end on creation.
 func (l List) IteratorAt(index uint64) ListIterator {
 	return ListIterator{newCursorAtIndex(l.seq, index)}
 }
 
-// Diff returns the diff of two different lists.
+// Diff streams the diff from last to the current list to the changes channel. Caller can close
+// closeChan to cancel the diff operation.
 func (l List) Diff(last List, changes chan<- Splice, closeChan <-chan struct{}) {
 	l.DiffWithLimit(last, changes, closeChan, DEFAULT_MAX_SPLICE_MATRIX_SIZE)
 }
 
-// DiffWithLimit returns a partial diff of two different lists.
+// DiffithLimit streams the diff from last to the current list to the changes channel. Caller can
+// close closeChan to cancel the diff operation.
 func (l List) DiffWithLimit(last List, changes chan<- Splice, closeChan <-chan struct{}, maxSpliceMatrixSize uint64) {
 	if l.Equals(last) {
 		return
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -151,7 +151,7 @@ func (l List) Append(vs ...Value) List {
 
 // Splice returns a new list where deleteCount values have been removed at idx and vs have been
 // inserted instead.
-// This function panics if idx is out of bounds or too many items are tried to be removed.
+// This function panics if idx or deleteCount is out of bounds.
 func (l List) Splice(idx uint64, deleteCount uint64, vs ...Value) List {
 	if deleteCount == 0 && len(vs) == 0 {
 		return l
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -9,6 +9,15 @@ import (
 	"github.com/attic-labs/noms/go/hash"
 )
 
+// List represents a list or an array of Noms values. A list can contain zero or more values of zero
+// or more types. The type of the list will reflect the type of the elements in the list. For
+// example:
+//
+//  l := NewList(Number(1), Bool(true))
+//  fmt.Println(l.Type().Describe())
+//  // outputs List<Bool | Number>
+//
+// Lists, like all Noms values are immutable so the "mutation" methods return a new list.
 type List struct {
 	seq indexedSequence
 	h   *hash.Hash
@@ -18,7 +27,8 @@ func newList(seq indexedSequence) List {
 	return List{seq, &hash.Hash{}}
 }
 
-// NewList creates a new List where the type is computed from the elements in the list, populated with values, chunking if and when needed.
+// NewList creates a new List where the type is computed from the elements in the list, populated
+// with values, chunking if and when needed.
 func NewList(values ...Value) List {
 	seq := newEmptySequenceChunker(nil, nil, makeListLeafChunkFn(nil), newIndexedMetaSequenceChunkFn(ListKind, nil), hashValueBytes)
 	for _, v := range values {
@@ -27,7 +37,9 @@ func NewList(values ...Value) List {
 	return newList(seq.Done().(indexedSequence))
 }
 
-// NewStreamingList creates a new List, populated with values, chunking if and when needed. As chunks are created, they're written to vrw -- including the root chunk of the list. Once the caller has closed values, she can read the completed List from the returned channel.
+// NewStreamingList creates a new List, populated with values, chunking if and when needed. As
+// chunks are created, they're written to vrw -- including the root chunk of the list. Once the
+// caller has closed values, the caller can read the completed List from the returned channel.
 func NewStreamingList(vrw ValueReadWriter, values <-chan Value) <-chan List {
 	out := make(chan List)
 	go func() {
@@ -42,10 +54,13 @@ func NewStreamingList(vrw ValueReadWriter, values <-chan Value) <-chan List {
 }
 
 // Collection interface
+
+// Len returns the number of elements in the list.
 func (l List) Len() uint64 {
 	return l.seq.numLeaves()
 }
 
+// Empty returns true if the list is empty (length is zero).
 func (l List) Empty() bool {
 	return l.Len() == 0
 }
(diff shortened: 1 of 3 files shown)


go/types/list.go
```diff
@@ -27,7 +37,9 @@ func NewList(values ...Value) List {
 	return newList(seq.Done().(indexedSequence))
 }
 
-// NewStreamingList creates a new List, populated with values, chunking if and when needed. As chunks are created, they're written to vrw -- including the root chunk of the list. Once the caller has closed values, she can read the completed List from the returned channel.
+// NewStreamingList creates a new List, populated with values, chunking if and when needed. As
+// chunks are created, they're written to vrw -- including the root chunk of the list. Once the
```
kalman: Thanks

go/types/list.go
```diff
@@ -99,6 +117,7 @@ func (l List) Get(idx uint64) Value {
 type MapFunc func(v Value, index uint64) interface{}
 
 func (l List) Map(mf MapFunc) []interface{} {
+	// TODO: This is bad API. It should have returned another List.
```
kalman: So, not documenting it because it's not a good API and we want to change it at some point? That... makes sense to me to some extent, but IMO it would be better to just say that in the godoc with a link to the issue (assuming an issue exists).

go/types/list.go
```diff
@@ -116,15 +135,21 @@ func (l List) elemType() *Type {
 	return l.seq.Type().Desc.(CompoundDesc).ElemTypes[0]
 }
 
+// Set returns a new list where the valie at the given index have been replaced with v. If idx is
+// out bounds then this panics.
 func (l List) Set(idx uint64, v Value) List {
 	d.Chk.True(idx < l.Len())
```
kalman: Incidentally I notice there are only 5 instances of d.Chk (True) in this file, if you wanted to replace them with PanicIfTrue now. Makes sense, sort of.

go/types/list.go
```diff
@@ -116,15 +135,21 @@ func (l List) elemType() *Type {
...
 	return l.Splice(idx, 1, v)
 }
 
+// Append returns a new list where vs have bben appended to the resulting list.
```
kalman: bben --> been

go/types/list.go
```diff
@@ -116,15 +135,21 @@ func (l List) elemType() *Type {
...
 func (l List) Append(vs ...Value) List {
 	return l.Splice(l.Len(), 0, vs...)
 }
 
+// Splice returns a new list where deleteCount values have been removed at idx and vs have been
+// inserted instead.
+// This function panics if idx is out of bounds or too many items are tried to be removed.
```
kalman: You could also say "This function panics if idx or idx + deleteCount is out of bounds."

go/types/list.go
```diff
@@ -146,21 +171,27 @@ func (l List) Splice(idx uint64, deleteCount uint64, vs ...Value) List {
 	return newList(ch.Done().(indexedSequence))
 }
 
+// Insert returns a new list where vs values have been inserted at idx.
 func (l List) Insert(idx uint64, vs ...Value) List {
 	return l.Splice(idx, 0, vs...)
 }
 
+// Remove returns a new list where the items at index start (inclusive) through end (exclusive) have
+// been removed.
```
kalman: panics if.... (we should be consistent)

go/types/list.go
```diff
@@ -175,7 +206,10 @@ func (l List) Iter(f listIterFunc) {
 
 type listIterAllFunc func(v Value, index uint64)
 
+// IterAll iterates over the list and calls f for every element in the list. Unlike Iter there is no
+// way to stop the iteration and all elements are visited no matter what.
```
kalman: I would leave out "and all elements are visited no matter what", or at least the "no matter what".

go/types/list.go
```diff
@@ -175,7 +206,10 @@ func (l List) Iter(f listIterFunc) {
...
 func (l List) IterAll(f listIterAllFunc) {
+	// TODO: Consider removing this and have Iter behave like IterAll.
```
kalman: What do you mean?

go/types/list.go
mikegray: How about using `Deprecated`? For example: https://golang.org/src/net/http/httputil/persist.go?s=918:1235#L31

go/types/list.go
```diff
@@ -185,18 +219,22 @@ func (l List) IterAll(f listIterAllFunc) {
 	})
 }
 
+// Iterator returns a ListIterator which can be used to iterate efficiently over a list.
 func (l List) Iterator() ListIterator {
 	return l.IteratorAt(0)
 }
 
+// IteratorAt returns a ListIterator starting at index.
```
kalman: You might want to say, "if index is out of bounds, returns an iterator which returns nil as its next value"

go/types/list.go
```diff
@@ -185,18 +219,22 @@ func (l List) IterAll(f listIterAllFunc) {
...
 func (l List) IteratorAt(index uint64) ListIterator {
 	return ListIterator{newCursorAtIndex(l.seq, index)}
 }
 
+// Diff returns the diff of two different lists.
```
kalman: Diff streams the diff from last to l to changes. Caller can close closeChan to cancel the diff operation.

go/types/list.go
```diff
@@ -185,18 +219,22 @@ func (l List) IterAll(f listIterAllFunc) {
...
 func (l List) Diff(last List, changes chan<- Splice, closeChan <-chan struct{}) {
 	l.DiffWithLimit(last, changes, closeChan, DEFAULT_MAX_SPLICE_MATRIX_SIZE)
 }
 
+// DiffWithLimit returns a partial diff of two different lists.
```
kalman: I'm not entirely sure what the "limit" is referring to here, but I don't think it's related to being a partial diff. Either way I would just duplicate the comment.

Or hide this method from the public interface, @rafael-atticlabs ?

go/types/list_iterator.go
```diff
@@ -8,12 +8,14 @@ import (
 	"github.com/attic-labs/noms/go/d"
 )
 
-// ListIterator can be used to efficiently iterate through a Noms List starting at an index of your choice.
+// ListIterator can be used to efficiently iterate through a Noms List starting at an index of your
```
kalman: I would drop the "starting at an index of your
 +// choice"

js/noms/src/list.js
```diff
@@ -35,6 +35,17 @@ function newListLeafChunkFn<T: Value>(vr: ?ValueReader): makeChunkFn<any, any> {
   };
 }
 
+/**
+ * List represents a list or an array of Noms values. A list can contain zero or more values of zero
+ * or more types. The type of the list will reflect the type of the elements in the list. For
+ * example:
+ *
+ *  const l = new List([1, true]);
+ *  console.log(l.type.escribe());
```
kalman: escribe -> describe

kalman:
I assume some of the comments I made on the Go docs apply to JS as well.

kalman:
Back to you

go/types/list.go
arv: Done.

go/types/list.go
mikegray: The `limit` is how big of an edit distance matrix we are willing to compute versus just saying the thing changed.

go/types/list.go
arv: I didn't want to go too much into specifics here.

kalman: The comment as it stands makes it sound like there is some limit to the number of items that can be removed. Like, you can only remove 42 elements at a time or something. My suggestion is trying to address that.

go/types/list.go
arv: Done.

go/types/list.go
arv: I can do it in another PR. I would prefer to keep this a documentation only change

go/types/list.go
arv: Done/

go/types/list.go
arv: Removed "no matter what".

go/types/list.go
arv: I don't like that we have two version here. One should be enough.

go/types/list.go
arv: Done.

go/types/list.go
arv: Done.

I copied the comment.

go/types/list_iterator.go
arv: Done.

arv:
PTAL

go/types/list.go
kalman: SGTM

go/types/list.go
kalman: What about just "This function panics if idx or deleteCount is out of bounds"

go/types/list.go
kalman: I agree, and have thought the same thing. IterAll saves you an extra line (a return statement).

Though I think an issue, rather than a TODO, is less likely to get lost.

go/types/list.go
```diff
@@ -224,17 +226,20 @@ func (l List) Iterator() ListIterator {
...
 
-// Diff returns the diff of two different lists.
+// Diff streams the diff from last to the current list to the changes channel. Caller can close
+// closeChan to cancel the diff operation.
 func (l List) Diff(last List, changes chan<- Splice, closeChan <-chan struct{}) {
 	l.DiffWithLimit(last, changes, closeChan, DEFAULT_MAX_SPLICE_MATRIX_SIZE)
 }
 
-// DiffWithLimit returns a partial diff of two different lists.
+// DiffithLimit streams the diff from last to the current list to the changes channel. Caller can
```
kalman: DiffithLimit -> Diff**W**ithLimit

js/noms/src/list.js
```diff
@@ -118,20 +165,30 @@ export default class List<T: Value> extends Collection<IndexedSequence<any>> {
...
+/**
+ * ListLeafSequence is used for the leaf lists of a list prolly-tree.
+ */
 export class ListLeafSequence<T: Value> extends IndexedSequence<T> {
   get chunks(): Array<Ref<any>> {
     return getValueChunks(this.items);
   }
 
   cumulativeNumberOfLeaves(idx: number): number {
+    // TODO: What does this do? Document me!
```
kalman: just rename this function to _cumulativeNumberOfLeaves ...?

kalman: Though to answer your question, this is an interface method which sequences implement, that returns how many leaf items there are up to an index within its sequence. For a leaf, this is trivially `idx + 1`.

For non-leaves it is more complex of course.

go/types/list.go
arv: Added an issue and linked to it from here.

js/noms/src/list.js
arv: Done.

js/noms/src/list.js
arv: Can't really rename it to _cumulativeNumberOfLeaves. We want _ to mean private and we might enforce this in babel/flow.

I ended up with:


```js
/**
 * This method is for internal use of sequences. It returns how many leaf items there are up to an
 * index within its sequence.
 */
```

kalman: SGTM

go/types/list.go
arv: Oops

js/noms/src/list.js
```diff
@@ -80,14 +114,24 @@ export default class List<T: Value> extends Collection<IndexedSequence<any>> {
...
+  /**
+   * Returns a new `AsyncIterator` starting at `i` which can be used to iterate over the list.
+   */
   iteratorAt(i: number): AsyncIterator<T> {
     return new IndexedSequenceIterator(this.sequence.newCursorAt(i));
   }
 
+  /**
+   * Diff returns the diff of two different lists. If `maxSpliceMatrixSize` is provided then a
+   * partial diff might be returned.
```
kalman: Remove the sentence about partial diff

arv:
Better?

js/noms/src/list.js
```diff
@@ -59,18 +78,33 @@ export default class List<T: Value> extends Collection<IndexedSequence<any>> {
...
     return this.splice(this.length, 0, ...values);
   }
 
+  /**
+   * This calls a function for every value in the list. This function is async and returns a promise
+   * that will be fulfilled when all the callback functions have been called. If the callback
+   * function returns a promise `forEach` will continue but it will not return until all of those
+   * promises have been fulfilled.
+   */
   async forEach(cb: (v: T, i: number) => ?Promise<void>): Promise<void> {
```
kalman: Unrelated, but this reminded me: we should make the callback type `(v: T, i: number) => ?Promise<any>` not `(v: T, i: number) => ?Promise<void>`

kalman:
lgtm with a last couple comments

kalman:
> better?
yesp

js/noms/src/list.js
arv: Agreed... Second follow up.

go/types/list.go
```diff
@@ -98,7 +116,10 @@ func (l List) Get(idx uint64) Value {
 
 type MapFunc func(v Value, index uint64) interface{}
 
+// Deprecated: This API may change in the future. Use IterAll or Iterator instead.
 func (l List) Map(mf MapFunc) []interface{} {
+	// TODO: This is bad API. It should have returned another List.
```
kalman: It turns out https://github.com/attic-labs/noms/issues/2559 might be a better issue for this

PR #2562, opened 2016-09-14
Change type of callback for List/Set/Map forEach

arv:
No need to restrict what type of promise the callback returns


arv:
Followup 1/2

kalman:
lgtm

PR #2563, opened 2016-09-14
Replace d.Chk.True with d.PanicIfFalse

arv:
And same for d.Chk.False

### DIFF
--- a/go/d/try.go
+++ b/go/d/try.go
@@ -33,22 +33,22 @@ func PanicIfError(err error) {
 	}
 }
 
+func argsToError(def string, args ...interface{}) error {
+	if len(args) == 0 {
+		return errors.New(def)
+	}
+	msg := args[0].(string)
+	if len(args) > 1 {
+		return fmt.Errorf(msg, args[1:]...)
+	}
+	return errors.New(msg)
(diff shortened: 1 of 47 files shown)


go/d/try.go
```diff
@@ -34,19 +33,37 @@ func PanicIfError(err error) {
...
 }
 
+// If b is false, creates an error using args, wraps it and panics. If args is provided the first
+// value must be a string.
+func PanicIfFalse(b bool, args ...interface{}) {
+	if !b {
+		if len(args) == 0 {
+			PanicIfTrue(!b, "Expected false")
+		} else {
+			PanicIfTrue(!b, args...)
```
kalman: The disadvantage of this is that the stack trace will include PanicIfTrue. IMO it would be better to have a helper function which turns `args` into an error, then have like
```
func PanicIfTrue(b bool, args ...interface{}) {
  if b {
    panic(makeError(args))
  }
}

func PanicIfFalse(b bool, args ...interface{}) {
  if !b {
    panic(makeError(args))
  }
}

arv: That would be better. Changing.

Issue #2348, opened 2016-08-11
Build script for noms binary should be in noms repo


mikegray:
Currently AWS CodePipeline is using our Jenkins install to build DemoServer which does also build attic-labs/noms repo, but does not publish the noms or other binaries. The code for building noms is in the private DemoServer repository and not in the public noms repo.

Issue #2306, opened 2016-08-06
Have (continuously deployed) pre-built binaries for download


rafael-atticlabs:

ahl:
This would be great to see.

PR #2564, opened 2016-09-14
Build go binaries for multiple platforms

mikegray:
Towards #2306 and #2348.


tools/run_go_bin_builder.py
```diff
@@ -0,0 +1,122 @@
...
+            'GOARCH': platform_match.group('arch'),
+            'GOPATH': workspace,
+            'PATH': cmd_path,
+        })
+
+        # Using 'go build' instead of 'go install' per the recommendation:
+        # http://dave.cheney.net/2015/08/22/cross-compilation-with-go-1-5
+        for pkg in PACKAGES:
+            # cmd: go build -o outputFile -ldFlags "-X file.Constant=value" package
+            cmd = ([get_go_bin_string(), 'build', '-o',
```
kalman: You shouldn't need the parens here.

tools/run_go_bin_builder.py
```diff
@@ -0,0 +1,122 @@
...
+PLATFORMS = [
+    'darwin-amd64',
+    'linux-amd64',
+    'linux-arm',
+]
+
+# The list of Go packages for which we should build binaries
+PACKAGES = [
+    './cmd/noms',
+    './cmd/noms-ui',
```
kalman: I wouldn't worry about noms-ui

tools/run_go_bin_builder.py
```diff
@@ -0,0 +1,122 @@
...
+def get_go_install_path():
+    """Returns the full path to the go installation location"""
+    go_install_path = '/usr/local/go/bin'
+    if platform.system() == "Darwin":
+        go_install_path = '/usr/local/bin'
+    assert os.path.exists(go_install_path)
+    return go_install_path
+
+def get_go_bin_string():
+    """Returns the full path to the go binary"""
```
kalman: Ideally I think we should look up go in the PATH (and hopefully there is a lib function for this), then fall back to get_go_install_path().

tools/run_go_bin_builder.py
```diff
@@ -0,0 +1,122 @@
...
+
+def get_go_install_path():
+    """Returns the full path to the go installation location"""
+    go_install_path = '/usr/local/go/bin'
+    if platform.system() == "Darwin":
+        go_install_path = '/usr/local/bin'
+    assert os.path.exists(go_install_path)
+    return go_install_path
+
+def get_go_bin_string():
```
kalman: Do you need this method? You're adding the go install path to the PATH before running the cmd anyway. I would think it's more correct to respect any other go binary in the path anyway.

tools/run_go_bin_builder.py
```diff
@@ -0,0 +1,122 @@
...
+
+    noms_output_dir = os.path.join(noms_src_dir, 'build_output')
+    if os.path.exists(noms_output_dir):
+        shutil.rmtree(noms_output_dir)
+    os.mkdir(noms_output_dir)
+    assert os.path.exists(noms_output_dir)
+
+    cmd_path = '%s:%s' % (os.getenv('PATH'), get_go_install_path())
+
+    platform_pattern = re.compile(r'^(?P<os>\w+)-(?P<arch>\w+)$')
```
kalman: Instead of using re can you just declare:
```python
PLATFORMS = [
    ('darwin', 'amd64'),
    ('linux', 'amd64'),
    ('linux', 'arm'),
]
```
then
```python
for os, arch in PLATFORMS:
  platform_output_dir = os.path.join(noms_output_dir, '%s-%s' % (os, arch))
  ...
```

tools/run_go_bin_builder.py
```diff
@@ -0,0 +1,122 @@
...
+        })
+
+        # Using 'go build' instead of 'go install' per the recommendation:
+        # http://dave.cheney.net/2015/08/22/cross-compilation-with-go-1-5
+        for pkg in PACKAGES:
+            # cmd: go build -o outputFile -ldFlags "-X file.Constant=value" package
+            cmd = ([get_go_bin_string(), 'build', '-o',
+                    os.path.join(platform_output_dir, os.path.basename(pkg)),
+                    '-ldflags',
+                    '-X github.com/attic-labs/noms/go/constants.NomsGitSHA=' + noms_rev,
```
kalman: Awesome :-)

tools/run_go_bin_builder.py
```diff
@@ -0,0 +1,122 @@
...
+                    pkg])
+            call_with_env_and_cwd(cmd, env, noms_src_dir)
+
+        platform_targz_file = os.path.join(noms_output_dir, 'noms-%s-%s' % (noms_rev, osarch))
+        shutil.make_archive(platform_targz_file, 'gztar', platform_output_dir)
```
kalman: Does this automatically add the .tar.gz extension, or do you need to add it yourself?

tools/run_go_bin_builder.py
```diff
@@ -0,0 +1,122 @@
...
+        shutil.rmtree(platform_output_dir)
```
kalman: You could also just rmtree the whole noms_output_dir at the end.

And that gives me an idea: it would be good to write an integration test for this. Just testing the existence of the tar.gz files would be enough.

tools/run_go_bin_builder.py
mikegray: it adds it

tools/run_go_bin_builder.py
mikegray: Hmm, don't want to remove the whole output dir since I need Jenkins to have access to the .tar.gz files generated by the build to upload to S3 (or Github) after this script finishes.

tools/run_go_bin_builder.py
mikegray: true, done. (learning python on the fly here.)

tools/run_go_bin_builder.py
mikegray: You are correct. Removed.

tools/run_go_bin_builder.py
mikegray: Sounds good. Done.

tools/run_go_bin_builder.py
```diff
@@ -74,6 +57,8 @@ def main():
     # Git SHA revision identifier to insert into built binaries
     noms_rev = os.getenv('NOMS_REVISION')
     assert noms_rev
+    if len(noms_rev) > 7:
```
kalman: You don't need this check (incidentally... why 7?)

mikegray: 'git describe --always' uses that so felt it was a good cut-off point

tools/run_go_bin_builder.py
```diff
@@ -84,34 +69,28 @@ def main():
...
         assert os.path.exists(platform_output_dir)
 
         env = copy.copy(os.environ)
         env.update({
-            'GOOS': platform_match.group('os'),
-            'GOARCH': platform_match.group('arch'),
+            'GOOS': osname,
+            'GOARCH': cpuarch,
             'GOPATH': workspace,
-            'PATH': cmd_path,
```
kalman: I think the PATH stuff was right though? It was using an absolute path to the binary that was unnecessary.

kalman: Eh, so long as Jenkins is configured properly, it shouldn't matter. Never mind.

mikegray: I think it is probably better to have the PATH setup outside of this script.

mikegray:
@kalman PTAL if you wouldn't mind. Added some more precise file/dir exists tests as well as checking the products were built after executing the build and archive commands.

PR #2566, opened 2016-09-14
Implement UTF-8 encode/decode

### LINKED ISSUE #2304: implement streaming utf8 decoder in js
i.e. `decodeUtf8(str: string, fn: (b: number) => void)`

if we had this would be unify a bunch of js code and avoid a bunch of allocations.

arv:
Fixes #2304


js/noms/package.json
```diff
@@ -42,7 +41,7 @@
   "scripts": {
     "pretest": "eslint src/ && flow src/",
     "ts": "mocha --ui tdd --timeout=4000 --reporter spec --slow=50 --invert --grep='LONG:' --compilers js:babel-core/register src/*-test.js",
-    "test": "mocha --ui tdd --timeout=6000 --reporter dot --compilers js:babel-core/register src/*-test.js",
+    "test": "mocha --ui tdd --timeout=10000 --reporter dot --compilers js:babel-core/register src/*-test.js",
```
arv: This is unrelated to this change but the slow tests were timing out on my machine (it was super slow due to Apple update :cry:)

js/noms/src/browser/bytes.js
```diff
@@ -20,11 +18,14 @@ export function fromValues(values: number[]): Uint8Array {
 }
 
 export function fromString(s: string): Uint8Array {
-  return encoder.encode(s);
+  const size = byteLength(s);
+  const buf = alloc(size);
```
arv: One other option would be to have a scratch buffer here so that we do not have to compute the `byteLength` in advance.

arv (commented):
PTAL

js/noms/src/bytes.js
```diff
@@ -75,16 +75,13 @@ export function readUtf8(buff: Uint8Array, start: number, end: number): string {
   return buff.toString('utf8', start, end);
 }
 
-export function encodeUtf8(str: string, buff: Uint8Array, dv: DataView, offset: number): number {
-  const size = Buffer.byteLength(str);
+export function encodeUtf8(str: string, buff: Uint8Array, offset: number): number {
```
arv: This used to write the length followed by the bytes. I felt like that was the wrong abstraction model.

arv (commented):
This is now ready for review

arv:
Ping!

js/noms/src/bytes.js
rafael-atticlabs: yup. this is cleaner. that was mostly about saving allocations (I forget how exactly).

js/noms/src/rolling-value-hasher.js
```diff
@@ -133,20 +133,23 @@ export default class RollingValueHasher {
...
-      this.bytesHashed += 4 + buff.byteLength;
+      this.bytesHashed += 4 + len;
       return;
     }
 
-    this.writeUint32(buff.byteLength);
-    for (let i = 0; i < buff.byteLength; i++) {
+    this.writeUint32(len);
+
+    const buff = getScratchBuffer(len);
```
rafael-atticlabs: nice.

ghost (approved):
lgtm

PR #2569, opened 2016-09-14
Move cumulativeNumberOfLeaves method into sequence

kalman:


go/types/indexed_sequences.go
```diff
@@ -4,20 +4,14 @@
...
-)
+import "github.com/attic-labs/noms/go/d"
 
 type indexedSequence interface {
 	sequence
-	cumulativeNumberOfLeaves(idx int) uint64 // returns the total number of leaf values reachable from this sequence for all sub-trees from 0 to |idx|
 }
 
 type indexedMetaSequence struct {
 	metaSequenceObject
```
kalman: It would be great to collapse the `indexedMetaSequence` and `orderedMetaSequence` into just the `metaSequence` struct, but `oMS` has `getKey()`.

Which, you could put on `metaSequence`, via putting on `sequence`, which *still* makes sense - keys are a sequence concept, not an ordered sequence concept - *but* they're used for binary searching, which indexed sequences don't support.

And *then* you could just make `metaSequence` a struct, not an interface.

Whew. Thoughts? All I really need now is `cumulativeNumberOfLeaves()` on `sequence`.

go/types/sequence.go
```diff
@@ -16,4 +16,5 @@ type sequence interface {
 	Chunks() []Ref
 	Type() *Type
 	getCompareFn(other sequence) compareFn
+	cumulativeNumberOfLeaves(idx int) uint64 // returns the total number of leaf values reachable from this sequence for all sub-trees from 0 to |idx|
```
kalman: Another thought... maybe we should have a `leafSequence` struct? You could save a bunch of boilerplate:
```go
type leafSequence struct {
  length int
  t *Type
  vr ValueReader
}

// implement seqLen, numLeaves, valueReader, Type, cumNumLeaves
```

kalman:
I should port this to JS in the same patch, though I'll wait for your feedback.

Issue #2565, opened 2016-09-14
What if the optimistic lock always fails?


cmasone-attic:
Every Database is ultimately backed by a ChunkStore, which has a single Root. Many clients might be committing to a Database at the same time, but only one can be moving the Root around at any given moment. It's possible that, given enough simultaneous committers, some client might be starved.

We should fix that.

FYI @rafael-atticlabs

PR #2576, opened 2016-09-15
Add comment referencing potential infinite Commit loop

cmasone-attic:
This comment was meant to be part of a previous patch. It points
out the code to which bug #2565 applies.


cmasone-attic:
TBR. Comment change only.

rafael-atticlabs:
lgtm

Issue #1605, opened 2016-05-23
Include code coverage output in builds


mikegray:
- not a goal: 100% code coverage
- not a goal: writing tests that don't make sense just to improve code coverage percentage
- is a goal: do not get worse over time with code coverage percentage
- is a goal: identify areas not being tested that should be

For go - code coverage information seems pretty easy to obtain and has a useful heatmap and source navigation to easily see what code is not being exercised:
go test -coverprofile=coverage.out
go tool cover -html=coverage.out

See [https://blog.golang.org/cover](https://blog.golang.org/cover) for additional information.

arv:
I did some quick research:

```
npm i nyc --save-dev
node_modules/.bin/nyc --require babel-core/register node_modules/.bin/mocha --ui tdd --timeout=60000 --reporter dot src/*-test.js
```

All tests pass and a table of code coverage is outputted. Also, a json file is written to disk.

PR #2577, opened 2016-09-15
Add codecov.io badge and submit coverage data from builds

mikegray:


tools/run_pr_builder.sh
```diff
@@ -18,7 +18,17 @@ node --version
 # go list is expensive, only do it once.
 GO_LIST="$(go list ./... | grep -v /vendor/ | grep -v /samples/js/)"
 go build ${GO_LIST}
-go test ${GO_LIST}
+
```
mikegray: Unfortunate to have to loop and test but it is required for building the coverage output across multiple golang packages.

tools/run_pr_builder.sh
```diff
@@ -18,7 +18,17 @@ node --version
...
+# go test plus build coverage data for upload codecov.io
+echo "" > coverage.txt
+for d in ${GO_LIST}; do
+    go test -coverprofile=profile.out -covermode=atomic $d
+    if [ -f profile.out ]; then
+        cat profile.out >> coverage.txt
+        rm profile.out
+    fi
+done
+bash <(curl -s https://codecov.io/bash) -t ${COVERALLS_TOKEN}
```
mikegray: I will update Jenkins to have this environment variable set appropriately in order to avoid having the credentials in this script.

tools/run_pr_builder.sh
```diff
@@ -18,7 +18,17 @@ node --version
 # go list is expensive, only do it once.
 GO_LIST="$(go list ./... | grep -v /vendor/ | grep -v /samples/js/)"
 go build ${GO_LIST}
-go test ${GO_LIST}
+
+# go test plus build coverage data for upload codecov.io
+echo "" > coverage.txt
```
kalman: probably don't need the "" here

tools/run_pr_builder.sh
kalman: you should also update jenkins (and this script) to only run this on the master builder, not on every PR.

mikegray: I actually do want to have it run for every PR. The codecov service offers (free) tracking and checking/notifications to ensure you are not reducing code coverage with a PR.

tools/run_pr_builder.sh
mikegray: ok

tools/run_pr_builder.sh
kalman: but this means that the code coverage badge on the front page will just show the results from the last PR builder?

mikegray: No. The badge shows results for the master branch - it is hard-coded - not updating the README.md links based on the git branch or PR. If you go to codecov.io, you can see the PRs listed and select the one you are reviewing or working on. I will also setup the Slack notifications once I have more data and verify it isn't too noisy.

mikegray:
jenkins test this

tools/run_pr_builder.sh
kalman: Ah gotcha - so the script magically does that. Could you comment that?

And honestly I'd rather we check in this script, the idea of just running remote code on a machine that has all of our credentials... kinda scares me.

mikegray: I could, but then we won't get updates as they change and improve the script for all of the various languages and code coverage tools. Somewhat ironically this is like the Google Chrome auto-updates arguments. I did have the same gut instinct as you on this, but they do have information on the codecov.io site that helped allay my fears. That plus it is open source, so it is something that gets reviewed. I am mildly against checking in the script. If you feel strongly let me know and I will do it.

mikegray:
retest this please

mikegray:
jenkins: build this

mikegray:
toward #1605

kalman:
Actually, I think this is fine. LGTM.

Issue #2509, opened 2016-09-02
Run python unit tests on bots


arv:
We are running go & js unit tests but not the python ones.

PR #2578, opened 2016-09-15
Run python unittests; configure codecov via YAML

mikegray:
fixing #2509, adding recommended YAML file for better codecov.io integration

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,53 @@
+codecov:
+  branch: master
+  bot: "mikegray"
+  ci:
+    - "jenkins.noms.io"
+
+coverage:
+  precision: 2         # how many decimal places to display in the UI: 0 <= value <= 4
+  round: down          # how coverage is rounded: down/up/nearest
+  range: 60...100      # custom range of coverage colors from red -> yellow -> green
+
+  notify:
+    slack:
+      default:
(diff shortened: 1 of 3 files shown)


codecov.yml
```diff
@@ -0,0 +1,53 @@
...
+
+coverage:
+  precision: 2         # how many decimal places to display in the UI: 0 <= value <= 4
+  round: down          # how coverage is rounded: down/up/nearest
+  range: 60...100      # custom range of coverage colors from red -> yellow -> green
+
+  notify:
+    slack:
+      default:
+        url: "secret:n+BYhIXTXsaCiMKB3vOf6yP68ytdKd3WpXtJFWPEUsEWXDiGnU5dTB5DO2yv8tR0COdxvs7K31hVpEfHEXdoXOaQhUw3FKf3fh8KZDLN7CGTbeDhw1uNGGyBr2d2TWnopzYtcXomdwMmuckARtiWQx0YXJiZY9YyCrIoDK9HIJQ="
```
arv: is this ok to keep in git?

mikegray: Yes, I generated it via the codecov.io website so only they know how to decrypt it to determine the real URL to use.

mikegray:
What about these failing python unit tests though?  It appears to work fine on my Mac, so maybe it is not deterministic? or ordering is based on something different for Linux vs. Darwin?

arv:
Maybe we need to sort these or use `set()`/`frozenset()`?

mikegray:
@arv PTAL, using sorted instead of frozenset. unittest2 package has `assertElementsEqual` but we aren't using that. Python 3.2+ unittest has another similar option, but we are using Python 2.7.

arv:
Still LGTM

Issue #2535, opened 2016-09-08
Move noms-merge into `noms commit` tool


cmasone-attic:
The new interface will be:

```
noms merge [--policy=l|r|a] [--out-ds-name=<name>] [--parent=<obj>] <db-spec> <left-obj> <right-obj>

  <db-spec>   : database in which named datasets live
  <left-obj>  : path to an object descending from <parent>
  <right-obj> : path to an object descending from <parent>

See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the <db-spec>, <left-obj>, <right-obj> and --parent arguments.

Flags:

--policy
    If unset, any merge conflict is fatal. Can be set to one of 'l', 'r', or 'a' which will resolve conflicts by choosing the Value from <left-obj>, <right-obj>, or asking the user respectively
--out-ds-name (= "")
    output dataset to write to - if empty, defaults to printing the hash of the merged object to stdout
--parent (= "")
    common ancestor of <left-obj> and <right-obj> (required unless left-obj and right-obj are Datasets)
--quiet  (= false)
    silence progress output
```

cmasone-attic:
@aboodman The above is the current, desired API, right?

aboodman:
Hm, the idea of this creating its own commit seems in conflict with factoring out `noms commit`. But it's required in the case where the inputs are datasets, if you want to reflect the merge by having two parents in the merge commit.

aboodman:
Ideas? @rafael-atticlabs ?

cmasone-attic:
I mean...it can still craft the merge Commit and emit the hash of that object. Then noms commit just needs to understand what to do if it sees that the hash is already a Commit?

aboodman:
I mean what metadata will it put there?

cmasone-attic:
hm. true. Open to ideas :-)

aboodman:
Also it wont actually work with `commit` because commit (as currently being
implemented) because both tools think they are the ones doing the commit.

cmasone-attic:
POR: `noms commit --merge` or similar. I'll flesh out what the command line args should look like once I look at noms commit in more detail

PR #2579, opened 2016-09-15
Add FindCommonAncestor for Commits

cmasone-attic:
Once we integrate noms-merge into the `noms commit` command, this
function will allow us to stop requiring users to pass in the common
ancestor to be used when merging. The code can just find it and merge
away.

Toward #2535

### DIFF
--- a/go/datas/commit.go
+++ b/go/datas/commit.go
@@ -5,6 +5,8 @@
 package datas
 
 import (
+	"sort"
+
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/types"
 )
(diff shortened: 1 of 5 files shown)


cmasone-attic:
PTAL

go/datas/commit.go
```diff
@@ -66,6 +68,50 @@ func CommitDescendsFrom(commit types.Struct, ancestor types.Ref, vr types.ValueR
...
+// one exists, setting ok to true. If there is no common ancestor, ok is set to false.
+func FindCommonAncestor(c1, c2 types.Struct, vr types.ValueReader) (a types.Struct, ok bool) {
+	c1Q, c2Q := &types.RefByHeight{types.NewRef(c1)}, &types.RefByHeight{types.NewRef(c2)}
+
+	makeSet := func(refs types.RefSlice) types.Set {
+		vals := make(types.ValueSlice, refs.Len())
+		for i, r := range refs {
+			vals[i] = r
+		}
+		return types.NewSet(vals...)
```
kalman: As said IRL: try using TargetHash().String() of these refs and put in a Go map, not types.Set.

go/datas/commit.go
```diff
@@ -66,6 +68,50 @@ func CommitDescendsFrom(commit types.Struct, ancestor types.Ref, vr types.ValueR
 	return true
 }
 
+// FindCommonAncestor returns the most recent common ancestor of c1 and c2, if
+// one exists, setting ok to true. If there is no common ancestor, ok is set to false.
+func FindCommonAncestor(c1, c2 types.Struct, vr types.ValueReader) (a types.Struct, ok bool) {
```
kalman: Are these structs because they are supposed to be Commits? Could you do `d.PanicIfFalse(IsCommitType(c1.Type())` then?

(maybe we should add an IsCommitStruct so you don't need to call .Type())

cmasone-attic:
PTAL

go/datas/commit.go
cmasone-attic: done

PR #2580, opened 2016-09-15
Add code coverage to js/noms

arv:


js/noms/package.json
```diff
@@ -37,12 +37,14 @@
     "documentation": "documentationjs/documentation#a71bb03681cd145723218910145c835dd5171a74",
     "flow-bin": "0.30.0",
     "fs-extra": "^0.30.0",
-    "mocha": "^2.5.3"
+    "mocha": "^2.5.3",
+    "nyc": "^8.3.0"
```
mikegray: Do we need to also add this to js/perf/*? and samples/js/*?

js/noms/package.json
```diff
@@ -37,12 +37,14 @@
...
   },
   "scripts": {
     "pretest": "eslint src/ && flow src/",
     "ts": "mocha --ui tdd --timeout=4000 --reporter spec --slow=50 --invert --grep='LONG:' --compilers js:babel-core/register src/*-test.js",
-    "test": "mocha --ui tdd --timeout=6000 --reporter dot --compilers js:babel-core/register src/*-test.js",
+    "test": "mocha --ui tdd --timeout=10000 --reporter dot --compilers js:babel-core/register src/*-test.js",
+    "code-cov": "nyc npm test && nyc report --reporter=text-lcov > coverage.lcov",
```
mikegray: Hmm, it must be expensive - is that why you are not running it whenever tests run? Also doesn't make sense to run it whenever any random person runs `npm test` probably. Maybe we just need a CI specific target?

arv: Yeah maybe we should rerun the tests?

js/noms/package.json
arv: Not at this time.

js/noms/package.json
arv: Actually, we already have `npm run ts` which skips long tests. I'll just make `npm test` run the code coverage.

mikegray:
lgtm

arv:
PTAL

PR #2584, opened 2016-09-16
Remove TypeDesc equals

arv:
This method was not being used.

Major Version Change: equals was part of exposed public API of *Desc


arv:
This was removed from the Go code a while ago but I guess I forgot to remove the js code.

PR #2585, opened 2016-09-16
Make codecov Slack integration not emit 'template string'

mikegray:


codecov.yml
```diff
@@ -14,9 +14,8 @@ coverage:
       default:
         url: "secret:n+BYhIXTXsaCiMKB3vOf6yP68ytdKd3WpXtJFWPEUsEWXDiGnU5dTB5DO2yv8tR0COdxvs7K31hVpEfHEXdoXOaQhUw3FKf3fh8KZDLN7CGTbeDhw1uNGGyBr2d2TWnopzYtcXomdwMmuckARtiWQx0YXJiZY9YyCrIoDK9HIJQ="
         branches: null
-        threshold: null
+        threshold: 1.15
```
arv: What is this magic number?

mikegray: For deciding how much change to allow in code coverage percentage before sending notification to Slack. Not sure of the correct value but this was in one of the samples codecov.io provided. I believe this can be thought of as "if code coverage difference is + /-1.15% of last build then send notification to Slack"

Issue #2037, opened 2016-07-13
We need to print progress when committing to a remote store


aboodman:
We've gotten to the point where just sending the bytes to the server is the slowest part of most writes. Yay. But now, the progress for e.g., `csv-import`, `sync`, and other things is not as useful because we get near 100% and then it pauses there for awhile during upload.

rafael-atticlabs:
I'm gonna close this one. 1) Because the "commit" phase is quite a bit faster than it was when we wrote this and 2) because I think it's superseded by implementing batched writing which there is now a clear path to.

PR #2588, opened 2016-09-16
Let Database expose a progress channel

arv:
The underlying HTTP batch store posts progress reports to this channel
during commit. At this point the numbers only makes sense if there is a
single commit going on.

Towards #2037


arv:
Is this better.

I'm still not happy with how this is exposed. I'll think of ways to make this less hacky.

cmasone-attic:
I'm not super happy with how it's ultimately exposed either (meaning, via the method on Database :-/)

But I don't have better ideas. That said, if we're going to do this I think that the BatchStore interface should just expose this channel as well. That way you don't have to hang on to an extra httpBatchStore pointer in RemoteDatabaseClient, because all batch stores will have a ProgressChannel() method on them.

arv:
Closing. We can revisit later.

PR #2595, opened 2016-09-19
Add unit tests for commit metadata

mikegray:
The codecov results pointed out to me last week that I had never written specific tests for this. Adding now. It is being tested through noms_commit too, but this is better and more thorough.


mikegray:
Because of the dependencies.  It can't be in go/dataset/ because it needs to use go/spec/ for resolving absolute paths and it creates a circular Go path dependency.  If you can think of a better place for it I don't have strong feelings about it.

arv:
I see. This tests behavior that is defined in spec. For some reason I thought it was testing something from datas (where Commit lives).

willhite:
LGTM

Issue #2131, opened 2016-07-22
create a 'noms alias' command


willhite:
This would be **vastly** superior to a 'noms use' command and easier to implement. The following command:
```noms alias local http://localhost:8000```
would save the alias to a .noms file in the user's home directory. From that point on, users could use the command:
```noms ds local```

The noms alias command would also have the following args: -d (for delete), -s (for show) and -ls (list all).
```
prompt> noms alias -s local
local: http://localhost:8000

prompt> noms alias -ls
local:                http://localhost:8000
cli-tour:             http://demo.noms.io/cli-tour
demo-sfcrime:         http://demo.noms.io/cli-tour::sf-crime

prompt> noms alias -d local
alias 'local' deleted
```

**Comment by kalman** [comment]
> the place we save the configuration should in fact be a noms repo

drive by: I like how .gitconfig is a text file, not git repositories.

**Comment by aboodman** [comment]
>> the place we save the configuration should in fact be a noms repo
> drive by: I like how .gitconfig is a text file, not git repositories.

yeah I had that gut reaction too. I suggested using noms as a kind of strawman.

**Comment by ehalpern** [comment]
Shooting from the hip and taking inspiration from git clone, what if noms defined a mapping between the local and remote repo implicitly when sync was issued. This combined with some prudent defaulting could be a convenient way to work with local and remote datasets.

Here's a rough sketch:

Issuing a sync like this: 
```
noms sync http://demo.noms.io/cli-tour::sf-film-locations ldb:.noms/tour::films
```

would implicitly create some definitions like this: 
```
# .nomsconfig
[db.tour]
  url = ldb:.noms/tour
  [db.tour.remote.origin]
    url = http://demo.noms.io/cli-tour
```    

The `use` command would activate the mappings. (Note that `use` could be implicit if there was only one db defined in the local .nomsconfig). 
```
noms use tour    # activates definitions under db.tour
```

Other noms commands would be modified to make use of the current db mappings: 

```
noms ds          # noms ds ldb://.noms/tour
noms ds origin   # noms ds http://demo.noms.io/cli-tour

noms log films   # noms log ldb://tmp/noms/tour::films
noms log origin::sf-film-locations   # noms log http://demo.noms.io/cli-tour::sf-film-locations

noms show 'it8pop...'         # noms show ldb://.noms/tour::'...'
noms show origin::'it8pop...'  # noms show http://demo.noms.io/cli-tour::'...'

noms sync films origin::sf-film-locations
[...]

PR #2597, opened 2016-09-20
Noms configuration for default and aliases

ehalpern:
This is the evolving proposal for noms config support discussed in issue #2131. Terminology has changed slightly as follows:
 
- [local] -> [default]
- [remote.<alias>] -> [db.<alias>]

Also the location of the noms config has changed from .noms/config to .nomsconfig. 

For details and a guide, see the REAMDE and sample .gitconfig in samples/cli/nomsconfig.
  
toward #2131

### DIFF
--- a/go/spec/config_test.go
+++ b/go/spec/config_test.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/testify/assert"
 )
 
@@ -45,20 +44,19 @@ type paths struct {
 	config string
 }
 
-func getPaths(base string) paths {
+func getPaths(assert *assert.Assertions, base string) paths {
 	abs, err := filepath.Abs(ctestRoot)
-	d.PanicIfError(err)
+	assert.NoError(err)
 	abs, err = filepath.EvalSymlinks(ctestRoot)
-	d.PanicIfError(err)
+	assert.NoError(err)
 	home := filepath.Join(abs, base)
 	config := filepath.Join(home, NomsConfigFile)
 	return paths{ home, config }
 }
 
-
-func qualifyFilePath(path string) string {
+func qualifyFilePath(assert *assert.Assertions, path string) string {
 	p, err := filepath.Abs(path)
-	d.PanicIfError(err)
+	assert.NoError(err)
 	return p
 }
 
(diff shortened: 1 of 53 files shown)


samples/cli/nomsconfig/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+When *.nomsconfig* is present in the current directory or ancestor, it's
+definitions will be active when running noms cli. 
+
+This directory contains an [example](.gitconfig) config to try out. If you lke what 
+you see once you've kicked the tires, you can create your own *.nomsconfig*'s wherever 
+you need one.
+
+A *.gitconfig* file can contain a single default database definition and 
+any number of database alaises. Here's are the definitons from the example:
```
willhite: alaises => aliases

go/spec/config_test.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+type paths struct {
+	home string
+	config string
+}
+
+func getPaths(base string) paths {
+	abs, err := filepath.Abs(ctestRoot)
+	d.PanicIfError(err)
```
willhite: I know our error handling is kind of messed up, but I'm not sure it's worth it to use PanicIfError in test cases since that wraps the error with stuff that might not be that useful in test case failure reporting.

go/spec/resolver.go
```diff
@@ -0,0 +1,83 @@
...
+			db = dsr.resolveDatabaseString(split[0])
+			ds = split[1]
+		}
+		return db + separator + ds
+	}
+	return str
+}
+
+// Resolve string to database spec. If a config is present,
+//   - resolve a remote alias to its db spec
```
willhite: All these comments refer to remote and local db's. Are they still relevant?

ehalpern:
Thanks for your detailed review and proofreading @willhite. I think I've addressed all of your review comments. PTAL.

ehalpern:
@willhite: History is clean and squashed, and hopefully easier to read. PTAL.

go/spec/resolver.go
```diff
@@ -39,45 +39,42 @@ func (dsr *Resolver) resolveDatabaseString(str string) string {
...
 	}
 	return str
 }
 
 // Resolve string to database spec. If a config is present,
-//   - resolve a remote alias to its db spec
-//   - resolve "" to the local db spec
+//   - resolve a db alias to its db spec
+//   - resolve "" to the default db spec
 func (dsr *Resolver) GetDatabase(str string) (datas.Database, error) {
```
willhite: You might be able to get your code coverage stats up if you have tests for these four functions.

ehalpern: I think you're right. All these cases are covered in the test but they're not run through the wrapper functions because require more setup. I think the clean way to handle this is to get rid of the wrappers and have the command line programs call the resolver first and that get the db, ds or object with a separate call.

PR #2604, opened 2016-09-22
Update Babel dependencies

arv:
The motivation is to try to see if this fixes a bug where it thinks
we are exporting something twice.


arv:
I'll TBR this if all the tests pass.

kalman:
Seems perfectly fine to me. LGTM when tests pass :-)

PR #2607, opened 2016-09-22
Support -perf.run flag for perf tests, filters by regexp

kalman:


kalman:
Yes that is the if statements. We can't _always_ go up in code coverage.

kalman:
Hey @willhite could you have another quick look? I realised that if you have a test Test42Bar then filtering for Test42Bar should run that test.

PR #2610, opened 2016-09-23
Add photo-index: a simple photo indexer. For now only indexes by tag.

aboodman:
Will add indexing by face/geo in subsequent patches.

### DIFF
--- a/samples/go/photo-index/main.go
+++ b/samples/go/photo-index/main.go
@@ -7,6 +7,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"path"
 
 	"github.com/attic-labs/noms/go/dataset"
 	"github.com/attic-labs/noms/go/spec"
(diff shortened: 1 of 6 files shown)


samples/go/photo-index/main.go
```diff
@@ -0,0 +1,116 @@
...
+				p := cv.(types.Struct)
+				tags := p.Get("tags").(types.Set)
+				tags.IterAll(func(t types.Value) {
+					gb.SetInsert([]types.Value{t}, p)
+				})
+				// Can't be any photos inside photos, so we can save a little bit here.
+				stop = true
+			}
+			return
+		}, 12)
```
aboodman: I should add a concurrency flag.

samples/go/photo-index/main.go
```diff
@@ -0,0 +1,116 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
```
aboodman: I should add tests.

samples/go/photo-index/main.go
```diff
@@ -0,0 +1,116 @@
...
+		return
+	}
+
+	win = true
+	return
+}
+
+func usage() {
+	fmt.Fprintf(os.Stderr, "photo-index indexes photos by common attributes\n\n")
+	fmt.Fprintf(os.Stderr, "Usage: %s -db=<db-spec> -out-ds=<name> [input-datasets...]\n\n", os.Args[0])
```
willhite: Probably want to do path.Base(os.Args[0])

samples/go/photo-index/main.go
```diff
@@ -0,0 +1,116 @@
...
+	fmt.Fprintf(os.Stderr, "  <db>             : Database to work with\n")
+	fmt.Fprintf(os.Stderr, "  <out-ds>         : Dataset to write index to\n")
+	fmt.Fprintf(os.Stderr, "  <input-datasets> : Input datasets to crawl\n\n")
```
willhite: can this be paths rather than datasets?

samples/go/photo-index/main.go
```diff
@@ -0,0 +1,116 @@
...
+	if !dataset.DatasetFullRe.MatchString(*outDSStr) {
+		fmt.Fprintf(os.Stderr, "Invalid output dataset name: %s\n", *outDSStr)
+		return
+	} else {
+		outDS = dataset.NewDataset(db, *outDSStr)
+	}
+
+	inputs := []types.Value{}
+	for i := 0; i < flag.NArg(); i++ {
+		if dataset.DatasetFullRe.MatchString(flag.Arg(i)) {
```
willhite: Should we allow paths to be specified as well as datasets? This is useful, for example when I want do say <dsname>.value so all the history isn't scanned.

aboodman: Oh yeah, great point. Will change.

samples/go/photo-index/main.go
aboodman: Will do.

samples/go/photo-index/main.go
aboodman: Good idea, done.

aboodman:
PTAL

samples/go/photo-index/main.go
```diff
@@ -55,38 +56,59 @@ func index() (win bool) {
...
 	inputs := []types.Value{}
 	for i := 0; i < flag.NArg(); i++ {
-		if dataset.DatasetFullRe.MatchString(flag.Arg(i)) {
-			ds := dataset.NewDataset(db, flag.Arg(i))
-			v, ok := ds.MaybeHeadValue()
-			if ok {
-				inputs = append(inputs, v)
-				continue
-			}
+		p, err := spec.NewAbsolutePath(flag.Arg(i))
```
willhite: I think this could be useful as a method in the specs package. It seems like there might be lots of programs that want to take a number of paths as arguments. Something like: 
```ResolvePaths(db datas.Database, paths []string) (ValueSlice, err)```

aboodman: Good point. I will do this as a separate patch.

aboodman:
PTAL

samples/go/photo-index/main.go
aboodman: Done

samples/go/photo-index/main.go
aboodman: Done

PR #2612, opened 2016-09-23
Implement types.Blob.Concat and types.NewBlobP

kalman:
Concat is a simple use of the sequence concat code that List.Concat
uses. NewBlobP uses Concat to construct a Blob in parallel.

Perf tests for NewBlobP write N temporary files then constructs a Blob
from them, so there is some I/O, but it appears to be mostly CPU bound.
NewBlobP doesn't get much more than 50% faster with any P >= 2.


kalman:
@willhite or @rafael-atticlabs PTAL

go/types/list_test.go
```diff
@@ -1107,3 +1107,17 @@ func TestListConcat(t *testing.T) {
 	run(6, 1e4, 0, 1000, 10)
 	run(7, 1e4, 1e4-1000, 1e4, 10)
 }
+
+func TestListConcatDifferentTypes(t *testing.T) {
```
kalman: Not related to this patch in any way, it just occurred to me to test it.

go/types/blob.go
```diff
@@ -172,10 +186,46 @@ func chunkBlobLeaf(vr ValueReader, buff []byte) (Collection, orderedKey, uint64)
 	return blob, orderedKeyFromInt(len(buff)), uint64(len(buff))
 }
 
+// NewBlob creates a Blob by reading from r.
 func NewBlob(r io.Reader) Blob {
 	return NewStreamingBlob(r, nil)
 }
 
+// NewBlobP creates a Blob by reading from every Reader in rs in parallel, then
```
kalman: I'm considering just making NewBlob/NewStreamingBlob do the parallelisation automatically, and not having NewBlobP. WDYT?

go/types/list_test.go
rafael-atticlabs: nice

rafael-atticlabs:
very cool. lgtm.

go/types/blob.go
rafael-atticlabs: makes sense to me.

kalman: done

PR #2615, opened 2016-09-26
Update FB and Flickr importers to be similar in names and descriptions

mikegray:


samples/js/fb/find-photos/src/main.js
```diff
@@ -22,7 +22,7 @@ import {
 const args = argv
   .usage(
     'Finds photos in slurped Facebook metadata\n\n' +
-    'Usage: fb/find-photos <in-object> <out-dataset>')
+    'Usage: noms-facebook-find-photos <in-dataset> <out-dataset>')
```
arv: Is this correct? We do not have a bin section in package.json so there is no binary installed.l To run this we need to do:

```sh
node . <in-dataset> <out-dataset>
```

mikegray: ok, fixed all 4 places that have this issue. PTAL

samples/js/fb/find-photos/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "noms-fb-find-photos",
-  "description": "Finds photos in slurped flickr data",
+  "name": "noms-facebook-find-photos",
```
aboodman: I think this can stay as `fb`.

mikegray: ok, changed it back, but then changed noms-facebook-slurp to be noms-fb-slurp to match.

mikegray:
PTAL

PR #2622, opened 2016-09-27
Refactor exit mockery into go/util/exit

aboodman:
Ignore the first commit, it is being reviewed separately.

fyi: @zcstarr, @willhite


go/util/exit/exit.go
```diff
@@ -0,0 +1,19 @@
+// Package exit provides a mockable implementation of os.Exit.
```
cmasone-attic: copyright thingy

willhite:
Nice.

PR #2633, opened 2016-09-28
Add integration test for sample: aggregate.

willhite:
I'm just getting my feet wet with the first integration test.


samples/js/aggregate/integration_test.go
```diff
@@ -0,0 +1,78 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package counter
+
+import (
+	"testing"
+
+    "github.com/attic-labs/noms/go/marshal"
```
arv: gofmt

samples/js/aggregate/integration_test.go
```diff
@@ -0,0 +1,78 @@
...
+	s.NoError(err)
+}
+
+func (s *testSuite) Teardown() {
+    db := s.Database()
+    defer db.Close()
+    outData := db.GetDataset(outDsName).HeadValue()
+    expectedMap := map[string]int {
+        "2012": 4,
+        "2013": 4,
```
arv: Maybe it would be better if not all were 4 ;-)

PR #2636, opened 2016-09-28
JS: Make Database a mutable API that vends immutable Datasets

### LINKED ISSUE #2589: Make Database a mutable API that vends immutable Datasets
Noms SDK users frequently shoot themselves in the foot because they're holding onto an "old" Database object. That is, they have a Database tucked away in some internal state, they call Commit() on it, and don't replace the object in their internal state with the new Database returned from Commit. The initial discussion is still pretty accurate from a high level, despite being fairly old and about the Javascript SDK. It’s pretty easy to wind up with an out-of-date Database object for a couple of reasons:

First, because usually a new, up-to-date Database gets created as a side-effect of calling Dataset.Commit(), and the only way to get at it is by extracting it from the new Dataset returned by that method. I think that DatasetSpecs exacerbate this issue, because it's not uncommon to get a DatasetSpec, pull the backing Database out of it, and then pass it around or hang onto it for value writing. Then you call Commit on the original _Dataset_ you got, and now the Database you've been hanging onto is out of date and out of sync with the new Dataset returned from Commit().

The second issue is that only half the Database API Is written in immutable-style. WriteValue, which one tends to use more often than Commit, mutates internal state, and so it's weird that sometimes you need to replace the Database object you have, but other times you don't.

So, the goal is to make Database a mutable-style API that vends Datasets, which are immutable-style.

```go

type Database interface {
	// To implement types.ValueWriter, Database implementations
	// provide WriteValue(). WriteValue() writes v to this
	// Database, though v is not guaranteed to be be persistent
	// until after a subsequent Commit(). The return value is the
	// Ref of v.
	types.ValueReadWriter
	io.Closer

	// Datasets returns the root of the database which is a
	// MapOfStringToRefOfCommit where string is a datasetID.
	Datasets() types.Map

	// GetDataset returns a Dataset struct containing the current
	// mapping of datasetID in the above Datasets Map.
	GetDataset(datasetID string) Dataset

	// Commit updates the Commit that ds.ID() in this database
	// points at. All Values that have been written to this
	// Database are guaranteed to be persistent after Commit().
	// The new Commit struct is constructed using `v`,
	// `opts.Parents`, and `opts.Meta`.
	// If `opts.Parents` is the zero value (`types.Set{}`) then the 
	// current head is used.
	// If `opts.Meta is the zero value (`types.Struct{}`) then a
	// fully initialized empty Struct is passed to NewCommit.
	// If the update cannot be performed, e.g., because of a
	// conflict, Commit returns an 'ErrMergeNeeded' error and
	// the current snapshot of the Dataset so that the client can
	// merge the changes and try again.
	// Regardless, Datasets() is updated to match backing storage
	// upon return.
	Commit(ds Dataset, v types.Value, opts CommitOpts) (Dataset, error)
	CommitValue(ds Dataset, v types.Value) (Dataset, error)

	// Delete removes the Dataset named ds.ID() from the map at
	// the root of the Database. The Dataset data is not
	// necessarily cleaned up at this time, but may be garbage
	// collected in the future.
	// If the update cannot be performed, e.g., because of a
	// conflict, Commit returns an 'ErrMergeNeeded' error and
	// the current snapshot of the Dataset so that the client can
	// decide whether to try again.
	// Regardless, Datasets() is updated to match backing storage
	// upon return.
	Delete(ds Dataset) (Dataset, error)

	// SetHead ignores any lineage constraints (e.g. the current Head being in
	// commit’s Parent set) and force-sets a mapping from datasetID: commit in
	// this database.
	// All Values that have been written to this
	// Database are guaranteed to be persistent after SetHead(). If
	// the update cannot be performed, e.g., because of a conflict,
	// error will be non-nil. Regardless, Datasets() is updated to
	// match backing storage upon return.
	SetHead(ds Dataset, newHeadRef types.Ref) error

	// FastForward takes a types.Ref to a Commit object and makes it the new
	// Head of ds iff it is a descendant of the current Head. Intended to be
	// used e.g. after a call to Pull(). If the update cannot be performed,
	// e.g., because another process moved the current Head out from under
	// you, err will be non- nil. The newest snapshot of the Dataset is always
	// returned, so the caller an easily retry using the latest.
	FastForward(ds Dataset, newHeadRef types.Ref) (Dataset, error)
}

// Pull objects that descend from sourceRef from srcDB to sinkDB. sinkHeadRef
// should point to a Commit (in sinkDB) that's an ancestor of sourceRef. This
// allows the algorithm to figure out which portions of data are already
// present in sinkDB and skip copying them.
func Pull(srcDB, sinkDB Database, sourceRef, sinkHeadRef types.Ref, concurrency int, progressCh chan PullProgress) {
...
}
```

```go
package datas

var DatasetRe = regexp.MustCompile(`[a-zA-Z0-9\-_/]+`)
var DatasetFullRe = regexp.MustCompile("^" + DatasetRe.String() + "$")

// Dataset represents a “named Commit” within a Database. Each Dataset instance represents a single moment in history.
type Dataset struct {
	store datas.Database
	id    string
	headRef types.Ref  
}


func (ds *Dataset) Database() datas.Database {
	return ds.store
}

func (ds *Dataset) ID() string {
	return ds.id
}

// MaybeHead returns the current Head Commit of this Dataset, which contains the current root of the Dataset's value tree, if available. If not, it returns a new Commit and 'false'.
func (ds *Dataset) MaybeHead() (types.Struct, bool) {
	...
}

func (ds *Dataset) MaybeHeadRef() (types.Ref, bool) {
	...
}

// Head returns the current head Commit, which contains the current root of the Dataset's value tree.
func (ds *Dataset) Head() types.Struct {
	...
}

func (ds *Dataset) HeadRef() types.Ref {
	...
}

// HeadValue returns the Value field of the current head Commit.
func (ds *Dataset) HeadValue() types.Value {
	...
}

// MaybeHeadValue returns the Value field of the current head Commit, if available. If not it
// returns nil and 'false'.
func (ds *Dataset) MaybeHeadValue() (types.Value, bool) {
	...
}
```

cmasone-attic:
Noms SDK users frequently shoot themselves in the foot because they're
holding onto an "old" Database object. That is, they have a Database
tucked away in some internal state, they call Commit() on it, and
don't replace the object in their internal state with the new Database
returned from Commit.

This PR fixes #2589 by changing the Database and Dataset JS API to be
in line with the proposal there.

### DIFF
--- a/js/noms/src/database.js
+++ b/js/noms/src/database.js
@@ -19,13 +19,11 @@ import {equals} from './compare.js';
 /**
  * Database provides versioned storage for noms values. While Values can be
  * directly read and written from a Database, it is generally more appropriate
- * to read data by inspecting the Head of a Dataset and write new data by
- * updating the Head of a Dataset via Commit() or similar. Particularly, new
- * data is not guaranteed to be persistent until after a Commit (Delete,
- * SetHead, or FastForward) operation completes.
- * The Database API is stateful, meaning that calls to GetDataset() or
- * Datasets() occurring after a call to Commit() (et al) will represent the
- * result of the Commit().
+ * to read data by inspecting the head of a Dataset and write new data by
+ * updating the head of a Dataset via commit(). Particularly, new
+ * data is not guaranteed to be persistent until after a commit operation completes.
+ * The Database API is stateful, meaning that calls to getDataset() or
+ * datasets() occurring after a call to commit() will represent the result of the commit().
  */
 export default class Database {
   _vs: ValueStore;
@@ -49,8 +47,7 @@ export default class Database {
   }
 
   /**
-   * datasets returns the root of the database which is a
-   * Map<String, Ref<Commit>> where string is a datasetID.
+   * datasets returns the root of the database.
    */
   datasets(): Promise<Map<string, Ref<Commit<any>>>> {
     return this._datasets;
@@ -58,7 +55,7 @@ export default class Database {
 
   /**
    * getDataset returns a Dataset struct containing the current mapping of
-   * datasetID in the above Datasets Map.
+   * id in the above Datasets Map.
    */
   getDataset(id: string): Dataset {
     return new Dataset(this, id, this.datasets().then(sets => sets.get(id)));
@@ -93,12 +90,12 @@ export default class Database {
   }
 
   /**
-   * commit updates the commit that ds.id() points at. If parents is provided then the promise
+   * commit updates the commit that ds.id points at. If parents is provided then the promise
    * is rejected if the commit does not descend from the parents.
    * The returned Dataset is always the newest snapshot, regardless of
    * success or failure, and datasets() is updated to match backing storage
    * upon return as well. If the update cannot be performed, e.g., because
-   * of a conflict, Commit throws an error containing 'Merge Needed'.
+   * of a conflict, commit() throws an error containing 'Merge Needed'.
    */
   async commit(ds: Dataset, v: Value, parents: ?Array<Ref<Commit<any>>> = undefined):
   Promise<Dataset> {
@@ -127,7 +124,7 @@ export default class Database {
       const currentHeadRef = await currentDatasets.get(datasetId);
       if (currentHeadRef) {
         if (equals(commitRef, currentHeadRef)) {
-          return Promise.resolve(commitRef);
+          return commitRef;
         }
         if (!await this._descendsFrom(commit, currentHeadRef)) {
           throw new Error('Merge needed');
@@ -138,7 +135,7 @@ export default class Database {
     currentDatasets = await currentDatasets.set(datasetId, commitRef);
     const newRootRef = this.writeValue(currentDatasets).targetHash;
     if (await this._rt.updateRoot(newRootRef, currentRootRef)) {
-      return Promise.resolve(commitRef);
+      return commitRef;
     }
 
     throw new Error('Optimistic lock failed');
--- a/js/noms/src/database.js
+++ b/js/noms/src/database.js
@@ -16,6 +16,17 @@ import Dataset from './dataset.js';
 import Commit from './commit.js';
 import {equals} from './compare.js';
 
+/**
+ * Database provides versioned storage for noms values. While Values can be
+ * directly read and written from a Database, it is generally more appropriate
+ * to read data by inspecting the Head of a Dataset and write new data by
+ * updating the Head of a Dataset via Commit() or similar. Particularly, new
+ * data is not guaranteed to be persistent until after a Commit (Delete,
+ * SetHead, or FastForward) operation completes.
+ * The Database API is stateful, meaning that calls to GetDataset() or
+ * Datasets() occurring after a call to Commit() (et al) will represent the
+ * result of the Commit().
+ */
 export default class Database {
   _vs: ValueStore;
   _rt: RootTracker;
@@ -37,19 +48,35 @@ export default class Database {
     });
   }
 
+  /**
+   * datasets returns the root of the database which is a
+   * Map<String, Ref<Commit>> where string is a datasetID.
+   */
   datasets(): Promise<Map<string, Ref<Commit<any>>>> {
     return this._datasets;
   }
 
+  /**
+   * getDataset returns a Dataset struct containing the current mapping of
+   * datasetID in the above Datasets Map.
+   */
   getDataset(id: string): Dataset {
     return new Dataset(this, id, this.datasets().then(sets => sets.get(id)));
   }
 
   // TODO: This should return Promise<Value | null>
+  /**
+   * readValue returns a Promise of the value with the hash `hash` in this Database, or
+   * null if it's not present.
+   */
   async readValue(hash: Hash): Promise<any> {
     return this._vs.readValue(hash);
   }
 
+  /**
+   * writeValue writes v to this Database and returns its Ref. v is not guaranteed to be
+   * durable until a subsequent call to commit().
+   */
   writeValue<T: Value>(v: T): Ref<T> {
     return this._vs.writeValue(v);
   }
@@ -65,8 +92,14 @@ export default class Database {
     return true;
   }
 
-  // Commit updates the commit that ds points at. If parents is provided then the promise
-  // is rejected if the commit does not descend from the parents.
+  /**
+   * commit updates the commit that ds.id() points at. If parents is provided then the promise
+   * is rejected if the commit does not descend from the parents.
+   * The returned Dataset is always the newest snapshot, regardless of
+   * success or failure, and datasets() is updated to match backing storage
+   * upon return as well. If the update cannot be performed, e.g., because
+   * of a conflict, Commit throws an error containing 'Merge Needed'.
+   */
   async commit(ds: Dataset, v: Value, parents: ?Array<Ref<Commit<any>>> = undefined):
   Promise<Dataset> {
     if (!parents) {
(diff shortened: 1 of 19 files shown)


js/noms/src/database.js
```diff
@@ -89,7 +93,8 @@ export default class Database {
       const currentHeadRef = await currentDatasets.get(datasetId);
       if (currentHeadRef) {
         if (equals(commitRef, currentHeadRef)) {
-          return this;
+          // $FlowIssue: thinks commitRef is a Promise
+          return commitRef;
```
cmasone-attic: I spent an hour trying to figure out what I'm doing wrong, here. Without these, flow complains that commitRef is a Promise, and that's incompatible with Ref. But...commitRef is the return value of `this.writeValue()`, which returns...a Ref! Can you help me figure out what the problem is?

cmasone-attic:
PTAL

js/noms/src/database.js
```diff
@@ -78,7 +65,24 @@ export default class Database {
     return true;
   }
 
-  async commit(datasetId: string, commit: Commit<any>): Promise<Database> {
+  // Commit updates the commit that ds points at. If parents is provided then the promise
```
arv: Can you make these JSDoc comments?

```js
/**
 * ...
 */
```

We should document all public methods.

js/noms/src/database.js
```diff
@@ -78,7 +65,24 @@ export default class Database {
...
+  async commit(ds: Dataset, v: Value, parents: ?Array<Ref<Commit<any>>> = undefined):
+  Promise<Dataset> {
+    if (!parents) {
+      const headRef = await ds.headRef();
+      parents = headRef ? [headRef] : [];
+    }
+    const commit = new Commit(v, new Set(parents));
+    try {
+      const commitRef = await this._doCommit(ds.id, commit);
+      return new Dataset(this, ds.id, Promise.resolve(commitRef));
```
arv: Can this be written as?

```js
const commitRefPromise = this._doCommit(ds.id, commit);
return new Dataset(this, ds.id, commitRefPromise);
```

js/noms/src/database.js
```diff
@@ -78,7 +65,24 @@ export default class Database {
...
+    } finally {
+      this._datasets = this._datasetsFromRootRef(this._rt.getRoot());
+    }
+  }
+
+  async _doCommit(datasetId: string, commit: Commit<any>): Ref<any> {
```
arv: The return type of any async function needs to be annotated as `Promise<T>`

js/noms/src/database.js
arv: Does changing the return type to `Promise<Ref<any>>` help?

js/noms/src/database.js
arv: OK. I like what you had before better. If we are using async functions we should minimize the number of `then` usages.

arv: You can do:

```js
const commitRefPromise = this._doCommit(ds.id, commit);
await commitRefPromise;
return new Dataset(this, ds.id, commitRefPromise);
```

that way you don't need to create a new promise.

js/noms/src/database.js
```diff
@@ -16,6 +16,17 @@ import Dataset from './dataset.js';
 import Commit from './commit.js';
 import {equals} from './compare.js';
 
+/**
+ * Database provides versioned storage for noms values. While Values can be
+ * directly read and written from a Database, it is generally more appropriate
+ * to read data by inspecting the Head of a Dataset and write new data by
```
arv: "Head" -> "head" maybe since JS does not use PascalCase

js/noms/src/database.js
```diff
@@ -37,19 +48,35 @@ export default class Database {
     });
   }
 
+  /**
+   * datasets returns the root of the database which is a
+   * Map<String, Ref<Commit>> where string is a datasetID.
```
arv: We have the return type already so we can remove it from the comment.

js/noms/src/database.js
```diff
@@ -65,8 +92,14 @@ export default class Database {
     return true;
   }
 
-  // Commit updates the commit that ds points at. If parents is provided then the promise
-  // is rejected if the commit does not descend from the parents.
+  /**
+   * commit updates the commit that ds.id() points at. If parents is provided then the promise
```
arv: `ds.id()` -> `ds.id`

js/noms/src/database.js
```diff
@@ -93,8 +127,7 @@ export default class Database {
       const currentHeadRef = await currentDatasets.get(datasetId);
       if (currentHeadRef) {
         if (equals(commitRef, currentHeadRef)) {
-          // $FlowIssue: thinks commitRef is a Promise
-          return commitRef;
+          return Promise.resolve(commitRef);
```
arv:
```js
return commitRef;
```

async functions already wrap non promises.

js/noms/src/database.js
```diff
@@ -105,8 +138,7 @@ export default class Database {
     currentDatasets = await currentDatasets.set(datasetId, commitRef);
     const newRootRef = this.writeValue(currentDatasets).targetHash;
     if (await this._rt.updateRoot(newRootRef, currentRootRef)) {
-      // $FlowIssue: thinks commitRef is a Promise
-      return commitRef;
+      return Promise.resolve(commitRef);
```
arv: same

js/noms/src/dataset.js
```diff
@@ -34,20 +35,38 @@ export default class Dataset {
...
+  /** ID returns the name of this Dataset. */
   get id(): string {
     return this._id;
   }
 
   // TODO: This should return Promise<Ref<Commit> | null>.
+  /**
+   * HeadRef returns the Ref of the current head Commit, which contains the
+   * current root of the Dataset's value tree.
+   * If the named Dataset doesn't exist in this._database, the returned Promise may resolve to null.
```
arv: I would not mention `this._database`. Maybe "doesn't exist in the underlying database"?

js/noms/src/dataset.js
```diff
@@ -34,20 +35,38 @@ export default class Dataset {
...
+   */
   headRef(): Promise<?Ref<Commit<any>>> {
     return this._headRef;
   }
 
   // TODO: This should return Promise<Commit | null>
+  /**
+   * Head returns the current head Commit, which contains the current root of
+   * the Dataset's value tree.
+   * If the named Dataset doesn't exist in this._database, the returned Promise may resolve to null.
```
arv: same

js/noms/src/dataset.js
```diff
@@ -34,20 +35,38 @@ export default class Dataset {
...
+   * The returned Promise may be rejected if errors occur while getting the Head.
+   */
   head(): Promise<?Commit<any>> {
     return this._headRef.then(hr => hr ? hr.targetValue(this._database) : null);
   }
 
+  /**
+   * HeadValue returns the Value field of the current head Commit.
+   * If the named Dataset doesn't exist in this._database, the returned Promise may resolve to null.
```
arv: and here

js/noms/src/database.js
cmasone-attic: done

cmasone-attic: done

js/noms/src/database.js
cmasone-attic: done

js/noms/src/database.js
cmasone-attic: done

js/noms/src/database.js
cmasone-attic: done and done

js/noms/src/dataset.js
cmasone-attic: done in all the places

cmasone-attic:
Ok, PTAL. I think I updated the right package.json as well.

PR #2637, opened 2016-09-28
try with mem

kalman:


kalman:
don't review this yet.

jenkins: perf this.

PR #2644, opened 2016-09-28
Run url-fetch perf tests against a fresh database each time

kalman:
The subsequent runs of url-fetch on jenkins are way faster, and this
appears to be because commiting is much faster on subsequent runs. The
perf tests now use a new database each time.

### DIFF
--- a/go/perf/suite/suite.go
+++ b/go/perf/suite/suite.go
@@ -269,7 +269,24 @@ func Run(datasetID string, t *testing.T, suiteT perfSuiteT) {
 	for repIdx := 0; repIdx < *perfRepeatFlag; repIdx++ {
 		testReps[repIdx] = testRep{}
 
-		serverHost, stopServerFn := suite.startServer()
+		// This is the temporary database for tests to use.
+		//
+		// === Why not use a local database + memory store?
+		// Firstly, because the spec would be "mem", and the spec library doesn't
+		// know how to reuse stores.
+		// Secondly, because it's an unrealistic performance measurement.
+		//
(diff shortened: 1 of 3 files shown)


kalman:
jenkins: perf this

kalman:
PTAL

go/perf/suite/suite.go
```diff
@@ -464,20 +481,13 @@ func (suite *PerfSuite) getGitHead(dir string) string {
...
-	// Again, leveldb is more realistic than memory, and in common cases disk space > memory space.
-	// However, on this developer's laptop, there is actually very little disk space, and a lot of memory;
-	// plus making the test run a little bit faster locally is nice.
+// StartRemoteDatabase creates a new remote database on an arbitrary free port,
+// running on a separate goroutine. Returns the hostname that that database was
+// started on, and a callback to run to shut down the server.
+//
+// If the -perf.mem flag is specified, the remote database is hosted in memory,
+// not on disk (in a temporary leveldb directory).
+func (suite *PerfSuite) StartRemoteDatabase() (host string, stopFn func()) {
```
kalman: I tried unit testing this, but it was running out of file descriptors.

ehalpern: Did you mean to put the docs describing the choice of db options in the StartRemoteDatabase block?

ehalpern: Otherwise LGTM.

kalman: It was deliberate but now that I look again it would make sense to all be in the docs, I think.

PR #2646, opened 2016-09-29
Make jsonToNoms handle some null values

willhite:
Null values in struct field are handled by not skipping that field in
the noms type. Null values as elements in array cause an exception.

### DIFF
--- a/js/noms/src/json-convert-test.js
+++ b/js/noms/src/json-convert-test.js
@@ -23,6 +23,10 @@ suite('jsonToNoms', () => {
     assert.isTrue(equals(new List([true]), jsonToNoms([true])));
     assert.isTrue(equals(new List([true, 42]), jsonToNoms([true, 42])));
     assert.isTrue(equals(new List([new List([88.8])]), jsonToNoms([[88.8]])));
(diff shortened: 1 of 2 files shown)


js/noms/src/json-convert.js
```diff
@@ -16,8 +16,21 @@ type JSON = string | number | boolean | null | JSONObject | JSONArray;
 type JSONObject = { [key:string]: JSON };
 type JSONArray = Array<JSON>;
 
+type NullableValue = Value | null;
+
+// Values in json can sometimes by null. If a field in a struct is null we
```
aboodman:
```
// JSON supports `null`, but Noms doesn't.
// If a field in a struct is null ...
```

js/noms/src/json-convert-test.js
```diff
@@ -46,6 +46,21 @@ suite('jsonToNoms', () => {
...
+      },
+      newStruct('', {
+        bool: true,
+        num: 42,
+        list: new List([]),
+        struct: newStruct('', {
+          key: 'val',
+        }),
+      }),
+      {
```
aboodman: Add a case of a list with a null item.

PR #2651, opened 2016-09-29
Add tx-regex, tx-replace, and tx-convert args.

willhite:
Also, make help text more complete.

### DIFF
--- a/samples/go/nomdex/nomdex_find.go
+++ b/samples/go/nomdex/nomdex_find.go
@@ -18,42 +18,54 @@ import (
 	flag "github.com/juju/gnuflag"
 )
 
-var longHelp = `Find retrieves and prints objects that satisfy the 'query' argument.
+var longFindHelp = `'nomdex find' retrieves and prints objects that satisfy the 'query' argument.
 
-Indexes are built using the 'nomdex up' command. Once built, the indexes can be referenced
-in the 'query' arg to select objects matching certain criteria. For example, if there are
-objects in the database that contain a personId and a gender field, 'nomdex up' can scan all
-the objects in a given dataset and build an index on the specified field with the following
-commands:
-   nomdex up --by gender --in-path <dsSpec> --out-ds gender-index
-   nomdex up --by personId --in-path <dsSpec> --out-ds personId-index
+Indexes are built using the 'nomdex up' command. For information about building
+indexes, see: nomdex up -h
 
-Once these indexes are built, objects can be retrieved quickly and efficiently using the
-nomdex query language. For example, the followign query could be used to find all people
-with with a personId between 1 and 2000 and who are female:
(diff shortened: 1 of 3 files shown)


willhite:
@mikegray These are the last few changes to nomdex that I'll be making for  a while. The 'tx' args were necessary for the video presentation that I'm doing. They are a bit hacky but get the job done for what I need and may be useful in other cases.

samples/go/nomdex/nomdex_find.go
```diff
@@ -18,42 +18,54 @@ import (
...
+struct Person {
+  name String,
+  geopos struct GeoPos {
+     latitude Number,
+     longitude Number,
+  }
+}
 
+Objects of this type can be indexed on the name, latitude and longitude fields
+with the following commands:
```
mikegray: Does it matter if these are in a List, Set, or Map?  You don't really say what people.value is below - does it need to be a Collection? Might be useful to specify what you are assuming in this example - e.g. "Consider objects of the following type in a Noms list".

samples/go/nomdex/nomdex_update.go
```diff
@@ -23,16 +26,63 @@ import (
...
+   nomdex up --in-path <dsSpec>.value --by .address.city --out-ds personId-index
+
+The previous commands can be understood as follows. The first command updates or
+builds an index by scanning all the objects that are reachable from |in-path| that
+have a string or number value reachable using |by| and stores the root of the
+resulting index in a dataset specified by |out-ds|.
+
+Notice that the --in-path argument has a value of '<dsSpec>.value'. The '.value'
+is not strictly necessary but it's normally useful when indexing. Since datasets
+generally point to Commit objects in NOMS, they usually have parents which are
```
mikegray: Change NOMS to Noms.

samples/go/nomdex/nomdex_update.go
```diff
@@ -126,8 +187,35 @@ func addElementsToGraphBuilder(gb *types.GraphBuilder, db datas.Database, rootOb
...
+			k2 = txRe.ReplaceAllString(string(k1), txReplaceArg)
+		} else {
+			matches := txRe.FindStringSubmatch(string(k1))
+			if len(matches) > 0 {
+				k2 = matches[len(matches)-1]
+			}
+		}
+		if txConvertArg == "number" {
+			if k2 == "" {
+				//fmt.Println("can't convert empty string, returning")
```
mikegray: Remove commented out code.

samples/go/nomdex/nomdex_find.go
mikegray: nevermind, i saw the answer below.

mikegray:
lgtm modulo spelling and typo nits

PR #2652, opened 2016-09-29
Add --force to run-all-js-tests.py

arv:
When the flag is present we force update @attic/noms. This is useful
since a lot of the samples use file: dependencies which lack a version
so they do not get updated as normal dependencies do.


arv:
@cmasone-attic PTAL

@ehalpern FYI... We talked about this last week.

PR #2655, opened 2016-09-29
JS: Auto-merge concurrent commits to noms datasets

cmasone-attic:
This patch brings JS into line with Go's handling of concurrent
commits to separate Datasets.

Fixes issue #2524

### DIFF
--- a/js/noms/src/remote-batch-store.js
+++ b/js/noms/src/remote-batch-store.js
@@ -156,11 +156,13 @@ export function makeTestingRemoteBatchStore(): BatchStore {
   return new RemoteBatchStore(3, new TestingDelegate(new MemoryStore()));
 }
 
-class TestingDelegate {
+export class TestingDelegate {
   _cs: ChunkStore;
+  preUpdateRootHook: () => Promise<void>;
 
   constructor(cs: ChunkStore) {
     this._cs = cs;
+    this.preUpdateRootHook = () => Promise.resolve();
   }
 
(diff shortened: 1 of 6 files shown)


js/noms/src/remote-batch-store.js
```diff
@@ -156,11 +156,13 @@ export function makeTestingRemoteBatchStore(): BatchStore {
...
 }
 
-class TestingDelegate {
+export class TestingDelegate {
   _cs: ChunkStore;
+  preUpdateRootHook: () => Promise<void>;
 
   constructor(cs: ChunkStore) {
     this._cs = cs;
+    this.preUpdateRootHook = () => Promise.resolve();
```
cmasone-attic: It seems like this is probably the wrong no-op function to put here, but I couldn't get anything with different syntax to work.

cmasone-attic:
PTAL

js/noms/src/remote-batch-store.js
arv: Seems fine to me.

js/noms/src/remote-batch-store.js
```diff
@@ -156,11 +156,13 @@ export function makeTestingRemoteBatchStore(): BatchStore {
   return new RemoteBatchStore(3, new TestingDelegate(new MemoryStore()));
 }
 
-class TestingDelegate {
+export class TestingDelegate {
```
arv: Is this for testing? Can this be moved to some test file?

arv:
Thanks.

LGTM

PR #2658, opened 2016-09-30
fixing part of js-tour.md

fcrick:
fixed some lines that I guess were how it used to work.


doc/js-tour.md
```diff
@@ -51,7 +51,7 @@ See [Spelling in Noms](spelling.md) for more information on database spec string
 Datasets are the main interface you'll use to work with Noms. A dataset is just a named value in the database that you can update:
 
 ```js
-let ds = new noms.Dataset(db, 'people');
+let ds = new db.getDataset('people');
```
cmasone-attic: `new db.getDataset()` isn't correct, it's just

```js
let ds = db.getDataset('people');
```

PR #2659, opened 2016-09-30
Fix JS and Go tour documents to use new Database/Dataset API

cmasone-attic:
Whoops, shoulda fixed these with the patches that updated the
APIs.


cmasone-attic:
PTAL :-)

Also, FYI @frick, @sandalwing

sandalwing:
👍

PR #2661, opened 2016-09-30
Make config.Resolve methods public and make photo-index .nomsconfig aware

ehalpern:
PTAL @willhite 
FYI @kalman, @aboodman

### DIFF
--- a/go/config/resolver.go
+++ b/go/config/resolver.go
@@ -17,7 +17,6 @@ import (
 
 type Resolver struct {
 	config      *Config
-	deferredErr error  // non-nil if error occurred during construction
 	dotDatapath string // set to the first datapath that was resolved
 }
 
@@ -29,11 +28,11 @@ func NewResolver() *Resolver {
 	c, err := FindNomsConfig()
 	if err != nil {
 		if err != NoConfig {
-			return &Resolver{deferredErr: err}
+			panic(fmt.Errorf("Failed to read .nomsconfig due to: %v", err))
 		}
 		return &Resolver{}
 	}
-	return &Resolver{c, nil, ""}
+	return &Resolver{c, ""}
 }
 
 // Print replacement if one occurred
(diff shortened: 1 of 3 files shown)


go/config/resolver.go
```diff
@@ -29,11 +28,11 @@ func NewResolver() *Resolver {
 	c, err := FindNomsConfig()
 	if err != nil {
 		if err != NoConfig {
-			return &Resolver{deferredErr: err}
+			panic(fmt.Sprintf("Failed to read .nomsconfig due to: %s", err))
```
kalman: fmt.Errorf?

samples/go/photo-index/main.go
```diff
@@ -27,21 +29,13 @@ func index() (win bool) {
...
 
 	flag.Usage = usage
 	flag.Parse(false)
 
-	if flag.NArg() == 0 {
-		flag.Usage()
-		return
-	}
-
-	if flag.NArg() == 0 {
```
willhite: Do you want to keep one of these checks around?

kalman: how were you able to remove these?

ehalpern: It's legal to have no db argument since a default might be defined in .nomsconfig. If not cfg.GetDatabase will return an error. Error reporting could be better, but my first cut here tried to allow integration with existing code with minimal changes.

ehalpern: See above

kalman: got it. Yeah that's tricky, if you want to keep the behavior the same you'd need the `--db` argument, so the value would be that you could say `--db default` or `--db photos` or whatever, instead of `--db http://demo.noms.io/...`.

Because otherwise it seems ambiguous. There can be an arbitrary number of arguments after the db and out-ds flags which are the datasets, inside db, to index.

ehalpern: My mistake. I misread the code. I restored one check. (The second was dead code).

PR #2662, opened 2016-10-01
Add a photo viewing web UI

kalman:


kalman:
We can walk through this on Monday.

It's about 90% old code, 50% of which has been modified to work with the new noms SDK and index structure.

samples/js/photos/index.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<!--
```
aboodman: doesn't it work to put the comment above the doctype?

aboodman: huh. stack overflow says no.

samples/js/photos/src/main.js
```diff
@@ -0,0 +1,79 @@
...
+import Nav from './nav.js';
+import PhotosPage from './photos-page.js';
+import Viewport from './viewport.js';
+import {createPhoto} from './photo.js';
+import type {PhotoIndex, NomsPhoto} from './types.js';
+import {Path, PathSpec, Struct} from '@attic/noms';
+
+// Cache of index paths to indices. Otherwise calls to render are pretty slow,
+// which is noticeable when resizing, toggling between full screen photos, etc.
+const indexMap: Map<string, PhotoIndex> = new Map();
```
aboodman: Why is that? Is it actually going back to http each time? I thought there was an in-memory cache?

samples/js/photos/src/main.js
```diff
@@ -0,0 +1,79 @@
...
+    try {
+      indexSpec = PathSpec.parse(indexStr);
+    } catch (e) {
+      ReactDOM.render(<div>{indexStr} is not a valid path. {e.message}.</div>, main);
+      return;
+    }
+
+    const [, indexValue] = await indexSpec.value();
+    if (!(indexValue instanceof Struct)) {
+      ReactDOM.render(<div>{indexStr} is not a valid index.</div>, main);
```
arv: Can you refactor this so that you have a function that returns a react element and then have a single call to ReactDOM.render?

samples/js/photos/src/nav.js
```diff
@@ -0,0 +1,76 @@
+// @flow
+
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+import {invariant} from '@attic/noms';
+import {Window} from './dom.js';
```
arv:
```js
import type {Window} from ...
```

samples/js/photos/src/nav.js
```diff
@@ -0,0 +1,76 @@
...
+
+export default class Nav {
+  _wnd: Window;
+  _listener: () => mixed;
```
arv: mixed?

samples/js/photos/src/nav.js
```diff
@@ -0,0 +1,76 @@
...
+
+  constructor(wnd: Window) {
+    this._wnd = wnd;
+    this._listener = () => undefined;
```
arv:
```js
() => {}
```

seems clearer.

samples/js/photos/src/nav.js
```diff
@@ -0,0 +1,76 @@
...
+
+import {invariant} from '@attic/noms';
+import {Window} from './dom.js';
+
+export default class Nav {
+  _wnd: Window;
+  _listener: () => mixed;
+
+  constructor(wnd: Window) {
+    this._wnd = wnd;
```
arv: _window

samples/js/photos/src/nav.js
```diff
@@ -0,0 +1,76 @@
...
+
+  _handleClick(e: Event) {
+    invariant(e instanceof MouseEvent);
+    if (e.button !== 0 || e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) {
+      // Modifiers to the click event may mean creating a new tab or window. Don't intercept these.
+      return;
+    }
+
+    // Find target anchor element, which may be nested e.g. <a><img></a>.
+    let anchor: ?HTMLAnchorElement = null;
```
arv: no need to assign null.

samples/js/photos/src/nav.js
```diff
@@ -0,0 +1,76 @@
...
+    for (let elem = e.target; elem; elem = elem.parentElement) {
+      invariant(elem instanceof Element);
```
arv: If you change parentElement to parentNode the invariant is no longer needed.

samples/js/photos/src/photo-grid.js
```diff
@@ -0,0 +1,355 @@
...
+import React from 'react';
+import ReactDOM from 'react-dom';
+import {
+  AsyncIterator,
+  notNull,
+} from '@attic/noms';
+import cache from './cache.js';
+import {searchToParams, paramsToSearch} from './dom.js';
+import Nav from './nav.js';
+import {default as Photo, createPhoto} from './photo.js';
```
arv: import Photo, {createPhoto} ...

samples/js/photos/src/photo-grid.js
```diff
@@ -0,0 +1,355 @@
...
+
+class PhotoComponent extends React.Component<void, PhotoCompProps, PhotoCompState> {
+  state: PhotoCompState;
+  _parentTop: number;
+  _parentLeft: number;
+  _shouldTransition: boolean;
+
+  constructor(props: PhotoCompProps) {
+    super(props);
+    this.state = {size: props.gridSize, sizeIsBest: false, url: props.url};
```
arv: line break when multiple properties

samples/js/photos/src/photo-grid.js
```diff
@@ -0,0 +1,355 @@
...
+        top: 0, right: 0, bottom: 0, left: 0,
+        zIndex: 1, // above photo grid, below fullscreen img because it's before in the DOM
+      }}/>;
+      if (!sizeIsBest) {
+        // The fullscreen image is low-res, fetch the hi-res version.
+        // disabled to avoid glaring errors, though it's still not perfect - if an image loads in
+        // less than transitionDelay then it will look a bit wonky.
+        // TODO: Also fade in the hi-res image.
+        const [bestSize, bestUrl] = photo.getBestSize(viewport.clientWidth, viewport.clientHeight);
+        cache(bestUrl).then(() => {
```
arv: Does this imply that we wait for the fullscreen image to load before we start the transition?

samples/js/photos/src/photo-grid.js
```diff
@@ -0,0 +1,355 @@
...
+  }
+
+  _getImgFullscreenStyle(): {[key: string]: any} {
+    return {
+      transition: this._maybeTransition(`transform ${transitionDelay} ease`),
+    };
+  }
+
+  _maybeTransition(transition: string): string {
+    return this._shouldTransition === true ? transition : '';
```
arv: === true?

samples/js/photos/src/photo-grid.js
```diff
@@ -0,0 +1,355 @@
...
+    if (fullscreen) {
+      params.delete('photo');
+    } else {
+      params.set('photo', photo.path);
+    }
+    nav.push(location.pathname + paramsToSearch(params));
+  }
+}
+
+type PhotoGridProps = {
```
arv: Can you split this into its own file?

samples/js/photos/src/photo-set.js
```diff
@@ -0,0 +1,134 @@
...
+
+type PhotoSetIteratorResult = AsyncIteratorResult<[number /* -date */, NomsPhoto]>
+
+export class EmptyIterator extends AsyncIterator<[number, NomsPhoto]> {
+  next(): Promise<PhotoSetIteratorResult> {
+    return Promise.resolve({done: true});
+  }
+}
+
+export class PhotoSetIterator extends AsyncIterator<[number, NomsPhoto]> {
```
arv: rename file to photo-set-iterator and make this default?

samples/js/photos/src/photo.js
```diff
@@ -0,0 +1,94 @@
...
+    return equals(this.nomsPhoto, p.nomsPhoto);
+  }
+
+  /**
+   * getBestSize gets the best photo that fits in dimensions (width, height),
+   * where best is the smallest photo which is larger than the dimensions, or
+   * failing that the widest - and one presumes therefore the largest.
+   */
+  getBestSize(width: number, height: number): SizeEntry {
+    const bestWidth = this._smallestImageAtLeast('width', width);
```
arv: I had to fix `_sizes` to work around the orientation bug in Flickr

https://github.com/arv/attic/blob/react-native-test/fe/react-native/Attic/src/photo.js#L77

samples/js/photos/src/photo.js
```diff
@@ -0,0 +1,94 @@
...
+      invariant(typeof axisValue === 'number' && typeof smallestValue === 'number');
+      if (axisValue >= atLeast && (smallest === null || axisValue < smallestValue)) {
+        smallest = entry;
+        smallestValue = axisValue;
+      }
+    }
+    return smallest;
+  }
+
+  _widestImage(): SizeEntry {
```
arv: No need for a method. Use a function if you do not access this.

samples/js/photos/src/photos-page.js
```diff
@@ -0,0 +1,426 @@
...
+  }
+
+  async _fetchFaces(): Promise<FaceState[]> {
+    const {index} = this.props;
+    const mostCommon = await this._getKeysByCount(index.facesByCount, 4);
+    const byFace = index.byFace;
+
+    // Get the most recent photo corresponding to each.
+    const self = this;
+    const faceListP = mostCommon.map(name =>
```
arv: maybe use async function here instead of all those callbacks?

samples/js/photos/src/types.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Without this file, webpack will fail.
```
arv: more info please.

samples/js/photos/src/types.js.flow
```diff
@@ -0,0 +1,47 @@
+// @flow
```
arv: Don't check in generated files.

samples/js/photos/src/main.js
kalman: done

samples/js/photos/src/nav.js
kalman: done

samples/js/photos/src/nav.js
kalman: Yeah, dunno, it was like that. `any` works too.

samples/js/photos/src/nav.js
kalman: done

samples/js/photos/src/nav.js
kalman: Except `{}` isn't a function...?

samples/js/photos/src/nav.js
kalman: done

kalman: done

samples/js/photos/src/nav.js
kalman: The problem is that e.target has no type, hence the invariant, but I can move it out of the loop.

samples/js/photos/src/photo-grid.js
kalman: done

samples/js/photos/src/photo-grid.js
kalman: done

samples/js/photos/src/photo-grid.js
kalman: Yes, load into the cache. You could probably structure it to immediately zoom the low-res version out, but that would be more transform wrangling, and this looks reasonable.

samples/js/photos/src/photo-grid.js
kalman: Yes, when I remember, I try to do this. Flow doesn't catch misspelled object properties when you use them as conditionals, if you do `=== true` then it will.

samples/js/photos/src/photo-grid.js
kalman: done (photo-grid-item.js)

samples/js/photos/src/photo-set.js
kalman: done

samples/js/photos/src/photo.js
kalman: Is there an example of this somewhere? It's not as easy as taking that code because the SDK has changed; these sizes are structs, so it's a tiny bit more involved to implement fixOrientation, but I'd like to check that it doesn't somehow figure itself out.

samples/js/photos/src/photo.js
kalman: I don't really understand this comment (I've had it before), seems fine to have it on the object and it encapsulates the function's use nicely.

But in this case, I actually do use `this`, see line below your comment.

samples/js/photos/src/photos-page.js
kalman: done

samples/js/photos/src/types.js
kalman: I've changed it to a more literal comment

samples/js/photos/src/types.js.flow
kalman: this isn't generated

samples/js/photos/src/photo.js
kalman: haha I literally just accidentally found one :-)

samples/js/photos/src/types.js.flow
arv: Then please name it `.js`. The `.flow.js` is a hack that I have a plan to fix.

samples/js/photos/src/photo.js
kalman: Ok, spent some time on this and it's not as easy as just picking the best size. We need to update all the rendering code because it assumes that the sizes it gets from getBestSize match the real sizes in the photos - which it doesn't any more.

I think we should just accept sideways photos for now, and fix them some other way, whether that is adding code everywhere to deal with sideways photos, or something else.

samples/js/photos/src/types.js.flow
kalman: As discussed: it needs to be in a .js.flow file because just .js is being picked up by eslint, which doesn't like the `declare` keyword.

... except it turns out it just doesn't like `declare export`, so a workaround is just not exporting them inline and instead have an `export {Face...}` at the end.

kalman:
PTAL I removed the flow files

samples/js/photos/src/main.js
kalman: Mm I messed up here. The answer is actually no, there is no in-memory cache, and that patch won't help.

The problem is that the spec parsing code recreates the database each time, based on the spec. There is no way to re-use a database.

Solution 1: a similar trick in the JS code that we use for leveldb, and re-use the databases. However, this is kind of icky because it means subsequent calls to database() would ignore the cache size parameter. I'd also need to do a bit more work to thread through the cacheSize parameter through the value() call, which I missed.

Solution 2: a factory pattern: refactor spec parsing into a SpecParser class which is constructed with a cacheSize parameter, then has parsePath, parseDataset, and parseDatabase methods which pass the correct cacheSize parameter.

@aboodman @arv but right now, I think we should leave this cache in place and just submit this PR.

kalman:
what is holding this up?

aboodman:
SLGTM ?

samples/js/photos/src/photos-page.js
```diff
@@ -0,0 +1,423 @@
...
+        {faces}
+        {tags}
+        {when}
+      </div>
+      <div style={{flexGrow: 1}}>{photoGrid}</div>
+    </div>;
+  }
+
+  _getTags(): React.Element<*>[] {
+    return this.state.allTags.map(tag => <label key={tag} style={{display:'block'}}>
```
arv: ws

samples/js/photos/src/photos-page.js
```diff
@@ -0,0 +1,423 @@
...
+    </label>);
+  }
+
+  _getFaces(): React.Element<*>[] {
+    const {allFaces, selectedFaces} = this.state;
+
+    const divStyle = {
+      display: 'inline-block',
+      height: faceSize,
+      overflow:'hidden',
```
arv: ws

arv: why didn't eslint catch this?

samples/js/photos/src/photos-page.js
```diff
@@ -0,0 +1,423 @@
...
+    }
+    const photoSets = (await Promise.all(photoSetPs)).filter(Boolean);
+
+    switch (photoSets.length) {
+      case 0:
+        return new EmptyIterator();
+      case 1:
+        return PhotoSetIterator.at(photoSets[0], selectedKey, null);
+      default:
+        return new PhotoSetIntersectionIterator(photoSets, selectedKey);
```
arv: Why doesn't this work for length 1?

samples/js/photos/src/types.js
```diff
@@ -0,0 +1,57 @@
...
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// implemented in types.js.flow
+
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+import {
```
arv: Why not `import type ...`?

samples/js/photos/src/photo-grid.js
```diff
@@ -0,0 +1,184 @@
...
+  availWidth: number,
+  photo: ?Photo,
+  photosIter: AsyncIterator<[number, NomsPhoto]>,
+  nav: Nav,
+  viewport: Viewport,
+}
+
+type State = {
+  photos: Photo[],
+  photosIter: AsyncIterator<[number, NomsPhoto]>,
```
kalman: done (sort of), PhotoSetIterator is now a base class, falls out reasonably

samples/js/photos/src/photo-grid.js
```diff
@@ -0,0 +1,184 @@
...
+      this._getMorePhotos(photos);
+    }
+
+    return <div style={{position: 'relative', height: top}}>{children}</div>;
+  }
+
+  async _getMorePhotos(current: Photo[]): Promise<void> {
+    const {photosIter} = this.props;
+    const moreP = [];
+    let lastNegdate = Number.NEGATIVE_INFINITY;
```
kalman: done

samples/js/photos/src/photo.js
```diff
@@ -0,0 +1,94 @@
...
+  equals,
+  invariant,
+  notNull,
+} from '@attic/noms';
+export type SizeEntry = MapEntry<PhotoSize, string>;
+
+/**
+ * createPhoto asynchronously derives a Photo instance from a NomsPhoto.
+ */
+export async function createPhoto(path: string, p: NomsPhoto): Promise<Photo> {
```
kalman: mapping back to the photo in the URL

samples/js/photos/src/photos-page.js
kalman: what is the issue here?

samples/js/photos/src/photos-page.js
kalman: good question :-\

samples/js/photos/src/photos-page.js
kalman: it does but it's a lot of unnecessary machinery.

samples/js/photos/src/types.js
kalman: cos I usually only do this if flow complains :-)

but done.

kalman:
ptal :-)

samples/js/photos/src/photos-page.js
arv: whitespace after `:`

samples/js/photos/src/types.js
arv: We don't want hard dependencies on types. Only compile time deps.

arv:
Should there be a README.md explaining how to use this?

PR #2664, opened 2016-10-03
Let DatabaseSpec.database take a cacheSize parameter

kalman:


kalman:
...

kalman:
i.e. I think this 1-line change which plumbs through a single parameter would be far harder to test than is worthwhile.

kalman:
PTAL

PR #2671, opened 2016-10-05
Remove dead deps from vendor/

cmasone-attic:
Removed:
github.com/bradfitz
github.com/gregjones
github.com/lsegal
github.com/mikkyang
github.com/nfnt
github.com/petar
github.com/peterbourgon
golang.org/x/net
golang.org/x/tools


cmasone-attic:
as far as I can tell, nothing refers to this stuff anymore.
I built and tested everything, but it's possible that I have local checkouts of some of this stuff, so we'll see what Jenkins says.

Have any objections?

kalman:
No objections, lgtm if Jenkins agrees with me.

PR #2672, opened 2016-10-06
Better error messages when index path is wrong in photo viewer

kalman:
I keep leaving out the .value from paths.


kalman:
Just something.

kalman:
Getting back to this. PTAL, I changed it a bit to make the relevant bits more obvious.

arv:
LGTM

PR #2677, opened 2016-10-06
JS: Add CommitOptions to commit method

arv:
This allows setting the meta field when you commit.

This is a version change because the signature of Database commit
changed in a non backwards compatible way. It now takes an optional
third options object instead of an optional array.


arv:
@willhite PTAL
@aboodman FYI

PR #2678, opened 2016-10-06
JS: Change FlowIssue style comment

arv:
This is a work around to make code using React Native work.

### DIFF
--- a/js/noms/src/list.js
+++ b/js/noms/src/list.js
@@ -155,8 +155,10 @@ export default class List<T: Value> extends Collection<IndexedSequence<any>> {
   /**
    * Returns a new JS array with the same values as list.
    */
-  // $FlowIssue: ???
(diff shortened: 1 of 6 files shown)


arv:
PTAL

js/noms/src/list.js
```diff
@@ -155,7 +155,7 @@ export default class List<T: Value> extends Collection<IndexedSequence<any>> {
   /**
    * Returns a new JS array with the same values as list.
    */
-  // $FlowIssue
+  // $FlowIssue: ???
```
kalman: it should be
```javascript
toJS(start: number = 0, end: (number | undefined) = undefined) {
  if (end === undefined) {
    end = this.length;
  }
  ...
}
```
it doesn't like `this` inside a default parameter list, which makes sense in a way.

PR #2680, opened 2016-10-07
Update to Facebook API v2.8

mikegray:
Upgrading to v2.8 to support tokens generated from Facebook SDK for iOS.

### DIFF
--- a/samples/js/fb/slurp/src/main.js
+++ b/samples/js/fb/slurp/src/main.js
@@ -49,7 +49,6 @@ const query = [
   'place',
   'name',
   'backdated_time',
-  'last_used_time',
   'link',
   'name_tags',
   'updated_time',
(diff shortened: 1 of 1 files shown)


samples/js/fb/slurp/src/main.js
```diff
@@ -49,7 +49,6 @@ const query = [
   'place',
   'name',
   'backdated_time',
-  'last_used_time',
```
mikegray: Field is no longer available in v2.8 API.

samples/js/fb/slurp/src/main.js
```diff
@@ -133,13 +132,19 @@ async function getPhotos(): Promise<List<any>> {
...
   let result = await new List();
-  let url = 'v2.7/me/photos/uploaded?limit=1000&fields=' + query.join(',') +
+  let url = 'v2.8/me/photos/uploaded?limit=1000&fields=' + query.join(',') +
       '&date_format=U';
   while (url) {
     const photosJSON = await callFacebook(url);
     result = await result.append(await jsonToNoms(photosJSON));
-    url = photosJSON.paging.next;
-    seen += photosJSON.data.length;
+    if (photosJSON.paging !== undefined) {
```
mikegray: If using a key generated from v2.8 and calling v2.7 API, paging and data are not returned and this crashes - so trying to protect this a little better.

aboodman:
lgtm, thanks

PR #2681, opened 2016-10-07
Fix Flickr importer to get an access token

arv:
We were using the short lived access token when we should have
exchanged that for a long term access token.

Also refactor the code a bit. Sorry!

### DIFF
--- a/samples/js/flickr/slurp/src/flickr.js
+++ b/samples/js/flickr/slurp/src/flickr.js
@@ -50,7 +50,7 @@ export default class Flickr {
     });
   }
 
-  callFlickr(method: string, params: ?{[key: string]: string}): Promise<*> {
+  callApi(method: string, params: ?{[key: string]: string}): Promise<*> {
     return new Promise((res, rej) => {
       flickrAPI.callApiMethod({
         method: method,
@@ -72,20 +72,6 @@ export default class Flickr {
         },
       });
     });
-  }
-
-  async getPhotoset(id: string): Promise<*> {
-    const json = await this.callFlickr('flickr.photosets.getPhotos', {
-      'photoset_id': id,
-      extras: 'license, date_upload, date_taken, owner_name, icon_server, original_format, ' +
-        'last_update, geo, tags, machine_tags, o_dims, views, media, path_alias, url_sq, url_t, ' +
-        'url_s, url_m, url_o',
-    });
-    return json.photoset;
-  }
-
-  getPhotosets(): Promise<*> {
-    return this.callFlickr('flickr.photosets.getList').then(v => v.photosets.photoset);
   }
 }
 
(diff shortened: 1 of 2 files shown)


arv:
PTAL

samples/js/flickr/slurp/src/main.js
```diff
@@ -65,105 +58,29 @@ async function main(): Promise<void> {
...
-  } else {
-    [authToken, authSecret, authURL] = await getAuthToken();
-    await promptForAuth(authURL);
+  const flickr = new Flickr(args['api-key'], args['api-secret'],
+                            args['access-token'], args['access-token-secret']);
+
+  if (!args['access-token'] || !args['access-token-secret']) {
+    await flickr.authenticate();
+    process.stdout.write(`Authenticated. Next time run:\n${process.argv.join(' ')
+  } --access-token=${flickr.accessToken} --access-token-secret=${flickr.accessTokenSecret}\n\n`);
```
kalman: This doesn't look right?

samples/js/flickr/slurp/src/flickr.js
```diff
@@ -0,0 +1,131 @@
+// @flow
```
kalman: Sooo how much of this changed?

arv: I added a class to keep track of the state (instead of globals). Then there is the whole two stage auth-token to access-token dance, which include getting the verification code.

samples/js/flickr/slurp/src/flickr.js
```diff
@@ -0,0 +1,131 @@
...
+
+  async authenticate(): Promise<void> {
+    const [token, secret, url] = await getAuthToken(this.apiKey, this.apiSecret);
+    const verificationCode = await promptForVerificationCode(url);
+    // $FlowIssue: Flow does not understand destructuring assignment.
+    [this.accessToken, this.accessTokenSecret] =
+        await this.getAccessToken(token, secret, verificationCode);
+  }
+
+  getAccessToken(oauthToken: string, oauthTokenSecret: string, oauthVerifier: string):
```
kalman: Could you:
 * make this private
 * have it set `this.accessToken` and `this.accessTokenSecret` itself, rather than assigning from `authenticate`.
 * not return anything (well, Promise<void> I suppose)
 * rename to `initAccessToken` or something

samples/js/flickr/slurp/src/flickr.js
```diff
@@ -0,0 +1,131 @@
...
+            } else {
+              res(data);
+            }
+          }
+        },
+      });
+    });
+  }
+
+  async getPhotoset(id: string): Promise<Struct> {
```
kalman: IMO these specific API calls should stay in main.js, but NBD.

arv: Yes. I thought about that on my to the office. flickr.js should not have any deps on noms.

arv: Done.

PR #2683, opened 2016-10-07
Update @attic/eslint

arv:
To better check flow syntax


arv:
PTAL

I have to npm publish before I update the clients but I manually checked/fixed js/noms.

PR #2684, opened 2016-10-07
Add concurrency to use of ValidatingBatchingSink

### LINKED ISSUE #1935: Increase parallelism in `ValidatingBatchSink`
There's low-hanging fruit here in two directions:

1) Loading hint chunks is done before anything else and can be done in parallel.
2) The major work in `Enqueue` is decoding the chunk and then re-encoding to check the hash. We can now use the `OrderedParallel` utility to parallelize this work to multiple cpu's while retaining the logic of linear ordering.

cmasone-attic:
There are two places where ValidatingBatchingSink could be more
concurrent: Prepare(), where it's reading in hints, and Enqueue().

Making Prepare() handle many hints concurrently is easy because the
hints don't depend on one another, so that method now just spins up
a number of goroutines and runs them all at once.

Enqueue() is more complex, because while Chunk decoding and validation
of its hash can proceed concurrently, validating that a given Chunk is
'ref-complete' requires that the chunks in the writeValue payload all
be processed in order. So, this patch uses orderedparallel to run the
new Decode() method on chunks in parallel, but then return to serial
operation before calling the modified Enqueue() method.

Fixes #1935

### DIFF
--- a/cmd/noms/noms_serve.go
+++ b/cmd/noms/noms_serve.go
@@ -37,6 +37,7 @@ func setupServeFlags() *flag.FlagSet {
 	serveFlagSet.IntVar(&port, "port", 8000, "port to listen on for HTTP requests")
 	spec.RegisterDatabaseFlags(serveFlagSet)
 	verbose.RegisterVerboseFlags(serveFlagSet)
+	profile.RegisterProfileFlags(serveFlagSet)
 	return serveFlagSet
 }
 
--- a/go/datas/remote_database_handlers.go
+++ b/go/datas/remote_database_handlers.go
@@ -18,6 +18,7 @@ import (
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
 	"github.com/attic-labs/noms/go/types"
+	"github.com/attic-labs/noms/go/util/orderedparallel"
 	"github.com/golang/snappy"
 )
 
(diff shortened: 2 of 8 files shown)


cmasone-attic:
PTAL

go/datas/remote_database_handlers.go
```diff
@@ -88,18 +89,25 @@ func handleWriteValue(w http.ResponseWriter, req *http.Request, ps URLParams, cs
 		io.Copy(ioutil.Discard, reader)
 		reader.Close()
 	}()
-	tc := types.NewTypeCache()
-	vbs := types.NewValidatingBatchingSink(cs, tc)
+	vbs := types.NewValidatingBatchingSink(cs)
 	vbs.Prepare(deserializeHints(reader))
 
-	chunkChan := make(chan *chunks.Chunk, 16)
+	chunkChan := make(chan interface{}, 16)
```
rafael-atticlabs: what if we make this buffer bigger? like 1024 or something. 16 seems like an odd number. a bigger buffer here means possibly more memory burned with pre-loading chunks off the network, but 64k seems a bit small. what about 256 (==1MB @ 4k/chunk)?

go/types/validating_batching_sink.go
```diff
@@ -16,28 +19,65 @@ type ValidatingBatchingSink struct {
...
+	pool  sync.Pool
 }
 
-func NewValidatingBatchingSink(cs chunks.ChunkStore, tc *TypeCache) *ValidatingBatchingSink {
-	return &ValidatingBatchingSink{vs: newLocalValueStore(cs), cs: cs, tc: tc}
+func NewValidatingBatchingSink(cs chunks.ChunkStore) *ValidatingBatchingSink {
+	return &ValidatingBatchingSink{
+		vs:   newLocalValueStore(cs),
+		cs:   cs,
+		pool: sync.Pool{New: func() interface{} { return NewTypeCache() }},
```
rafael-atticlabs: i think the problem with using a pool is that each of these typcaches is going to grow over time (as they accumulate more & more different types).

was it a perf problem to just create a new one for each validating batching sink?

cmd/noms/noms_serve.go
```diff
@@ -37,6 +37,7 @@ func setupServeFlags() *flag.FlagSet {
 	serveFlagSet.IntVar(&port, "port", 8000, "port to listen on for HTTP requests")
 	spec.RegisterDatabaseFlags(serveFlagSet)
 	verbose.RegisterVerboseFlags(serveFlagSet)
+	profile.RegisterProfileFlags(serveFlagSet)
```
rafael-atticlabs: what happened here?

go/types/validating_batching_sink.go
rafael-atticlabs: nevermind. i get it. the validatingbatchingsink only lives the lifetime of the handleWrite and then all of the typeCaches get dropped, right?

cmasone-attic: Yep

cmd/noms/noms_serve.go
cmasone-attic: I wanted to be able to profile the noms server, and figured that's a thing we'd want to keep.

go/types/validating_batching_sink.go
```diff
@@ -16,28 +19,65 @@ type ValidatingBatchingSink struct {
...
+// DecodedChunk holds a pointer to a Chunk and the Value that results from
+// calling DecodeFromBytes(c.Data()).
+type DecodedChunk struct {
+	Chunk *chunks.Chunk
+	Value *Value
 }
 
-// Enequeue adds a Chunk to the queue of Chunks waiting to be Put into vbs' backing ChunkStore. The instance keeps an internal buffer of Chunks, spilling to the ChunkStore when the buffer is full. If an attempt to Put Chunks fails, this method returns the BackpressureError from the underlying ChunkStore.
-func (vbs *ValidatingBatchingSink) Enqueue(c chunks.Chunk) chunks.BackpressureError {
+// DecodeUnqueued decodes c and checks that the hash of the resultng value
```
cmasone-attic: done

ghost (approved):
lgtm

PR #2687, opened 2016-10-08
photo-index: tag commit with run date

aboodman:
Would be nice to also tag with pinned path, but that's more work


aboodman:
TBR

arv:
LGTM

There is a _test.go file in here. It feels sad that it is not getting enough attention :cry:

PR #2689, opened 2016-10-10
demo-server: Remove cachingReadThroughChunkStoreFactory

### LINKED ISSUE #2688: demo-server: `noms sync` reports no work needed even to new database
In the case where `t5` is a brand new database:

```
!? noms sync http://demo.noms.io/t3::photos http://demo.noms.io/t5?access_token=xxx::photos/susan

All chunks already exist at destination! Created new dataset http://demo.noms.io/t6?access_token=xxx::photos/susan.
```

This doesn't happen when syncing from demo-server to ldb or to a localhost server.

cmasone-attic:
demo-server was using a read-through cache to allow it to serve more
concurrent requests more quickly by reducing disk I/O. As seen in issue
across Databases, leading to incorrect sync results in some instances.

Since we're not worried about demo-server load right now, simply delete
the cache.

Fixes #2688


cmasone-attic:
PTAL

aboodman:
shrug, lgtm

PR #2690, opened 2016-10-11
Remove the orderedSequence interface, move getKey to sequence

kalman:
It makes the JS port easier. It's slightly less compile time typesafe,
but the hoops aren't really worth it.


arv (approved):
LGTM

PR #2691, opened 2016-10-11
fb: fix paging logic

aboodman:
We didn't notice this before because none of us had enough fb data to cause it to page.

### DIFF
--- a/samples/js/fb/slurp/src/main.js
+++ b/samples/js/fb/slurp/src/main.js
@@ -21,6 +21,8 @@ import {
 } from '@attic/noms';
 import {stripIndent} from 'common-tags';
 
+const graphAPIHost = 'https://graph.facebook.com/';
+
 const args = argv
   .usage(stripIndent`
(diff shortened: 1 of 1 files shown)


samples/js/fb/slurp/src/main.js
```diff
@@ -142,9 +145,10 @@ async function getPhotos(): Promise<List<any>> {
 
   // Sadly we cannot issue these requests in parallel because Facebook doesn't give you a way to
   // get individual page URLs ahead of time.
+  // Note: Even though the documentation says that the max value for 'limit' is 1000, aa@ observed
+  // errors from fb servers past about 500.
   let result = await new List();
-  let url = 'v2.8/me/photos/uploaded?limit=1000&fields=' + query.join(',') +
-      '&date_format=U';
+  let url = `${graphAPIHost}v2.8/me/photos/uploaded?limit=500&date_format=U&fields=${query.join(',')}`;
```
kalman: eslint not happy because this line is too long

PR #2702, opened 2016-10-13
Introduce samples/js/dropbox/slurp

aboodman:
Doing the filter step will be a bit different because Dropbox does
not provide secret URLs for every photo by default. Requests have
to be oauth-authenticated so we'll have to figure out how to do
that.


samples/js/dropbox/slurp/src/main.js
```diff
@@ -0,0 +1,101 @@
...
+  await db.commit(out, jsonToNoms(data), {
+    meta: newStruct('', {
+      date: new Date().toISOString(),
+    }),
+  });
+  await db.close();
+  process.stdout.write(clearLine);
+}
+
+async function callDropbox(url: string, body: ?any = null): Promise<any> {
```
arv: How about `body: Object` since there is no caller that does not pass an object.

samples/js/dropbox/slurp/src/main.js
```diff
@@ -0,0 +1,101 @@
...
+  const params = {
+    method: 'POST',
+    headers: new Headers({
```
arv: At least in the web version of fetch you do not need to convert the object to a Headers object.

samples/js/dropbox/slurp/src/main.js
```diff
@@ -0,0 +1,101 @@
...
+      'Content-type': 'application/json',
+    }),
+    body: '',
+  };
+
+  if (body) {
+    params.body = JSON.stringify(body);
```
arv: maybe compute body before you set it in the params so you don't have to update it after creation.

samples/js/dropbox/slurp/src/main.js
```diff
@@ -0,0 +1,101 @@
...
+  newStruct,
+} from '@attic/noms';
+import {stripIndent} from 'common-tags';
+
+const apiHost = 'https://api.dropboxapi.com/2/';
+const clearLine = '\x1b[2K\r';
+
+const args = argv
+  .usage(stripIndent`
+    Downloads metadata from the Dropbox API
```
arv: Expand instructions.

aboodman: Done

samples/js/dropbox/slurp/src/main.js
aboodman: ok

aboodman:
PTAL

arv:
LGTM

PR #2703, opened 2016-10-13
Add some info about the  package to CONTRIBUTING.md

cmasone-attic:
Discuss some guidelines for how we do Go error reporting
in Noms code

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,12 @@ Due to legal reasons, all contributors must sign a contributor agreement, either
 
 ### Go error reporting
 
+In general, for Public API in Noms, we use the Go-style of returning errors by default.
+
+For non-exposed code, we do provide, and use, some wrappers to do Exception-style error handling. There *must* be an overriding rationale for using this style, however. One reason to use the Exception-style is that the current code doesn't know how to proceed and needs to panic, but you want to signal that a calling function somewhere up the stack might be able to recover from the failure and continue.
+
+Test code rarely, if ever, needs to or should use the following Exception-style functions.
+
 * For problems that are conceptually `asserts`, meaning that the program cannot meaningfully continue, use the following calls:
 	* d.Chk.Error()
 	* d.Chk.NoError()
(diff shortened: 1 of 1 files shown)


cmasone-attic:
PTAL, and make sure I'm capturing things accurately :-)

CONTRIBUTING.md
```diff
@@ -21,6 +21,18 @@ Due to legal reasons, all contributors must sign a contributor agreement, either
 * Tag PRs with either `toward: #<bug>` or `fixes: #<bug>` to help establish context for why a change is happening
 * Commit messages follow [Chris Beam's awesome commit message style guide](http://chris.beams.io/posts/git-commit/)
 
+### Go error reporting
+
+* For problems that are conceptually `asserts`, meaning that the program cannot meaningfully continue, use the following calls:
+	* d.Chk.Error()
+	* d.Chk.NoError()
+	* d.Chk.True()
+	* d.Chk.False()
```
willhite: I'm not sure if you're intent is to limit people to whqt you have listed here, but I find d.Chk.Nil() and d.Chk.NotNil() to be pretty useful.

CONTRIBUTING.md
```diff
@@ -21,6 +21,18 @@ Due to legal reasons, all contributors must sign a contributor agreement, either
...
+* For problems that might occur during correct processing, you can choose the Go-style of returned errors, or use the following to 'raise' a 'catchable' error (see [go/d/try.go](https://godoc.org/github.com/attic-labs/noms/go/d)):
+	* d.PanicIfError()
```
willhite: You might want to add two additional points:
In general, for Public API in Noms, we use the Go-style of returned errors by default.
* There must be an overriding rational for using the Exception-style. One reason to use the Exception style is that the current code doesn't know how to proceed and needs to panic, but we want to signal that a calling function somewhere up the stack might be able to recover from the panic and continue on.
* Test code rarely, if ever, needs or should use these Exception-style functions

CONTRIBUTING.md
cmasone-attic: talked offline; turns out anything that takes `interface{}` is not good for production code :-/

cmasone-attic:
Ok, PTAL

PR #2704, opened 2016-10-13
Make AbsolutePath's private fields public.

willhite:
Also, add String() method to PathSpec.

### DIFF
--- a/go/spec/absolute_path.go
+++ b/go/spec/absolute_path.go
@@ -18,9 +18,9 @@ import (
 var datasetCapturePrefixRe = regexp.MustCompile("^(" + datas.DatasetRe.String() + ")")
 
 type AbsolutePath struct {
-	dataset string
-	hash    hash.Hash
-	path    types.Path
+	Dataset string
+	Hash    hash.Hash
+	Path    types.Path
 }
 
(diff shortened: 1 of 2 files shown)


go/spec/absolute_path.go
```diff
@@ -18,9 +18,9 @@ import (
 var datasetCapturePrefixRe = regexp.MustCompile("^(" + datas.DatasetRe.String() + ")")
 
 type AbsolutePath struct {
-	dataset string
-	hash    hash.Hash
-	path    types.Path
+	Dataset string
```
arv: GoDoc?

kalman:
You should have @arv review I'm about to fly out

willhite:
@arv PTAL

go/spec/absolute_path.go
arv: Don't you need comments on the fields for godoc to show things about them?

willhite: I'm not sure what you mean. It seems like commenting the fields would be a little overkill on this and we have plenty of places where public types and/or their fields are not documented and they show up in godocs. Do we have a requirement that all public fields in public structs need doc? If so, I'll do it.

kalman: I think we should comment them.

PR #2705, opened 2016-10-13
Enhance TestValidatingBatchingSinkPrepare

cmasone-attic:
The test now actually validates that processing the provided
hints populates the VBS' chunk cache


cmasone-attic:
Tiny patch, PTAL

PR #2709, opened 2016-10-14
dropbox/slurp: add ability to restrict import to a path prefix

aboodman:
TBR


samples/js/dropbox/slurp/src/main.js
```diff
@@ -55,7 +59,7 @@ async function main(): Promise<void> {
   const data = [];
   let count = 0;
   let resp = await callDropbox(apiHost + 'files/list_folder', {
-    path: '',
+    path: args['path'],
```
arv: Does this default to `''`?

PR #2710, opened 2016-10-14
dropbox/slurp: path is optional

aboodman:
TBR


samples/js/dropbox/slurp/src/main.js
```diff
@@ -59,7 +59,7 @@ async function main(): Promise<void> {
   const data = [];
   let count = 0;
   let resp = await callDropbox(apiHost + 'files/list_folder', {
-    path: args['path'],
+    path: args['path'] || '',
```
arv: That answered that question.

I think you could also have provided a default value in the option.

PR #2711, opened 2016-10-15
Improve sequence read performance using read-ahead

ehalpern:
This change implements read-ahead in the sequence_cursor to speed up sequence reads.

Some before and after numbers reading sf-registered-businesses/raw:

| Origin | Before | After |
| --- | --- | --- |
| ldb | 45 MB/s | 60 MB/s |
| localhost | 7.5 MB/s | 27 MB/s |
| demo.noms.io | 140 KB/s | 2.1 MB/s |

The approach:

For each meta-sequence that contains leaf sequences, start reading ahead in
parallel and deliver in order to a buffered channel. Each advance of the cursor gets
the next sequence in the read-ahead channel.

Since this is operating on the sequence_cursor, it benefits all sequence types.

### DIFF
--- a/go/types/sequence_cursor.go
+++ b/go/types/sequence_cursor.go
@@ -34,7 +34,6 @@ func newSequenceCursor(parent *sequenceCursor, seq sequence, idx int) *sequenceC
 	return &sequenceCursor{parent, seq, idx, nil}
 }
 
-
 func (cur *sequenceCursor) length() int {
 	return cur.seq.seqLen()
 }
@@ -169,17 +168,15 @@ func newCursorAtIndex(seq sequence, idx uint64) *sequenceCursor {
 	return cur
 }
 
-// enableReadAhead turns on chunk read-ahead for this cursor.
+// enableReadAhead turns on chunk read-ahead for a leaf sequence cursor.
+// It is only intended to be called by sequenceIterator.
 //
 // Read-ahead should only be used in cases where the caller is iterating sequentially
 // forward through all chunks starting at the current position.
-//
-// should only be call by sequenceIterator
 func (cur *sequenceCursor) enableReadAhead() {
 	_, meta := cur.seq.(metaSequence)
 	d.PanicIfTrue(meta)
-	if cur.parent != nil { // ignore if there are multiple chunks
+	if cur.parent != nil { // only operative if sequence is chunked
 		cur.readAhead = newSequenceReadAhead(cur.parent, readAheadParallelism)
 	}
-
 }
--- a/go/types/sequence_cursor.go
+++ b/go/types/sequence_cursor.go
@@ -5,20 +5,22 @@
 package types
 
 import (
-	"math"
-	"os"
 	"runtime"
 
 	"github.com/attic-labs/noms/go/d"
 )
 
+// The number of go routines to devote to read-ahead.
+// The current setting provides good throughput on on a
+// 2015 MacBook Pro/2.7 GHz i5 /8 GB.
+var readAheadParallelism = runtime.NumCPU() * 64
+
 // sequenceCursor explores a tree of sequence items.
 type sequenceCursor struct {
-	parent        *sequenceCursor
-	seq           sequence
-	idx           int
-	readAheadMap  map[int]chan sequence
-	readAheadNext int
+	parent    *sequenceCursor
+	seq       sequence
+	idx       int
+	readAhead *sequenceReadAhead
 }
 
 // newSequenceCursor creates a cursor on seq positioned at idx.
@@ -29,17 +31,9 @@ func newSequenceCursor(parent *sequenceCursor, seq sequence, idx int) *sequenceC
 		idx += seq.seqLen()
 		d.PanicIfFalse(idx >= 0)
 	}
-
-	var readAheadMap map[int]chan sequence
-	// enable read-ahead if the cursor represents to a meta-sequence containing leaves
-	ms, ok := seq.(metaSequence)
-	if os.Getenv("NOMS_NO_READAHEAD") == "" && ok {
-		if ms.tuples[0].ref.height == 1 {
-			readAheadMap = make(map[int]chan sequence)
-		}
-	}
-	return &sequenceCursor{parent, seq, idx, readAheadMap, 0}
+	return &sequenceCursor{parent, seq, idx, nil}
 }
+
 
 func (cur *sequenceCursor) length() int {
 	return cur.seq.seqLen()
@@ -53,59 +47,19 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
 // It's called whenever the cursor advances/retreats to a different chunk.
 func (cur *sequenceCursor) sync() {
 	d.PanicIfFalse(cur.parent != nil)
+	if cur.readAhead != nil {
+		v := cur.parent.current()
+		hash := v.(metaTuple).ref.TargetHash()
+		if cs, ok := cur.readAhead.get(cur.parent.idx, hash); ok {
+			cur.seq = cs
+			return
+		}
+	}
 	cur.seq = cur.parent.getChildSequence()
-	if cur.readAheadMap != nil {
-		// if this cursor uses read-ahead, reset the cache for the next chunk
-		cur.readAheadMap = make(map[int]chan sequence)
-		cur.readAheadNext = 0
-	}
-}
-
-var readAheadParallelism = runtime.NumCPU() * 8
-
-func minInt(x, y int) int {
-	return int(math.Min(float64(x), float64(y)))
-}
-
-func maxInt(x, y int) int {
-	return int(math.Max(float64(x), float64(y)))
-}
-
-// readAhead (called when read-ahead is enabled) primes the next entries in the
-// read-ahead cache. It ensures that go routines have been allocated for reading
-// the next n entries in the current sequence. N is either readAheadParallelism
-// or the number of entries left in the sequence if smaller.
-func (cur *sequenceCursor) readAhead() {
-	d.PanicIfTrue(cur.readAheadMap == nil)
-	// the next position to be primed
-	cur.readAheadNext = maxInt(cur.idx, cur.readAheadNext)
-	// the last position to be primed
-	last := minInt(cur.idx+readAheadParallelism, cur.seq.seqLen()) - 1
-	// prime all positions between next and last
-	for i := cur.readAheadNext; i <= last; i += 1 {
-		futureSeq := make(chan sequence)
-		cur.readAheadMap[i] = futureSeq
-		go func(seq sequence, idx int) {
-			defer close(futureSeq)
-			futureSeq <- seq.getChildSequence(idx)
-		}(cur.seq, i)
-		cur.readAheadNext += 1
-	}
 }
 
 // getChildSequence retrieves the child at the current cursor position.
 func (cur *sequenceCursor) getChildSequence() sequence {
-	if cur.readAheadMap != nil {
-		// keep priming the cache
-		cur.readAhead()
-
-		// read the child from channel if it's there
-		if raChan, ok := cur.readAheadMap[cur.idx]; ok {
-			result := <-raChan
-			delete(cur.readAheadMap, cur.idx)
-			return result
-		}
-	}
 	return cur.seq.getChildSequence(cur.idx)
 }
 
@@ -179,8 +133,6 @@ func (cur *sequenceCursor) retreatMaybeAllowBeforeStart(allowBeforeStart bool) b
 }
 
 // clone creates a copy of the cursor
-//
-// The clone does not inherit read-ahead behavior
 func (cur *sequenceCursor) clone() *sequenceCursor {
 	var parent *sequenceCursor
 	if cur.parent != nil {
@@ -192,10 +144,9 @@ func (cur *sequenceCursor) clone() *sequenceCursor {
 type cursorIterCallback func(item interface{}) bool
 
 // iter iterates forward from the current position
+// TODO: replace calls to this with direct calls to IterSequence()
 func (cur *sequenceCursor) iter(cb cursorIterCallback) {
-	for cur.valid() && !cb(cur.getItem(cur.idx)) {
-		cur.advance()
-	}
+	iterSequence(cur, cb)
 }
 
 // newCursorAtIndex creates a new cursor over seq positioned at idx.
@@ -214,7 +165,21 @@ func newCursorAtIndex(seq sequence, idx uint64) *sequenceCursor {
 		}
 		seq = cs
 	}
-
 	d.PanicIfTrue(cur == nil)
 	return cur
+}
+
+// enableReadAhead turns on chunk read-ahead for this cursor.
+//
+// Read-ahead should only be used in cases where the caller is iterating sequentially
+// forward through all chunks starting at the current position.
+//
+// should only be call by sequenceIterator
+func (cur *sequenceCursor) enableReadAhead() {
+	_, meta := cur.seq.(metaSequence)
+	d.PanicIfTrue(meta)
+	if cur.parent != nil { // ignore if there are multiple chunks
+		cur.readAhead = newSequenceReadAhead(cur.parent, readAheadParallelism)
+	}
+
 }
--- a/go/types/sequence_cursor.go
+++ b/go/types/sequence_cursor.go
@@ -8,19 +8,17 @@ import (
 	"math"
 	"os"
 	"runtime"
-	"time"
 
 	"github.com/attic-labs/noms/go/d"
-	"github.com/attic-labs/noms/go/util/orderedparallel"
 )
 
 // sequenceCursor explores a tree of sequence items.
 type sequenceCursor struct {
-	parent      *sequenceCursor
-	seq         sequence
-	idx         int
-	readAhead   bool
-	readAheadCh chan interface{}
+	parent        *sequenceCursor
+	seq           sequence
+	idx           int
+	readAheadMap  map[int]chan sequence
+	readAheadNext int
 }
 
 // newSequenceCursor creates a cursor on seq positioned at idx.
@@ -32,21 +30,15 @@ func newSequenceCursor(parent *sequenceCursor, seq sequence, idx int) *sequenceC
 		d.PanicIfFalse(idx >= 0)
 	}
 
-	cur := &sequenceCursor{parent, seq, idx, false, nil}
-	return cur
-}
-
-// enableReadAhead turns on read-ahead.
-//
-// When enabled, the cursor will assumes it'll be moving forward and can
-// benefit from optimistically reading leaves before they are requested.
-//
-// This is a hack to minimize API changes. A better approach would be to
-// declare a cursor as "forward" at construction time, and only enable
-// read-ahead for "forward" cursors.
-func (cur *sequenceCursor) enableReadAhead() {
-	// Environment variable is temporary to allow for before & after perf tests.
-	cur.readAhead = os.Getenv("NOMS_NO_READAHEAD") == ""
+	var readAheadMap map[int]chan sequence
+	// enable read-ahead if the cursor represents to a meta-sequence containing leaves
+	ms, ok := seq.(metaSequence)
+	if os.Getenv("NOMS_NO_READAHEAD") == "" && ok {
+		if ms.tuples[0].ref.height == 1 {
+			readAheadMap = make(map[int]chan sequence)
+		}
+	}
+	return &sequenceCursor{parent, seq, idx, readAheadMap, 0}
 }
 
 func (cur *sequenceCursor) length() int {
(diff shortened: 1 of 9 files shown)


go/types/sequence_cursor.go
```diff
@@ -32,16 +62,115 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
...
+ * getChildSequence retrieves the child at the current cursor position.
+ *
+ * If the the read-ahead channel is enabled, read the child from the channel.
+ */
 func (cur *sequenceCursor) getChildSequence() sequence {
+	if cur.doReadAhead() {
+		cur.readAheadLeaves()
+	}
+	if cur.readAheadCh != nil {
+		FOR: for {
```
arv: Is this a common naming scheme for labels? This is the first time I see all caps names in go. I would probably just name it `loop`.

go/types/sequence_cursor.go
```diff
@@ -32,16 +62,115 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
...
+					// Keep looking
+				} else if cur.idx == ra.idx {
+					// Match
+					return ra.leafSeq
+				} else {
+					// Leaf missing; read directly. This can occur if the cursor is
+					// retreating.
+					break FOR
+				}
+			case <-time.After(time.Second * 5):
```
arv: Is this reasonable on a cellular network? FWIW, HTTP uses 30s.

go/types/sequence_cursor.go
```diff
@@ -85,6 +215,10 @@ func (cur *sequenceCursor) advanceMaybeAllowPastEnd(allowPastEnd bool) bool {
 	return false
 }
 
+func (cur *sequenceCursor) String() string {
```
arv: I'm not sure we want to include these?

go/types/sequence_cursor.go
```diff
@@ -32,16 +62,115 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
...
+		parallelism := int(math.Min(float64(readAheadParallelism), float64(tuplesToRead)))
+		input := make(chan interface{}, parallelism)
+
+		cur.readAheadCh = orderedparallel.New(input, func(item interface{}) interface{} {
+			i := item.(int)
+			return &readAheadRec{i, ms.getChildSequence(i)}
+		}, parallelism)
+
+		go func() {
+			for i := startIdx; i < len(ms.tuples); i++ {
```
cmasone-attic:
```go
for i := range ms.tuples {
  input <- i
}
```
It's more idiomatic

go/types/sequence_cursor.go
```diff
@@ -109,26 +243,49 @@ func (cur *sequenceCursor) retreatMaybeAllowBeforeStart(allowBeforeStart bool) b
...
+ * Implemented by searching down the tree to the leaf sequence containing idx. Each
+ * sequence cursor includes a back pointer to its parent so that it can follow the path
+ * to the next leaf chunk when the cursor exhausts the entries in the current chunk.
+ */
 func newCursorAtIndex(seq sequence, idx uint64) *sequenceCursor {
 	var cur *sequenceCursor
 	for {
 		cur = newSequenceCursor(cur, seq, 0)
+		// Assume cursor starting at 0 will be moving forward and benefit from read-ahead
+		// TODO: Inferring that a 0 idx means the cursor will be moving forward is
```
arv: Agreed. We should expose two different cursors.

go/types/sequence_cursor.go
```diff
@@ -32,16 +62,115 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
...
+	} else {
+		parallelism := int(math.Min(float64(readAheadParallelism), float64(tuplesToRead)))
+		input := make(chan interface{}, parallelism)
+
+		cur.readAheadCh = orderedparallel.New(input, func(item interface{}) interface{} {
+			i := item.(int)
+			return &readAheadRec{i, ms.getChildSequence(i)}
+		}, parallelism)
+
+		go func() {
```
cmasone-attic: I think the biggest problem here is that you have no way to stop this goroutine if the caller decides to only iterate part of the sequence. That's a problem because it means the buffered channel and this goroutine will stick around forever. Depending on what's actually captured in the closure above, the cursor could wind up living forever as well.

Now, I don't think this API has any way to indicate that the caller is done with the cursor either, so that might be something we need to add :-/ I'd want @kalman to weigh in on that

go/types/sequence_cursor.go
cmasone-attic: definitely run gofmt on this...I think labels like this are supposed to be not-indented, and maybe on a line by themselves?

go/types/sequence_cursor.go
```diff
@@ -32,16 +62,115 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
...
+ */
 func (cur *sequenceCursor) getChildSequence() sequence {
+	if cur.doReadAhead() {
+		cur.readAheadLeaves()
+	}
+	if cur.readAheadCh != nil {
+		FOR: for {
+			select {
+			case item, more := <-cur.readAheadCh:
+				if !more {
```
cmasone-attic: These cases that handle stuff which only happen if the cursor is retreating (a case in which we kinda wouldn't want to be in this loop at all anyway) make me think it might be good to have an explicit readAheadCursor or forwardCursor or something and confine the readahead logic to that.

ehalpern:
Thanks for your reviews @cmasone-attic & @arv. I've tried address all the small issues.

I need to think a bit more about how to best address the problem of more tightly constraining cursors used for read ahead. I wanted to first build a solution that would work with minimal API changes to prove it out. API changes are going to be a little hairy since the sequence_cursor is used from many call sights and it's current abstraction is a little leaky.

CC: @kalman & @rafael-atticlabs

go/types/sequence_cursor.go
kalman: The only time retreat is even used is priming the sequence chunker.

If that doesn't mean anything to you: when modifying a prolly-tree, we use a sequence chunker, which needs to
(a) set the rolling hash to the correct value (requires retreating the hash window) and
(b) re-start from the beginning of the chunk at the modification point (also requires retreating).

Straight after doing that, we advance back to where we started. Furthermore, after that PR I just sent you, we later use that same cursor to read to the end of the chunk.

We should read that ahead.

Point being: just always read ahead.

go/types/sequence_cursor.go
```diff
@@ -109,26 +232,43 @@ func (cur *sequenceCursor) retreatMaybeAllowBeforeStart(allowBeforeStart bool) b
...
 	}
-	return &sequenceCursor{parent, cur.seq, cur.idx}
+	return newSequenceCursor(parent, cur.seq, cur.idx)
 }
 
 type cursorIterCallback func(item interface{}) bool
 
+// iter iterates forward from the current position
 func (cur *sequenceCursor) iter(cb cursorIterCallback) {
+	cur.enableReadAhead() // ensure read-ahead is enabled to
```
kalman: unfinished comment?

go/types/sequence_cursor.go
```diff
@@ -109,26 +232,43 @@ func (cur *sequenceCursor) retreatMaybeAllowBeforeStart(allowBeforeStart bool) b
 	return false
 }
 
+// clone creates a copy of the cursor
+//
+// The clone does not inherit read-ahead behavior
```
kalman: Why not?

go/types/sequence_cursor.go
kalman: Yeah, that's a great point. I'm finding this a little bit hard to follow - I'm sure I could figure it out - but I wonder if a more robust approach would be to *not* try to read ahead in a gorouting that outlives the call to advance. 

Instead, have a buffer that is filled in parallel for the next N calls to advance.
* if the next chunk has already been buffered, advance can increment an index and return immediately
* if the buffer is full, read in the next N chunks in parallel

Generally speaking, trying to have goroutines that outlive a call is super complicated. Other issues include error handling.

ehalpern:
PTAL @rafael-atticlabs, @kalman
cc: @cmasone-attic, @arv 

Checkpointing the review comments, there's one remaining issue: what is the best way to manage the background read-ahead work? 

**Current approach:**

 - Implement read-ahead in the sequence cursor, limiting the read-ahead to leaf chunks
 - Whenever the cursor moves to a new metaSequence chunk (that references leaf chunks), 
   it initiates read-ahead by creating an orderedparallel goroutine pool whose input channel is 
   leaf chunk idx's and output channel is (chunk idx, chunk byte) pairs. 
 - When the cursor moves, it checks the output channel to see if the chunk it's looking for is in
   the queue. 

Trade-offs:
  - + Resource usage is bounded by the size of the goroutine pool
  - + Works well if cursor is scanning forward
  - - Tricky to handle cases where cursor isn't moving forward
  - - Tricky to managing the lifetime of the goroutine pool. To be robust, this requires 
       implementing a close() method on the cursor to ensure the goroutines are always 
       terminated. If we go this route, I'd propose introducing a new, higher level type for 
       forward iteration (sequenceIterator? or sequenceReader?)  that includes a Close()
       method. 

** Pool-less approach **

@kalman suggested in this thread and subsequent discussion considering a solution that 
doesn't require a goroutine pool. The simplest example of this would be just reading ahead 
one chunk by starting a new goroutine on each advance. Here the goroutine could read the 
chunk, deliver it on the read ahead channel and terminate. This way, goroutines are always short lived and don't require termination. 

This is an elegant idea, but I don't think there's a good way to extend it to read-ahead by more than
one chunk if we continue to deliver the read-ahead chunks on a channel. The problem is that as soon as you introduce more that one goroutine working in parallel, you need a mechanism to order the read-ahead channel (so that chunks are delivered in index order). To do this requires introducing at least one long running goroutine that reads from the unordered channel and and writes to an ordered channel. 

It's possible to handle the ordering problem without a long-lived goroutine, but I can't yet think of a solution that isn't significantly more complicated.

** Read-thru cache approach **

A third approach would be do the caching at a lower level. Here's the idea: 
 - Ensure that the ChunkStore is always fronted by ReadThruStore
 - In the sequence cursor, initiate n read-ahead by starting n goroutines to read n chunks ahead 
 - Each goroutine will just simply read a chunk. By doing so, it will make it likely that the next 
   read of the chunk will find it in the CachingChunkStore into the underlying cache.
 - On each advance, start one more goroutine to read the next chunk into the cache.

Trade-offs:
 - + Simple solution that requires very little code
 - + Doesn't require managing long-lived goroutines
 - - Requires always using CachingChunkStore to get read-ahead performance

** Next steps **

 - The ReadThruStore approach seems very attractive on it's face. It's pool-less and it's 
    simple. Is it practical to enable the the store cache by default? 
 - If the ReadThruStore approach isn't viable, I'm leaning toward improving on the 
   current approach rather than implementing (what seems like) a more complicated 
   pool-less approach. My sense is that a more constrained api for forward scanning is a 
   good direction regardless of the read-ahead implementation; and that once that exists, 
   including a Close method is not too onerous.

Look forward to your feedback.

kalman:
Nice writeup.

Could there be a 4th approach, which uses the same idea as 3 but implementing the cache on the sequenceCursor itself? A sort of hybrid between 2 and 3.

Very simplistic strawman, and without an attempt to deal with tricky issues (like overlowing into the next chunk, retreating, initialization, etc), instead of:
```go
func (c *sequenceCursor) getChildSequence() sequence {
  return c.seq.getChildSequence(c.idx)
}
```
give the cursor a map of index -> channel, like:
```go
type sequenceCursor struct {
  ...
  raCache map[int]chan sequence
}
```
and implement getChildSequence like (simplistically):
```go
func (c *sequenceCursor) getChildSequence() sequence {
  // Set up next readahead.
  raIdx = c.idx + raSize
  raChan = make(chan sequence)
  c.raCache[raIdx] = raChan
  go func() {
    raChan <- c.seq.getChildSequence(raIdx)
  }()

  // Fetch the current child sequence from the readahead cache.
  return <-c.seq.raCache[c.idx]
}
```

I also agree that either way we'll want a cursor-esque API that doesn't suffer from really subtle issues like this.

ehalpern:
Thanks @kalam: Yes, I think a hybrid is possible and your sketch looks like a good start. Your use of channels in the map is an interesting simplification over storing the actual sequence there. Makes me realize that I've only scratched the surface of what can be done with them.

I'm still keen on the idea of a shared chunk cache if practical. It simplifies this solution further by better separating concerns and it may have more far reaching performance advantages. 

If not practical, however, your suggestion seems like a good direction to go.

kalman:
Makes sense. I think @cmasone-attic and @rafael-atticlabs have thought much more about caching at the chunk store layer than I have, I'll let them speak for that.

ehalpern:
PTAL @rafael-atticlabs: 
FYI @kalman, @cmasone-attic 

I've changed the read-ahead mechanism to use short-lived go routines and unbuffered channels based on @kalman's suggestion. This removes the need to close at the end of iteration.

I'm uncertain whether to proceed with restricting read-ahead to a more limited iteration API ( sequenceCursor.iter++) or keeping it enabled for all users of sequence_cursor. 

A new, more restricted API is the clean thing to do, but it will require some surgery since there are many call sites that need to change. 

Enabling read-ahead for all callers of sequence cursor (or enabling it by a switch) has the pragmatic benefit that it works in all cases where it's needed without any code changes. It has the drawbacks that it (1) wastes resources in the cases where it's not needed (which are not numerous) and (2) may hinder more aggressive optimizations.

go/types/sequence_cursor.go
```diff
@@ -4,24 +4,41 @@
...
 		idx += seq.seqLen()
 		d.PanicIfFalse(idx >= 0)
 	}
 
-	cur := &sequenceCursor{parent, seq, idx}
-	return cur
+	var readAheadMap map[int]chan sequence
+	// enable read-ahead if the cursor represents to a meta-sequence containing leaves
+	ms, ok := seq.(metaSequence)
+	if os.Getenv("NOMS_NO_READAHEAD") == "" && ok {
```
kalman: I remember us talking about this back in the original code review, but an environment variable here isn't great. Especially since we should port this code to JS, and we can't access environment variables from JS without some contortions (webpack tricks).

When do you expect this to be used? If there is actually a use case immediately, can we make it a global variable or something and set it from clients. Like, with a command line argument even.

If it's not needed immediately, can we just take it out, and if it's needed later, add something like it back?

go/types/sequence_cursor.go
```diff
@@ -4,24 +4,41 @@
...
+		if ms.tuples[0].ref.height == 1 {
```
kalman: Ref height isn't enough unfortunately, since it's the height of the ref, not the prolly-tree.

This can be different from the tree height if the leaves are refs have heights > 1. You could get this with any nested collections, or just a collection of refs.

One way you could fix this would be to have a `seqHeight` property on `sequenceCursor` which you set from the 2 places sequenceCursors are constructed from: `newCursorAt` and `newCursorAtIndex` (and `clone()` I suppose). By stepping up the cursor chain after it's constructed.

go/types/sequence_cursor.go
```diff
@@ -32,16 +49,67 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
 	return cur.seq.getItem(idx)
 }
 
+// sync loads the sequence that the cursor index points to.
+// It's called whenever the cursor advances/retreats to a different chunk.
 func (cur *sequenceCursor) sync() {
 	d.PanicIfFalse(cur.parent != nil)
 	cur.seq = cur.parent.getChildSequence()
+	if cur.readAheadMap != nil {
```
kalman: Also check if the old sequence != new cursor?

go/types/sequence_cursor.go
```diff
@@ -32,16 +49,67 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
...
+		// if this cursor uses read-ahead, reset the cache for the next chunk
+		cur.readAheadMap = make(map[int]chan sequence)
+		cur.readAheadNext = 0
+	}
+}
+
+var readAheadParallelism = runtime.NumCPU() * 8
```
kalman: constants are usually declared at top of file.

go/types/sequence_cursor.go
```diff
@@ -32,16 +49,67 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
...
+
+func minInt(x, y int) int {
+	return int(math.Min(float64(x), float64(y)))
```
kalman: Can we add a util/math package? This function gets written every month or so. Also it seems like it would be better to save a cast and do the min/max check yourself instead of going through math.min/max.

go/types/sequence_cursor.go
kalman: I believe this comment is incorrect now?

go/types/sequence_cursor.go
```diff
@@ -109,22 +178,31 @@ func (cur *sequenceCursor) retreatMaybeAllowBeforeStart(allowBeforeStart bool) b
...
 type cursorIterCallback func(item interface{}) bool
 
+// iter iterates forward from the current position
 func (cur *sequenceCursor) iter(cb cursorIterCallback) {
 	for cur.valid() && !cb(cur.getItem(cur.idx)) {
 		cur.advance()
 	}
 }
 
+// newCursorAtIndex creates a new cursor over seq positioned at idx.
```
kalman: Thanks for adding/fixing all of these comments.

go/types/sequence_cursor.go
```diff
@@ -4,24 +4,41 @@
...
 	}
 
-	cur := &sequenceCursor{parent, seq, idx}
-	return cur
+	var readAheadMap map[int]chan sequence
+	// enable read-ahead if the cursor represents to a meta-sequence containing leaves
+	ms, ok := seq.(metaSequence)
+	if os.Getenv("NOMS_NO_READAHEAD") == "" && ok {
+		if ms.tuples[0].ref.height == 1 {
+			readAheadMap = make(map[int]chan sequence)
```
kalman: Did you think about calling readAhead here immediately? Or is it automatically happening somehow outside the diff of this PR? It certainly seems like after clone() it's not initiated immediately.

It seems reasonable to me that we don't until the first call I suppose, but you could comment about that.

go/types/sequence_cursor.go
```diff
@@ -32,16 +49,67 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
...
+
+// getChildSequence retrieves the child at the current cursor position.
 func (cur *sequenceCursor) getChildSequence() sequence {
+	if cur.readAheadMap != nil {
+		// keep priming the cache
+		cur.readAhead()
+
+		// read the child from channel if it's there
+		if raChan, ok := cur.readAheadMap[cur.idx]; ok {
+			result := <-raChan
```
kalman: With respect to readAhead (potentially) not being initiated until the first call to getChildSequence(), an optimisation here is to only readahead from index `cur.idx + 1` instead of `cur.idx` and calling `cur.seq.getChildSequence` on this thread immediately.

Because... looking at the implementation of readAhead, it seems like if cur.idx and readAheadNext are both 0 then we immediately create channels for indices `0 ... P` where P is parallelism. Then the `ok` check here will be true, and we immediately wait on the channel.

If we instead created channels for indices `1 ... P` then we wouldn't block. Hopefully this can be done without adding any complexity?

go/types/sequence_cursor.go
```diff
@@ -32,16 +49,67 @@ func (cur *sequenceCursor) getItem(idx int) sequenceItem {
...
+// or the number of entries left in the sequence if smaller.
+func (cur *sequenceCursor) readAhead() {
+	d.PanicIfTrue(cur.readAheadMap == nil)
+	// the next position to be primed
+	cur.readAheadNext = maxInt(cur.idx, cur.readAheadNext)
+	// the last position to be primed
+	last := minInt(cur.idx+readAheadParallelism, cur.seq.seqLen()) - 1
+	// prime all positions between next and last
+	for i := cur.readAheadNext; i <= last; i += 1 {
+		futureSeq := make(chan sequence)
```
kalman: I actually think we want a buffer size of 1 here, otherwise the goroutines won't terminate until the main thread reads off the channel, which might not happen if the caller never finishes iteration.

If it's 1 then the goroutine will immediately exit, and the channel's buffer will just be GC'd with the sequence.

ehalpern:
PTAL @rafael-atticlabs 
cc: @kalman 

I've reworked read-ahead to use a cursor above the leaves so that it isn't limited to the current chunk. The performance diff is significant:

| Origin | Before | One chunk at a time | Across chunks |
| --- | --- | --- |---|
| ldb | 45 MB/s | 60 MB/s | 60 MB/s |
| localhost | 7.5 MB/s | 27 MB/s | 38 MB/s |
| demo.noms.io | 140 KB/s | 2.1 MB/s | NA* |

* demo.noms.io has inconsistent throughput from office 

I compared the shorted-lived go-routine approach to the long-lived pool based approach and found that short-lived approach to have comparable performance. This result combined with the big advantage of not requiring the caller to shut down the threadpool (with a close) convinced me to take the first approach. 

I've also creating a new sequenceIterator class which encapsulates forward iteration. It's currently only used for BlobReader and sequenceCursor.iter, but I'd recommend we refactor all iterators to use this as their foundation. This reduces code and reduces the surface of sequence_iterator.

go/types/sequence_iterator.go
```diff
@@ -0,0 +1,68 @@
...
+// newSequenceIterator creates an iterator for reading a sequence
+func newSequenceIterator(seq sequence, idx uint64) *sequenceIterator {
+	sc := newCursorAtIndex(seq, idx)
+	sc.enableReadAhead()
+	return &sequenceIterator{sc}
+}
+
+
+// hasMore return true if there's more to iterate
+func (si sequenceIterator) hasMore() bool {
```
kalman: What do you think about adhering to the Set/ListIterator style of interface, and only providing a `Next()` method, which returns a Value or nil if there are no more values?

ehalpern:
Thanks for comments @kalman & @rafael-atticlabs. 

Regarding @kalman's comments around the semantics of sequenceCursor.iter and sequenceIterator.iterSequence:

The root complication here is that iter() is a method on sequenceCursor. This requires the caller to manage the cursor and raises questions about what can be done with the cursor before and after iteration. This flexibility introduces complexity without a clear benefit.

For this patch, I'll implement, your suggestions, but I'd like to follow it with a patch that removes
sequenceCursor.iter and replace it with sequenceIterator.iterSequence... functions that iterate
over a sequence without requiring the caller to create a cursor.  

> What do you think about adhering to the Set/ListIterator style of interface, and only providing a Next() method, which returns a Value or nil if there are no more values?

Unfortunately, the Next() method is not enough for the needs of BlobReader since it needs to advance the cursor (multiple increments) in one place and read the current value in another.

ehalpern:
PTAL @kalman

zcstarr:
jenkins: test this

PR #2714, opened 2016-10-15
picasa/find-photos

aboodman:


aboodman:
@kalman can you go ahead and review this -- I'll see if I have enough motivation to write tests. Espec if I can figure out some easy way to do "golden file" type tests.

aboodman:
Landing... will fix any comments async.

samples/js/picasa/find-photos/src/main.js
```diff
@@ -0,0 +1,248 @@
...
+  if (!input) {
+    throw `Invalid input spec: ${inSpec.toString()}`;
+  }
+  const outSpec = DatasetSpec.parse(args._[1]);
+  const [outDB, output] = outSpec.dataset();
+  const result = [];
+
+  await walk(input, db, async (v: any) => {
+    if (v instanceof Struct && isSubtype(sourceType, getTypeOfValue(v))) {
+      const w = parseInt(v['gphotoQ24width']['Q24t'], 10);
```
kalman: you could use `.` notation here if you wanted. Everywhere, basically.

samples/js/picasa/find-photos/src/main.js
```diff
@@ -0,0 +1,248 @@
...
+
+  const makeURL = s => {
+    thumbURL[sizePart] = `s${s}`;
+    return thumbURL.join('/');
+  };
+
+  const tuples = sizes.map(s => {
+    const r = fit(origWidth, origHeight, s, s);
+    if (r.scale > 1) {
+      // $FlowIssue: Does not understand that filter removes all null values.
```
kalman: weird, but curious - what if you changed it from `filter(t => t)` to `filter(t => !!t)`? I remember having to do that in the past.

samples/js/picasa/find-photos/src/main.js
```diff
@@ -0,0 +1,248 @@
...
+    makeURL('d'),
+  ]);
+
+  return new Map(tuples);
+}
+
+async function getFaces(photo: Object, origWidth: number, origHeight: number)
+    : Promise<?Set<Struct>> {
+  const parsePos = (s) => {
+    const pattern = /^(\d+) (\d+)$/;
```
kalman: define pattern outside so that you don't need to reparse it every time?

samples/js/picasa/find-photos/src/main.js
```diff
@@ -0,0 +1,248 @@
...
+
+async function main(): Promise<void> {
+  const inSpec = PathSpec.parse(args._[0]);
+  const pinnedSpec = await inSpec.pin();
+  if (!pinnedSpec) {
+    throw `Invalid input dataset: ${inSpec.path.dataset}`;
+  }
+  const [db, input] = await pinnedSpec.value();
+  if (!input) {
+    throw `Invalid input spec: ${inSpec.toString()}`;
```
kalman: more accurately it couldn't find a value at that spec

samples/js/picasa/find-photos/src/main.js
aboodman: I decided not to because many of them are invalid identifiers and it looked more consistent to use same pattern everywhere in this file.

samples/js/picasa/find-photos/src/main.js
aboodman: Ah, 'll try that (or `t => Boolean(t)`).

samples/js/picasa/find-photos/src/main.js
aboodman: These expressions are constant in JS, they aren't reparsed everytime.

samples/js/picasa/find-photos/src/main.js
aboodman: Yeah I don't have a succinct way to say that. I think 'invalid' covers it.

samples/js/picasa/find-photos/src/main.js
aboodman: nawp. We had to do this in other places too.

kalman: dang, https://github.com/facebook/flow/issues/1026

PR #2722, opened 2016-10-17
Extract print functionality from Diff function.

willhite:
The Diff function now returns Difference objects that can be
used in different contexts (e.g. print_diff)

### DIFF
--- a/cmd/noms/diff/diff.go
+++ b/cmd/noms/diff/diff.go
@@ -5,121 +5,138 @@
 package diff
 
 import (
-	"io"
-
 	"github.com/attic-labs/noms/go/types"
-	"github.com/attic-labs/noms/go/util/writers"
-	humanize "github.com/dustin/go-humanize"
-)
-
-type prefixOp string
-
-const (
-	ADD = "+   "
-	DEL = "-   "
 )
 
 type (
 	diffFunc     func(changeChan chan<- types.ValueChanged, stopChan <-chan struct{})
-	lineFunc     func(w io.Writer, op prefixOp, key, val types.Value) error
 	pathPartFunc func(v types.Value) types.PathPart
 	valueFunc    func(k types.Value) types.Value
 )
 
-type diffWriter struct {
-	w         io.Writer
+// Difference represents a "diff" between two Noms graphs.
+type Difference struct {
+	// Path to the Value that has changed
+	Path types.Path
+	// ChangeType indicates the type of diff: modified, added, deleted
+	ChangeType types.DiffChangeType
+	// OldValue is Value before the change, can be nil if Value was added
+	OldValue types.Value
+	// NewValue is Value after the change, can be nil if Value was removed
+	NewValue types.Value
+}
+
+// differ is used internally to hold information necessary for diffing two graphs.
+type differ struct {
+	// Channel used to send Difference objects back to caller
+	diffChan chan<- Difference
+	// Channel that caller should close() to terminate Diff function.
+	stopChan chan struct{}
+	// Use LeftRight diff as opposed to TopDown
 	leftRight bool
 }
 
-func (w diffWriter) Write(p []byte) (n int, err error) {
-	return w.w.Write(p)
-}
-
-func shouldDescend(v1, v2 types.Value) bool {
-	kind := v1.Type().Kind()
-	return !types.IsPrimitiveKind(kind) && kind == v2.Type().Kind() && kind != types.RefKind
-}
-
-// Diff writes the diff from |v1| to |v2| to |w|.
-// If |leftRight| is true then the left-right diff is used for ordered sequences - see Diff vs DiffLeftRight in Set and Map.
-func Diff(w io.Writer, v1, v2 types.Value, leftRight bool) error {
-	if v1.Equals(v2) {
-		return nil
+// Diff traverses two graphs simultaneously looking for differences. It returns
+// two channels: a DiffReceiveChan that the caller can use to iterate over the
+// diffs in the graph and a StopSendChanel that a caller can use to signal the
+// Diff function to stop processing.
+// Diff returns the Differences in depth-first first order. A 'diff' is defined
+// as one of the following conditions:
+//  * a Value is Added or Removed from a node in the graph
+//  * the type of a Value has changed in the graph
+//  * a primitive (i.e. Bool, Number, String, Ref or Blob) Value has changed
+//
+// A Difference is not returned when a non-primitive value has been modified. For
+// example, a struct field has been changed from one Value of type Employee to
+// another. Those modifications are accounted for by the Differences described
+// above at a lower point in the graph.
+//
+// If leftRight is true then the left-right diff is used for ordered sequences
+// - see Diff vs DiffLeftRight in Set and Map.
+//
+// Note: the function sends messages on diffChan and checks whether stopChan has
+// been closed to know if it needs to terminate diffing early. To function
+// properly it needs to be executed concurrently with code that reads values from
+// diffChan. The following is a typical invocation of Diff():
+//    dChan := make(chan Difference)
+//    sChan := make(chan struct{})
+//    go func() {
+//        d.Diff(s3, s4, dChan, sChan, leftRight)
+//        close(dChan)
+//    }()
+//    for dif := range dChan {
+//        <some code>
+//    }
+func Diff(v1, v2 types.Value, dChan chan<- Difference, stopChan chan struct{}, leftRight bool) {
+	d := differ{diffChan: dChan, stopChan: stopChan, leftRight: leftRight}
+	if !v1.Equals(v2) {
+		if !shouldDescend(v1, v2) {
+			d.sendDiff(Difference{Path: nil, ChangeType: types.DiffChangeModified, OldValue: v1, NewValue: v2})
+		} else {
+			d.diff(nil, v1, v2)
+		}
 	}
-
-	dw := diffWriter{w, leftRight}
-
-	if !shouldDescend(v1, v2) {
-		line(dw, DEL, nil, v1)
-		return line(dw, ADD, nil, v2)
-	}
-
-	return diff(dw, types.Path{}, nil, v1, v2)
 }
 
-func diff(w diffWriter, p types.Path, key, v1, v2 types.Value) error {
+func (d differ) diff(p types.Path, v1, v2 types.Value) bool {
 	switch v1.Type().Kind() {
 	case types.ListKind:
-		return diffLists(w, p, v1.(types.List), v2.(types.List))
+		return d.diffLists(p, v1.(types.List), v2.(types.List))
 	case types.MapKind:
-		return diffMaps(w, p, v1.(types.Map), v2.(types.Map))
+		return d.diffMaps(p, v1.(types.Map), v2.(types.Map))
 	case types.SetKind:
-		return diffSets(w, p, v1.(types.Set), v2.(types.Set))
+		return d.diffSets(p, v1.(types.Set), v2.(types.Set))
 	case types.StructKind:
-		return diffStructs(w, p, v1.(types.Struct), v2.(types.Struct))
+		return d.diffStructs(p, v1.(types.Struct), v2.(types.Struct))
 	default:
 		panic("Unrecognized type in diff function")
 	}
 }
 
-func diffLists(w diffWriter, p types.Path, v1, v2 types.List) (err error) {
+func (d differ) diffLists(p types.Path, v1, v2 types.List) (stop bool) {
 	spliceChan := make(chan types.Splice)
-	stopChan := make(chan struct{}, 1) // buffer size of 1, so this won't block if diff already finished
+	stopChan := make(chan struct{}, 1) // buffer size of 1s, so this won't block if diff already finished
 
 	go func() {
 		v2.Diff(v1, spliceChan, stopChan)
 		close(spliceChan)
 	}()
 
-	wroteHdr := false
-
 	for splice := range spliceChan {
-		if err != nil {
+		if stop {
 			break
 		}
-
 		if splice.SpRemoved == splice.SpAdded {
 			// Heuristic: list only has modifications.
-			for i := uint64(0); i < splice.SpRemoved && err == nil; i++ {
+			for i := uint64(0); i < splice.SpRemoved; i++ {
 				lastEl := v1.Get(splice.SpAt + i)
 				newEl := v2.Get(splice.SpFrom + i)
 				if shouldDescend(lastEl, newEl) {
 					idx := types.Number(splice.SpAt + i)
-					writeFooter(w, &wroteHdr)
-					err = diff(w, append(p, types.NewIndexPath(idx)), idx, lastEl, newEl)
+					stop = d.diff(append(p, types.NewIndexPath(idx)), lastEl, newEl)
 				} else {
-					writeHeader(w, p, &wroteHdr)
-					line(w, DEL, nil, v1.Get(splice.SpAt+i))
-					err = line(w, ADD, nil, v2.Get(splice.SpFrom+i))
+					p1 := copyAppend(p, types.NewIndexPath(types.Number(splice.SpAt+i)))
+					dif := Difference{p1, types.DiffChangeModified, v1.Get(splice.SpAt + i), v2.Get(splice.SpFrom + i)}
+					stop = !d.sendDiff(dif)
 				}
 			}
 			continue
 		}
 
 		// Heuristic: list only has additions/removals.
-		for i := uint64(0); i < splice.SpRemoved && err == nil; i++ {
-			writeHeader(w, p, &wroteHdr)
-			err = line(w, DEL, nil, v1.Get(splice.SpAt+i))
+		for i := uint64(0); i < splice.SpRemoved && !stop; i++ {
+			p1 := copyAppend(p, types.NewIndexPath(types.Number(splice.SpAt+i)))
+			dif := Difference{Path: p1, ChangeType: types.DiffChangeRemoved, OldValue: v1.Get(splice.SpAt + i), NewValue: nil}
+			stop = !d.sendDiff(dif)
 		}
-		for i := uint64(0); i < splice.SpAdded && err == nil; i++ {
-			writeHeader(w, p, &wroteHdr)
-			err = line(w, ADD, nil, v2.Get(splice.SpFrom+i))
+		for i := uint64(0); i < splice.SpAdded && !stop; i++ {
+			p1 := copyAppend(p, types.NewIndexPath(types.Number(splice.SpFrom+i)))
+			dif := Difference{Path: p1, ChangeType: types.DiffChangeAdded, OldValue: nil, NewValue: v2.Get(splice.SpFrom + i)}
+			stop = !d.sendDiff(dif)
 		}
 	}
 
-	err = writeFooter(w, &wroteHdr)
-
-	if err != nil {
+	if stop {
 		stopChan <- struct{}{}
 		// Wait for diff to stop.
 		for range spliceChan {
@@ -128,11 +145,11 @@ func diffLists(w diffWriter, p types.Path, v1, v2 types.List) (err error) {
 	return
 }
 
-func diffMaps(w diffWriter, p types.Path, v1, v2 types.Map) error {
-	return diffOrdered(w, p, line,
+func (d differ) diffMaps(p types.Path, v1, v2 types.Map) bool {
+	return d.diffOrdered(p,
 		func(v types.Value) types.PathPart { return types.NewIndexPath(v) },
 		func(cc chan<- types.ValueChanged, sc <-chan struct{}) {
-			if w.leftRight {
+			if d.leftRight {
 				v2.DiffLeftRight(v1, cc, sc)
 			} else {
 				v2.Diff(v1, cc, sc)
@@ -144,11 +161,11 @@ func diffMaps(w diffWriter, p types.Path, v1, v2 types.Map) error {
 	)
 }
 
-func diffStructs(w diffWriter, p types.Path, v1, v2 types.Struct) error {
+func (d differ) diffStructs(p types.Path, v1, v2 types.Struct) bool {
 	str := func(v types.Value) string {
 		return string(v.(types.String))
 	}
-	return diffOrdered(w, p, field,
+	return d.diffOrdered(p,
 		func(v types.Value) types.PathPart { return types.NewFieldPath(str(v)) },
 		func(cc chan<- types.ValueChanged, sc <-chan struct{}) {
 			v2.Diff(v1, cc, sc)
@@ -159,23 +176,28 @@ func diffStructs(w diffWriter, p types.Path, v1, v2 types.Struct) error {
 	)
 }
 
-func diffSets(w diffWriter, p types.Path, v1, v2 types.Set) error {
-	return diffOrdered(w, p, line,
-		func(v types.Value) types.PathPart { return types.NewIndexPath(v) },
+func (d differ) diffSets(p types.Path, v1, v2 types.Set) bool {
+	return d.diffOrdered(p,
+		func(v types.Value) types.PathPart {
+			if types.ValueCanBePathIndex(v) {
+				return types.NewIndexPath(v)
+			}
+			return types.NewHashIndexPath(v.Hash())
+		},
 		func(cc chan<- types.ValueChanged, sc <-chan struct{}) {
-			if w.leftRight {
+			if d.leftRight {
 				v2.DiffLeftRight(v1, cc, sc)
 			} else {
 				v2.Diff(v1, cc, sc)
 			}
 		},
-		func(k types.Value) types.Value { return nil },
+		func(k types.Value) types.Value { return k },
 		func(k types.Value) types.Value { return k },
 		func(k types.Value) types.Value { return k },
 	)
 }
 
-func diffOrdered(w diffWriter, p types.Path, lf lineFunc, ppf pathPartFunc, df diffFunc, kf, v1, v2 valueFunc) (err error) {
+func (d differ) diffOrdered(p types.Path, ppf pathPartFunc, df diffFunc, kf, v1, v2 valueFunc) (stop bool) {
 	changeChan := make(chan types.ValueChanged)
 	stopChan := make(chan struct{}, 1) // buffer size of 1, so this won't block if diff already finished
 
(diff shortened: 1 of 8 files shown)


cmd/noms/diff/diff_test.go
```diff
@@ -311,11 +346,217 @@ func TestNomsRefDiff(t *testing.T) {
...
+		PrintDiff(buf, r1, r2, leftRight)
 
 		test.EqualsIgnoreHashes(t, expected, buf.String())
 	}
 
 	tf(true)
 	tf(false)
 }
+
+func TestNomsDiffMap(t *testing.T) {
```
mikegray: Wondering if you can't combine these print vs. non-print versions rather than duplicating the creation and expected stuff?

willhite:
fyi @aboodman

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
+	// NewValue is Value after the change, can be nil if Value was removed
+	NewValue types.Value
 }
 
-func (w diffWriter) Write(p []byte) (n int, err error) {
-	return w.w.Write(p)
-}
+type (
+	// Differences is a slice of Difference
+	Differences []Difference
```
kalman: Never used?

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
-		line(dw, DEL, nil, v1)
-		return line(dw, ADD, nil, v2)
-	}
+// Diff traverses two graphs simultaneously looking for differences. It returns
+// two channels: a DiffReceiveChan that the caller can use to iterate over the
+// diffs in the graph and a StopSendChanel that a caller can use to signal the
+// Diff function to stop processing.
+// Diff returns the Differences in depth-first first order. A 'diff' is defined
+// as one of the following conditions:
+// * a Value is Added or Removed from a node in the graph
```
kalman: You need to indent these bullet points 1 space for godoc to render properly.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
+// * the type of a Value has changed in the graph
+// * a primitive (i.e. Bool, Number, String, Ref or Blob) Value has changed
+//
+// A Difference is not returned when a non-primitive value has been modified. For
+// example, a struct field has been changed from one Value of type Employee to
+// another. Those modifications are accounted for by the Differences described
+// above at a lower point in the graph.
+//
+// If |leftRight| is true then the left-right diff is used for ordered sequences
```
kalman: FYI comments like `|leftRight|` look pretty weird in godoc, idiom is just to use `leftRight` I think.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
+	DiffReceiveChan <-chan Difference
 
-// Diff writes the diff from |v1| to |v2| to |w|.
-// If |leftRight| is true then the left-right diff is used for ordered sequences - see Diff vs DiffLeftRight in Set and Map.
-func Diff(w io.Writer, v1, v2 types.Value, leftRight bool) error {
-	if v1.Equals(v2) {
-		return nil
-	}
+	// StopSendChan is used by the caller to signal that Diff() should stop processing
+	StopSendChan chan<- struct{}
```
kalman: Likewise, call this `StopChan` and remove `stopReceiveChan`.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
+type (
+	// Differences is a slice of Difference
+	Differences []Difference
 
-func shouldDescend(v1, v2 types.Value) bool {
-	kind := v1.Type().Kind()
-	return !types.IsPrimitiveKind(kind) && kind == v2.Type().Kind() && kind != types.RefKind
-}
+	// DiffReceiveChan is used by caller to process diffs
+	DiffReceiveChan <-chan Difference
```
kalman: How about you just call this DiffChan, then remove the `diffSendChan` type.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
+//
+// A Difference is not returned when a non-primitive value has been modified. For
+// example, a struct field has been changed from one Value of type Employee to
+// another. Those modifications are accounted for by the Differences described
+// above at a lower point in the graph.
+//
+// If |leftRight| is true then the left-right diff is used for ordered sequences
+// - see Diff vs DiffLeftRight in Set and Map.
+//
+func Diff(v1, v2 types.Value, leftRight bool) (DiffReceiveChan, StopSendChan) {
```
kalman: The idiom we've been using is to pass the channels into functions, not have them create channels themselves. Likewise, don't create a goroutine here.

That pushes parallelism decisions to the caller.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
+	diffChan := make(chan Difference, 256)
```
kalman: And passing channels into this function rather than creating them lets the caller decide the buffer size. 256 is probably unnecessarily large.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
+	stopChan := make(chan struct{}, 1)
```
kalman: FYI: a mistake I made when implementing this was to make stopChan a buffered channel and have it read off the value to finish. Instead it should be possible for caller to just `close` the stop channel and have that do the right thing.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
 
-	return diff(dw, types.Path{}, nil, v1, v2)
+	go func() {
+		if !v1.Equals(v2) {
+			if !shouldDescend(v1, v2) {
+				diffChan <- Difference{Path: types.Path{}, ChangeType: types.DiffChangeModified, OldValue: v1, NewValue: v2}
```
kalman: I would leave `Path` unallocated, saves an allocation. I think it's go idiom to avoid allocating empty slices, `len` and `append` work fine on nil.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,136 @@
...
+			}
+		}
+		close(diffChan)
+	}()
+
+	return diffChan, stopChan
 }
 
-func diff(w diffWriter, p types.Path, key, v1, v2 types.Value) error {
+func diff(dc diffSendChan, sc stopReceiveChan, p types.Path, v1, v2 types.Value, lr bool) bool {
```
kalman: Could you leave a struct like `diffWriter` in place, and add the channels to that, instead of passing 3 parameters everywhere?

cmd/noms/diff/diff.go
```diff
@@ -184,40 +207,33 @@ func diffOrdered(w diffWriter, p types.Path, lf lineFunc, ppf pathPartFunc, df d
 		close(changeChan)
 	}()
 
-	wroteHdr := false
-
 	for change := range changeChan {
-		if err != nil {
+		if stop = stopSent(sc); stop {
```
kalman: I prefer the way that the ordered diff does this. Remove this code, then instead of
```
dc <- Difference{...}
```
have
```
if !sendDiff(dc, sc, Difference{...}) {
  break Range
}
```
(you'll need a Range label on the outer `for` to be able to break out of the switch... or use some variable I suppose)

Even better: if you leave that struct in place that bundles state, you can put the `sendDiff` function on that, and not need to pass in those 2 parameters separately.

cmd/noms/diff/diff_test.go
```diff
@@ -5,12 +5,13 @@
 package diff
 
 import (
-	"bytes"
 	"strings"
 	"testing"
 
+	"bytes"
```
kalman: Move back?

cmd/noms/diff/diff_test.go
```diff
@@ -129,6 +141,10 @@ func TestNomsSetDiff(t *testing.T) {
 -   }
   }
 `
+	expectedPaths2 := []string{
+		`[#3rsij03n33f3qefnfcf9kmh9qubst0km]`,
```
kalman: Could you construct these strings instead of hard coding them? It's a weaker assertion otherwise.

cmd/noms/diff/print_diff.go
```diff
@@ -0,0 +1,153 @@
...
+
+type (
+	printFunc func(w io.Writer, op prefixOp, key, val types.Value) error
+)
+
+// PrintDiff writes a textual reprensentation of the diff from |v1| to |v2|
+// to |w|. If |leftRight| is true then the left-right diff is used for ordered
+// sequences - see Diff vs DiffLeftRight in Set and Map.
+func PrintDiff(w io.Writer, v1, v2 types.Value, leftRight bool) (err error) {
+	if !shouldDescend(v1, v2) {
```
kalman: Why do you need this check? it should just fall out of the diff API you wrote.

cmd/noms/diff/print_diff.go
```diff
@@ -0,0 +1,153 @@
...
+
+		var key types.Value
+		var pfunc printFunc
+
+		switch parentEl.(type) {
+		case types.Map:
+			key = lastPart.(types.IndexPath).Index
+			pfunc = line
+		case types.Set:
+			key = nil
```
kalman: don't need this assignment (not the other `nil` one)

cmd/noms/diff/print_diff.go
```diff
@@ -0,0 +1,153 @@
...
+			pfunc = line
```
kalman: maybe assign `pfunc` to line initially.

willhite:
@kalman PTAL

cmd/noms/diff/diff.go
```diff
@@ -5,134 +5,146 @@
...
+// above at a lower point in the graph.
+//
+// If |leftRight| is true then the left-right diff is used for ordered sequences
+// - see Diff vs DiffLeftRight in Set and Map.
+//
+
+func (d Differ) Diff(v1, v2 types.Value) {
+	if !v1.Equals(v2) {
+		if !shouldDescend(v1, v2) {
+			d.sendDiff(Difference{Path: nil, ChangeType: types.DiffChangeModified, OldValue: v1, NewValue: v2})
```
kalman: You don't need the `{Path:nil}`, `nil` is the default value.

cmd/noms/diff/diff.go
```diff
@@ -5,134 +5,146 @@
...
 			// Heuristic: list only has modifications.
-			for i := uint64(0); i < splice.SpRemoved && err == nil; i++ {
+			for i := uint64(0); i < splice.SpRemoved; i++ {
 				lastEl := v1.Get(splice.SpAt + i)
 				newEl := v2.Get(splice.SpFrom + i)
 				if shouldDescend(lastEl, newEl) {
 					idx := types.Number(splice.SpAt + i)
-					writeFooter(w, &wroteHdr)
-					err = diff(w, append(p, types.NewIndexPath(idx)), idx, lastEl, newEl)
+					if !d.diff(append(p, types.NewIndexPath(idx)), lastEl, newEl) {
```
kalman: I would prefer you didn't make cosmetic changes like this. Transforming the old code into the new should be:
```
cont = d.diff(append(p, types.NewIndexPath(idx)), lastEl, newEl)
```
and keep the `err != nil` check. It would make it easier to review this patch if you changed it back.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,126 @@
...
 	case types.StructKind:
-		return diffStructs(w, p, v1.(types.Struct), v2.(types.Struct))
+		return d.diffStructs(p, v1.(types.Struct), v2.(types.Struct))
 	default:
 		panic("Unrecognized type in diff function")
 	}
 }
 
-func diffLists(w diffWriter, p types.Path, v1, v2 types.List) (err error) {
+func (d Differ) diffLists(p types.Path, v1, v2 types.List) (cont bool) {
```
kalman: I'm aware of avoiding more churn with this PR (so feel free not to do this) *but*:

Consistency wise, the other APIs we have tend to have a `stop` variable not a `cont` variable. E.g. Iter/IterAll, Walk. The advance is that you don't need to initialize `cont` to the common case - that you're *not* stopping.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,126 @@
...
 				if shouldDescend(lastEl, newEl) {
 					idx := types.Number(splice.SpAt + i)
-					writeFooter(w, &wroteHdr)
-					err = diff(w, append(p, types.NewIndexPath(idx)), idx, lastEl, newEl)
+					cont = d.diff(append(p, types.NewIndexPath(idx)), lastEl, newEl)
 				} else {
-					writeHeader(w, p, &wroteHdr)
-					line(w, DEL, nil, v1.Get(splice.SpAt+i))
-					err = line(w, ADD, nil, v2.Get(splice.SpFrom+i))
+					p1 := append(p, types.NewIndexPath(types.Number(splice.SpAt+i)))
```
kalman: This is subtle. Short story: you need to copy `p`.

Long story:

If you have a slice `p := {1, 2, 3}` and then do `p1 := append(p, 42)` then it may modify the underlying storage of `p`. If you then did `p2 := append(p, 43)` then `p1` will actually have the value `{1, 2, 3, 43}`.

You can see this happen: https://play.golang.org/p/gcigJxy5CH

The old code did this as well - so, subtle there too - but I think introducing channels into the equation makes it actually incorrect. E.g. if the channel is buffered and has `p1` in it, then creating that `p2` variable will modify it.

You can see this happen: https://play.golang.org/p/8nJIeTBxMA

cmd/noms/diff/diff.go
```diff
@@ -159,23 +164,29 @@ func diffStructs(w diffWriter, p types.Path, v1, v2 types.Struct) error {
 	)
 }
 
-func diffSets(w diffWriter, p types.Path, v1, v2 types.Set) error {
-	return diffOrdered(w, p, line,
-		func(v types.Value) types.PathPart { return types.NewIndexPath(v) },
+func (d Differ) diffSets(p types.Path, v1, v2 types.Set) bool {
+	return d.diffOrdered(p,
+		func(v types.Value) types.PathPart {
+			if types.ValueCanBePathIndex(v) {
```
kalman: Oh, nice.

cmd/noms/diff/diff_test.go
```diff
@@ -70,7 +72,13 @@ func createStruct(name string, kv ...interface{}) types.Struct {
 	return types.NewStruct(name, fields)
 }
 
-func TestNomsMapdiff(t *testing.T) {
+func newDiffer(leftRight bool) (Differ, chan Difference) {
+	dChan := make(chan Difference, 64)
```
kalman: why buffer?

cmd/noms/diff/diff_test.go
```diff
@@ -70,7 +72,13 @@ func createStruct(name string, kv ...interface{}) types.Struct {
...
+	sChan := make(chan struct{}, 1)
+	return Differ{DiffChan: dChan, StopChan: sChan, LeftRight: leftRight}, dChan
```
kalman: Caller could just read `dChan` off the `Differ`?

cmd/noms/diff/diff_test.go
```diff
@@ -137,19 +164,68 @@ func TestNomsSetDiff(t *testing.T) {
...
+		PrintDiff(buf, s1, s2, leftRight)
 		assert.Equal(expected1, buf.String())
 
+		d, dChan := newDiffer(leftRight)
+		go func() {
+			d.Diff(s1, s2)
+			close(dChan)
+		}()
+		paths := pathsFromDiffResults(dChan)
+		assert.Equal(expectedPaths1, paths)
```
kalman: This pattern occurs in a lot of tests. Could you write a function which gets the diff paths, so that you can write:
```
// same as before
buf := &bytes.Buffer{}
PrintDiff(buf, s1, s2, leftRight)
assert.Equal(expected1, buf.String())
// new
paths := PathDiff(s1, s2, leftRight)
assert.Equal(expectedPath1, paths)
```

cmd/noms/diff/print_diff.go
```diff
@@ -0,0 +1,163 @@
...
+func PrintDiff(w io.Writer, v1, v2 types.Value, leftRight bool) (err error) {
+	// In the case where the diff involves two simple values, just print out the
+	// diff and return. This is needed because the code below assumes that the
+	// values being compared have a parent.
+	if !shouldDescend(v1, v2) {
+		line(w, DEL, nil, v1)
+		return line(w, ADD, nil, v2)
+	}
+
+	dChan := make(chan Difference, 128)
```
kalman: This buffer seems very large, it's expensive to allocate. I don't imagine buffering helps much anyway.

cmd/noms/diff/print_diff.go
```diff
@@ -0,0 +1,163 @@
...
+	var lastParentPath types.Path
+	wroteHdr := false
+	firstTime := true
+
+	for d := range dChan {
+		parentPath := d.Path[:len(d.Path)-1]
+		parentPathChanged := !parentPath.Equals(lastParentPath)
+		lastParentPath = parentPath
+		if parentPathChanged && wroteHdr {
+			writeFooter(w, &wroteHdr)
```
kalman: You need to make sure that `err` is assigned if writing fails. If you have a chain of write calls, just assigning on the last one should be enough:
```
write()
write()
err = write()
```
but here all of the writes are behind conditionals, so I think you need to assign on every call. I.e. `err = writeFooter`. Same everywhere else basically (and writeHeader etc).

cmd/noms/diff/print_diff.go
```diff
@@ -0,0 +1,163 @@
...
+		case types.Struct:
+			key = types.String(lastPart.(types.FieldPath).Name)
+			pfunc = field
+		case types.List:
+			// default values are ok
+		}
+
+		if d.OldValue != nil {
+			err = pfunc(w, DEL, key, d.OldValue)
+			if err != nil {
```
kalman: I'd move this to the end of the range loop so you only check `err` once, especially since err will be assigned on every call to writeFooter/writeHeader.

cmd/noms/diff/diff.go
```diff
@@ -101,30 +112,30 @@ func (d Differ) diffLists(p types.Path, v1, v2 types.List) (cont bool) {
 				newEl := v2.Get(splice.SpFrom + i)
 				if shouldDescend(lastEl, newEl) {
 					idx := types.Number(splice.SpAt + i)
-					cont = d.diff(append(p, types.NewIndexPath(idx)), lastEl, newEl)
+					stop = !d.diff(append(p, types.NewIndexPath(idx)), lastEl, newEl)
 				} else {
-					p1 := append(p, types.NewIndexPath(types.Number(splice.SpAt+i)))
+					p1 := copyPath(append(p, types.NewIndexPath(types.Number(splice.SpAt+i))))
```
kalman: you need to copy `p`, not the result of `append(p, ...)`.
```go
p1 := append(copyPath(p), types.NewIndexPath(...))
```

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,137 @@
...
 			// Heuristic: list only has modifications.
-			for i := uint64(0); i < splice.SpRemoved && err == nil; i++ {
+			for i := uint64(0); i < splice.SpRemoved; i++ {
 				lastEl := v1.Get(splice.SpAt + i)
 				newEl := v2.Get(splice.SpFrom + i)
 				if shouldDescend(lastEl, newEl) {
 					idx := types.Number(splice.SpAt + i)
-					writeFooter(w, &wroteHdr)
-					err = diff(w, append(p, types.NewIndexPath(idx)), idx, lastEl, newEl)
+					stop = !d.diff(append(p, types.NewIndexPath(idx)), lastEl, newEl)
```
kalman: should this be `stop = d.diff(...)` not `stop = !d.diff(...)`?

cmd/noms/diff/diff.go
```diff
@@ -184,102 +205,66 @@ func diffOrdered(w diffWriter, p types.Path, lf lineFunc, ppf pathPartFunc, df d
...
-			writeHeader(w, p, &wroteHdr)
-			err = lf(w, DEL, k, v1(change.V))
+			dif := Difference{Path: p1, ChangeType: types.DiffChangeRemoved, OldValue: v1(change.V), NewValue: nil}
+			stop = !d.sendDiff(dif)
 		case types.DiffChangeModified:
 			c1, c2 := v1(change.V), v2(change.V)
 			if shouldDescend(c1, c2) {
-				writeFooter(w, &wroteHdr)
-				err = diff(w, append(p, ppf(k)), change.V, c1, c2)
+				stop = !d.diff(p1, c1, c2)
```
kalman: and here

cmd/noms/diff/diff.go
```diff
@@ -184,102 +205,66 @@ func diffOrdered(w diffWriter, p types.Path, lf lineFunc, ppf pathPartFunc, df d
...
 	}
-	write(w, []byte("Blob ("))
-	write(w, []byte(humanize.Bytes(v.(types.Blob).Len())))
-	return write(w, []byte(")"))
-}
 
-func write(w io.Writer, b []byte) error {
-	_, err := w.Write(b)
-	return err
+	d.diffChan <- dif
```
kalman: This isn't quite right.

Consider if the consumer has not sent a message to the stop chan when this code runs. It falls through the select here, and this code blocks trying to send on the diffChan. *Then* the consumer sends on the stop chan. This will deadlock.

Instead you need to do:
```go
select {
case <-d.stopChan:
  return false
case d.diffChan <- dif:
  return true
}
```

cmd/noms/diff/diff_test.go
```diff
@@ -70,7 +72,32 @@ func createStruct(name string, kv ...interface{}) types.Struct {
 	return types.NewStruct(name, fields)
 }
 
-func TestNomsMapdiff(t *testing.T) {
+func pathsFromDiff(v1, v2 types.Value, leftRight bool) []string {
+	dChan := make(chan Difference, 64)
```
kalman: Please see comment about buffers. I believe buffered channels should be the exception not the rule. The algorithm is only going to be as fast as the diff is produced or consumed anyway, and adding a buffer isn't going to change that, it's just going to use up more memory and make creating the buffer slow.

It helps to respond to comments with "done" or something else if you want to discuss something further. Otherwise I need to go through the PR every time to find what hasn't changed.

cmd/noms/diff/print_diff.go
```diff
@@ -0,0 +1,158 @@
...
+func PrintDiff(w io.Writer, v1, v2 types.Value, leftRight bool) (err error) {
+	// In the case where the diff involves two simple values, just print out the
+	// diff and return. This is needed because the code below assumes that the
+	// values being compared have a parent.
+	if !shouldDescend(v1, v2) {
+		line(w, DEL, nil, v1)
+		return line(w, ADD, nil, v2)
+	}
+
+	dChan := make(chan Difference, 16)
```
kalman: Buffer comment.

cmd/noms/diff/diff_test.go
kalman: > buffer slow

Oh, and the other big disadvantage of buffered channels is that they can hide deadlock bugs.

cmd/noms/diff/diff.go
kalman: Same for the other instances of course.

cmd/noms/diff/diff.go
willhite: yes, good eye. This actually caused a test to fail and I tracked it down from there.

cmd/noms/diff/diff.go
willhite: same here

cmd/noms/diff/diff.go
willhite: Good point. I was trying to ensure that we stopped immediately when stopChan got closed but I didn't think about the possible deadlock.

cmd/noms/diff/print_diff.go
willhite: It seems like a small amount of buffering (at a minimum would be good here so that the processing of the diff can move forward in the event that the printing is stopped or slow for some reason. I wasn't aware that creating a buffer for channels was any more expensive than creating an array. These objects are pretty small.

cmd/noms/diff/diff.go
```diff
@@ -184,102 +206,64 @@ func diffOrdered(w diffWriter, p types.Path, lf lineFunc, ppf pathPartFunc, df d
...
 
-func write(w io.Writer, b []byte) error {
-	_, err := w.Write(b)
-	return err
+// stopSent returns true if a message has been sent to this StopChannel
+func (d differ) sendDiff(dif Difference) bool {
+	select {
+	case <-d.stopChan:
+		return false
+    case d.diffChan <- dif:
```
kalman: Indentation looks wrong.

cmd/noms/diff/diff.go
```diff
@@ -5,121 +5,138 @@
...
 			// Heuristic: list only has modifications.
-			for i := uint64(0); i < splice.SpRemoved && err == nil; i++ {
+			for i := uint64(0); i < splice.SpRemoved; i++ {
 				lastEl := v1.Get(splice.SpAt + i)
 				newEl := v2.Get(splice.SpFrom + i)
 				if shouldDescend(lastEl, newEl) {
 					idx := types.Number(splice.SpAt + i)
-					writeFooter(w, &wroteHdr)
-					err = diff(w, append(p, types.NewIndexPath(idx)), idx, lastEl, newEl)
+					stop = d.diff(append(p, types.NewIndexPath(idx)), lastEl, newEl)
```
kalman: You need to copy this too.

I agree with your observation that this is a reallocation, maybe instead of `append(copyPath(p), foo)` it chould be `copyAppend(p, foo)` then you can allocate a buffer of size +1.

cmd/noms/diff/diff_test.go
```diff
@@ -70,7 +72,32 @@ func createStruct(name string, kv ...interface{}) types.Struct {
...
+func pathsFromDiff(v1, v2 types.Value, leftRight bool) []string {
+	dChan := make(chan Difference)
+	sChan := make(chan struct{})
+
+	go func() {
+		Diff(v1, v2, dChan, sChan, leftRight)
+		close(dChan)
+	}()
+
+	paths := []string{}
```
kalman: You should probably declare this like `var paths []string` so that it's nil when there are no paths. Or even name the return value.

cmd/noms/diff/diff_test.go
```diff
@@ -237,85 +342,131 @@ func TestNomsListDiff(t *testing.T) {
...
 	tf(true)
 	tf(false)
 }
 
-func TestNomsTypeDiff(t *testing.T) {
+func TestNomsDiffPrintType(t *testing.T) {
 	assert := assert.New(t)
 
 	expected1 := "-   List<Number>\n+   List<String>\n"
+	expectedPaths1 := []string{``}
```
kalman: Using a raw string instead of a normal string (i.e. `''`) is a bit weird but ok.

cmd/noms/diff/diff_test.go
```diff
@@ -237,85 +342,131 @@ func TestNomsListDiff(t *testing.T) {
...
+	testCases := []string{
+		``,
+		`["key"]`,
+		`["key"].field1`,
+		`["key"]@key.field1`,
+	}
+
+	for _, s1 := range testCases {
+		p := mustParsePath(assert, s1)
+		p1 := copyPath(p)
```
kalman: You should really also modify `p` after this to make sure it's an actual copy. Like, `p[0] = <something>`

cmd/noms/diff/print_diff.go
kalman: Ok you win.

cmd/noms/diff/diff.go
willhite: done

cmd/noms/diff/diff.go
willhite: fixed

cmd/noms/diff/diff_test.go
willhite: done, question, should we always use declarations rather than an empty object initialization. For example, if I have a struct foo, should I use:
var f foo
or
f := foo{}
?

cmd/noms/diff/diff_test.go
willhite: changed

cmd/noms/diff/diff_test.go
kalman: Only for slices I think.

PR #2729, opened 2016-10-18
Use the same cursor when initializing and finalizing the chunker

kalman:
Previously we would clone them from the original cursor, to (a) not
modify the original cursor, and (b) have initialization and finalization
not interfere with each other.

However, this isn't necessary and it just creates unnecessary churn. For
example, when we read-ahead, it would be wasteful to re-read the
read-head chunks from initialization.


kalman:
@ehalpern PTAL, I'm going to reference this PR in the PR you sent me.

@rafael-atticlabs FYI if you're around.

ehalpern:
@kalman: LGTM

PR #2731, opened 2016-10-18
Handle NaN and Infinity in number encoding

arv:
We were hitting iloops for these non finite numbers

### DIFF
--- a/go/types/value_encoder.go
+++ b/go/types/value_encoder.go
@@ -6,6 +6,7 @@ package types
 
 import (
 	"fmt"
+	"math"
 
 	"github.com/attic-labs/noms/go/d"
 )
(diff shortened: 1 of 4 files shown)


go/types/value_encoder.go
```diff
@@ -129,7 +130,12 @@ func (w *valueEncoder) writeValue(v Value) {
 	case BoolKind:
 		w.writeBool(bool(v.(Bool)))
 	case NumberKind:
-		w.writeNumber(v.(Number))
+		n := v.(Number)
+		f := float64(n)
+		if math.IsNaN(f) || math.IsInf(f, 0) {
```
mikegray: What about negative zero? Okay to just silently continue and probably encode as zero?

mikegray:
lgtm

go/types/value_encoder.go
arv: We want to treat `-0` as `0`. But I don't see an actual test for that. Like you said `0 === -0` (in both Go and JS) so the assertRoundTrips test is not testing that.

PR #2734, opened 2016-10-19
Flow header after copyright

arv:
This also fixes a few invalid/missing headers.

I've split this into logical commits for ease of review.

### DIFF
--- a/tools/licensify.py
+++ b/tools/licensify.py
@@ -82,7 +82,11 @@ def buildLicensePattern(ext):
     # Allow flow annotations
     flow = r'(?P<flow>// @flow\n+|)'
 
-    return '^' + shebang + '(' + prefix + head + body + suffix + r'\n)?' + flow
+    # Doctype header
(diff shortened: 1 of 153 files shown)


tools/licensify.py
```diff
@@ -16,83 +16,89 @@
...
-      pattern = buildLicensePattern(ext)
-      if pattern != None:
-        with open(n, 'r+') as f:
-          processFile(f, ext, pattern)
-
-
-# Updates the license block in file |f|.
+    files = subprocess.check_output(['git', 'ls-files']).split('\n')
+    for n in files:
+        if n != '' and n.find('vendor/') == -1:
```
kalman: `'/vendor/'`?

arv: I didn't change this but I agree. I might also add a check for `.min.js` since the script now adds a header to js/perf/viewer/modules/Chart.bundle.min.js

PR #2735, opened 2016-10-19
Followup to licensify.py

arv:
Use startwith and endswith

### DIFF
--- a/tools/licensify.py
+++ b/tools/licensify.py
@@ -34,7 +34,7 @@ def main():
     files = subprocess.check_output(['git', 'ls-files']).split('\n')
     for n in files:
         if n != '' and not n.startswith('vendor/') and (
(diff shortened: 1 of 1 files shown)


arv:
PTAL

tools/licensify.py
```diff
@@ -33,7 +33,8 @@
 def main():
     files = subprocess.check_output(['git', 'ls-files']).split('\n')
     for n in files:
-        if n != '' and 'vendor/' not in n and '.min.js' not in n:
+        if n != '' and not n.startswith('vendor/') and (
+            not n.endswith('.min.js')):
```
kalman: is this really the indentation that pylint wants? sseems like it should be +4

arv: pylint is a happy camper... I'm not. I'll fix

PR #2749, opened 2016-10-24
Remove commit de-duping

### LINKED ISSUE #2720: `db.Commit()` ignores duplicate commits
cc @rafael-atticlabs 

Early on in Noms we made a decision that if you do something like:

```
let ds = await db.getDataset('foo');

// note: these both are extending the same parent
await db.Commit(42, {parents: [ds]})
await db.Commit(42, {parents: [ds]})
```

... that the second commit would be silently deduplicated. This is essentially a very basic merge heuristic: two different callers committed the same value against the same parent, so it frequently makes sense to merge them.

We now have a much more fully-featured merge function. It would handle this case. So I think it makes sense to just remove this feature.

cmasone-attic:
Now that we're going to put in real commit-merge logic, we
can take out the very-old hack that deduplicated identical
commits.

Fixes #2720

### DIFF
--- a/go/datas/database_test.go
+++ b/go/datas/database_test.go
@@ -191,6 +191,20 @@ func (suite *DatabaseSuite) TestDatabaseCommit() {
 	newDB.Close()
 }
 
+func (suite *DatabaseSuite) TestDatabaseDuplicateCommit() {
+	datasetID := "ds1"
+	ds := suite.db.GetDataset(datasetID)
+	datasets := suite.db.Datasets()
+	suite.Zero(datasets.Len())
+
+	v := types.String("Hello")
+	_, err := suite.db.CommitValue(ds, v)
+	suite.NoError(err)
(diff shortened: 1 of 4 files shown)


go/datas/database_test.go
```diff
@@ -191,6 +191,20 @@ func (suite *DatabaseSuite) TestDatabaseCommit() {
...
+	datasets := suite.db.Datasets()
+	suite.Zero(datasets.Len())
+
+	v := types.String("Hello")
+	_, err := suite.db.CommitValue(ds, v)
+	suite.NoError(err)
+
+	_, err = suite.db.CommitValue(ds, v)
+	suite.Error(err)
+}
```
aboodman: maybe just double-check that they're both in there after this?

aboodman:
Nevermind previous comment. LGTM.

PR #2750, opened 2016-10-24
Adds face merge functionality to photos

zcstarr:
it takes face center points and face rectangles and returns the
set of faces in which the face rectangle contains the face
center point. We return a photo object with a set of faces, that
have the names from the face centerpoints and the rectangles from the
face rectangles.

### DIFF
--- a/samples/go/face-merge/main.go
+++ b/samples/go/face-merge/main.go
@@ -68,8 +68,20 @@ func mergeFaces() {
 		return
 	}
 
+	sizeType := types.MakeStructTypeFromFields("", types.FieldMap{
+		"width":  types.NumberType,
+		"height": types.NumberType,
+	})
+
+	dateType := types.MakeStructTypeFromFields("Date", types.FieldMap{
+		"nsSinceEpoch": types.NumberType,
+	})
+
 	photoType := types.MakeStructTypeFromFields("Photo", types.FieldMap{
-		"title": types.StringType,
+		"sizes":         types.MakeMapType(sizeType, types.StringType),
+		"title":         types.StringType,
+		"datePublished": dateType,
+		"dateUpdated":   dateType,
 	})
 
 	annotatedPhotoSet := types.NewGraphBuilder(db, types.SetKind, true)
--- a/samples/go/face-merge/main.go
+++ b/samples/go/face-merge/main.go
@@ -147,7 +147,7 @@ func centeredFaceAligned(facePts facePoints) bool {
 }
 
 func usage() {
-	fmt.Fprintf(os.Stderr, "face-merge merges face data to create comprehensive face view\n\n")
+	fmt.Fprintf(os.Stderr, "face-merge merges face photo data to create comprehensive face view\n\n")
 	fmt.Fprintf(os.Stderr, "Usage: %s -db=<db-spec> -out-ds=<name> [input-paths...]\n\n", path.Base(os.Args[0]))
 	fmt.Fprintf(os.Stderr, "  <db>             : Database to work with\n")
 	fmt.Fprintf(os.Stderr, "  <out-ds>         : Dataset to write to\n")
--- a/samples/go/face-merge/main.go
+++ b/samples/go/face-merge/main.go
@@ -105,9 +105,11 @@ func mergeFaces() {
 	for _, v := range inputs {
 		walk.WalkValues(v, db, func(cv types.Value) bool {
 			if types.IsSubtype(photoType, cv.Type()) {
-				photo := cv.(types.Struct).Set("faces", types.NewSet())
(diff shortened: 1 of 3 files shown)


zcstarr:
@aboodman  PTAL

samples/go/face-merge/main.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func main() {
+	mergeFaces()
+}
+
+func mergeFaces() {
+
+	var dbStr = flag.String("db", "", "input database spec")
+	var outDSStr = flag.String("out-ds", "", "output dataset to write to - if empty, defaults to input dataset")
```
aboodman: The description is incorrect - this parameter is not optional.

samples/go/face-merge/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+	flag.Usage = usage
+	flag.Parse(false)
+
+	if flag.NArg() == 0 {
+		flag.Usage()
+		return
+	}
+
+	cfg := config.NewResolver()
```
aboodman: Nit: as `cfg` is not used anywhere else, it can be inline.

samples/go/face-merge/main.go
```diff
@@ -0,0 +1,169 @@
...
+	db, err := cfg.GetDatabase(*dbStr)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Invalid input database '%s': %s\n", flag.Arg(0), err)
+		return
+	}
+	defer db.Close()
+
+	var outDS datas.Dataset
```
aboodman: Nit: you can omit this line by changing line 64 to `outDS := ...`

samples/go/face-merge/main.go
```diff
@@ -0,0 +1,169 @@
...
+	return filteredPhoto
+}
+
+func getMergedFaces(photo types.Struct) types.Set {
+	facesRect, rectExists := photo.MaybeGet("facesRect")
+	facesCentered, centerExists := photo.MaybeGet("facesCentered")
+	faceSet := types.NewSet()
+
+	if rectExists && centerExists {
+		facesCentered.(types.Set).Iter(func(faceCentered types.Value) bool {
```
aboodman: Right here there is no guarantee that `facesCentered` is `types.Set`. In practice, this will work with our data, but it would be more self-documenting to actually do a type-check. The pattern I have been doing is like:

```
withFaces := types.NewStructWithFields("Photo", types.FieldData{
  "faces": ...
})
...
if (types.IsSubtype(photo.Type, withFaces) {
  ...
}
```

samples/go/face-merge/main.go
```diff
@@ -89,10 +105,11 @@ func mergeFaces() {
 	for _, v := range inputs {
 		walk.WalkValues(v, db, func(cv types.Value) bool {
 			if types.IsSubtype(photoType, cv.Type()) {
-				s := cv.(types.Struct)
-				faces := getMergedFaces(s)
-				photo := filterExtraFields(s)
-				photo = photo.Set("faces", faces)
+				photo := cv.(types.Struct).Set("faces", types.NewSet())
```
aboodman: I mean the `Set` call specifically.

samples/go/face-merge/main.go
```diff
@@ -117,29 +134,27 @@ func filterExtraFields(photo types.Struct) types.Struct {
...
+			facePts := getFacePoints(faceRect.(types.Struct), faceCentered.(types.Struct))
+			if centeredFaceAligned(facePts) {
+				faceSet = faceSet.Insert(types.NewStruct("", types.StructData{
+					"x":    facePts.x,
+					"y":    facePts.y,
+					"w":    facePts.w,
+					"h":    facePts.h,
+					"name": facePts.name,
+				}))
+				return true
```
aboodman: For this signature we usually name the return value, like this, so that it's clear what the meaning is while reading:

```
func (faceRect types.Value) (stop bool) {

}
```

In that case, at this line you'd do:

`stop = true`

and the only return would be at line 154, and it'd just be:

`return`

samples/go/face-merge/main.go
```diff
@@ -34,9 +34,8 @@ func main() {
 }
 
 func mergeFaces() {
-
 	var dbStr = flag.String("db", "", "input database spec")
-	var outDSStr = flag.String("out-ds", "", "output dataset to write to - if empty, defaults to input dataset")
+	var outDSStr = flag.String("out-ds", "", "output dataset to write to")
```
aboodman: In usage, please expand the commentary to describe why we're doing this:

```
face-merge merges the face rectangles detected by <program> with the labeled center points from Facebook to produce a rectangle that is labeled, which is what other jobs need.
```

PR #2751, opened 2016-10-25
Update eslint to 3.8.1, noms was broken at 3.8.1

kalman:
Also remove roll.py (not used anymore, we use file path) and make it
pinned to a specific version (so that this doesn't happen again.


js/noms/src/json-convert.js
```diff
@@ -13,7 +13,7 @@ import {
 import type Value from './value.js';
 
 type JSON = string | number | boolean | null | JSONObject | JSONArray;
-type JSONObject = { [key:string]: JSON };
+type JSONObject = { [key: string]: JSON };
```
arv: Ha! This still does not catch the lint error.

kalman: wat?

kalman: By "this" do you mean 3.7.1? Not sure what your comment means exactly :-) anyhow, submitting.

arv:
I meant that there should be no space after {.

Let me see if I can find a rule for that.

PR #2753, opened 2016-10-25
Better codecov usage

mikegray:


codecov.yml
```diff
@@ -7,15 +7,15 @@ codecov:
 coverage:
   precision: 2         # how many decimal places to display in the UI: 0 <= value <= 4
   round: down          # how coverage is rounded: down/up/nearest
-  range: 60...100      # custom range of coverage colors from red -> yellow -> green
+  range: 70...100      # custom range of coverage colors from red -> yellow -> green
```
mikegray: We are at 88% and have been since we started, so updating the range to more closely align with our current state.

codecov.yml
```diff
@@ -7,15 +7,15 @@ codecov:
...
 
   notify:
     slack:
       default:
         url: "secret:n+BYhIXTXsaCiMKB3vOf6yP68ytdKd3WpXtJFWPEUsEWXDiGnU5dTB5DO2yv8tR0COdxvs7K31hVpEfHEXdoXOaQhUw3FKf3fh8KZDLN7CGTbeDhw1uNGGyBr2d2TWnopzYtcXomdwMmuckARtiWQx0YXJiZY9YyCrIoDK9HIJQ="
         branches: null
-        threshold: 1.15
-        attachments: "sunburst, diff"
+        threshold: 5.0
```
mikegray: From my reading of the Codecov docs, it appears threshold is a percentage - so this is meant to only notify Slack channel if the code coverage changes by +/-5%.  Removing sunburst because while it looks pretty it is not very useful.

codecov.yml
```diff
@@ -47,6 +47,6 @@ coverage:
         if_ci_failed: error
 
 comment:
-  layout: "header, diff, changes, sunburst, suggestions, tree"
```
mikegray: Limiting the amount of information being posted into PR comments - just the changes in the files being added/removed/modified are interesting for these.  If files aren't being touched in this PR then the percentage change is mostly meaningless.

codecov.yml
```diff
@@ -47,6 +47,6 @@ coverage:
...
+  layout: "tree"
   branches: null
-  behavior: once
+  behavior: default
```
mikegray: This behavior modification is meant to add a new comment to the PR when a new commit is added rather than updating the old comment.  This will be more useful in seeing the history - especially if the later commits are to fix the coverage.

codecov.yml
arv: Can we do 80% then?

codecov.yml
arv: I agree. Sunburst be gone!

codecov.yml
mikegray: We could, but I didn't want to put us into the "red".

arv: Didn't you say we are at 88%? 88 > 80

PR #2757, opened 2016-10-25
Remove calls to d.Chk.True. Fixes #2741

### LINKED ISSUE #2741: Remove usages of d.Chk.True/False which box strings to interfaces in graph_builder

willhite:


rafael-atticlabs:
This doesn't really fix the problem because PanicIfTrue also takes a var args of interface{}, so it'll box the string param (and all following)

Can we change panic if true to be

`func PanicIfTrue(cond bool, message string)`

so as to make the point that you shouldn't be doing work in an assertion?

willhite:
Yep. Even as I was making that change, I didn't understand the benefit of it either. Now I see what you are asking. Also, I woke up last night realizing that d.Chk.True() and d.PanicIfTrue() are opposites. I'll fix these and send it to you again.

willhite:
PTAL @rafael-atticlabs

willhite:
@rafael-atticlabs btw, I ran go fmt on the entire source tree. So you may see a couple of files with just formatting changes.

rafael-atticlabs:
nice. lgtm. watch the benchmark bots. i'll bet some benchmark somewhere will tick up slightly.

PR #2758, opened 2016-10-25
Use a newer version of @attic/eslintrc

arv:


arv:
Still testing locally...

Issue #2534, opened 2016-09-08
Add merge options to Dataset.CommitOptions


cmasone-attic:
In code, we want callers to be able to specify optional merge behavior when they commit a new Head for a Dataset -- including disallowing any merging. They should be able to provide a callback for custom conflict resolution, and several defaults will be provided: none, 'left', and 'right'.

This should be implemented in Go and JS

PR #2759, opened 2016-10-26
Go: add heuristic merge option to db.Commit()

cmasone-attic:
This patch adds an optional MergePolicy field to CommitOptions. It's a
callback. If the caller sets it, then the commit code will look for a
common ancestor between the Dataset HEAD and the provided Commit. If
the caller-provided Commit descends from HEAD, then Commit proceeds as
normal.
If it does not, but there is a common ancestor, the code runs
merge.ThreeWay() on the values of the provided Commit, HEAD, and the
common ancestor, invoking the MergePolicy callback to resolve
conflicts. If merge succeeds, a merge Commit is created that descends
from both HEAD and the caller-provided Commit. This becomes the new
HEAD of the Dataset.

Towards #2534

### DIFF
--- a/go/datas/commit_options.go
+++ b/go/datas/commit_options.go
@@ -11,12 +11,17 @@ import (
 
 // CommitOptions is used to pass options into Commit.
 type CommitOptions struct {
-	// Parents, if provided is the parent commits of the commit we are creating.
+	// Parents, if provided is the parent commits of the commit we are
+	// creating.
 	Parents types.Set
 
-	// Meta is a Struct that describes arbitrary metadata about this Commit, e.g. a timestamp or descriptive text.
+	// Meta is a Struct that describes arbitrary metadata about this Commit,
+	// e.g. a timestamp or descriptive text.
 	Meta types.Struct
 
-	// Policy will be called to attempt to merge this Commit with the current Head, if this is not a fast-forward. If Policy is nil, no merging will be attempted.
+	// Policy will be called to attempt to merge this Commit with the current
+	// Head, if this is not a fast-forward. If Policy is nil, no merging will
+	// be attempted. Note that because Commit() retries in some cases, Policy
+	// might also be called multiple times with different values.
 	Policy merge.Policy
 }
--- a/go/datas/commit_options.go
+++ b/go/datas/commit_options.go
@@ -4,11 +4,19 @@
 
 package datas
 
-import "github.com/attic-labs/noms/go/types"
+import (
+	"github.com/attic-labs/noms/go/merge"
+	"github.com/attic-labs/noms/go/types"
+)
 
 // CommitOptions is used to pass options into Commit.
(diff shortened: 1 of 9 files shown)


cmasone-attic:
PTAL

aboodman:
Can we change it like:

```
type CommitOptions struct {
  ...
  Policy: MergePolicy
}

type MergePolicy interface {
  func Merge(a, b, parent types.Value, vrw types.ValueReadWriter, progress chan struct{})
}

func NewThreeway(resolve ResolveFunc) MergePolicy {
  ...
}
```

aboodman:
That way one could actually do the counter resolution correctly, and all the retry logic would remain inside Commit(). WDYT?

cmasone-attic:
I don't see a reason for `MergePolicy` to be an interface, really. I think it can also be a function type

```go
package merge

type Policy func(a, b, parent types.Value, vrw types.ValueReadWriter, progress chan struct{})

func NewThreeWay(resolve ResolveFunc) Policy {
	return func(a, b, parent types.Value, vrw types.ValueReadWriter, progress chan struct{}) {
		ThreeWay(a, b, parent, vrw, resolve, progress)
	}
}
```
```go
package datas

type CommitOptions struct {
...
	// Policy will be called to attempt to merge this Commit with the current Head, if this is not a fast-forward. If Policy is nil, no merging will be attempted.
	Policy merge.Policy
}
```

aboodman:
Good point. LGTM. Thanks.

cmasone-attic:
Ok, PTAL

go/datas/commit_options.go
```diff
@@ -4,11 +4,19 @@
...
 // CommitOptions is used to pass options into Commit.
 type CommitOptions struct {
 	// Parents, if provided is the parent commits of the commit we are creating.
 	Parents types.Set
-	Meta    types.Struct
+
+	// Meta is a Struct that describes arbitrary metadata about this Commit, e.g. a timestamp or descriptive text.
+	Meta types.Struct
+
+	// Policy will be called to attempt to merge this Commit with the current Head, if this is not a fast-forward. If Policy is nil, no merging will be attempted.
```
aboodman:
```
.append("Note that because Commit() retries in some cases, Policy might also be called multiple times with different values.\n")
```

go/merge/three_way.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/attic-labs/noms/go/types"
 )
 
+// Policy functors are used to merge two values (a and b) against a common
+// ancestor. All three Values and their must by wholly readable from vrw.
+// Whenever a change is merged, implementations should send a struct{} over
+// progress.
+type Policy func(a, b, ancestor types.Value, vrw types.ValueReadWriter, progress chan struct{}) (merged types.Value, err error)
```
aboodman: Can this be defined in `datas`?

PR #2766, opened 2016-10-27
Make our eslint setup less hacky

arv:
Rename the package to eslint-config-noms as documented on the eslint
site.

This allows other npm modules to use and `extends: 'noms'` in their
eslint rc file.


arv:
I will do `npm deprecate @attic/eslintrc "Please use eslint-config-noms instead"` once this has landed and been pushed to npm.

WDYT?

kalman:
I like this, lgtm.

PR #2769, opened 2016-10-27
Make pin return unchanged spec if it's already pinned

willhite:
The current implementation of pin means that programs like find-photos can't take a pinned path as it's <in-path> argument. This should re-enable that functionality.

Do I need to bump up the noms version for this change?


aboodman:
whoops. I think you bump the 'patch' (third component) for something like this

aboodman:
LGTM w/ nit

PR #2771, opened 2016-10-27
Add vendor/golang.org/x/crypto

kalman:


kalman:
TBR

aboodman:
Next time try to only vendor what you need directly ... we don't use most
of these crypto subpackages.

aboodman:
It's only 3MB on disk so not a huge deal.

kalman:
Hm, I just used roll.py which didn't seem to work with subdirectories.

aboodman:
It doesn't, you have to do it manually.

kalman:
Got it. Next time I do this I might add a feature to roll.py to support subdirectories.

PR #2772, opened 2016-10-27
Implement private databases for the demo server

kalman:
Private databases begin with "/p/" - for example, "/kalman" is not
private, but "/p/kalman" is private. They are not the same database.

The bulk of this work is the receipt infrastructure.

A receipt is form data that gives access to a database, encrypted using
secretbox. For example, "Database=/p/kalman&Date=12345678" might encrypt
to "SFH5bcIJ3_XgEbtmi_AdCKTItW20fl90czVl5_pF5PAXhNQ366U1yOpYGAjT".
- A new tool receiptkey generates random receipt (secretbox) keys.
- A new tool receipttool generates receipts for databases.
- demo-server has been updated to check for a receipt in the
  Authorization header to access private databases.

receipttool and demo-server must be given the same receipt key.

### DIFF
--- a/go/util/receipts/receipts.go
+++ b/go/util/receipts/receipts.go
@@ -6,34 +6,30 @@ package receipts
 
 import (
 	"crypto/rand"
+	"crypto/sha512"
 	"encoding/base64"
 	"fmt"
 	"net/url"
 	"time"
 
 	"github.com/attic-labs/noms/go/d"
-	"github.com/attic-labs/noms/go/hash"
 	"golang.org/x/crypto/nacl/secretbox"
 )
 
 // Data stores parsed receipt data.
 type Data struct {
-	Database string
-	Date     time.Time
+	Database  string
+	IssueDate time.Time
 }
 
-// KeySize is the size in bytes of receipt keys.
-const KeySize = 32 // secretbox
+// keySize is the size in bytes of receipt keys.
+const keySize = 32 // secretbox uses 32-byte keys
 
 // Key is used to encrypt receipt data.
-type Key [KeySize]byte
+type Key [keySize]byte
 
-// Don't use a nonce when sealing/opening secretbox because these receipts need
-// to be used across sessions.
-var emptyNonce [24]byte
-
-// Force all receipts to have the same size, to avoid size attacks.
-const receiptSize = 256
+// nonceSize is the size in bytes that secretbox uses for nonces.
+const nonceSize = 24
 
 // DecodeKey converts a base64 encoded string to a receipt key.
 func DecodeKey(s string) (key Key, err error) {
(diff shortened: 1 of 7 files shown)


aboodman:
@rafael-atticlabs should also review. I'll take a first pass.

go/util/receipts/receipts.go
```diff
@@ -0,0 +1,130 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
```
aboodman: I'm going to review it for now here though.

go/util/receipts/receipts.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/hash"
+	"golang.org/x/crypto/nacl/secretbox"
+)
+
+// Data stores parsed receipt data.
+type Data struct {
+	Database string
+	Date     time.Time
```
aboodman: `IssueDate` maybe to make it more self-documenting?

go/util/receipts/receipts.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// KeySize is the size in bytes of receipt keys.
+const KeySize = 32 // secretbox
```
aboodman: Can you expand more on the inline comment -- why is this 32 exactly?

go/util/receipts/receipts.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Key is used to encrypt receipt data.
+type Key [KeySize]byte
+
+// Don't use a nonce when sealing/opening secretbox because these receipts need
+// to be used across sessions.
+var emptyNonce [24]byte
+
+// Force all receipts to have the same size, to avoid size attacks.
```
aboodman: I'm not familiar with what you mean here by 'size attack'

go/util/receipts/receipts.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	data.Date = date
+	return nomsHash(data.Database) == database, nil
+}
+
+func nomsHash(s string) string {
+	return hash.FromData([]byte(s)).String()
```
aboodman: One side-effect of this is that if the hash function used in Noms changes, this will change. I guess we'd notice, so not that big a deal. But consider just depending on the exact hash function you want specifically.

go/util/receipts/receipts.go
```diff
@@ -0,0 +1,130 @@
...
+	copy(key[:], keySlice)
+	return
+}
+
+// Generate returns a receipt for Data, which is an encrypted query string
+// encoded as base64.
+func Generate(key Key, data Data) (string, error) {
+	d.PanicIfTrue(data.Database == "" || data.Date == (time.Time{}))
+
+	receiptPlainOrig := []byte(url.Values{
```
aboodman: Why use querystring as the encoding? it seems somewhat wasteful to repeat the key names in every receipt. Why not using more like a straight binary encoding ... assuming that first thing is hash of db name, second thing date, etc?

go/util/receipts/receipts.go
```diff
@@ -0,0 +1,130 @@
...
+
+// KeySize is the size in bytes of receipt keys.
+const KeySize = 32 // secretbox
+
+// Key is used to encrypt receipt data.
+type Key [KeySize]byte
+
+// Don't use a nonce when sealing/opening secretbox because these receipts need
+// to be used across sessions.
+var emptyNonce [24]byte
```
aboodman: This is surprising. I would have expected us to use random nonces... and to do that instead of the salt that you have. Were you following some specific instructions to do it this way?

go/util/receipts/receipts.go
aboodman: ok

go/util/receipts/receipts.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	data.IssueDate = date
+	return hash(data.Database) == database, nil
+}
+
+func hash(s string) string {
+	h := sha512.Sum512([]byte(s))
```
aboodman: the length of the sha512 digest is 64 bytes - probably far longer than most of the inputs. We would be fine with the 224 variant: https://golang.org/pkg/crypto/sha512/#Sum512_224.

samples/go/demo-server/main.go
```diff
@@ -6,32 +6,43 @@ package main
...
-	flag.PrintDefaults()
-}
-
 func main() {
 	chunks.RegisterLevelDBFlags(flag.CommandLine)
 	dynFlags := chunks.DynamoFlags("")
 
-	flag.Usage = usage
+	flag.Usage = func() {
+		fmt.Println("Usage: demo-server --authkey <authkey> [options]")
```
aboodman: Why does the `--authkey` flag get special treatment

kalman: Because it's required apparently.

go/util/receipts/receipts.go
```diff
@@ -6,34 +6,30 @@ package receipts
...
 	"github.com/attic-labs/noms/go/d"
-	"github.com/attic-labs/noms/go/hash"
 	"golang.org/x/crypto/nacl/secretbox"
 )
 
 // Data stores parsed receipt data.
 type Data struct {
-	Database string
-	Date     time.Time
+	Database  string
```
kalman: Are you saying to add a `User` field?

I can imagine doing that, but it's not immediately useful. That's why the query string thing is nice because I can do it later.

kalman:
PTAL

aboodman:
LGTM

PR #2773, opened 2016-10-27
Mark non standalone packages as private

arv:
This prevents us from doing npm publish on them.

https://docs.npmjs.com/files/package.json#private


arv:
Motivation: To make publish-js-sdk.py npm publish all non private packages.

arv:
I was planning to do `git ls-files | grep package.json | ...`

We should have used `private: true` from the beginning.

arv:
Another option is to process all `package.json` in `/js/`

PR #2776, opened 2016-10-28
Fix babel-preset-noms

arv:
It wasn't working on the code coverage tests. Now I restored this
to what it was before these changes (minus removing some features
that chrome and node now supports out of the box)

### DIFF
--- a/js/babel-preset-noms/index.js
+++ b/js/babel-preset-noms/index.js
@@ -12,6 +12,18 @@ function plugin(n) {
   return require('babel-plugin-' + n);
 }
 
+const commonPlugins = [
+  plugin('syntax-async-functions'),
+  plugin('syntax-flow'),
+  plugin('transform-class-properties'),
+  [
+    plugin('transform-runtime'), {
+      polyfill: false,
+      regenerator: true,
+    },
+  ],
+];
+
 const production = {
   presets: [
     preset('es2015'),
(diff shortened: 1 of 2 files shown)


js/babel-preset-noms/package.json
```diff
@@ -15,9 +15,7 @@
     "babel-plugin-syntax-async-functions": "^6.13.0",
     "babel-plugin-transform-async-to-generator": "^6.8.0",
     "babel-plugin-transform-class-properties": "^6.11.5",
-    "babel-plugin-transform-es2015-destructuring": "^6.9.0",
```
arv: These two are no longer needed in dev mode.

js/babel-preset-noms/index.js
```diff
@@ -39,8 +38,17 @@ const development = {
     preset('react'),
   ],
   plugins: [
+    plugin('syntax-async-functions'),
+    plugin('syntax-flow'),
+    plugin('transform-async-to-generator'),
+    plugin('transform-class-properties'),
     plugin('transform-es2015-modules-commonjs'),
```
kalman: Was the problem the spread? Because... you could also just make this:
```
plugins: [
  plugin('transform-es2015-modules-commonjs'),
].concat(commonPlugins)
```

arv: The ordering matters. I can play around with it a bit more to see if I can unify the ordering a bit more and if that works, extract the shared plugins to its own array.

PR #2778, opened 2016-10-28
Add resources field to find-photo's Photo object

willhite:


samples/js/fb/find-photos/src/main.js
```diff
@@ -143,6 +144,16 @@ async function getSizes(input): Promise<Map<Struct, string>> {
   return result;
 }
 
+async function getResources(input): Promise<Map<Struct, Struct>> {
+  let result = Promise.resolve(new Map());
+  await input.images.forEach(v => {
+    result = result.then(m => m.set(
+      newStruct('', {width: v.width, height: v.height}),
+      v.source));
```
aboodman: Isn't `v.source` a string? The signature says that this method returns `Map<Struct, Struct>`.

samples/js/picasa/find-photos/src/main.js
```diff
@@ -204,6 +205,38 @@ async function getSizes(input: Object, origWidth: number, origHeight: number)
   return new Map(tuples);
 }
 
+async function getResources(input: Object, origWidth: number, origHeight: number)
+: Promise<Map<Struct, Struct>> {
```
aboodman: Indent 4.

samples/js/picasa/find-photos/src/main.js
```diff
@@ -204,6 +205,38 @@ async function getSizes(input: Object, origWidth: number, origHeight: number)
...
+  }).filter(t => t);
+
+  // The original file.
+  tuples.push([
+    newStruct('', {width: origWidth, height: origHeight}),
+    newStruct('RemoteResource', {url: makeURL('d')}),
+  ]);
+
+  return new Map(tuples);
+}
```
aboodman: I think since this is so similar to getURLs it would be worth sharing the code ..

samples/js/picasa/find-photos/src/main.js
```diff
@@ -204,6 +205,38 @@ async function getSizes(input: Object, origWidth: number, origHeight: number)
...
+
 async function getFaces(photo: Object, origWidth: number, origHeight: number)
     : Promise<?Set<Struct>> {
   const parsePos = (s) => {
```
aboodman: There's also Dropbox. It's fine to do it incrementally though, as we discussed.

samples/js/fb/find-photos/src/main.js
willhite: Oops this got lost in a merge. Fixed now.

samples/js/picasa/find-photos/src/main.js
willhite: Done

samples/js/picasa/find-photos/src/main.js
willhite: Do you mean getSizes(). If so, I thought about it and given that the return types in this code are different, it would be kind of hard to factor most of it out without resorting to Map<Struct, Any>. This might be worth it but my thinking was that getSizes was going to go away once the UI started using resources rather than sizes. Let me know if you disagree with that.

willhite:
Re-implemented getSizes() in terms of getResources() for dropbox and picasa. PTAL.

samples/js/dropbox/find-photos/src/main.js
```diff
@@ -126,7 +128,15 @@ async function main(): Promise<void> {
   });
 }
 
-function getSizes(input: Object): Map<Struct, string> {
+async function getSizes(resources: Map<Struct, Struct>): Promise<Map<Struct, string>> {
+  const tuples = [];
+  await resources.forEach((v: any, k: Struct) => {
+    tuples.push([k, v.url]);
+  });
+  return Promise.resolve(new Map(tuples));
```
aboodman: You don't have to manually wrap `new Map(tuples)` here. It is implied in an `async` function.

samples/js/fb/find-photos/src/main.js
```diff
@@ -143,6 +144,16 @@ async function getSizes(input): Promise<Map<Struct, string>> {
   return result;
 }
 
+async function getResources(input): Promise<Map<Struct, Struct>> {
+  let result = Promise.resolve(new Map());
+  await input.images.forEach(v => {
+    result = result.then(m => m.set(
```
aboodman: Since you've marked this function async, why not get rid of the promises inside the loop:

```
let m = new Map();
await input.images.forEach(v => {
  m = await m.set(...);
});
return m;
```

samples/js/picasa/find-photos/src/main.js
```diff
@@ -172,10 +174,19 @@ function getGeo(input): ?Struct {
...
 }
 
-async function getSizes(input: Object, origWidth: number, origHeight: number)
+async function getSizes(resources: Map<Struct, Struct>)
     : Promise<Map<Struct, string>> {
+  const tuples = [];
+  await resources.forEach((v: any, k: Struct) => {
+    tuples.push([k, v.url]);
+  });
+  return Promise.resolve(new Map(tuples));
```
aboodman: Same thing as above - when the function is `async`, the `Promise` wrapper is done automatically.

PR #2779, opened 2016-10-28
Add delete field to structs

arv:


arv:
@zcstarr I think you asked for this?

js/noms/src/struct.js
```diff
@@ -184,12 +191,42 @@ export class StructMirror<T: Struct> {
...
+    const type = removeFieldFromType(this.type, idx);
+    return newStructWithValues(type, values);
+  }
+}
+
+function removeFieldFromType(type: Type<StructDesc>, idx: number): Type<StructDesc> {
+  const {desc} = type;
+  const fieldMap = Object.create(null);
+  let i = 0;
+  desc.forEachField((n, t) => {
```
zcstarr: nit. should we type lambda functions here ?

zcstarr:
@arv lgtm. with one nit, not sure what the policy is, see comment.  BTY

js/noms/src/struct.js
arv: Not really. Only declare types on exported bindings and where it makes the code more readable. We hardly ever annotate local variables.

PR #2782, opened 2016-10-28
Add --subdir argument to roll.py

kalman:
If you want to roll just the go/ directory of noms, you can do:

$ roll.py https://github.com/attic-labs/noms --subdir go

Or, the nacl/secretbox/ directory of https://go.googlesource.com/crypto:

$ roll.py https://go.googlesource.com/crypto --subdir nacl/secretbox

### DIFF
--- a/tools/roll.py
+++ b/tools/roll.py
@@ -10,10 +10,16 @@ import argparse, os, os.path, subprocess, sys, shutil, urlparse
 def main():
   parser = argparse.ArgumentParser(description='Dependency snapshotter')
   parser.add_argument('url')
-  parser.add_argument('--path', help='path to store the dependency at, defaults to vendor/[url without protocol]')
-  parser.add_argument('--incl', help='subdirectory of the dependency to check out, defaults to root. Evaluated before --excl.')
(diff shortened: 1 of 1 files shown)


tools/roll.py
```diff
@@ -29,7 +35,7 @@ def main():
     path = path[1:]
 
   depdir = args.path
-  if depdir == None:
+  if depdir is None:
```
kalman: It's a good habit to do `is None` not `== None` because classes can override `__eq__` and not deal with `None`, in theory.

tools/roll.py
```diff
@@ -52,7 +58,6 @@ def main():
 
   shutil.rmtree('.git')
 
-  # TODO: Should we also do Godeps?
```
kalman: Presumably not.

kalman:
Hold up, I'm going to change this to be multiple `--include` and `--exclude` flags instead, since I want to be able to exclude directories as well.

kalman:
Ok, PTAL.

I didn't go with the multiple argument thing because it was too complicated, but this gets me a lot of the way to cherrypicking logical vendor stuff for the attic repo.

kalman:
We don't really need this if we don't end up vendoring noms into attic, so hold off.

zcstarr:
jenkins:test this

PR #2784, opened 2016-10-28
Add omitempty to Go marshal

arv:
If you provide an omitempty tag and the Go value is the zero/empty value
then we do not include that field in the resulting Noms struct.

Towards #2376


arv:
@willhite Does this do what you want?

Issue #2787, opened 2016-10-31
photos - dedupe


aboodman:
We should try @ehalpern's fancy new dedupe algorithm on the photos sample

aboodman:
cc @mikegray @arv

PR #2789, opened 2016-10-31
Add jobs for grouping similar photos in PhotoGroups

ehalpern:
Outline:
- The first photo-hash job adds a dhash field to each photo. The [dhash](http://www.hackerfactor.com/blog/?/archives/529-Kind-of-Like-That.html) is a 128 bit
  downsampled representation of the photo that works well for visual similarity comparisions.
- The second photo-dedup job groups photos that have similar dhash's into PhotoGroups.

toward: #2787

### DIFF
--- a/samples/go/photo-dedup/cmd/photo-dedup/main.go
+++ b/samples/go/photo-dedup/cmd/photo-dedup/main.go
@@ -20,18 +20,19 @@ import (
 )
 
 func usage() {
-	fmt.Fprintf(os.Stderr, "photo-dedup group photos in groups of duplicates\n\n")
-	fmt.Fprintf(os.Stderr, "Usage: %s -db=<db-spec> -out-ds=<name> [input-paths...]\n\n", path.Base(os.Args[0]))
+	fmt.Fprintf(os.Stderr, "Usage: %s -db=<db-spec> -out-ds=<name> <input-paths...>\n\n", path.Base(os.Args[0]))
+	fmt.Fprintf(os.Stderr, "Groups photos into PhotoGroups based on similarity determined by their dhashes\n\n")
 	fmt.Fprintf(os.Stderr, "  <db>             : Database to work with\n")
 	fmt.Fprintf(os.Stderr, "  <out-ds>         : Dataset to write photos groups to\n")
-	fmt.Fprintf(os.Stderr, "  [input-paths...] : One or more paths within <db-spec> to crawl\n\n")
+	fmt.Fprintf(os.Stderr, "  <input-paths...> : One or more input paths within <db-spec>\n\n")
 	fmt.Fprintln(os.Stderr, "Flags:\n")
 	flag.PrintDefaults()
 }
 
 func main() {
 	var dbStr = flag.String("db", "", "input database spec")
-	var outDSStr = flag.String("out-ds", "", "output dataset to write to - if empty, defaults to input dataset")
+	var outDSStr = flag.String("out-ds", "", "output dataset to write to")
+	var threshold = flag.Int("threshold", 10, "photo's whose dhash distance is < threshold are grouped together")
 	verbose.RegisterVerboseFlags(flag.CommandLine)
 
 	flag.Usage = usage
@@ -64,7 +65,7 @@ func main() {
 		return
 	}
 
-	err = job.DeduplicateJob(db, inputs, outDS)
+	err = job.DeduplicateJob(db, inputs, outDS, *threshold)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "%s\n", err)
 		exit.Fail()
--- a/samples/go/photo-dedup/cmd/photo-dedup/main.go
+++ b/samples/go/photo-dedup/cmd/photo-dedup/main.go
@@ -1,0 +1,73 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/attic-labs/noms/go/config"
+	"github.com/attic-labs/noms/go/datas"
+	"github.com/attic-labs/noms/go/spec"
+	"github.com/attic-labs/noms/go/util/exit"
+	"github.com/attic-labs/noms/go/util/verbose"
+	"github.com/attic-labs/noms/samples/go/photo-dedup/job"
+
+	flag "github.com/juju/gnuflag"
+)
+
+func usage() {
+	fmt.Fprintf(os.Stderr, "photo-dedup group photos in groups of duplicates\n\n")
+	fmt.Fprintf(os.Stderr, "Usage: %s -db=<db-spec> -out-ds=<name> [input-paths...]\n\n", path.Base(os.Args[0]))
+	fmt.Fprintf(os.Stderr, "  <db>             : Database to work with\n")
+	fmt.Fprintf(os.Stderr, "  <out-ds>         : Dataset to write photos groups to\n")
+	fmt.Fprintf(os.Stderr, "  [input-paths...] : One or more paths within <db-spec> to crawl\n\n")
+	fmt.Fprintln(os.Stderr, "Flags:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var dbStr = flag.String("db", "", "input database spec")
+	var outDSStr = flag.String("out-ds", "", "output dataset to write to - if empty, defaults to input dataset")
+	verbose.RegisterVerboseFlags(flag.CommandLine)
+
+	flag.Usage = usage
+	flag.Parse(false)
+
+	if flag.NArg() == 0 {
+		flag.Usage()
+		return
+	}
+
+	cfg := config.NewResolver()
+	db, err := cfg.GetDatabase(*dbStr)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Invalid input database '%s': %s\n", flag.Arg(0), err)
+		return
+	}
+	defer db.Close()
+
+	var outDS datas.Dataset
+	if !datas.IsValidDatasetName(*outDSStr) {
+		fmt.Fprintf(os.Stderr, "Invalid output dataset name: %s\n", *outDSStr)
+		return
+	} else {
+		outDS = db.GetDataset(*outDSStr)
+	}
+
+	inputs, err := spec.ReadAbsolutePaths(db, flag.Args()...)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		return
+	}
+
+	err = job.DeduplicateJob(db, inputs, outDS)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		exit.Fail()
+	}
+}
+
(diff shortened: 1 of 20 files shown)


samples/go/photo-dedup/cmd/photo-dedup/main.go
```diff
@@ -0,0 +1,73 @@
...
+	fmt.Fprintf(os.Stderr, "  <db>             : Database to work with\n")
+	fmt.Fprintf(os.Stderr, "  <out-ds>         : Dataset to write photos groups to\n")
+	fmt.Fprintf(os.Stderr, "  [input-paths...] : One or more paths within <db-spec> to crawl\n\n")
+	fmt.Fprintln(os.Stderr, "Flags:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var dbStr = flag.String("db", "", "input database spec")
+	var outDSStr = flag.String("out-ds", "", "output dataset to write to - if empty, defaults to input dataset")
```
aboodman: There is no input dataset according to `usage`.

samples/go/photo-dedup/cmd/photo-dedup/main.go
```diff
@@ -0,0 +1,73 @@
...
+	cfg := config.NewResolver()
+	db, err := cfg.GetDatabase(*dbStr)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Invalid input database '%s': %s\n", flag.Arg(0), err)
+		return
+	}
+	defer db.Close()
+
+	var outDS datas.Dataset
+	if !datas.IsValidDatasetName(*outDSStr) {
```
aboodman: Looks like the output dataset is required.

samples/go/photo-dedup/cmd/photo-hash/main.go
```diff
@@ -0,0 +1,72 @@
...
+	fmt.Fprintf(os.Stderr, "  <db>             : Database to work with\n")
+	fmt.Fprintf(os.Stderr, "  <out-ds>         : Dataset to write annotated photos to\n")
+	fmt.Fprintf(os.Stderr, "  [input-paths...] : One or more paths within <db-spec> to crawl\n\n")
+	fmt.Fprintln(os.Stderr, "Flags:\n")
+	flag.PrintDefaults()
+}
+
+func main() {
+	var dbStr = flag.String("db", "", "input database spec")
+	var outDSStr = flag.String("out-ds", "", "output dataset to write to - if empty, defaults to input dataset")
```
aboodman: Same comments about flags as elsewhere.

samples/go/photo-dedup/job/dedupjob.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package job
```
aboodman: Whatever code is being shared between these two jobs seems like it would be just as useful to other jobs, and should be outside this directory.

samples/go/photo-dedup/job/dedupjob.go
```diff
@@ -0,0 +1,60 @@
...
+	grouper = newPhotoGrouper()
+	for _, set := range photoSets {
+		walk.WalkValues(set, db, func(cv types.Value) (stop bool) {
+			if photo, ok := model.UnmarshalPhoto(cv); ok {
+				grouper.insertPhoto(photo)
+			}
+			return false
+		})
+	}
+	photoGroupsCh := make(chan types.Struct)
```
aboodman: This is an unbuffered channel, probably not what you want.

samples/go/photo-dedup/job/dedupjob.go
```diff
@@ -0,0 +1,60 @@
...
+	go func() {
```
aboodman: I'm curious why the walk above is blocking, but this part here is async... why not have the entire contents of this function inside the goroutine?

samples/go/photo-dedup/job/grouper.go
```diff
@@ -0,0 +1,63 @@
...
+// duplicate photos. Otherwise it creates a new group.
+//
+// The current implementation is a brute force n^2 comparision. A more efficient
+// implementation would be to build an VP/MVP tree. A VP tree is a binary search
+// tree that works in a geometric space. Each node defines a center point and a
+// radius. Dhashes within the radius can be found to the left; those outside the
+// radius can be found to the right. An MVP is the k-tree equivalent.
+func (g *photoGrouper) insertPhoto(photo model.Photo) {
+	//
+	const similarityThreshold = 10
```
aboodman: Maybe pull this out into a flag. Seems like something we'd want to tinker with.

samples/go/photo-dedup/job/hashjob.go
```diff
@@ -0,0 +1,143 @@
...
+	var wg sync.WaitGroup
+	wg.Add(numWorkers)
+	for i := 0; i < numWorkers; i++ {
+		go func() {
+			defer wg.Done()
+			for photo := range photosToHashCh {
+				hashed, err := addHashToPhoto(photo)
+				if err != nil {
+					fmt.Fprintf(os.Stderr, "\nSkipping: %s\n", err)
+					hashedPhotosCh <- photo.Marshal()
```
aboodman: is there a reason to keep the failures?

samples/go/photo-dedup/job/hashjob.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, err
+	}
+	photo.SetDhash(dhash.New(img))
+	return photo, nil
+}
+
+// pickBestImage returns the image URL corresponding to the size
+// closest to but not below the ideal size. If there is no image
+// large enough, it returns the next smallest image.
+func pickBestImage(photo model.Photo) string {
```
aboodman: This should be factored out somewhere ... I feel like we have it, or something like it, in several places.

samples/go/photo-dedup/model/commitmeta.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/attic-labs/noms/go/types"
+	"github.com/attic-labs/noms/go/d"
+)
+
+type CommitMeta interface {
+	Marshal() types.Struct
+}
+
+type commitMeta struct {
+	Date string
```
aboodman: Other things that would be cool to track:

- distance required by grouping function
- path to input data (ask @willhite about "pinned paths")

samples/go/photo-dedup/model/photo.go
```diff
@@ -0,0 +1,73 @@
...
+
+type Photo interface {
+	ID() ID
+	Dhash() dhash.Hash
+	SetDhash(dhash.Hash)
+	IterSizes(cb func(width int, height int, url string))
+	Marshal() types.Struct
+}
+
+type lazyPhoto struct {
```
aboodman: I don't think you gain much from doing this lazy unmarshalling. The built-in unmarhsal doesn't traverse very far into the tree. And since you typically setup the struct having the shape you want ahead of time, you by definition only unmarshal what you want. I think you should remove this and just use the built-in Unmarshal().

samples/go/photo-dedup/model/photogroup.go
```diff
@@ -0,0 +1,74 @@
...
+
+package model
+
+import (
+	"github.com/attic-labs/noms/go/types"
+	"github.com/attic-labs/noms/samples/go/photo-dedup/dhash"
+	"github.com/attic-labs/noms/go/d"
+)
+
+type PhotoGroup interface {
```
aboodman: I think we want to avoid these kind of wrapper datatypes. If we need to improve something in marshaling or elsewhere, let's do that, but we don't want to have to create these.

aboodman:
Super excited to see this by the way!

samples/go/photo-dedup/cmd/photo-dedup/main.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
```
ehalpern: Both jobs depend on a common code (model and dhash). It would make more sense to put (at least some of) the model code in a common location shared by all jobs working on photos, but there's no place to put it yet.

samples/go/photo-dedup/job/dedupjob.go
ehalpern: Agreed. Need a place to put it.

samples/go/photo-dedup/job/dedupjob.go
ehalpern: I didn't think a buffered channel would help much here because reading from the channel is serialized, but on second thought it'd be good to allow the io to read ahead.

samples/go/photo-dedup/job/hashjob.go
ehalpern: I think so. In this case, there are failures that occur because of formats we don't handle, images that aren't reachable and possibly rate limiting. It makes sense to skip past them but we don't want to lose track of the fact that they happened. I think we need a mechanism for logging errors and warning in plugins in general.

samples/go/photo-dedup/job/hashjob.go
ehalpern: Agreed.

samples/go/photo-dedup/model/photo.go
ehalpern: Here's the problem: 

If I unmarshal into a go Photo struct that doesn't define all the fields in the noms struct, those fields will be lost when marshalling back into a noms struct.

Any job that is adding a field to a struct will have this problem. Not sure what the best solution to this problem is yet. One would be to have an option that forces the go struct to define all the fields in the noms struct. This would fix the problem, but it means that job code has to change whenever somethings added to the schema. Another solution would be to provide a way to merge go struct into a noms struct on unmarshal.

samples/go/photo-dedup/job/dedupjob.go
```diff
@@ -6,26 +6,27 @@ package job
...
-	"github.com/attic-labs/noms/go/util/status"
 )
 
 var grouper *photoGrouper
 
 // DeduplicateJob reads Set<Photo>'s (annotated with dhash) and writes Set<PhotoGroup> to
 // outDS where each group contains all duplicates.
-func DeduplicateJob(db datas.Database, photoSets []types.Value, outDS datas.Dataset) error {
-	return commitPhotoGroups(db, outDS, groupPhotos(db, photoSets))
+func DeduplicateJob(db datas.Database, photoSets []types.Value, outDS datas.Dataset, similarityThreshold int) error {
```
aboodman: "s...d" -> "threshold", or even "thresh" would be more go-ey.

you'll get used to it :)

samples/go/photo-dedup/job/hashjob.go
```diff
@@ -33,51 +33,50 @@ func HashPhotosJob(db datas.Database, photoSets []types.Value, outDS datas.Datas
 
 // hashPhotos adds a dhash to each photo in photoSets and delivers them on the returned channel
 func hashPhotos(db datas.Database, photoSets []types.Value) <-chan types.Struct {
-	photosToHashCh := make(chan model.Photo)
-	hashedPhotosCh := make(chan types.Struct)
+	numWorkers := runtime.NumCPU() * 4
```
aboodman: I see both `NumCPU()*4` and `NumCPU() * 4` in this code. I'm surprised `gofmt` doesn't have an opinion on this...

samples/go/photo-dedup/dhash/LICENSE
```diff
@@ -0,0 +1,24 @@
+Copyright (c) 2015, Andy Balholm
```
aboodman: Let's just do the right thing here. It will only take a few mins ... I'll help.

samples/go/photo-dedup/job/hashjob.go
ehalpern: It does! This is the result for gofmt. The idea is to remove spaces around hi precedence operators when lower precedence operators are also present.

PR #2793, opened 2016-11-02
Noms, demo-server, and photos UI changes to support auth

kalman:
The big change here is adding a new Spec class in spec.js. This replaces
DatabaseSpec/DatasetSpec/PathSpec in specs.js, but I'm leaving those in
and moving code over in a later patch. For now, only photos UI.

The photos UI change is to plumb through the authorization token through
the Spec code. For now, it's reading it from a URL parameter, but soon
I'll make it session based (probably localStorage).

The demo-server change is to add the Authorization header into CORS.

### DIFF
--- a/js/noms/src/http-batch-store.js
+++ b/js/noms/src/http-batch-store.js
@@ -19,6 +19,7 @@ import HTTPError from './http-error.js';
 import {notNull} from './assert.js';
 import nomsVersion from './version.js';
 
+export const DEFAULT_MAX_READS = 5;
 const HTTP_STATUS_CONFLICT = 409;
 const versionHeader = 'x-noms-vers';
 
@@ -52,7 +53,8 @@ const readBatchOptions = {
 export default class HttpBatchStore extends RemoteBatchStore {
   _rpc: RpcStrings;
 
-  constructor(urlparam: string, maxReads: number = 5, fetchOptions: FetchOptions = {}) {
+  constructor(
+      urlparam: string, maxReads: number = DEFAULT_MAX_READS, fetchOptions: FetchOptions = {}) {
     const [url, params] = separateParams(urlparam);
     const rpc = {
       getRefs: url + '/getRefs/' + params,
--- a/js/noms/src/spec.js
+++ b/js/noms/src/spec.js
@@ -1,0 +1,315 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// @flow
+
+import AbsolutePath from './absolute-path.js';
+import {invariant} from './assert.js';
+import {BatchStoreAdaptor} from './batch-store.js';
+import Dataset, {datasetRe} from './dataset.js';
+import Database from './database.js';
+import HttpBatchStore, {DEFAULT_MAX_READS} from './http-batch-store.js';
+import MemoryStore from './memory-store.js';
+import type Value from './value.js';
+
+export type SpecOptions = {
+  /**
+   * Authorization token for requests. For example, if the Database is backed by
+   * HTTP this will used for an `Authorization: Bearer ${authorization}` header.
+   */
+  authorization?: string;
+
+  /**
+   * If this is specified and > 0, the Database instances returned by
+   * `database()` will be backed by a cache.
+   */
+  cacheSize?: number;
+};
+
+/**
+ * Spec is a parsed specification for the location of a Noms database, dataset,
+ * or value.
+ *
+ * Spec has the same `Database` instance for its construction, and the
+ * accessors have the same mutability semantics as the `Database` interface,
+ * which is mutable. For example, if the underlying `Database` commits a
+ * dataset, then calls to `Spec.dataset()` will return the new one.
+ *
+ * For example:
+ *  - a database spec might be `mem` or `https://demo.noms.io/cli-tour`.
+ *  - a dataset spec might be `mem::photos` or `https://demo.noms.io/cli-tour::sf-crime`.
+ *  - a path spec might be `http://demo.noms.io::foo.bar` or `http://demo.noms.io::#abcdef.bar`
+ *
+ * See https://github.com/attic-labs/noms/blob/master/doc/spelling.md.
+ */
+export default class Spec {
+  _spec: string;
+  _opts: ?SpecOptions;
+  _protocol: string;
+  _databaseName: string;
+
+  // Lazily created.
+  _database: ?Database;
+
+  // Exists if the spec was constructed through `forDataset()`.
+  _datasetName: ?string;
+
+  // Exists if the spec was constructed through `fromValue()` or `pin()`.
+  _path: ?AbsolutePath;
+
+  constructor(spec: string, databaseSpec: string, opts: ?SpecOptions) {
+    this._spec = spec;
+    this._opts = opts;
+    const [protocol, databaseName] = parseDatabaseSpec(databaseSpec);
+    this._protocol = protocol;
+    this._databaseName = databaseName;
+  }
+
+  /**
+   * Creates a `Spec` from a spec to a database.
+   */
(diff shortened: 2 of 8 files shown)


kalman:
@willhite assigning to you because I think you know the most about Path code?

js/noms/src/spec.js
```diff
@@ -0,0 +1,315 @@
...
+  /**
+   * Returns a `Spec` in which the dataset component, if any, has been replaced
+   * with the hash of the HEAD of that dataset. This "pins" the path to the
+   * state of the database at the current moment in time.
+   *
+   * Returns this if the PathSpec is already "pinned".
+   *
+   * Throws a `TypeError` if this spec does not have an associated dataset.
+   */
+  async pin(): Promise<?Spec> {
```
kalman: The only real difference between spec.js and specs.js is the implementation of `pin()`. It now supports also pinning Datasets, as opposed to an AbsolutePath from a dataset. This didn't make sense before because `pin()` was on `PathSpec`.

Tests have been added accordingly.

kalman:
The motivating reason for this change is because I need to pass the authentication header into the HTTP stack. There is no way with the paths to do this, because of the wacky way that PathSpec embeds a DatabaseSpec, creates a new Database each call to value(), etc.

I had this problem with cacheSize as well - now I'll be able to clean up a bit more of the photos UI to properly cache data.

I think combining the different Spec types into a single class is cleaner anyway. It does mean it's slightly less type safe, but given the usage of Spec so far has been to access the fields immediately, this seems fine.

js/noms/src/spec.js
```diff
@@ -0,0 +1,315 @@
...
+
+  /**
+   * Returns the `Dataset` from the spec.
+   *
+   * Throws a `TypeError` if this spec does not have a associated dataset.
+   *
+   * Make sure to call `close()`.
+   */
+  dataset(): Dataset {
+    return this.database().getDataset(this.datasetName());
```
kalman: I'm debating whether it makes sense to make this also work for paths, *if* the path is relative to a dataset. This would be equivalent to pulling out that logic from pin() which figures it out.

So far I think no, it's more consistent for this simply to only work if there is a datasetName.

js/noms/src/http-batch-store.js
```diff
@@ -52,7 +53,8 @@ const readBatchOptions = {
 export default class HttpBatchStore extends RemoteBatchStore {
   _rpc: RpcStrings;
 
-  constructor(urlparam: string, maxReads: number = 5, fetchOptions: FetchOptions = {}) {
+  constructor(
```
willhite: Does this change belong in this PR? I'm not sure how this affects specs..

js/noms/src/spec.js
```diff
@@ -0,0 +1,315 @@
...
+   * Creates a `Spec` from a spec to a database.
+   */
+  static forDatabase(spec: string, opts: ?SpecOptions): Spec {
+    return new Spec(spec, spec, opts);
+  }
+
+  /**
+   * Creates a `Spec` from a spec to a dataset.
+   */
+  static forDataset(spec: string, opts: ?SpecOptions): Spec {
```
willhite: Do we still need forDatabase, forDataset, and forPath? Can't we just have a parse()? It seems like we can determine what fields to fill out from the spec string.

js/noms/src/spec.js
```diff
@@ -0,0 +1,315 @@
...
+    }
+    return this._database;
+  }
+
+  /**
+   * Returns the dataset name, for example `"sf-crime"`.
+   *
+   * Throws a `TypeError` if this spec does not have an associated value.
+   */
+  datasetName(): string {
```
willhite: Why throw an error rather than just return nil?

js/noms/src/spec.js
```diff
@@ -0,0 +1,315 @@
...
+  }
+
+  /**
+   * Returns the path with this spec's `Database`. Throws a `TypeError` if this
+   * spec does not have an associated path.
+   */
+  path(): AbsolutePath {
+    if (!this._path) {
+      throw new TypeError(`${this._spec} is not a path spec`);
+    }
```
willhite: Same comment here, can we just return nil?

js/noms/src/spec.js
```diff
@@ -0,0 +1,315 @@
...
+   */
+  close(): Promise<void> {
+    if (!this._database) {
+      return Promise.resolve();
+    }
+    return this._database.close();
+  }
+
+  _createDatabase(): Database {
+    let batchStore;
```
willhite: There are no tests that exercise this code.

samples/go/demo-server/web_server.go
```diff
@@ -171,6 +171,7 @@ func corsHandle(f httprouter.Handle) httprouter.Handle {
 		// Can't use * when clients are using cookies.
 		w.Header().Add("Access-Control-Allow-Origin", r.Header.Get("Origin"))
 		w.Header().Add("Access-Control-Allow-Methods", "GET, POST")
+		w.Header().Add("Access-Control-Allow-Headers", "Authorization")
```
willhite: I think that the allowed headers are normally listed rather than put into separate header lines: https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Access-Control-Allow-Headers. I'm not sure if it makes a difference.

js/noms/src/spec.js
kalman: I was thinking because a Spec has a single purpose, but I think returning null is good too. Prefer it right now, so I'll change.

js/noms/src/spec.js
kalman: (sanity checking)

But I'll change it.

js/noms/src/http-batch-store.js
kalman: Spec needs to set FetchOptions, and to pass that in it needs to pass through a maxReads value as well (just because of parameter ordering). This change is just exposing that `5` constant so that spec can pass in `5` itself.

js/noms/src/spec.js
kalman: Oh right, and the other reason is that because the caller (presumably) knows that they want a path vs dataset vs database spec, it'd be annoying to null check these fields when they know it should be set.

js/noms/src/spec.js
kalman: I think you at least need to distinguish between a database and dataset/path. For example: IPV6 addresses contain `::` and you can't tell if that's a database separator or just part of the database.

But we could combine forDataset/forPath for parsing sake. The use cases are typically quite different, though: you want a dataset if you want to commit to a dataset, for example. You want a path if you want to extract a single value.

If they were combined, code would need to look like:

```javascript
const spec = Spec.parse('http://demo.noms.io/cli-tour::myDataset');
if (!spec.isDataset()) {
  console.error('must be a dataset');
  return;
}
```

The advantage of separate forDataset/forPath methods is that the sanity checking is done for you.

kalman:
> I think that the allowed headers are normally listed rather than put into separate header lines

I don't know why github won't let me reply to this comment inline, but `Add` actually does append (confusingly): https://golang.org/pkg/net/http/#Header.Add

kalman:
After going through your comments (no changes uploaded yet) I think what makes sense is either:

1. intrinsic error checking: go with what I have with separate forDatabase/forDataset/forPath methods, and have the methods which only make sense for one of the other (e.g. `dataset()`) throw errors if they don't match.
2. extrinsic error checking: you still need forDatabase (I think?) but have a combined forPath method, then make callers check themselves like `if (!spec.dataset())`.

Instinctively it seems like if we go with (2) then people will just write wrappers that provide an interface like (1).

Any other ideas/thoughts?

js/noms/src/spec.js
willhite: Given the convincing arguments that you make against having one parse method(i.e. ipv6 and different use cases for path vs dataset), it seems like programmers will always know what 'get' methods should be called. It's probably best to continue throwing errors so that callers don't have to constantly check for null.

kalman:
Cool, thanks.

PR #2794, opened 2016-11-02
Danloader

willhite:


samples/go/downloader/downloader.go
```diff
@@ -0,0 +1,326 @@
...
+	Url        string
+	Downloaded bool
+	Err        string    `noms:",omitempty"`
+	BlobRef    types.Ref `noms:",omitempty"`
+}
+
+func download() (win bool) {
+	var inPathArg = flag.String("in-path", "", "root of object hierarchy to scan for RemotePhoto objects")
+	var outDsArg = flag.String("out-ds", "", "name of output dataset")
+	var cacheDsArg = flag.String("cache-ds", "", "name of photo-cache dataset")
```
aboodman: Having a separate cache-ds will be less compatible with a future event loop world. Consider using just one dataset now?

samples/go/downloader/downloader.go
```diff
@@ -0,0 +1,322 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
```
aboodman: Seems like this should really go in other repo.

samples/go/downloader/treewalker.go
```diff
@@ -0,0 +1,83 @@
...
+// TreeWalkCallback defines a function prototype submitting successive values
+// from TreeWalk to the caller. At any point, the function can return true
+// to prevent traversing any of the current value's children.
+type TreeWalkCallback func(p types.Path, parent, v types.Value) (stop bool)
+
+// TreeWalk walks over a noms graph starting at 'v' and calls 'twcb' for each
+// value it encounters. It calls 'twcb' with a path from the original value to
+// the current value, the current value's parent, and the current value. TreeWalk
+// also takes a ValueReader so that it can traverse values across refs.
+//
```
aboodman: Remove blank line.

samples/go/downloader/treewalker.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/attic-labs/noms/go/hash"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// TreeWalkCallback defines a function prototype submitting successive values
+// from TreeWalk to the caller. At any point, the function can return true
+// to prevent traversing any of the current value's children.
+type TreeWalkCallback func(p types.Path, parent, v types.Value) (stop bool)
+
+// TreeWalk walks over a noms graph starting at 'v' and calls 'twcb' for each
```
aboodman: Should we just update the built-in noms walk to maintain the path and pass it to callee?

samples/go/downloader/downloader.go
willhite: ha ha

willhite: i'll leave it here until we're done with the Code Review and then move it to attic at that point.

samples/go/downloader/resource_cache.go
```diff
@@ -0,0 +1,67 @@
...
+	"errors"
+	"sync"
+
+	"github.com/attic-labs/noms/go/datas"
+	"github.com/attic-labs/noms/go/spec"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// ResourceCache is a Map<String, Ref<Blob>>
+type ResourceCache struct {
```
aboodman: I sorta feel like this should all be private. WDYT?

samples/go/downloader/resource_cache.go
```diff
@@ -0,0 +1,67 @@
...
+	cache types.Map
+	orig  types.Map
+	mutex sync.Mutex
+}
+
+func GetResourceCache(db datas.Database, dsname string) (*ResourceCache, error) {
+	m, ok := db.GetDataset(dsname).MaybeHeadValue()
```
aboodman: What is `m` short for? Maybe `root`?

samples/go/downloader/resource_cache.go
```diff
@@ -0,0 +1,67 @@
...
+	if ok {
+		refOfBlobType := types.MakeRefType(types.BlobType)
```
aboodman: I think we should try to lean on marshaling for this kind of use case ... building up noms types for subtype checking is pretty laborious. Can you instead do:

- try to unmarshal onto `Map`
- check the type of the key and value manually with just equality

?

samples/go/downloader/resource_cache.go
```diff
@@ -0,0 +1,67 @@
...
+	"sync"
+
+	"github.com/attic-labs/noms/go/datas"
+	"github.com/attic-labs/noms/go/spec"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// ResourceCache is a Map<String, Ref<Blob>>
+type ResourceCache struct {
+	cache types.Map
```
aboodman: Can you comment the difference between `cache` and `orig` ?

samples/go/downloader/resource_cache.go
```diff
@@ -0,0 +1,67 @@
...
+	orig  types.Map
+	mutex sync.Mutex
```
aboodman: Is there a reader/writer lock you could use? Might be worthwhile.

go/diff/patch.go
```diff
@@ -0,0 +1,146 @@
...
+func (r Patch) Len() int {
+	return len(r)
+}
+
+func (r Patch) Less(i, j int) bool {
+	if r[i].Path.Equals(r[j].Path) {
+		if r[i].ChangeType == r[j].ChangeType {
+			return false
+		}
+		if r[i].ChangeType == types.DiffChangeRemoved {
```
aboodman: To make this easier to read, you could have a map from these constants to numeric values, then just compare those directly.

go/diff/patch.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		if r[j].ChangeType == types.DiffChangeAdded {
+			return true
+		}
+	}
+	return pathIsLess(r[i].Path, r[j].Path)
+}
+
+// Utility methods on path
+// Todo: should these be on types.Path & types.PathPart
```
aboodman: TODO

aboodman: .append("?")

aboodman: Capitalize "Should"

go/diff/patch.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		switch pathPartCompare(pp1, p2[i]) {
+		case -1:
+			return true // p1 < p2
+		case 1:
+			return false // p1 > p2
+		}
+	}
+
+	if len(p2) > len(p1) {
```
aboodman: Nit: can just say: `return len(p2) > len(p1)`

go/diff/apply_patch.go
```diff
@@ -0,0 +1,287 @@
...
+//    ApplyPatch(a1, Diff(a1, a2)) == a2
+// This is useful for IncrementalUpdate() and possibly other problems. See
+// Updater.go for more information.
+//
+// This function uses a patchStack to maintain state of the graph as it cycles
+// through the diffs in a patch, applying them to 'root' one by one. It minimizes
+// the number of changes to the graph by making subsequent changes locally and
+// incorporating lower level changes into the higher level once all the lower
+// level changes have been made.
+func ApplyPatch(root types.Value, patch Patch) types.Value {
```
aboodman: We either need to go to the whiteboard for you to explain how this works, or @arv / @rafael-atticlabs need to review, or you need to just land it w/o review :).

samples/go/downloader/downloader.go
```diff
@@ -0,0 +1,326 @@
...
+	shouldUpdateCb := func(p types.Path, root, parent, v types.Value) (res bool) {
+		shouldCnt.Increment()
+		return v != nil && v.Type().Kind() == types.StructKind && v.Type().Desc.(types.StructDesc).Name == "RemoteResource"
+	}
+
+	updateCnt := &counter{}
+	failedCnt := &counter{}
+	rwMutex := sync.RWMutex{}
+
+	// use info from dif to create a new RemoteResource and return it.
```
aboodman: Capitalize 'Use'.

go/diff/apply_patch.go
```diff
@@ -0,0 +1,287 @@
...
+	"sort"
+
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// ApplyPatch applies a Patch (list of diffs) to a graph. It makes the following
+// contract:
+//  Given 2 Noms graphs: a1 and a2:
+//    ApplyPatch(a1, Diff(a1, a2)) == a2
```
willhite: Diff() doesn't actually return anything, but if it did, i would return []Difference which I've typedef'd to path: 'type Patch []Difference'.
I think ApplyPatch() gives some useful context and Apply() is too general. Let me know if you still want this changed.

go/diff/apply_patch.go
```diff
@@ -0,0 +1,287 @@
...
+// This is useful for IncrementalUpdate() and possibly other problems. See
+// Updater.go for more information.
```
willhite: fixed

go/diff/apply_patch.go
willhite: What!? The comments make is self-explanatory. :-)

go/diff/apply_patch.go
```diff
@@ -0,0 +1,287 @@
...
+		p := dif.Path
+
+		// idx will hold the index of the last common element between p and
+		// lastPath (p from the last iteration).
+		var idx int
+
+		// p can be identical to lastPath in certain cases. For example, when
+		// one item gets removed from a list at the same place another item
+		// is added to it. In this case, we need pop the last operation of the
+		// stack early and set the idx to be the len(p) - 1
```
willhite: done

go/diff/patch.go
willhite: I don't understand this comment. Can you show me an example?

go/diff/patch.go
willhite: ok

willhite: ok

willhite: ok

go/diff/patch.go
willhite: done

samples/go/downloader/downloader.go
```diff
@@ -0,0 +1,322 @@
...
+)
+
+func main() {
+	if !download() {
+		exit.Fail()
+	}
+}
+
+func usage() {
+	fmt.Println("usage: downloader --in-path <pathspec> --out-ds <dsname> --cache-ds <dsname>")
```
willhite: doing this now...

samples/go/downloader/downloader.go
willhite: Having a global cache allow us to commit independently from the data dataset, share the cache among multiple datasets, and have an exact mirror of the source. Per our conversation earlier, I'll leave it in it's own dataset for the time being.

samples/go/downloader/downloader.go
```diff
@@ -0,0 +1,326 @@
...
+	Downloaded bool
+	Err        string    `noms:",omitempty"`
+	BlobRef    types.Ref `noms:",omitempty"`
+}
+
+func download() (win bool) {
+	var inPathArg = flag.String("in-path", "", "root of object hierarchy to scan for RemotePhoto objects")
+	var outDsArg = flag.String("out-ds", "", "name of output dataset")
+	var cacheDsArg = flag.String("cache-ds", "", "name of photo-cache dataset")
+	var concurrencyArg = flag.Uint("concurrency", 4, "number of concurrent http calls to retrieve remote resources")
```
willhite: done

samples/go/downloader/downloader.go
```diff
@@ -0,0 +1,326 @@
...
+	verbose.RegisterVerboseFlags(flag.CommandLine)
+	flag.Usage = usage
+	flag.Parse(false)
+
+	if flag.NArg() != 0 {
+		flag.Usage()
+		return
+	}
+
+	// Check that concurrency is at least 1
```
willhite: ok, I was just tagging each paragraph with a comment because arg checking is so laborious. I removed this one.

samples/go/downloader/downloader.go
```diff
@@ -0,0 +1,326 @@
...
+	check := []string{"in-path", *inPathArg, "out-ds", *outDsArg, "cache-ds", *cacheDsArg}
+	for i := 0; i < len(check)-1; i += 2 {
+		if check[i+1] == "" {
+			fmt.Fprintln(os.Stderr, "error: missing required argument:", check[i])
+			flag.Usage()
+			return
+		}
+	}
+
+	// Resolve the in-path arg and get the inRoot.
```
willhite: done

samples/go/downloader/downloader.go
```diff
@@ -0,0 +1,326 @@
...
+	db, inRoot, err := cfg.GetPath(*inPathArg)
+	if err != nil || inRoot == nil {
+		if err == nil {
+			err = errors.New("Could not find referenced value.")
+		}
+		fmt.Fprintf(os.Stderr, "Invalid input path '%s': %s\n", *inPathArg, err)
+		return
+	}
+
+	if datas.IsCommitType(inRoot.Type()) {
```
willhite: Not having this here seems pretty bad. I'll leave it for now with the hope that this all gets standardized some day.

samples/go/downloader/downloader.go
willhite: Done

samples/go/downloader/resource_cache.go
willhite: fine with that.

samples/go/downloader/resource_cache.go
willhite: done

samples/go/downloader/resource_cache.go
willhite: the cache is a Map<string, Ref<Blob>>

samples/go/downloader/resource_cache.go
willhite: I'm not sure I understand your point here. I don't want to unmarshal the cache into a go map because it could be very large. Can you explain this in more detail?

samples/go/downloader/treewalker.go
willhite: This also passes in the parent of the current value along with the path. I like the idea of having one WalkValues function that hands all of this info to the client but I'm not sure if the cost of allocating a new path for every value would be prohibitive.

samples/go/downloader/treewalker.go
willhite: done.

willhite:
@aboodman Please take a look at these comments and give me your responses.

go/diff/apply_patch.go
aboodman: It's a minor thing but I still `Apply` is best because this is in the `diff` package so it's very clear what you're applying.

go/diff/apply_patch.go
aboodman: I guess the reason this comment wasn't clear to me was because I don't understand the last sentence at all. Words like "minimizes" and "subsequent" and "locally" are too abstract ... maybe there is a way to explain the strategy the algorithm uses with more specific words while still staying high enough level to not write a book.

Anyway, it's not a criticism.

go/diff/patch.go
aboodman:
```
vals := map[int]int{types.DiffChangeRemoved: 0, types.DiffCahngeAdded: 1}
return vals[r[i].ChangeType] < vals[r[j].CahngeType]
```

samples/go/downloader/treewalker.go
aboodman: It wouldn't be an allocation because the path objects would be on the stack.

go/diff/patch.go
willhite: Yes. This is much better. Thanks.

willhite:
@aboodman PTAL

aboodman:
lgtm!

let's get with the danloading

aboodman:
!!

aboodman:
hook it up hook it up hook it up

PR #2798, opened 2016-11-03
Add year, month, day to photo indexing

mikegray:


mikegray:
@aboodman PTAL, wasn't exactly sure if you just wanted byYear then the photo set ordered by date, or if given UI needs we might want to be able to dig as needed into year, then month, and then day.

aboodman:
We shouldn't need this because the existing byDate index is sorted. So it
seems like it can already answer all three of these questions, right?

mikegray:
What 3 questions are these? So you think we should walk or scan across the byDate index to find how many items we have for a given year? and for a given year-month combination?

aboodman:
The existing `byDate` index can answer the questions:

* what are all the items for a given year
* what are all the items for a given year/month
* what are all the items for a given year/month/day

In exact same number of seeks as an index from `year->set<photo>`. Because the index is sorted, you can seek directly to the start of a particular year (or year/month, or year/month/day). Then scanning over that is same cost as scanning a specialized index. Yay noms.

aboodman:
It can also answer the questions:

* how many items are there for a given year?
* how many items are there for a given year/month?
* how many items are there for a given year/month/day?

... in not quite same efficiency as a specialized index, but still really good (logarithmic in size of byDate index).

aboodman:
If what you want is the answer to the latter set of questions (how many items, vs actually fetching the items), you should talk to @kalman about how best to do that. I don't know if the right API is exposed anywhere, but the index definitely contains the info you want.

mikegray:
I believe the efficiency of "how many items" will not be good enough for UI responsiveness. No reason to optimize this prematurely though so I have closed this PR and we can easily recreate it if needed somewhere down the line.

kalman:
I can comment on an issue about this when it comes up again.

We might have to do a bit of work to expose more innards of the sequence API to figure this out. 

For example, like @aboodman says, you can figure out how many items there are in a given year by getting the index at year-in-nanoseconds, the index at (year+1)-in-nanoseconds, then comparing the offsets. However I don't believe we actually expose the offsets as part of the public API. @rafael-atticlabs

PR #2800, opened 2016-11-03
Don't use invalid Root in bad version test

cmasone-attic:
The httpBatchStore test TestVersionMismatch() expects a panic,
but the test was actually potentially causing MULTIPLE panics.
One due to the version mismatch, and another due to using an
invalid root.


cmasone-attic:
PTAL

PR #2803, opened 2016-11-04
Update to Flow 0.34

arv:


arv:
this got screwed up...

js/noms/src/type.js
```diff
@@ -16,7 +16,7 @@ import search from './binary-search.js';
 import {staticTypeCache} from './type-cache.js';
 
 export interface TypeDesc {
-  kind: NomsKind;
```
kalman: whoa what does this mean

arv:
Fixed.

js/noms/src/type.js
arv: https://flowtype.org/blog/2016/10/04/Property-Variance.html

arv:
jenkins: test this

PR #2804, opened 2016-11-04
Clobber ValueStore cache entry on WriteValue

### LINKED ISSUE #2802: JS vs Go memory stores cache nil values differently
I noticed this while writing spec tests for Go and JS.

I've found that the following JS code behaves like:
```javascript
const s = "hello";
const db = new Database(new BatchStoreAdaptor(new MemoryStore()));
await db.readValue(getHash(s)); // undefined
db.writeValue(s);
await db.readValue(getHash(s)); // "hello"
```

Whereas the equivalent Go code behaves like:
```go
s := types.String("hello")
db := datas.NewDatabase(chunks.NewMemoryStore())
db.ReadValue(s.Hash()) // nil
db.WriteValue(s)
db.ReadValue(s.Hash()) // nil
```

The difference is that the final line in the Go code returns `nil`, but I would expect it to return `s` like JS. This is pretty surprising behavior.

@cmasone-attic

cmasone-attic:
ValueStore caches Values that are read out of it, but it doesn't
do the same for Values that are written. This is because we expect
that reading Values shortly after writing them is an uncommon usage
pattern, and because the Chunks that make up novel Values are
generally efficiently retrievable from the BatchStore that backs
a ValueStore. The problem discovered in issue #2802 is that ValueStore
caches non-existence as well as existence of read Values. So, reading
a Value that doesn't exist in the DB would result in the ValueStore
permanently returning nil for that Value -- even if you then go and
write it to the DB.

This patch drops the cache entry for a Value whenever it's written.

Fixes #2802

### DIFF
--- a/go/types/value_store_test.go
+++ b/go/types/value_store_test.go
@@ -12,6 +12,17 @@ import (
 	"github.com/attic-labs/testify/assert"
 )
 
+func TestValueReadWriteRead(t *testing.T) {
+	assert := assert.New(t)
+
+	s := String("hello")
+	vs := NewTestValueStore()
+	assert.Nil(vs.ReadValue(s.Hash())) // nil
+	r := vs.WriteValue(s)
+	v := vs.ReadValue(r.TargetHash()) // non-nil
+	assert.True(s.Equals(v))
+}
+
 func TestCheckChunksInCache(t *testing.T) {
 	assert := assert.New(t)
 	cs := chunks.NewTestStore()
@@ -89,7 +100,7 @@ func TestCacheOnReadValue(t *testing.T) {
 	v := cvs2.ReadValue(r.TargetHash())
 	assert.True(bref.Equals(v))
 	assert.True(cvs2.isPresent(b.Hash()))
-	assert.True(cvs2.isPresent(b.Hash()))
+	assert.True(cvs2.isPresent(bref.Hash()))
 }
 
 func TestHintsOnCache(t *testing.T) {
--- a/go/util/sizecache/size_cache.go
+++ b/go/util/sizecache/size_cache.go
@@ -70,7 +70,7 @@ func (c *SizeCache) Add(key interface{}, size uint64, value interface{}) {
 		defer c.mu.Unlock()
 
 		if _, ok := c.entry(key); ok {
-			// this value is already in the cache just return
+			// this value is already in the cache; just return
 			return
 		}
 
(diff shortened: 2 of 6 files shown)


cmasone-attic:
PTAL

go/types/value_store_test.go
```diff
@@ -89,7 +100,7 @@ func TestCacheOnReadValue(t *testing.T) {
 	v := cvs2.ReadValue(r.TargetHash())
 	assert.True(bref.Equals(v))
 	assert.True(cvs2.isPresent(b.Hash()))
-	assert.True(cvs2.isPresent(b.Hash()))
+	assert.True(cvs2.isPresent(bref.Hash()))
```
kalman: `.TargetHash()`?

go/util/sizecache/size_cache.go
```diff
@@ -92,3 +92,15 @@ func (c *SizeCache) Add(key interface{}, size uint64, value interface{}) {
 		}
 	}
 }
+
+// Drop will remove the element associated with the given key from the cache.
+func (c *SizeCache) Drop(key interface{}) {
```
kalman: Kinda bummer this class needs to use interface{} everywhere. Just an observation.

js/noms/src/value-store-test.js
```diff
@@ -87,6 +90,31 @@ suite('ValueStore', () => {
...
+    const v1 = await vs.readValue(hash);
+    assert.equal(null, v1);
+    (bs: any).get = () => { throw new Error(); };
+    const v2 = await vs.readValue(hash);
+    assert.equal(null, v2);
+    await vs.close();
+  });
+
+  test('write clobbers cached nonexistence', async () => {
+    const vs = new ValueStore(new BatchStoreAdaptor(new MemoryStore()), 1e6);
```
kalman: You should be able to use a much smaller cache here, you're only storing "hello".

js/noms/src/value-store.js
```diff
@@ -173,14 +175,24 @@ export class SizeCache<T> {
       }
     }
   }
+
+  drop(hash: Hash): void {
+    const key = hash.toString();
+    const entry = this._cache.get(key);
+    if (entry) {
+      this._size -= entry.size;
```
kalman: Don't you need to delete it from _cache?

go/types/value_store_test.go
cmasone-attic: Nope, the test is verifying that, as a result of reading the Value `bref` out of `cvs2`, both `b` and `bref` are now in cvs2's cache of 'chunks I know about'

go/util/sizecache/size_cache.go
cmasone-attic: It's the Golang Way!

js/noms/src/value-store.js
cmasone-attic: yeah, I had that but in deleting my sweet printf debugging it got nuked. Javascript is...

cmasone-attic:
PTAL

rafael-atticlabs:
deferring to @kalman since he already approved.

PR #2806, opened 2016-11-04
Close and destroy databases being left open during tests

cmasone-attic:
There were several tests in the Database suites that were failing to
close test Databases that had orderedChunkCaches in them (backed by
levelDBs). Close them.

I was ALSO failing to destroy the cache used in LocalDatabase
instances only while testing Pull(). That's cleared up now as well.


cmasone-attic:
PTAL

This seems to have cleared up the issues when I run the tests locally

PR #2807, opened 2016-11-05
Port new Spec API to Go from JS

kalman:
This is a side-by-side port, taking inspiration from the old dataspec.go
code. Notably:

- LDB support has been added in Go. It wasn't needed in JS.
- There is an Href() method on Spec now.
- Go now handles IPV6.
- Go no longer treats access_token specially.
- Go now has Pin.
- I found some issues in the JS while doing this, I'll fix later.

I've also updated the config code to use the new API so that basically
all the Go samples use the code, even if they don't really change.

### DIFF
--- a/go/spec/spec.go
+++ b/go/spec/spec.go
@@ -2,8 +2,6 @@
 // Licensed under the Apache License, version 2.0:
 // http://www.apache.org/licenses/LICENSE-2.0
 
-// TODO: Make sure that removing access_token still works.
-
 // Package spec provides builders and parsers for spelling Noms databases,
 // datasets and values.
 package spec
--- a/go/spec/spec.go
+++ b/go/spec/spec.go
@@ -1,0 +1,297 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// TODO: Make sure that removing access_token still works.
+
+// Package spec provides builders and parsers for spelling Noms databases,
+// datasets and values.
+package spec
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/attic-labs/noms/go/chunks"
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/datas"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// SpecOptions customize Spec behavior.
+type SpecOptions struct {
+	// Authorization token for requests. For example, if the database is HTTP
+	// this will used for an `Authorization: Bearer ${authorization}` header.
+	Authorization string
+}
+
+// Spec describes a Database, Dataset, or a path to a Value. They should be
+// constructed by parsing strings, either through ForDatabase, ForDataset, or
+// ForPath (or their Opts variations).
+type Spec struct {
+	// Spec is the spec string this was parsed into.
+	Spec string
+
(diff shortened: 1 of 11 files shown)


kalman:
This is pretty epic, however:
* `config` changes were mechanical. The formatting was just gofmt.
* The spec changes are basically a side-by-side translation, you can review it that way if you like.
* Hopefully the PR description is descriptive.

js/noms/src/spec-test.js
```diff
@@ -70,8 +71,24 @@ suite('Spec', () => {
 
   test('database spec', async () => {
     const invalid = [
-      'mem:stuff', 'mem::', 'mem:', 'http:', 'https:', 'random:', 'random:random',
-      'local', './local', 'ldb', 'ldb:', 'ldb:local',
+      'mem:stuff',
```
kalman: I found it easier to port side-by-side, and easier to read regardless, with these on their own line.

kalman:
Realised a couple of problems, will fix ASAP and ping after that.
* Forgot to re-run the JS tests.
* I need to use that refCountingLeveldbStore class.

kalman:
PTAL, turns out this was fine. I did add a test for the LDB thing though.

kalman:
There is a failing poke test, the string error has changed. I'll fix but it shouldn't hold up the review.

go/spec/spec.go
```diff
@@ -0,0 +1,295 @@
...
+	db := *sp.db
+	if db == nil {
+		return nil
+	}
+
+	*sp.db = nil
+	return db.Close()
+}
+
+func (sp Spec) createDatabase() datas.Database {
```
willhite: Does this result in a panic if it gets called when the database it's trying to connect to doesn't exist? If so, that would be unfortunate. An error would be preferable.

kalman: Only if the protocol isn't recognised, but it shouldn't be possible because `parseDatabaseSpec` will fail (with an error) if it isn't.

However you bring up a good point which is trying to use a Spec which did fail construction, e.g.:
```go
sp, err := ParseDatabaseSpec(...)
sp.GetDatabase()
```
if there actually was an error then `sp` will be `Spec{}` and `GetDatabase()` will `panic("unreachable")`. I do think that panic is the correct behavior here rather than silently returning nil, but I'll test for it.

go/spec/spec.go
```diff
@@ -0,0 +1,295 @@
...
+
+	sp.Path = path
+	return sp, nil
+}
+
+// GetDatabase returns the Database instance that this Spec's DatabaseName
+// describes. The same Database instance is returned every time, unless Close
+// is called. If the Spec is closed, it is re-opened with a new Database.
+func (sp Spec) GetDatabase() datas.Database {
+	if *sp.db == nil {
```
aboodman: I think you meant `sp.db` here. `*sp.db` can also be `nil` because `*sp.db` is of type `Database`, which is interface. But I'm not sure what it would mean for the pointer to be OK, but the db to be a nil interface.

kalman: I meant `*sp.db`. This is so that when you copy a Spec they still reference the underlying Database and can null it out when it closes.

aboodman:
Hm. I noticed this because I tried to construct a spec like this:

sp := Spec{Protocol: "nbs", DatabaseName: "/path/to/thing"}

sp.Database() crashes then.

kalman:
I see. We could make GetDatabase have a pointer receiver and correct this case.

PR #2813, opened 2016-11-08
Don't crash perf suite test if partition cannot be read

kalman:


kalman:
If you can figure out why this is failing, don't worry about it @zcstarr

kalman:
(don't worry about reviewing this CL I mean)

PR #2816, opened 2016-11-09
Change to use interface instead of declare

arv:


kalman:
Interface can extend class? Interesting.

kalman:
lgtm

arv:
The docs are a bit thin but it makes sense. A class implicitly creates an interface.

PR #2817, opened 2016-11-10
Remove dependency on bash shell from package

zcstarr:
Change removes dependency on bash, pushd/popd breaks installation when installing in a non bash env


zcstarr:
@arv ptal

PR #2819, opened 2016-11-10
Export HttpError and SpecOptions to Noms JS API

kalman:
I renamed HTTPError to HttpError for consistency with HttpBatchStore
(and XMLHttpRequest).


js/noms/src/noms.js
```diff
@@ -59,7 +60,7 @@ export {
   getTypeOfValue,
 } from './type.js';
 export {equals, less} from './compare.js';
-export {default as Spec} from './spec.js';
+export {default as Spec, SpecOptions} from './spec.js';
```
arv: Inc minor version for this?

kalman: I did in a 2nd patchset (pushed with -f), did you see that?

arv: See it now.

arv:
Still LGTM

PR #2820, opened 2016-11-11
Add auth parameter to Splore

kalman:


samples/js/splore/src/main.js
```diff
@@ -12,20 +12,20 @@ import {
...
   Hash,
   IndexedMetaSequence,
-  invariant,
-  kindToString,
   List,
   Map,
   OrderedMetaSequence,
   Ref,
   Set,
+  Spec,
```
kalman: lemme fix this import order thing, 1 sec.

PR #2830, opened 2016-11-14
Better test fix. Need to obtain raw bytes before reading from buffer.

ehalpern:
TBR: @zcstarr


zcstarr:
jenkins: test this

zcstarr:
lgtm 🐑 it!

PR #2838, opened 2016-11-16
Change flickr/find-photoss to store urls under RemoteResource.url

ehalpern:
fixes: #2837


aboodman:
This will break current samples. What @willhite did in for a previous example was generate both the old schema and the new one. Unsure whether it's worth trying to keep this stuff working for now.

aboodman:
LGTM

PR #2839, opened 2016-11-17
Remove tagging_user from fb query

aboodman:
This was causing an auth error for @rafael-atticlabs. Unclear why, but it's not necessary in first place.

### DIFF
--- a/samples/js/fb/slurp/src/main.js
+++ b/samples/js/fb/slurp/src/main.js
@@ -65,7 +65,7 @@ const query = [
   'width',
   'height',
   'likes.limit(1000){id,name}',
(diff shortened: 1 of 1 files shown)


samples/js/fb/slurp/src/main.js
```diff
@@ -166,8 +166,13 @@ async function getPhotos(): Promise<List<any>> {
   return result;
 }
 
-function callFacebook(path: string): Promise<any> {
```
aboodman: The argument was actually already a complete url, not just a path.

aboodman:
TBR

Issue #2836, opened 2016-11-16
OID is too slow


arv:
The OID (order id) is only used to determine the order of union types. It has one reqiurement: it needs to be deterministic.

Currently we use a sha512 over the encoded bytes. This is very expensive when there is a deep type hierarchy since at every level where we need to compute the OID we have to serialize the whole sub tree [1].

A simpler and faster method would be to compute an OID as a function of the OIDs of the elements of the type. That way we would only need to visit each type once as we compute the OID. This would however lead to a change in the ordering of the union elements which requires a Noms version change.

[1] That is not entirely true. When we hit a union type further down we can compute the OID for its elements and cache them.

arv:
@rafael-atticlabs

rafael-atticlabs:
Thanks to type simplification, we can just remove it now.

PR #2840, opened 2016-11-17
Cache OIDs as we descend

arv:
Remove validation/normalization of union order and struct field order as we decode a chunk into a type.

Instead the validation happens in ValidatingBatchSink.

We still normalize the union order when a struct type is created directly (not from a chunk) using  makeStructType.

The motivation for this change is that computing the OID (order ID) is expensive and it used to be a O(n^2) since we kept recomputing it as we traversed the type hierarchy.

Towards #2836

### DIFF
--- a/go/types/encoding_test.go
+++ b/go/types/encoding_test.go
@@ -134,7 +134,7 @@ func assertEncoding(t *testing.T, expect []interface{}, v Value) {
 	assert.EqualValues(t, expect, tw.a)
 
 	ir := &nomsTestReader{expect, 0}
-	dec := valueDecoder{ir, vs, staticTypeCache, 0, false}
+	dec := newValueDecoder(ir, vs, staticTypeCache)
 	v2 := dec.readValue()
 	assert.True(t, ir.atEnd())
 	assert.True(t, v.Equals(v2))
--- a/go/types/encoding_test.go
+++ b/go/types/encoding_test.go
@@ -620,3 +620,40 @@ func TestBogusValueWithUnresolvedCycle(t *testing.T) {
 		EncodeValue(g, nil)
 	})
 }
+
+func TestInvalidStructFieldOrder(t *testing.T) {
+	data := []interface{}{
+		uint8(TypeKind),
+		uint8(StructKind), "S", uint32(2) /* len */, "b", uint8(NumberKind), "a", uint8(NumberKind),
+	}
+
+	vs := NewTestValueStore()
+	r := &nomsTestReader{data, 0}
+	dec := valueDecoder{r, vs, staticTypeCache}
+	assert.Panics(t, func() {
+		dec.readValue()
+	})
+}
+
+func TestInvalidUnionOrder(t *testing.T) {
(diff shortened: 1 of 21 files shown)


arv:
@rafael-atticlabs PTAL

I will do the map/forEach/etc in a separate patch.

I also want to refactor this to not sort the elemTypes in readType but throw an error if the order is not correct.

arv:
jenkins: Test this

arv:
jenkins: test this

ghost (approved):
lgtm

js/noms/src/type-cache.js
```diff
@@ -297,84 +320,139 @@ function walkType(t: Type<any>, parentStructTypes: Type<any>[],
...
-
-        const data = alloc(hashByteLength);
-        oids.forEach((oid: Hash) => {
-          const digest = oid.digest;
-          for (let i = 0; i < hashByteLength; i++) {
-            data[i] ^= digest[i];
+          encodeForOID(elemType, mbuf, allowUnresolvedCycles, root, parentStructTypes);
+          h = Hash.fromData(mbuf.data);
+          if (parentStructTypes.indexOf(elemType) === -1) {
+            t.updateOID(h);
```
arv: This is wrong! Changing this to elemType and the tests fail. Will have to go back to the drawing board.

arv:
@rafael-atticlabs This is now ready to be reviewed again.

I'll update the PR description.

js/noms/src/type-cache.js
```diff
@@ -297,84 +310,143 @@ function walkType(t: Type<any>, parentStructTypes: Type<any>[],
...
 
 function generateOID(t: Type<any>, allowUnresolvedCycles: boolean) {
-  const buf = new BinaryWriter();
-  encodeForOID(t, buf, allowUnresolvedCycles, t, []);
-  t.updateOID(Hash.fromData(buf.data));
+  if (!hasOID(t)) {
+    const buf = new BinaryWriter();
+    encodeForOID(t, buf, allowUnresolvedCycles, t, []);
+    const oid = Hash.fromData(buf.data);
+    if (t._oid) {
```
rafael-atticlabs: how can this ever be true?

go/types/encoding_test.go
```diff
@@ -134,7 +134,7 @@ func assertEncoding(t *testing.T, expect []interface{}, v Value) {
 	assert.EqualValues(t, expect, tw.a)
 
 	ir := &nomsTestReader{expect, 0}
-	dec := valueDecoder{ir, vs, staticTypeCache}
+	dec := valueDecoder{ir, vs, staticTypeCache, 0, false}
```
rafael-atticlabs: does an existing ctor not work here?

ghost (approved):
LGTM again (with questions).

Also, what about only generating an OID for types that are directly inside of a union. The three cases that I think we care about this are:

-Sorting types in a union (this recursively comes up when generating an oid)
-Validating sort order of union types

PR #2841, opened 2016-11-18
Teach photos UI about LocalResource

aboodman:
The animation is still not working. I temporarily disabled it. This is so finicky :-/.


samples/js/photos/src/main.js
```diff
@@ -50,7 +51,6 @@ async function getRenderElement(nav: Nav): Promise<React.Element<any>> {
       };
     }
 
-    let indexSpec;
```
kalman: This change doesn't look right?

samples/js/photos/src/main.js
```diff
@@ -87,6 +87,7 @@ async function getRenderElement(nav: Nav): Promise<React.Element<any>> {
   const viewport = new Viewport(window, document.body);
 
   return <PhotosPage
+    db={index.byDate.sequence.vr}
```
kalman: This should be `indexSpec.database()`?

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
-      size: props.gridSize,
-      sizeIsBest: false,
-      url: props.url,
+      blob: this.getBlob(props),
+      tx: false,
```
kalman: `tx` doesn't exist anymore?

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
     };
     this._parentTop = 0;
     this._parentLeft = 0;
-    this._shouldTransition = true;
+    this.load(this.state.blob);
   }
 
-  componentWillUpdate() {
+  componentWillReceiveProps(nextProps) {
```
kalman: `nextProps: Props`

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+    const nextBlob = this.getBlob(nextProps);
+    if (nextBlob == this.state.blob) {
```
kalman: `===`

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+      return;
+    }
+    this.setState({
+      blob: nextBlob,
+    });
+    this.load(nextBlob);
+  }
+
+  componentWillUpdate(nextProps, nextState) {
+    if (this.state.url != nextState.url) {
```
kalman: `!==`

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+      this.releaseBlobURL();
+    }
+
     const rect = ReactDOM.findDOMNode(this).parentElement.getBoundingClientRect();
     this._parentTop = rect.top;
     this._parentLeft = rect.left;
   }
 
+  getBlob(props: Props): NomsBlob {
```
kalman: `_getBlob`

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+    if (nextBlob == this.state.blob) { 
+      return;
+    }
+    this.setState({
+      blob: nextBlob,
+    });
+    this.load(nextBlob);
+  }
+
+  componentWillUpdate(nextProps, nextState) {
```
kalman: `nextProps: Props, nextState: State`

samples/js/photos/src/photo-grid-item.js
```diff
@@ -159,14 +206,19 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+      height: size.height,
+      transition: t,
     };
   }
 
   _maybeTransition(transition: string): string {
-    return this._shouldTransition === true ? transition : '';
+    return false;
+    //TODO
+    //return this._shouldTransition === true ? transition : '';
```
kalman: Could you file a bug and reference it or something so this doesn't get lost.

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+    this.setState({
+      url: URL.createObjectURL(new Blob(parts)),
+    });
+  }
+
+  componentWillUnmount() {
+    this.releaseBlobURL();
+  }
+
+  releaseBlobURL() {
```
kalman: `_releaseBlobURL`

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+    if (this.state.url) {
+      URL.revokeObjectURL(this.state.url);
```
kalman: Do you want to also clear the URL state?

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
 
+  getBlob(props: Props): NomsBlob {
+    let [w, h] = props.fullscreen ?
+      [props.viewport.clientWidth, props.viewport.clientHeight] :
+      [props.gridSize.width, props.gridSize.height];
+    const [_, blob] = props.photo.getBestSize(w, h);
+    return blob;
+  }
+
+  async load(blob: NomsBlob) {
```
kalman: `_load`

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+    const b = await blob.targetValue(this.props.db);
```
kalman: There is no `targetValue` on `NomsBlob`, did this used to be a ref at some point?

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+    const r = b.getReader();
+    const parts = [];
+    while (true) {
+      const n = await r.read();
+
+      // might have changed
+      if (this.state.blob != blob) {
```
kalman: `!==`

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+        return;
+      }
+
+      if (n.value) {
```
kalman: It's an invariant that there is a value if !done, Flow should even know that: https://github.com/attic-labs/noms/blob/master/js/noms/src/async-iterator.js#L11

So it would be better to structure this like
```javascript
while (true)
  const n = await.read();
  if (n.done) break;
  if (this.state.blob !== blob) return;
  parts.push(n.value);
}
```
modulo formatting.

samples/js/photos/src/photo.js
```diff
@@ -8,20 +8,21 @@ import type {NomsPhoto, PhotoSize} from './types.js';
...
+export type SizeEntry = MapEntry<PhotoSize, Blob>;
 
 /**
  * createPhoto asynchronously derives a Photo instance from a NomsPhoto.
  */
 export function createPhoto(path: string, p: NomsPhoto): Promise<Photo> {
   const sizeEntries = [];
-  return p.sizes.forEach((url, size) => {
-    sizeEntries.push([size, url]);
+  return p.resources.forEach((details, size) => {
```
kalman: For the sake of the mobile UI, we should probably have both the sizes and the resources on the photo.

aboodman:
Why can't we expect mobile UI to work this way?

samples/js/photos/src/photo.js
aboodman: The resources map contain a superset of info in sizes map.

kalman:
I was chatting to @arv about this over lunch. The image loading code in RN has a native path, but doing it this way side-steps that. I'm also not sure it supports Blob so it might need to be a data URL.

samples/js/photos/src/photo-grid-item.js
arv: also, missing url field

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +50,96 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+      this.releaseBlobURL();
+    }
+
     const rect = ReactDOM.findDOMNode(this).parentElement.getBoundingClientRect();
     this._parentTop = rect.top;
     this._parentLeft = rect.left;
   }
 
+  getBlob(props: Props): NomsBlob {
+    let [w, h] = props.fullscreen ?
```
arv: Do we need to load both sizes?

arv:
https://github.com/attic-labs/attic/issues/112

samples/js/photos/src/main.js
aboodman: not sure why, but moved due to other comments, so ptal

samples/js/photos/src/main.js
aboodman: had to move other stuff around, but done

samples/js/photos/src/photo-grid-item.js
aboodman: done

samples/js/photos/src/photo-grid-item.js
aboodman: done

samples/js/photos/src/photo-grid-item.js
aboodman: meh, I think either way it is an error

samples/js/photos/src/photo-grid-item.js
aboodman: Is it really going to get lost? There are visibly no transitions.

samples/js/photos/src/photo-grid-item.js
aboodman: This code gets run in both cases.

samples/js/photos/src/photo-grid-item.js
```diff
@@ -24,14 +29,17 @@ type Props = {
...
 type State = {
-  size: PhotoSize,
-  sizeIsBest: boolean,
-  url: string,
+  // The NomsBlob we are currently supposed to be displaying. When this changes, we start loading
+  // it asynchronously and eventually change |url| to match.
+  blob: Ref<NomsBlob> | null,
+
+  // A blob: URL referring to a DOM blob we are displaying. This is populated once |blob| has
+  // loaded.
```
kalman: The 2nd sentence is potentially misleading, because it looks like if the props change to a different blob, the url will stay the same as the original blob until the new one has loaded. Perhaps say "... else null or the previous loaded blob".

Speaking of which I wonder if this should just be string type and empty-string if there is no blob, up to you.

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +51,95 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+      return;
+    }
+    this.setState({
+      blob: nextBlob,
+    });
+    this._load(nextBlob);
+  }
+
+  componentWillUpdate(nextProps: Props, nextState: State) {
+    if (this.state.url !== nextState.url) {
```
kalman: It seems like it would be better to do this at the point the URL state actually updates, i.e. from _load.

samples/js/photos/src/photo-grid-item.js
```diff
@@ -43,53 +51,95 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
...
+    const r = b.getReader();
+    const parts = [];
+    for (;;) {
+      const n = await r.read();
+
+      if (n.done) {
+        break;
+      }
+
+      // might have changed
```
kalman: It also seems possible (or I'm missing it) that the component may have just unloaded, and you should check for that.

samples/js/photos/src/photo-grid-item.js
aboodman: I don't follow this -- the url does stay unchanged until load has completed.

samples/js/photos/src/photo-grid-item.js
kalman: I know. I was thinking because that's the point we explicitly drop the string on the floor, it makes more sense in my mental model. Not a big point.

aboodman:
P to the TAL

samples/js/photos/src/photo-grid-item.js
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Attic Labs, Inc. All rights reserved.
+  // Copyright 2016 Attic Labs, Inc. All rights reserved.
```
aboodman: will fix

samples/js/photos/src/photo-grid-item.js
```diff
@@ -56,6 +58,7 @@ export default class PhotoGridItem extends React.Component<void, Props, State> {
     };
     this._parentTop = 0;
     this._parentLeft = 0;
+    this._unmounted = false;
```
kalman: Shouldn't this be true?

I would use a mounted variable and set it to true on componentDidMount, etc.

samples/js/photos/src/photo-grid-item.js
aboodman: yeah I see why you'd say that. The thing is that then I'd have to have componentMounted just to track that. What I'm doing right now is the bare minimum to support what I actually need.

Issue #2842, opened 2016-11-19
Go Set marshalling


kalman:
Our Go code doesn't support types.Set marshalling. The Go idiom for sets is `map[T]struct{}`, we should do the same, but require a struct annotation to enable it:
```go
type struct Foo {
  ID        string
  Databases map[string]struct{} `noms:"set"`
}
```
For now, we should _only_ support `struct{}` as the value type. It's the idiomatic way to do sets.

@arv

kalman:
For supporting top-level marshalling of Sets we can have a separate function, but I was thinking, this is much less useful than having them embedded inside structs. You should just be iterating over a set if it's top level.

arv:
https://github.com/attic-labs/noms/issues/2376

kalman:
Fixed

PR #2845, opened 2016-11-21
Improve Set marshaling to add encoding support, and decoding to map

kalman:
The only support that marshal has for Set at the moment is decoding to
slice.

### DIFF
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -27,21 +27,28 @@ import (
 // length to zero and then appends each element to the slice. If the Go slice
 // was nil a new slice is created.
 //
-// To unmarshal a Noms list or set into a Go array, Unmarshal decodes Noms list
+// To unmarshal a Noms list into a Go array, Unmarshal decodes Noms list
 // elements into corresponding Go array elements.
 //
 // To unmarshal a Noms map into a Go map, Unmarshal decodes Noms key and values
 // into corresponding Go array elements. If the Go map was nil a new map is
 // created.
 //
-// When unmarshalling onto `interface{}` the following rules are used:
-//  - `types.Bool` -> `bool`
-//  - `types.List` -> `[]T`, where `T` is determined recursively using the same rules.
-//  - `types.Set` -> same as `types.List`
-//  - `types.Map` -> `map[T]V`, where `T` and `V` is determined recursively using the same rules.
-//  - `types.Number` -> `float64`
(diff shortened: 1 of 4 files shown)


kalman:
For #2842.

go/marshal/decode.go
```diff
@@ -386,32 +438,45 @@ func interfaceDecoder(t reflect.Type) decoderFunc {
...
 		return reflect.TypeOf(float64(0))
 	case types.StringKind:
 		return reflect.TypeOf("")
-	case types.ListKind, types.SetKind:
-		et := getGoTypeForNomsType(nt.Desc.(types.CompoundDesc).ElemTypes[0], rt, v)
-		return reflect.SliceOf(et)
+	case types.ListKind:
+		return listType()
+	case types.SetKind:
+		if shouldSetDecodeAsMap(rt, tags) {
```
kalman: This change looked right, but now I realise it isn't. Reverting (just "getGoTypeForNomsType").

kalman: Done, PTAL.

go/marshal/encode.go
```diff
@@ -33,7 +33,8 @@ import (
 //
 // Slices and arrays are encoded as Noms types.List.
 //
-// Maps are encoded as Noms types.Map.
+// Maps are encoded as Noms types.Map, or a types.Set if the value type is
+// struct{} and the field is tagged with `noms:"key"`.
```
arv: tagged with `noms:",set"`?

go/marshal/encode.go
```diff
@@ -277,6 +282,10 @@ type encoderCacheT struct {
 
 var encoderCache = &encoderCacheT{}
 
+// Separate Set encoder cache because the same type with and without the
+// `noms:",set"` tag encode differently (Set vs Map).
+var setEncoderCache = &encoderCacheT{}
```
arv: Maybe the cache could use a "tuple" as the key instead?

kalman: The tuple would probably be (reflect.Type, types.NomsKind). I like the way it is now but if you prefer that I can change it.

go/marshal/encode.go
kalman: Done.

kalman:
PTAL, test added.

and see response to comment.

kalman:
THanks

PR #2846, opened 2016-11-21
Add debug logging to HandleWriteValue

cmasone-attic:
This patch introduces optional debug logging in util/verbose, and adds
some usage of it to HandleWriteValue and the httpBatchStore
SchedulePut code path. It also modifies chunks.DeserializeToChan() so
that callers can better recover from panics in there.

https://github.com/attic-labs/attic/issues/103


cmasone-attic:
PTAL

ghost (approved):
lgtm

PR #2850, opened 2016-11-22
Type erasure for noms :)

aboodman:
cc @arv

### DIFF
--- a/samples/js/picasa/slurp/src/main.js
+++ b/samples/js/picasa/slurp/src/main.js
@@ -8,7 +8,10 @@ import argv from 'yargs';
 import {
   DatasetSpec,
   jsonToNoms,
+  makeStructType,
   newStruct,
+  newStructWithType,
+  valueType,
 } from '@attic/noms';
 import {
   default as fetch,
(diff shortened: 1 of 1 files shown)


samples/js/picasa/slurp/src/main.js
```diff
@@ -158,15 +161,15 @@ async function main(): Promise<void> {
...
 
   const j = (v: any) => jsonToNoms(v);
-  return db.commit(out, newStruct('', {
+  return db.commit(out, newStructWithType(makeStructType('', {value: valueType}), [newStruct('', {
     albums: j(albums),
     faces: j(faces),
     user: j(user),
-  }, {
-    meta: {
+  })]), {
```
aboodman: There was actually a bug here before. No idea why flow didn't catch it.

kalman: I think because Flow (unfortunately) doesn't care about extra arguments to functions, and the meta is optional.

samples/js/picasa/slurp/src/main.js
```diff
@@ -158,15 +161,15 @@ async function main(): Promise<void> {
   // necessary. Consider doing something about that.
 
   const j = (v: any) => jsonToNoms(v);
-  return db.commit(out, newStruct('', {
+  return db.commit(out, newStructWithType(makeStructType('', {value: valueType}), [newStruct('', {
```
kalman: Maybe explain why you need to do this, and in this particular way? Also assigning the `newStruct(...)` portion of this to a variable would make it easier for me to read, but up to you.

samples/js/picasa/slurp/src/main.js
arv: Not sure. If an optional argument is provided it needs to match. Maybe we declared meta as any?

kalman: (updated comment with correct code)

PR #2851, opened 2016-11-23
Add MustParsePath

kalman:


kalman:
Another tiny one.

I find it useful for addressing into some kind of data structure to use a path, like
```go
p := MustParsePath(fmt.Sprintf(`.accounts.byEmail["%s"]`, email))
```

Issue #2849, opened 2016-11-22
Segfault in photo dhash sample


kalman:
See https://atticlabs.slack.com/files/slackbot/F35AKF68Y/hashed_photos.value.bydate_

Not posting logs in case there is sensitive data in there.

kalman:
CC @ehalpern

ehalpern:
Tracked this to a race condition in encode.mapEncoder.  Here's the relevant code: 

```
func mapEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc {
2-->	e := encoderCache.get(t)
	if e != nil {
		return e
	}
	var keyEncoder encoderFunc
	var valueEncoder encoderFunc
	e = func(v reflect.Value) types.Value {
		keys := v.MapKeys()
		kvs := make([]types.Value, 2*len(keys))
		for i, k := range keys {
3---->>>	        kvs[2*i] = keyEncoder(k)
			kvs[2*i+1] = valueEncoder(v.MapIndex(k))
		}
		return types.NewMap(kvs...)
	}

1-->	encoderCache.set(t, e)
	keyEncoder = typeEncoder(t.Key(), parentStructTypes, nomsTags{})
	valueEncoder = typeEncoder(t.Elem(), parentStructTypes, nomsTags{})
	return e
}
```

The first go-routine calls mapEncode and creates an encoder for a particular map type and inserts it into the cache at (1). Note that at this point, keyEncoder and valueEncoder are not set. 

A second go-routine calls mapEncode for the same type and finds it in the cache at (2). If it does this before the first routine has completed initializing keyEncoder, it will segfault at (3).

kalman:
Wow, nice find. @arv

aboodman:
wrongeric

kalman:
No, it was the righteric. This is the marshal code.

aboodman:
Oh, I misread. I thought you said "nice find @arv". But you said "nice
find. @arv". :).

PR #2852, opened 2016-11-23
Fix race conditions in encoder to address photo-dedup segfault

ehalpern:
This change addresses race conditions revealed by a segfault while running photo-dedup in production. 

The root of the problem is described in #2849. The fix is to add use a RWMutex to block encoder function calls until the function is fully initialized. 

I've verified the fix by running photo-dedup under the race detector before and after. The race is detected in the before run and is gone in the after run.

An improvement over this fix would be to implement a fast path using an atomic int (see the sync.Once.Do() implementation for an example). 

-
toward: #2849

### DIFF
--- a/go/marshal/encode.go
+++ b/go/marshal/encode.go
@@ -153,9 +153,9 @@ func float64Encoder(v reflect.Value) types.Value {
 }
 
 func intEncoder(v reflect.Value) types.Value {
-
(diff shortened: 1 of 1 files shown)


ehalpern:
PTAL @arv 
CC: @kalman

PR #2856, opened 2016-11-29
Make splore param parser not strip trailing '=' characters

kalman:
Params must include trailing '=' characters for base64 receipts.


aboodman:
lgtm, thanks

PR #2857, opened 2016-11-29
fix .nomsconfig error: Near line 11

stormasm:
when you run 

noms config

in the directory samples/cli/nomsconfig 

you get the following error message.

error: Near line 11 (last key parsed 'db.temp.url'): Strings can not contain new lines...

line 11 and line 15 need quotes at the end of the line...


aboodman:
Michael, sorry, but in order to keep control of the IP we need to have you sign a contributor agreement in order to accept your patch.

Here is the [individual agreement](http://noms.io/ca_individual.html) and here is the [corporate one](http://noms.io/ca_corporation.html).

aboodman:
(thank you very much for the heads-up though)

PR #2860, opened 2016-11-29
Add `@type` support to paths in Go

aboodman:


go/types/path.go
```diff
@@ -66,13 +66,16 @@ func constructPath(p Path, str string) (Path, error) {
 		}
 		rem = rem[1:]
 
+		// TODO: Is there some prettier way to collapse this w/ the common case for annotations
+		// below?
```
kalman: Would it work to remove the code here, then see below...

go/types/path.go
```diff
@@ -91,6 +94,15 @@ func constructPath(p Path, str string) (Path, error) {
 		p = append(p, part)
 		return constructPath(p, rem)
 
+	case '@':
+		ann, rem := getAnnotation(tail)
+		switch ann {
+		case "type":
+			return constructPath(append(p, TypePart{}), rem)
```
kalman:
```go
// (Implemented by IndexPath and HashIndexPath)
type keyIndexable interface {
  setIntoKey(intoKey bool)
}
...
case "key":
  if len(p) == 0 {
    return Path{}, fmt.Errorf("Cannot @key index at the start of a path")
  }
  lastPart := p[len(p) - 1]
  if ki, ok := lastPart.(keyIndexable); ok {
    ki.setIntoKey(true)
  } else {
    return Path{}, fmt.Errorf("Cannot @key index into %s", lastPart.String())
  }
```
With appropriate other changes of course (e.g. removing the `NewHashIndexIntoKeyPath`-style methods). This would at least be equivalent with the existing code, and address the TODO.

aboodman:
Fab idea. PTAL.

go/types/path.go
```diff
@@ -65,29 +65,12 @@ func constructPath(p Path, str string) (Path, error) {
...
-					intoKey = true
-					rem = rem2
-				}
-			}
-		}
-
 		d.Chk.NotEqual(idx == nil, h.IsEmpty())
 
 		var part PathPart
 		switch {
```
kalman: `switch` here is overkill now, it can be much simpler:
```go
if idx != nil {
  p = append(p, NewIndexPath(idx))
} else {
  p = append(p, NewHashIndexPath(h))
}
```

go/types/path.go
```diff
@@ -65,32 +65,38 @@ func constructPath(p Path, str string) (Path, error) {
...
+		switch ann {
+		case "key":
+			if len(p) == 0 {
+				return Path{}, fmt.Errorf("Cannot use @key annotation at beginning of path")
+			}
+			lastPart := p[len(p)-1]
+			if ki, ok := lastPart.(keyIndexable); ok {
+				p[len(p)-1] = ki.setIntoKey(true).(PathPart)
+				return constructPath(p, rem)
+			} else {
```
kalman: Don't need the else

PR #2867, opened 2016-12-01
Make blob-get print to stdout if there is no file arg

willhite:
I'd like to use blob-get to print string blobs directly to stdout. Here's a proposed change to make that happen.

### DIFF
--- a/samples/go/blob-get/blob_get.go
+++ b/samples/go/blob-get/blob_get.go
@@ -32,7 +32,7 @@ func main() {
 	verbose.RegisterVerboseFlags(flag.CommandLine)
 	flag.Parse(true)
 
-	if len(flag.Args()) != 1 && len(flag.Args()) != 2 {
+	if flag.NArg() != 1 && flag.NArg() != 2 {
 		d.CheckError(errors.New("expected dataset and optional file flag"))
 	}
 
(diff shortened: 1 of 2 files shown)


samples/go/blob-get/blob_get.go
```diff
@@ -51,24 +51,32 @@ func main() {
...
 
-	// Note: overwrites any existing file.
-	file, err := os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE, 0644)
-	d.CheckErrorNoUsage(err)
-	defer file.Close()
-
 	expected := humanize.Bytes(blob.Len())
 	start := time.Now()
 
 	progReader := progressreader.New(blob.Reader(), func(seen uint64) {
```
kalman: I think it would be better to make `progReader` just `blob.Reader()` in the case that you're writing to stdout.

Like:
```go
file := os.Stdout
blobReader := blob.Reader()

if filePath != "" {
  file := os.OpenFile(...)
  blobReader = progressReader.New(blobReader, func(seen uint64) { ... })
}

io.Copy(file, blobReader)

File: samples/go/blob-get/blob_get.go [hunk 2]
```diff
@@ -24,16 +24,16 @@ import (
...
 		flag.PrintDefaults()
 	}
 
 	spec.RegisterDatabaseFlags(flag.CommandLine)
 	verbose.RegisterVerboseFlags(flag.CommandLine)
 	flag.Parse(true)
 
-	if len(flag.Args()) != 2 {
-		d.CheckError(errors.New("expected dataset and file flags"))
+	if len(flag.Args()) != 1 && len(flag.Args()) != 2 {
```
kalman: I think there is `flag.NArgs()`

samples/go/blob-get/blob_get.go
```diff
@@ -51,24 +51,32 @@ func main() {
 	}
 
 	filePath := flag.Arg(1)
+	var file *os.File
+	showProgress := true
 	if filePath == "" {
-		d.CheckErrorNoUsage(errors.New("Empty file path"))
+		file = os.Stdout
```
kalman: The only real problem here is non-printable characters. Best case they look super weird in bash, worst case they're control characters that do weird things.

I think you might want `unicode.IsGraphic`: https://golang.org/pkg/unicode/#IsGraphic, but it's not that simple, because io.Writer operates in bytes not runes. I had a look and didn't find anything pre-built that does this.

Hacky idea:
```go
// (In the case of writing to stdout... and fixing threading issues...
// and fitting into any other code... and fixing compilation...)
rdr := bufio.NewReader(buffer.Reader())

go func() {
  wtr := bufio.NewWriter(file)
  defer wtr.Close()

  for {
    r, _, err := rdr.ReadRune()
    if err != nil {
      break
    }
    if !unicode.IsGraphic(r) {
      r = '~'
    }
    wtr.WriteRune(r)
  }
}()
```

kalman: Though I think you'd only want to do this if stdout is TTY. It's fun to be able to go `blob-get foobar::mypath > file.txt`.

samples/go/blob-get/blob_get.go
willhite: Changed.

samples/go/blob-get/blob_get.go
willhite: We decided that, like cat, this is really a user responsibility.

samples/go/blob-get/blob_get.go
willhite: PTAL

willhite:
PTAL

Issue #2877, opened 2016-12-01
Abstract Manifest & Table such that durable media need not be local (i.e. File). Implement S3 Provider


rafael-atticlabs:
For example, for storage in AWS S3 or GCE Cloud Storage

PR #2886, opened 2016-12-01
Introduce more fakes for testing NomsBlockStore

cmasone-attic:
Added a "constructor" for NomsBlockStore that takes a tableManager
and a manifest instance, so that fakes can be injected for testing.
This also allows me to break the dependency on on-disk manifests in
root_tracker_test.go

Towards issue #2877

### DIFF
--- a/go/nbs/root_tracker_test.go
+++ b/go/nbs/root_tracker_test.go
@@ -5,30 +5,18 @@
 package nbs
 
 import (
-	"io/ioutil"
-	"os"
-	"path/filepath"
-	"strings"
+	"encoding/binary"
 	"testing"
 
 	"github.com/attic-labs/noms/go/chunks"
 	"github.com/attic-labs/noms/go/constants"
-	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
 	"github.com/attic-labs/testify/assert"
 )
 
-func makeStoreInTempDir(t *testing.T) (dir string, store *NomsBlockStore) {
-	dir, err := ioutil.TempDir("", "")
-	assert.NoError(t, err)
-	store = NewBlockStore(dir, defaultMemTableSize)
-	return
-}
-
 func TestChunkStoreZeroValue(t *testing.T) {
 	assert := assert.New(t)
-	dir, store := makeStoreInTempDir(t)
-	defer os.RemoveAll(dir)
+	_, _, store := makeStoreWithFakes(t)
 	defer store.Close()
 
 	// No manifest file gets written until the first call to UpdateRoot(). Prior to that, Root() will simply return hash.Hash{}.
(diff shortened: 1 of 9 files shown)


cmasone-attic:
PTAL

cmasone-attic:
sigh. Now the history is all screwed up. I need you to land https://github.com/attic-labs/noms/pull/2888 so I can rebase properly and re-upload

cmasone-attic:
so fresh, so clean. PTAL

go/nbs/table_manager.go
```diff
@@ -5,18 +5,18 @@
 package nbs
 
 type tableManager interface {
-	compact(mt *memTable, haver chunkReader) (name addr, chunkCount uint32)
-	open(name addr, chunkCount uint32) chunkSource
+	Compact(mt *memTable, haver chunkReader) (name addr, chunkCount uint32)
```
rafael-atticlabs: the `tableManager` interface is starting to feel a little weird. it's fine for now, but it smells a bit bad that the `fileTableManager`s two methods just immediately call into either a static method or a specific implementation of another interface.

go/nbs/test/manifest_clobber.go
```diff
@@ -29,7 +27,7 @@ func main() {
 	// lock released by closing l.
 	err = unix.Flock(int(l.Fd()), unix.LOCK_EX|unix.LOCK_NB)
 	if err != nil {
-		if !*takeLock && err == unix.EWOULDBLOCK {
+		if err == unix.EWOULDBLOCK {
```
rafael-atticlabs: nit:

```
if err == unix.EWOULDBLOCK {
  return
} 

if err != nil {
  log.Fatalln(err)
}
```

go/nbs/root_tracker_test.go
```diff
@@ -112,28 +137,75 @@ func assertDataInStore(slices [][]byte, store chunks.ChunkSource, assert *assert
 	}
 }
 
-func TestChunkStoreManifestFirstWriteByOtherProcess(t *testing.T) {
-	assert := assert.New(t)
-	dir, err := ioutil.TempDir(os.TempDir(), "")
-	assert.NoError(err)
-	defer os.RemoveAll(dir)
+type fakeManifest struct {
```
rafael-atticlabs: it's a little unclear to me how to think about the semantics of `fakeManifest`. `fakeChunkSource` is short enough that it's obviously just a chunk source which will report its name and chunk count and panic on any other interaction.

can you put a comment here explaining (at a high-level) the behavior of fakeManifest, so it's easier to think about the behavior of the tests which use this?

cmasone-attic: The behavior of fakeManifest is intended to be the same as fileManifest, so you can reason about it as though it were a fileManifest. I've added some big comments to the methods of the `manifest` interface to explain the contract there. This is intended to be a fake that fulfills that contract.

cmasone-attic:
B2U

cmasone-attic:
Closed in favor of #2905

PR #2888, opened 2016-12-02
Minor rework of hash.Hash API

rafael-atticlabs:
Define the hash.Hash type to *be* a 20-byte array, rather than embed one. Also remove the `DigestSlice` function and prefer just directly taking a slice of the value.

### DIFF
--- a/go/chunks/chunk.go
+++ b/go/chunks/chunk.go
@@ -34,7 +34,7 @@ func (c Chunk) IsEmpty() bool {
 
 // NewChunk creates a new Chunk backed by data. This means that the returned Chunk has ownership of this slice of memory.
 func NewChunk(data []byte) Chunk {
-	r := hash.FromData(data)
+	r := hash.Of(data)
 	return Chunk{r, data}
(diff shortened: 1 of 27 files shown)


rafael-atticlabs:
Note that this also changes some usages of `len(h.DigestSlice())` to be `hash.ByteLen`

cmasone-attic:
LGTM

cmasone-attic:
Heh, there's definite evidence here of how our intentions for Hash changed over time...some of this code hails from a time when we thought it made sense for there to be different kinds of hashes coexisting, potentially with different algorithms and lengths. So it was more future-proof to take the `len(h.DigestSlice())` instead of just using hash.ByteLen. We understand better, now :-P

rafael-atticlabs:
PTAL last minor API naming change. + @arv for js

cmasone-attic:
go still LGTM, but you apparently need to rebase to get the l8test package.json?

arv:
LGTM

PR #2890, opened 2016-12-02
Update Noms JS SDK to use Flow 0.35

arv:


arv:
Keeping the version in sync makes things smoother.

Issue #2889, opened 2016-12-02
Support Go marshaling for *types.Type


ehalpern:
Problem: 
 - Instances of `types.Type` are referred to by reference (i.e, `*types.Type`)
 - Go->Noms marshaling does not (currently) support marshaling from Go struct pointers to Noms values
 - This is makes it impossible to use Go marshaling to generate values containing type.Types.

The ideal solution would be to fix Go marshaling to handle pointers in the way that JSON marshaling does:
 - Marshaling a pointer to a Go value is produces the Noms value of the referent
 - Marshaling a nil pointer to a Go value produces a Nil Noms value
 - Unmarshaling a Noms value to a Go pointer produces a pointer to the Go value
 - Absence of a Noms value produces a nil pointer

See [json.Marshal](https://golang.org/pkg/encoding/json/#Marshal) and [json.Unmarshal](https://golang.org/pkg/encoding/json/#Unmarshal)

A useful first step would be to handle pointers to noms types (in particular `*types.Type`).

ehalpern:
CC: @arv

PR #2892, opened 2016-12-02
Support marshaling from and unmarshaling to *types.Type

ehalpern:
Marshaling from `*types.Type` produces a noms type:
```
type S struct { Type *types.Type }
go := S{ types.String }
noms := types.NewStruct("S", types.StructData{"type": types.String})
v, err := marshal.Marshal(go) 
assert.Equals(t, noms, v)
```

Unmarshaling to `*types.Type` produces a pointer to an instance of the value:
```
type S struct { Type *types.Type }
noms := types.NewStruct("S", types.StructData{"type": types.String})
go :=  S{ types.String }

var s S
err := marshal.Unmarshal(v, &s) 
assert.Equals(t, go, s)
```

fixes: #2889

### DIFF
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -48,6 +48,7 @@ import (
 //    same rules.
 //  - types.Number -> float64
 //  - types.String -> string
+//  - *types.Type -> *types.Type
 //  - types.Union -> interface
 //  - Everything else an error
 //
@@ -149,7 +150,7 @@ func typeDecoder(t reflect.Type, tags nomsTags) decoderFunc {
 		if t.Implements(nomsValueInterface) {
 			return nomsValueDecoder
 		}
-		panic(&UnsupportedTypeError{Type: t})
+		fallthrough
 	default:
 		panic(&UnsupportedTypeError{Type: t})
 	}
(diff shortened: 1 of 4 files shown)


ehalpern:
cc: @willhite

arv:
In the PR description: types.String -> types.StringType

go/marshal/decode.go
```diff
@@ -144,6 +144,12 @@ func typeDecoder(t reflect.Type, tags nomsTags) decoderFunc {
 			return mapFromSetDecoder(t)
 		}
 		return mapDecoder(t, tags)
+	case reflect.Ptr:
+		// Allow implementations of types.Value (like *types.Type)
+		if t.Implements(nomsValueInterface) {
+			return nomsValueDecoder
+		}
+		panic(&UnsupportedTypeError{Type: t})
```
arv: you can also use a `fallthrough` statement

aboodman:
Documentation at top of encode.go and decode.go needs updating.

ehalpern:
Addressed review suggestions. 
@aboodman: Thx for the catch on godocs. I updated decode, but left encode as is since it already indicates that noms values are marshaled as is.

Issue #2873, opened 2016-12-01
Allow NBS tables to have more than 2^32 chunks


rafael-atticlabs:
Right now, order is a uint32, which effectively limits the number of chunks in a table to 2^32 (or roughly 17TB assuming average chunk size of 4k).

Not a pressing problem, but it'd be nice to have tables be sized arbitrarily.

PR #2894, opened 2016-12-02
uint32ChunkCount

rafael-atticlabs:
Encode chunk counts consistently as uint32 until #2873 is addressed. This also fixes an error in passing chunkCounts resulting from compaction that don't account for dropped (duplicate) chunks.


cmasone-attic:
LGTM

PR #2895, opened 2016-12-02
Give Flow a helping hand

arv:
Flow didn't know which type to pick for addEventListener.


arv:
TBR... Letting Jenkins do its thing before landing.

PR #2897, opened 2016-12-02
Try to make path spec that references bare dataset default to .value

aboodman:
This got super gnarly because there are so many different variants of specs and paths.

After hacking at this, I think to be precise the behavior we want is:

absolute paths (as opposed to specs, and relative paths) that point to a bare dataset should default to `.value`.

specs would then inherit the behavior (because they use AbsolutePath)

and _dataset specs_ would be unaffected.

### DIFF
--- a/go/spec/absolute_path.go
+++ b/go/spec/absolute_path.go
@@ -8,11 +8,14 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/attic-labs/noms/go/datas"
 	"github.com/attic-labs/noms/go/hash"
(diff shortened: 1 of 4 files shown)


kalman:
Got it. I think AbsolutePath might be the wrong place to put this then.

Check out https://github.com/attic-labs/noms/pull/2896 - I prefer doing the annotation detection/parsing inside Path. However,
* AbsolutePath seems the wrong place to interpret it.
* The code in that PR is entirely wrong anyway.

A good way to solve this might be to:
* Put the `@commit` at the end, not the start (like I did in that CL).
* Change spec.go to resolve the Commit to its .value if it's a `ForPath`-constructed Spec.

PR #2900, opened 2016-12-02
marshal.Unmarshal(): introduce omitempty and original support

### LINKED ISSUE #2795: Support preservation of unnamed field values in go struct marshaling
Problem: 

When unmarshaling a noms struct into a go struct, fields omitted from the go struct but present in the noms struct are ignored. If the go struct is then marshaled back to noms, the ignored fields are lost. 

A concrete example of this problem can be seen in the photo-dhash job which adds a dhash field to every photo object. The noms struct looks something like:

```
struct Photo {
  id: String,
  title: String 
  sizes: Map<struct, {...}, String> 
  datePublished...
  dateTaken...
  ...
}
```

The go struct looks like: 

```
struct Photo {
  Id string,
  Sizes map[struct{...}]string, 
  Dhash String `noms:"-"`
}
```

Proposed Solution:

- Define a naming convention or annotation that marks a field to hold the original noms value.
- On unmarshal, the field is set to the value read
- On marshal, the other go struct fields are merged with the original value

The new go struct would look something like this:

```
struct Photo {
  Id string,
  Sizes map[struct{...}]string, 
  Dhash String `noms:"-"`
  Original types.Value `noms:"+"` 
}
```

### LINKED ISSUE #2796: Support optional fields in go struct marshaling
Problem: 

Suppose you want to add a field to a noms struct if it doesn't already exist using go struct marshaling. If you don't annotate the field, unmarshaling from noms -> go will fail in the 
case that the field doesn't exist in noms. If you ignore the field using `noms:"-"`, the field won't be set if it exists in noms.

Proposal: 

Introduce a new optional annotation `noms:"[noms-field-name]?"`. When unmarshaling from noms to go, an optional go field will be set if the noms field exists. When marshaling from go to noms, an optional go field will be marshaled to the noms values if its set.

aboodman:
marshal.Unmarshal(): introduce omitempty and original support

Fixes #2795
Fixes #2796

### DIFF
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -230,9 +230,11 @@ func (c *decoderCacheT) set(t reflect.Type, d decoderFunc) {
 }
 
 type decField struct {
-	name    string
-	decoder decoderFunc
-	index   int
+	name      string
+	decoder   decoderFunc
+	index     int
+	omitEmpty bool
+	original  bool
 }
 
 func structDecoder(t reflect.Type) decoderFunc {
@@ -256,9 +258,11 @@ func structDecoder(t reflect.Type) decoderFunc {
 		}
 
 		fields = append(fields, decField{
-			name:    tags.name,
-			decoder: typeDecoder(f.Type, tags),
-			index:   i,
+			name:      tags.name,
+			decoder:   typeDecoder(f.Type, tags),
+			index:     i,
+			omitEmpty: tags.omitEmpty,
+			original:  tags.original,
 		})
 	}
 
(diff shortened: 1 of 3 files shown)


go/marshal/decode_test.go
```diff
@@ -750,3 +750,66 @@ func TestDecodeSetWrongMapType(t *testing.T) {
 	assert.Error(err)
 	assert.Equal(`Cannot unmarshal Map<Number, struct {}> into Go value of type map[int]struct {}, field has "set" tag`, err.Error())
 }
+
+func TestDecodeOmitEmpty(t *testing.T) {
+	assert := assert.New(t)
+
+	type S struct {
+		Foo int `noms:",omitempty"`
```
arv: Why the nested struct in this test?

go/marshal/decode.go
```diff
@@ -270,11 +274,19 @@ func structDecoder(t reflect.Type) decoderFunc {
 
 		for _, f := range fields {
 			sf := rv.Field(f.index)
+			if f.original {
+				if sf.Type() != reflect.TypeOf(s) {
+					panic(&UnmarshalTypeMismatchError{v, rv.Type(), ", field with tag \"original\" must have type String"})
```
arv: Why does it have to have type String?

go/marshal/decode_test.go
aboodman: I just wanted to ensure that nested structs were handled properly because that was my use case. I guess given the structure of the code there's no special reason these would fail, so I can remove this test if you prefer.

aboodman:
done, ptal

go/marshal/decode.go
```diff
@@ -276,7 +280,7 @@ func structDecoder(t reflect.Type) decoderFunc {
 			sf := rv.Field(f.index)
 			if f.original {
 				if sf.Type() != reflect.TypeOf(s) {
-					panic(&UnmarshalTypeMismatchError{v, rv.Type(), ", field with tag \"original\" must have type String"})
+					panic(&UnmarshalTypeMismatchError{v, rv.Type(), ", field with tag \"original\" must have type Struct"})
```
arv: Ah, I should have been able to guess that that was a typo.

PR #2901, opened 2016-12-03
Make paths capable of using index to address set elements

willhite:
FYI only, not for review. I haven't looked at this in a while but it was working in go at one point.


aboodman:
Can I close this?

PR #2902, opened 2016-12-03
photo-index: output PhotoGroup instead of Photo

aboodman:
photo-index: output PhotoGroup instead of Photo.

This will allow UIs to display duplicate photos. Also, simplify design by creating groups even when the group has one item.

This change also maintains the old behavior via a flag so that the UI can keep working until it is updated.

### DIFF
--- a/samples/go/photo-index/main.go
+++ b/samples/go/photo-index/main.go
@@ -9,13 +9,11 @@ import (
 	"math"
 	"os"
 	"path"
-	"sync"
 	"time"
 
 	"github.com/attic-labs/noms/go/config"
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/datas"
-	"github.com/attic-labs/noms/go/hash"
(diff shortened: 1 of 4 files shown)


samples/go/photo-index/main.go
```diff
@@ -32,23 +30,39 @@ func main() {
...
 
+func (d Date) IsEmpty() bool {
+	return d.NsSinceEpoch == float64(0)
+}
+
 func index() (win bool) {
 	var dbStr = flag.String("db", "", "input database spec")
-	var groupsStr = flag.String("groups", "", "path within db to look for PhotoGroup structs")
 	var outDSStr = flag.String("out-ds", "", "output dataset to write to - if empty, defaults to input dataset")
+	var indexPhotos = flag.Bool("index-photos", false, "index Photo (rather than PhotoGroup)")
```
kalman: Maybe `covers` would be more descriptive.

PR #2903, opened 2016-12-03
Implement Set.At and Map.At in Go

kalman:
These get the set/map element at a specific index.
I haven't implemented it in JS yet because the JS code has no method to
create a cursor at an index. This exists in Go because a refactor was
done a few months ago to add it, but it hasn't been ported to JS.


aboodman:
Lgtm

PR #2906, opened 2016-12-04
Implement Set.At + Add position indices to Path

kalman:
This is https://github.com/attic-labs/noms/pull/2903 + a subsequent patch to add indices to Path.

If that other patch goes in first, I'll rebase, otherwise I'll squash. @willhite

### DIFF
--- a/go/types/path.go
+++ b/go/types/path.go
@@ -74,17 +74,38 @@ func constructPath(p Path, str string) (Path, error) {
 		if !strings.HasPrefix(rem, "]") {
 			return Path{}, errors.New("[ is missing closing ]")
 		}
-		rem = rem[1:]
-		d.Chk.NotEqual(idx == nil, h.IsEmpty())
+		d.PanicIfTrue(idx == nil && h.IsEmpty())
+		d.PanicIfTrue(idx != nil && !h.IsEmpty())
 
-		var part PathPart
 		if idx != nil {
-			part = NewIndexPath(idx)
+			p = append(p, NewIndexPath(idx))
 		} else {
-			part = NewHashIndexPath(h)
+			p = append(p, NewHashIndexPath(h))
 		}
-		p = append(p, part)
-		return constructPath(p, rem)
+		return constructPath(p, rem[1:])
+
+	case '{':
+		if len(tail) == 0 {
+			return Path{}, errors.New("Path ends in {")
+		}
+
+		closingIdx := strings.Index(tail, "}")
+		if closingIdx == -1 {
+			return Path{}, errors.New("{ is missing closing }")
+		}
+
+		pos := tail[:closingIdx]
+		if len(pos) == 0 {
+			return Path{}, fmt.Errorf("Empty position value")
+		}
+
+		idx, err := strconv.ParseInt(pos, 10, 64)
+		if err != nil {
+			return Path{}, fmt.Errorf("Invalid position: %s", pos)
+		}
+
+		p = append(p, NewPositionPath(idx))
+		return constructPath(p, tail[closingIdx+1:])
 
 	case '@':
 		ann, rem := getAnnotation(tail)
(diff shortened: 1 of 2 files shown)


go/types/path_test.go
```diff
@@ -62,14 +62,9 @@ func TestPathIndex(t *testing.T) {
...
 	var v Value
-	resolvesTo := func(exp, val Value, str string) {
-		// Indices resolve to |exp|.
-		assertResolvesTo(assert, exp, v, str)
-		// Keys resolve to themselves.
-		if exp != nil {
-			exp = val
-		}
-		assertResolvesTo(assert, exp, v, str+"@key")
+	resolvesTo := func(expVal, expKey Value, str string) {
```
kalman: Test cleanup, I preferred the way I wrote them just now, so I changed the existing ones to match.

kalman:
I slept on this and... don't hate me... I'm thinking that a prefix operator inside a square bracket would actually be more consistent with how we've/you've designed paths so far. For example:
* `[42]` to index by key.
* `[#abde]` to index by hash.
* `[&42]` to index by position (`&` chosen arbitrarily).

I.e. `[...]` means "look up index in collection" and the operator determines how the lookup occurs: by key, by hash, or by position.

I got onto this thinking by realising that we have `IndexPath`, `HashIndexPath`, and `PositionPath`; but better naming would be `KeyIndexPath`, `HashIndexPath`, and `PositionIndexPath`.

Sorry. I'll mull some more.

go/types/path.go
```diff
@@ -405,3 +433,74 @@ func getAnnotation(str string) (ann, rem string) {
...
+	var absIndex uint64
+	if col, ok := v.(Collection); !ok {
+		return nil
+	} else if absIndex, ok = getAbsoluteIndex(col, pp.Index); !ok {
+		return nil
+	}
+
+	switch v := v.(type) {
+	case List:
+		if pp.IntoKey {
```
kalman: For the record, your comment:

> I don't think supporting @key here is needed.

got lost in the rebase, sorry. My response was along the lines of

> I agree it's not that useful, and rather bizarre, but it is consistent with the key indexing.

I'll probably remove it.

aboodman:
OK, here's one last thought:

foo@pos(42)
foo@pos(-8)

The rationale would be that indexing into a map or set by position kind of
falls into the bucket of things like @type and @key -- things that need to
be possible but which are pretty rare and thus don't warrant dedicated
syntax.

===

FWIW, the way I thought about [...] was different: it was that [...] always
indexes by key. The special chars at beginning of some keys indicate the
type of the key, not that the mechanism of indexing is different:

[42] index by integer key
[88.8] index by real key
["foo"] index by string key
[#...] index by hash key

kalman:
> foo@pos(42)
> foo@pos(-8)

One problem with that is it means writing parsing logic for arguments to annotations. This does seem pretty useful, though this is the first instance of it. Note that it takes `(` and `)` out of the picture for use anywhere else basically.

> indexing into a map or set by position ...
> things that need to
> be possible but which are pretty rare and thus don't warrant dedicated
> syntax

I'm not so sure. I've wanted this way more often than I've wanted `@key` (never) and `@type` (never, though it's new so maybe I will). Starting out with at as an annotation could be a sensible interim step (modulo the extra effort in having parameterized annotations) but instinctively I feel indexing by position is actually pretty handy?

> [#...] index by hash key

Nit... it's not really indexing by hash key in the same way that `".."` is index by string key etc: it's "finding the Value which has the hash `#...`", a grey area on which my brain fell the other side of with respect to `[&42]` being an index by position. But now my brain is not so sure.

aboodman:
On Mon, Dec 5, 2016 at 10:33 AM, Ben Kalman <notifications@github.com>
wrote:

> Note that it takes ( and ) out of the picture for use anywhere else
> basically.
>
> I don't think so because we can know the annotations that support
arguments.

> indexing into a map or set by position ...
> things that need to
> be possible but which are pretty rare and thus don't warrant dedicated
> syntax
>
> I'm not so sure. I've wanted this way more often than I've wanted @key
> (never) and @type (never, though it's new so maybe I will).
>
Dan and I use @type frequently.

> Starting out with at as an annotation could be a sensible interim step
> (modulo the extra effort in having parameterized annotations) but
> instinctively I feel indexing by position is actually pretty handy?
>
> [#...] index by hash key
>
> Nit... it's not really indexing by hash key in the same way that ".." is
> index by string key etc: it's "finding the Value which has the hash #...",
> a grey area on which my brain fell the other side of with respect to [&42]
> being an index by position. But now my brain is not so sure.
>
The reason I like @pos is because we don't have to grab another piece of
syntax (presuming you agree with my argument about parens above).

PR #2907, opened 2016-12-05
Verify chunks using suffix index not computing address from data.

### LINKED ISSUE #2875: (NBS) Verifing Gets() by computing addr may be too expensive
Right now we locate a chunk by matching an address prefix. This creates the possibility of false positives. We verify the correct chunk by computing the full address from the content.

It looks as though we can compute addresses at about the rate of 300 MB/s on a single i5 core. This is roughly the raw read rate of a single ssd, so it effectively cuts the potential read rate of a single core about in half.

The obvious thing to do to avoid this compute expense is to check some previously encoded address suffix - either inline in the chunk, or in the suffix index. If we did that, we'll likely want to add a CRC32 in each chunk record as an integrity check (computing the address was serving that purpose).

rafael-atticlabs:
Revert to verifying chunks using the suffix index. Replace the inline 4-byte suffix used as integrity check with a more standard and efficient CRC32.

Fixes #2875

### DIFF
--- a/go/nbs/table.go
+++ b/go/nbs/table.go
@@ -23,9 +23,9 @@ import (
    +----------------+----------------+-----+----------------+-------+--------+
 
    Chunk Record:
-   +--------------------+-----------------------+
+   +---------------------------+----------------+
    | (Chunk Length) Chunk Data | (Uint32) CRC32 |
-   +--------------------+-----------------------+
+   +---------------------------+----------------+
 
      -Address suffix is the 4 least-significant bytes of the Chunk's address. Used (e.g. in place
       of CRC32) as a checksum and a filter against false positive reads costing more than one IOP.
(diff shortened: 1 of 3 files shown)


go/nbs/table.go
```diff
@@ -22,9 +23,9 @@ import (
    +----------------+----------------+-----+----------------+-------+--------+
 
    Chunk Record:
-   +--------------------+---------------------------+
-   | (4) Address suffix | (Chunk Length) Chunk Data |
-   +--------------------+---------------------------+
+   +--------------------+-----------------------+
```
cmasone-attic: nitnitnit: fix the placement of the middle `+` in this line and below

go/nbs/table.go
```diff
@@ -117,6 +118,13 @@ const (
 	maxChunkSize       uint64 = 0xffffffff // Snappy won't compress slices bigger than this
 )
 
+var crcTable = crc32.MakeTable(crc32.Castagnoli)
+
+// CORRECT?
```
cmasone-attic: UNSURE

go/nbs/table_reader.go
```diff
@@ -326,23 +334,13 @@ func (tr tableReader) getMany(reqs []getRecord) (remaining bool) {
...
-	chksum := binary.BigEndian.Uint32(buff)
-	if chksum != h.Checksum() {
-		return nil // false positive
-	}
-	buff = buff[checksumSize:]
-
-	// data
-	data, err := snappy.Decode(nil, buff)
+	dataLen := uint64(len(buff)) - checksumSize
+	data, err := snappy.Decode(nil, buff[:uint64(len(buff))-uint32Size])
```
cmasone-attic: shouldn't this be `buff[:dataLen]`?

go/nbs/table.go
rafael-atticlabs: done

go/nbs/table.go
rafael-atticlabs: assuming so. i put this here because the snappy library does some weird bit shifting of the result (for internal use).

go/nbs/table_reader.go
rafael-atticlabs: done

PR #2910, opened 2016-12-05
Add path @at for positional indexing, and negative indices

kalman:
This lets you do foo.bar@at(n) to get the nth element of a list, set, or
map (for lists, this is equivalent to foo.bar[n]). This patch also adds
support for negative indices to @at(-n) and [-n] to get the nth element
relative to the back of collections.

### DIFF
--- a/go/types/path.go
+++ b/go/types/path.go
@@ -17,7 +17,10 @@ import (
 	"github.com/attic-labs/noms/go/hash"
 )
 
-var annotationRe = regexp.MustCompile("^([a-z]+)")
+// For an annotation like @type, 1st capture group is the annotation.
+// For @at(42), 1st capture group is the annotation and 3rd is the parameter.
+// Note, @at() is valid under this regexp, code should deal with the error.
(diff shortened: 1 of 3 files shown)


go/types/path.go
```diff
@@ -17,7 +17,10 @@ import (
 	"github.com/attic-labs/noms/go/hash"
 )
 
-var annotationRe = regexp.MustCompile("^([a-z]+)")
+// For an annotation like @type, 1st capture group is the annotation.
+// For @at(42), 1st capture group is the annotation and 3rd is the parameter.
+// Note, @at() is valid under this regexp, code should deal with the error.
```
aboodman: OOC, why not make `at()` invalid?

kalman: So I can distinguish between `@at` and `@at()` for constructing error messages, i.e. "missing annotation" vs "not a number".

aboodman:
Makes sense. LGTM, modulo doc.

aboodman:
LGTM

Issue #2911, opened 2016-12-06
JS walk(...) fails when traversing Type 


ehalpern:
The walk function wrongly treats values of type Type as a Struct and fails. 

Fix on the way.

PR #2912, opened 2016-12-06
Fix JS walk() handling for values of type Type

ehalpern:
fixes: #2911

FYI: @kalman, @willhite

### DIFF
--- a/js/noms/src/walk-test.js
+++ b/js/noms/src/walk-test.js
@@ -21,6 +21,7 @@ import Blob from './blob.js';
 import List from './list.js';
 import Map from './map.js';
 import NomsSet from './set.js'; // namespace collision with JS Set
+import Struct from './struct.js';
 import walk from './walk.js';
 import type Value from './value.js';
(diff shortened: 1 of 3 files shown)


js/noms/src/walk-test.js
```diff
@@ -21,6 +21,7 @@ import Blob from './blob.js';
 import List from './list.js';
 import Map from './map.js';
 import NomsSet from './set.js'; // namespace collision with JS Set
+import Struct from './struct.js';
```
arv: not used

js/noms/src/walk.js
```diff
@@ -44,7 +45,7 @@ export default async function walk(v: Value, ds: Database, cb: walkCb): Promise<
       return;
   }
 
-  if (v instanceof Blob) {
+  if (v instanceof Blob || v instanceof Type) {
```
arv: We want to descend into typea.

arv:
I could not figure out how to add more commits to someone else's PR. I opened a new one: #2913 29

PR #2913, opened 2016-12-06
Fix JS walk handling for values of type Type

### LINKED ISSUE #2911: JS walk(...) fails when traversing Type

arv:
Fixes #2911

### DIFF
--- a/go/types/type.go
+++ b/go/types/type.go
@@ -91,7 +91,7 @@ func (t *Type) WalkValues(cb ValueCallback) {
 		desc.IterFields(func(name string, t *Type) {
 			cb(t)
 		})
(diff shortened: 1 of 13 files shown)


arv:
Replaces #2912

arv:
@zcstarr Maybe we should just add walkValues to the JS value class?

ehalpern:
LGTM

Agreed that it would nicer to handle this with walkValues on the value.

zcstarr:
This makes sense to me

arv:
PTAL

ehalpern:
Much better! LGTM @arv.

PR #2922, opened 2016-12-08
Change NBS s3TablePersister to use Multipart upload

cmasone-attic:
Instead of putting an entire table to S3 in a single request, split it
into 5MB parts (the smallest allowable) and send all the parts in
parallel.

### DIFF
--- a/go/nbs/table_set.go
+++ b/go/nbs/table_set.go
@@ -10,14 +10,18 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
+	"sync"
 
 	"github.com/attic-labs/noms/go/d"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/s3"
 )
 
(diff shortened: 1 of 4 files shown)


cmasone-attic:
PTAL

go/nbs/table_set.go
```diff
@@ -81,27 +85,141 @@ type tablePersister interface {
...
+			Body:       bytes.NewReader(data[start:end]),
+		})
+		if err != nil {
+			failed <- err
+			return
+		}
+
+		select {
+		case sent <- s3UploadedPart{int64(partNum), *result.ETag}:
+		case <-done:
```
rafael-atticlabs: is the point of reading from `done` here just to ensure that this goroutine exits (and doesn't sitting blocking on the send to `sent` forever?

rafael-atticlabs: maybe attempt read off of `done` above in to avoid uploading the part in the failure case

ghost (approved):
LGTM

go/nbs/table_set.go

PR #2923, opened 2016-12-09
JS: Do not validate struct type when reading a chunk

arv:
When we read a chunk and create structs we were validating that the
struct was of the type it claimed to be.

This now, no longer does that validation, which matches Go.


arv:
In the timeline for the app readStruct went from 33% to 22% of the total time.

aboodman:
ok

PR #2933, opened 2016-12-12
Add ChunkSource.GetMany(); RemoteBatchStore getRefs uses GetMany()

rafael-atticlabs:
Add `GetMany()`, which most ChunkStores implement by repeated calls to their own `Get()`, but creates the opportunity for stores to optimize reads of larger blocks of potentially sequential chunks (e.g. NBS).

Add RemoteBatchStore `getRefs` endpoint support for calling `GetMany()` rather than `Get()`

Remove `ReadThroughChunkStore` which was dead code.


go/datas/local_batch_store.go
```diff
@@ -44,6 +44,14 @@ func (lbs *localBatchStore) Get(h hash.Hash) chunks.Chunk {
 	return lbs.cs.Get(h)
 }
 
+func (lbs *localBatchStore) GetMany(hashes []hash.Hash) (batch []chunks.Chunk) {
+	batch = make([]chunks.Chunk, len(hashes))
+	for i, h := range hashes {
+		batch[i] = lbs.Get(h)
```
cmasone-attic: you could just return `lbs.cs.GetMany(hashes)`, right?

rafael-atticlabs: Excellent point, good sir.

PR #2934, opened 2016-12-12
Make compaction async

### LINKED ISSUE #2879: Allow compaction of NBS memtable to be async
That is, don't block new writes on compaction

cmasone-attic:
Introduce a 'compactingChunkStore', which knows how
to compact itself in the background. It satisfies
get/has requests from an in-memory table until compaction
is complete. Once compaction is done, it destroys the
in-memory table and switches over to using solely the
persistent table.

Fixes #2879

### DIFF
--- a/go/nbs/compacting_chunk_source.go
+++ b/go/nbs/compacting_chunk_source.go
@@ -1,0 +1,120 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package nbs
+
+import (
+	"sync"
+
+	"github.com/attic-labs/noms/go/d"
+)
+
+func newCompactingChunkSource(mt *memTable, haver chunkReader, p tablePersister, rl chan struct{}) *compactingChunkSource {
+	ccs := &compactingChunkSource{mt: mt}
+	ccs.wg.Add(1)
+	rl <- struct{}{}
+	go func() {
+		defer ccs.wg.Done()
+		var cs chunkSource = emptyChunkSource{}
+		if tableHash, chunkCount := p.Compact(mt, haver); chunkCount > 0 {
+			cs = p.Open(tableHash, chunkCount)
+		}
+		ccs.mu.Lock()
+		defer ccs.mu.Unlock()
+		ccs.cs = cs
+		ccs.mt = nil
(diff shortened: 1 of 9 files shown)


cmasone-attic:
PTAL

go/nbs/compacting_chunk_source.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+func newCompactingChunkSource(mt *memTable, haver chunkReader, p tablePersister, rl chan struct{}) *compactingChunkSource {
+	ccs := &compactingChunkSource{mt: mt}
+	ccs.wg.Add(1)
+	rl <- struct{}{}
+	go func() {
+		defer ccs.wg.Done()
+		var cs chunkSource = emptyChunkSource{}
+		if tableHash, chunkCount := p.Compact(mt, haver); chunkCount > 0 {
```
rafael-atticlabs: Doesn't need to be this patch, but I think it'd be good to have `Compact` return the `chunkSource`. We always immediately turn around and open it again and it's a bit silly to go read a bunch of bytes from s3 that we just wrote.

Maybe open a bug for that?

go/nbs/table_set.go
```diff
@@ -4,64 +4,94 @@
...
-	return tableSet{p: fsTablePersister{dir}}
+	return tableSet{p: fsTablePersister{dir}, rl: make(chan struct{}, 5)}
 }
 
 // tableSet is an immutable set of persistable chunkSources.
 type tableSet struct {
 	chunkSources
-	p tablePersister
+	p  tablePersister
+	rl chan struct{}
```
rafael-atticlabs: I'm ok with this for now, but it feels pretty at odds with the idea that the tableSet being an immutable thing, that it has a rate limiter attached to it.

ghost (approved):
LGTM.

I see why this is the best solution right, but I wish there was a way to retain table_set being more immutable and not have so much locking. Let's just open a bug to refactor table set to be more immutable. If sufficient time passes and neither of us has a good idea about it, we can just move on.

go/nbs/compacting_chunk_source.go
cmasone-attic: https://github.com/attic-labs/noms/issues/2935

go/nbs/table_set.go
cmasone-attic: fair!

PR #2937, opened 2016-12-12
ChunkSink.Flush()

### LINKED ISSUE #2876: Remove or Implement NomsBlockStore.Flush()
It's only there so that we can implement the types.BatchStore interface, but it's not clear it's needed.

rafael-atticlabs:
Add `ChunkSink.Flush()` which signals the ChunkSink that any previously `Put` chunks should be made durable.

Fixes #2876

### DIFF
--- a/go/types/batch_store.go
+++ b/go/types/batch_store.go
@@ -78,7 +78,10 @@ func (bsa *BatchStoreAdaptor) UpdateRoot(current, last hash.Hash) bool {
 func (bsa *BatchStoreAdaptor) AddHints(hints Hints) {}
 
 // Flush is a noop.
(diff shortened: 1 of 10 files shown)


go/types/batch_store.go
```diff
@@ -78,7 +78,10 @@ func (bsa *BatchStoreAdaptor) UpdateRoot(current, last hash.Hash) bool {
 func (bsa *BatchStoreAdaptor) AddHints(hints Hints) {}
 
 // Flush is a noop.
-func (bsa *BatchStoreAdaptor) Flush() {}
+func (bsa *BatchStoreAdaptor) Flush() {
```
cmasone-attic: The comment above is no longer accurate

PR #2939, opened 2016-12-12
tablePersister.Compact returns a chunkSource

### LINKED ISSUE #2935: NBS: Consider making Compact return a chunkSource
since we usually just go open the now-persisted table anyway

cmasone-attic:
It turns out the only caller of Compact() immediately
turns around and calls Open, so why don't I just do
that FOR you?

Fixes #2935

### DIFF
--- a/go/nbs/table_persister.go
+++ b/go/nbs/table_persister.go
@@ -21,7 +21,7 @@ import (
 const defaultS3PartSize = 5 * 1 << 20 // 5MiB, smallest allowed by S3
 
 type tablePersister interface {
-	Compact(mt *memTable, haver chunkReader) (name addr, chunkCount uint32)
+	Compact(mt *memTable, haver chunkReader) chunkSource
 	Open(name addr, chunkCount uint32) chunkSource
 }
 
@@ -40,7 +40,7 @@ type s3UploadedPart struct {
 	etag string
 }
 
-func (s3p s3TablePersister) Compact(mt *memTable, haver chunkReader) (name addr, chunkCount uint32) {
+func (s3p s3TablePersister) Compact(mt *memTable, haver chunkReader) chunkSource {
 	name, data, chunkCount := mt.write(haver)
 
 	if chunkCount > 0 {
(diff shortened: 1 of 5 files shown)


cmasone-attic:
PTAL

go/nbs/table_persister.go
```diff
@@ -69,8 +69,11 @@ func (s3p s3TablePersister) Compact(mt *memTable, haver chunkReader) (name addr,
 			UploadId:        aws.String(uploadID),
 		})
 		d.Chk.NoError(err)
+		s3tr := &s3TableReader{s3: s3p.s3, bucket: s3p.bucket, h: name}
+		s3tr.tableReader = newTableReader(data, s3tr)
```
rafael-atticlabs: nice. that worked out well.

go/nbs/table_persister.go
```diff
@@ -184,13 +187,13 @@ func (ftp fsTablePersister) Compact(mt *memTable, haver chunkReader) (name addr,
...
-		d.PanicIfError(err)
-	} else {
+	if chunkCount == 0 {
 		os.Remove(tempName)
+		return emptyChunkSource{}
 	}
-	return name, chunkCount
+	err := os.Rename(tempName, filepath.Join(ftp.dir, name.String()))
+	d.PanicIfError(err)
+	return newMmapTableReader(ftp.dir, name, chunkCount)
```
rafael-atticlabs: it's a mild bummer that file table reader is still forced to read the bytes it just wrote, but kind of neccessary because of the need to setup the mmapping. one idea is if the table reader constructor took two byte slices, one for the prefix excluding the suffix index (since that is all eager converted to in-memory structures, and one for the suffixes, because that is left as lazily accessed memory.

in that case, the mmap guy would pass the data he's holding onto as well as correctly setup the mapping.

i'm find leaving that as an open bug. your call.

ghost (approved):
lgtm

go/nbs/table_persister.go
cmasone-attic: I considered that and decided that it would ugly up the API for not enough gain. These bytes were just written, so I expect them to be in the OS filesystem cache, meaning that it'll be pretty cheap to read them again.

Issue #2948, opened 2016-12-14
NBS/AWS DemoServer takes ~1 sec to boot strap each request


rafael-atticlabs:
Prior to https://github.com/attic-labs/noms/commit/d7ee0025d6290f4cc85f1f356054416416f8e713, it took even longer.

The issue is loading the index data for all of the tables.

PR #2950, opened 2016-12-14
Factor tableIndex out of tableReader

### LINKED ISSUE #2949: (NBS) TablePersister leaks s3 table data
https://github.com/attic-labs/noms/blob/master/go/nbs/table_persister.go#L73 is attempting to avoid re-reading the index, but ends up keeping the entire table alive.

rafael-atticlabs:
Fixes #2949. Towards #2948.

### DIFF
--- a/go/nbs/table_reader.go
+++ b/go/nbs/table_reader.go
@@ -14,20 +14,22 @@ import (
 	"github.com/golang/snappy"
 )
 
-// tableReader implements get & has queries against a single nbs table. goroutine safe.
-type tableReader struct {
-	r                 io.ReaderAt
-	suffixes          []byte
+type tableIndex struct {
+	chunkCount        uint32
 	prefixes, offsets []uint64
 	lengths, ordinals []uint32
-	chunkCount        uint32
-	readAmpThresh     uint64
+	suffixes          []byte
 }
 
-// newTableReader parses a valid nbs table byte stream and returns a reader. buff must end with an NBS index and footer, though it may contain an unspecified number of bytes before that data. r should allow retrieving any desired range of bytes from the table.
-func newTableReader(buff []byte, r io.ReaderAt, readAmpThresh uint64) tableReader {
-	tr := tableReader{r: r, readAmpThresh: readAmpThresh}
+// tableReader implements get & has queries against a single nbs table. goroutine safe.
+type tableReader struct {
+	tableIndex
+	r             io.ReaderAt
+	readAmpThresh uint64
+}
 
+// parses a valid nbs table index byte stream and returns a reader. |buff| must end with an NBS index and footer, though it may contain an unspecified number of bytes before that data. |tableIndex| doesn't keep alive any references to |buff|.
+func parseTableIndex(buff []byte) tableIndex {
 	pos := uint64(len(buff))
 
 	// footer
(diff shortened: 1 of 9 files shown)


rafael-atticlabs:
The next step here will be to have a cache of `tableIndex` stored on AWSStoreFactory.

go/nbs/table_reader.go
```diff
@@ -80,6 +89,55 @@ func computePrefixes(count uint32, buff []byte) (prefixes []uint64, ordinals []u
 	return
 }
 
+func (ti tableIndex) prefixIdxToOrdinal(idx uint32) uint32 {
```
rafael-atticlabs: the only code that changed here is `lookupOrdinal`

go/nbs/table_reader.go
```diff
@@ -14,20 +14,22 @@ import (
...
-func newTableReader(buff []byte, r io.ReaderAt, readAmpThresh uint64) tableReader {
-	tr := tableReader{r: r, readAmpThresh: readAmpThresh}
+// tableReader implements get & has queries against a single nbs table. goroutine safe.
+type tableReader struct {
+	tableIndex
+	r             io.ReaderAt
+	readAmpThresh uint64
+}
 
+// parses a valid nbs table index byte stream and returns a reader. |buff| must end with an NBS index and footer, though it may contain an unspecified number of bytes before that data. |tableIndex| doesn't keep alive any references to |buff|.
```
cmasone-attic: This returns a tableIndex, not a reader

go/nbs/table_reader.go
```diff
@@ -38,21 +40,28 @@ func newTableReader(buff []byte, r io.ReaderAt, readAmpThresh uint64) tableReade
...
-	tr.chunkCount = binary.BigEndian.Uint32(buff[pos:])
+	chunkCount := binary.BigEndian.Uint32(buff[pos:])
 
 	// index
-	suffixesSize := uint64(tr.chunkCount) * addrSuffixSize
+	suffixesSize := uint64(chunkCount) * addrSuffixSize
 	pos -= suffixesSize
-	tr.suffixes = buff[pos : pos+suffixesSize]
+	suffixes := make([]byte, suffixesSize)
+	copy(suffixes, buff[pos:])
```
cmasone-attic: Per my comment in slack, "I do also wonder if the fact that different approaches are ideal for different persistence layers (e.g. mmap the index for files, copying for S3) means that we should have different code pathways for creating the tableReader"

There's no need to copy this data over when backed by files, though there may be a need when the data is already in memory.

go/nbs/table_reader.go
```diff
@@ -80,6 +89,55 @@ func computePrefixes(count uint32, buff []byte) (prefixes []uint64, ordinals []u
...
+	li := uint64(ordinal) * addrSuffixSize
+	return bytes.Compare(h[addrPrefixSize:], ti.suffixes[li:li+addrSuffixSize]) == 0
+}
+
+// returns the ordinal of |h| if present. returns |ti.chunkCount| if absent
+func (ti tableIndex) lookupOrdinal(h addr) uint32 {
+	prefix := h.Prefix()
+	idx := ti.prefixIdx(prefix)
+
+	for ; idx < ti.chunkCount && ti.prefixes[idx] == prefix; idx++ {
```
cmasone-attic: Why can't `idx` be initialized inline in the for loop?

go/nbs/table_reader.go
rafael-atticlabs: done

go/nbs/table_reader.go
rafael-atticlabs: yeah, i'm hoping we'll get to the point that a table index is just a wrapper around `[]byte` and at that point, it's much easier to have different approaches to managing the lifetime of those slices. but for now, i'd rather just have one approach which is that the index always makes a copy and keeps it in memory.

go/nbs/table_reader.go
rafael-atticlabs: done

rafael-atticlabs:
ptal

go/nbs/table_reader.go
cmasone-attic: Per offline conversation, it seems that the goal is to make the table index a wrapper around a memory-mapped piece of a file no matter what the ultimate storage backend is. That sounds fine to me.

PR #2951, opened 2016-12-14
JS: Make equals and compare exit early if ===

arv:


arv:
@aboodman FYI

arv:
jenkins: test this

PR #2959, opened 2016-12-15
NBS: tableSet must exclude empty tables during ToSpecs()

### LINKED ISSUE #2957: (NBS) AWS Manifest ends up with 0-table
Something is causing manifests to have references to tables with zero chunks and a hash of `00000000000000000000000000000000`. This results in the store being unusable because it attempts to fetch an non-present table

cmasone-attic:
Fixes #2957


cmasone-attic:
PTAL

ghost (approved):
lgtm. thx

PR #2961, opened 2016-12-17
Make datas.localFactory public

kalman:


kalman:
PTAL. If there is an established way to make memory database factories, let me know and I'll use that instead.

cmasone-attic:
There doesn't seem to be. I think this might technically be orphaned code. I think the only stuff that uses any Factory abstraction these days is server code that uses `chunks.Factory` directly.

This LGTM, but I'm curious what you're using it for

PR #2963, opened 2016-12-19
Add frag tool to measure nbs fragmentation

cmasone-attic:
Before we can defragment NBS stores, we need to understand how
fragmented they are. This tool provides a measure of fragmentation in
which optimal chunk-graph layout implies that ALL children of a given
parent can be read in one storage-layer operation (e.g. disk read, S3
transaction, etc).


cmasone-attic:
PTAL

go/nbs/table_reader.go
```diff
@@ -345,6 +301,72 @@ func (tr tableReader) getMany(reqs []getRecord) (remaining bool) {
 	return
 }
 
+// findOffsets iterates over |reqs| and |tr.prefixes| (both sorted by
+// address) to build the set of table locations which must be read in order to
+// find each chunk specified by |reqs|. If this table contains all requested
+// chunks remaining will be set to false upon return. If some are not here,
+// then remaining will be true.
+func (tr tableReader) findOffsets(reqs []getRecord) (ors offsetRecSlice, remaining bool) {
```
cmasone-attic: `findOffsets()` and `canReadAhead()` are intended to have the same semantics as the code they were refactored from, so please check me and make sure I didn't change some edge case that's not covered by tests or something

go/nbs/store.go
```diff
@@ -200,6 +200,32 @@ func (nbs *NomsBlockStore) GetMany(hashes []hash.Hash) []chunks.Chunk {
 	return resp
 }
 
+func (nbs *NomsBlockStore) CalcReads(hashes []hash.Hash) (reads int, split bool) {
+	reqs := make([]getRecord, len(hashes))
```
rafael-atticlabs: make a utility for this which is used here & GetMany() to produce this slice of getRecords from hash requests?

ghost (approved):
LGTM

go/nbs/frag/main.go
```diff
@@ -0,0 +1,133 @@
...
+	for i := 0; i < concurrency; i++ {
+		go func() {
+			for ref := range refs {
+				mu.Lock()
+				visited := visitedRefs[ref.TargetHash()]
+				visitedRefs[ref.TargetHash()] = true
+				mu.Unlock()
+
+				if !visited {
+					if ref.Height() >= uint64(2) {
```
cmasone-attic: Why, no! It cannot.

go/nbs/store.go

go/nbs/frag/main.go
```diff
@@ -0,0 +1,133 @@
...
+						numbers <- record{count: 1}
+					}
+
+					if ref.Height() > 1 {
```
cmasone-attic: done

PR #2966, opened 2016-12-20
Remove unnessary invariants

rafael-atticlabs:
Title says it all


arv:
Please run `npm version patch` so that we can pick up this version

PR #2973, opened 2016-12-21
Marshal: Use nil for empty Go collections

### LINKED ISSUE #2969: Marshal: use nil for empty go collections
When unmarshalling  an empty Noms collection on onto a Go collection we currently create an empty collection (allocation) if the Go collection is `nil`

https://github.com/attic-labs/noms/blob/master/go/marshal/decode.go#L423-L425

We should defer the allocation until we actually add a value to the collection:

https://github.com/attic-labs/noms/blob/master/go/marshal/decode.go#L443

And similarly for slices/lists.

arv:
This is a breaking change!

We used to create empty Go collections `[]int{}` when unmarshalling an
empty Noms collection onto a Go collection that was `nil`. Now we keep
the Go collection as `nil` which means that you will get `[]int(nil)`
for an empty Noms List.

Fixes #2969

### DIFF
--- a/go/marshal/decode_test.go
+++ b/go/marshal/decode_test.go
@@ -456,6 +456,28 @@ func TestDecodeSlice(t *testing.T) {
 	assert.Equal([]string{"a", "b", "c"}, s)
 }
 
+func TestDecodeSliceEmpty(t *testing.T) {
+	assert := assert.New(t)
+	var s []string
+
+	err := Unmarshal(types.NewList(), &s)
+	assert.NoError(err)
+	assert.Equal([]string(nil), s)
+
+	err = Unmarshal(types.NewSet(), &s)
+	assert.NoError(err)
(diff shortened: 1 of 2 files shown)


go/marshal/decode_test.go
```diff
@@ -456,6 +456,28 @@ func TestDecodeSlice(t *testing.T) {
 	assert.Equal([]string{"a", "b", "c"}, s)
 }
 
+func TestDecodeSliceEmpty(t *testing.T) {
+	assert := assert.New(t)
+	var s []string
+
+	err := Unmarshal(types.NewList(), &s)
+	assert.NoError(err)
+	assert.Equal([]string(nil), s)
```
kalman: `assert.Nil(s)` would also work, but I like the way you have it, being explicit with the nil-ness here.

PR #2974, opened 2016-12-21
Temporarily remove kalman from coffee rotation

kalman:


kalman:
PTAL & submit if LGTY.

aboodman:
not lgtm

kalman:
Huh? Seriously?

It means figuring out amongst yourselves who will take my rotation, which there is an example of it ending up not getting cleaned.

aboodman:
Joke.

PR #2975, opened 2016-12-22
Marshal: Be more lenient when unmarshal a struct

### LINKED ISSUE #2971: Marshal: Unmarshal onto Go struct should be more lenient
Given:

```go
type S struct {
  X string
  Y int
}
var s S
marshal.Unmarshal(types.NewStruct("S", types.StructData{
  "x": types.String("s"),
}, &s)
```

Today this fails with `missing field "Y"`. This is inconsistent with encoding/json as well as inconsistent with how we Unmarshal onto slices and maps. We should just copy over `x` and be done with it.

Pros:
 - This allows upgrading structs
 - More consistent with encoding/json
 - More consistent with how we handle maps and slices

Cons:
 - Pretty much any Noms struct can be marshalled onto any Go struct

arv:
This is a potentially breaking change!

Before this change we required all the fields in a Go struct to be
present in the Noms struct when we unmarshal the Noms struct onto the
Go struct. This is no longer the case, which means that all fields in
the Go struct that are present in the Noms struct will be copied over.

This also means that `omitempty` is useless in Unmarshal and it has been
removed.

This might break your code if expected to get errors when the field
names did not match!

Fixes #2971

### DIFF
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -20,10 +20,7 @@ import (
 //
 // To unmarshal a Noms struct into a Go struct, Unmarshal matches incoming
 // object fields to the fields used by Marshal (either the struct field name or
-// its tag).  Unmarshal will only set exported fields of the struct.  The name
-// of the Go struct must match (ignoring case) the name of the Noms struct. All
-// exported fields on the Go struct must be present in the Noms struct, unless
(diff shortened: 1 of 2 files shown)


arv:
Adding @willhite in case @kalman has to leave.

PR #2978, opened 2016-12-22
Add support for `original` in marshal.Marshal

### LINKED ISSUE #2795: Support preservation of unnamed field values in go struct marshaling

aboodman:
Now it roundtrips

Fixes #2795

### DIFF
--- a/go/marshal/encode.go
+++ b/go/marshal/encode.go
@@ -218,7 +218,7 @@ func structEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc
 	}
 
 	parentStructTypes = append(parentStructTypes, t)
-	fields, structType := typeFields(t, parentStructTypes)
+	fields, structType, originalFieldIndex := typeFields(t, parentStructTypes)
 	if structType != nil {
 		e = func(v reflect.Value) types.Value {
 			values := make([]types.Value, len(fields))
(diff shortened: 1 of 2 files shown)


ehalpern:
Thx!

arv:
If I understand how this works it allows you to extract a subset of a Noms struct into a Go struct and still have that round trip?

ehalpern:
If helpful, the feature is described in
https://github.com/attic-labs/noms/issues/2795.

aboodman:
PTAL @arv

aboodman:
also the unit tests give a reasonable overview of how to use the feature

PR #2979, opened 2016-12-22
Parallel s3 Slice reads

rafael-atticlabs:
GetMany() calls can now be serviced by <= N goroutines, where N is the number of physical reads the request in broken down into.

This patch also adds a `maxReadSize` param to the code which decides how to break chunk reads into physical reads, and sets the s3 blockSize to 5MB, which experimentally resulted to lower total latency.

Lastly, some small refactors.

### DIFF
--- a/go/nbs/s3_table_reader.go
+++ b/go/nbs/s3_table_reader.go
@@ -17,7 +17,7 @@ const (
 	s3RangePrefix   = "bytes"
 	s3ReadAmpThresh = uint64(5)
 	s3BlockSize     = (1 << 20) * 5  // 8MiB
-	s3MaxReadSize   = (1 << 20) * 20 // 32MiB
+	s3MaxReadSize   = (1 << 20) * 20 // 20MiB
 )
 
 type s3TableReader struct {
--- a/go/nbs/s3_table_reader.go
+++ b/go/nbs/s3_table_reader.go
@@ -16,7 +16,8 @@ import (
 const (
 	s3RangePrefix   = "bytes"
 	s3ReadAmpThresh = uint64(5)
-	s3BlockSize     = 0
+	s3BlockSize     = (1 << 20) * 5  // 8MiB
+	s3MaxReadSize   = (1 << 20) * 20 // 32MiB
 )
 
 type s3TableReader struct {
@@ -24,6 +25,7 @@ type s3TableReader struct {
 	s3     s3svc
 	bucket string
 	h      addr
+	readRl chan struct{}
 }
 
 type s3svc interface {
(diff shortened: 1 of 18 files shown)


rafael-atticlabs:
@cmasone-attic Note, in this case, it might be helpful to do a first pass over the individual commits contained

go/nbs/table_reader.go
```diff
@@ -351,24 +349,26 @@ func canReadAhead(fRec offsetRec, fLength uint32, readStart, readEnd, readAmp, b
 
 	// only consider "wasted" bytes ABOVE tr.BlockSize to be read amplification.
 	fReadAmp := fRec.offset - readEnd
-	if fReadAmp < blockSize {
+	if fReadAmp < tr.blockSize {
```
cmasone-attic: The reason I had `canReadAhead` and `calcReads` take `blockSize` and `ampThresh` as parameters was because I figured we'd want the `frag` tool to be able to report a sort of "max" fragmentation, assuming no amplification and that ANY undesired byte is a wasted byte, as well as fragmentation given a certain block size and allowed amplification.  This change would make it only report the latter. Perhaps that's actually the number we care about?

go/nbs/table_reader.go
```diff
@@ -314,6 +305,10 @@ func (tr tableReader) findOffsets(reqs []getRecord) (ors offsetRecSlice, remaini
 	// Iterate over |reqs| and |tr.prefixes| (both sorted by address) and build the set
 	// of table locations which must be read in order to satisfy |reqs|.
 	for i, req := range reqs {
+		if req.found {
```
cmasone-attic: There's a race here that we discussed offline. We should be OK for the immediate future, but will fix ASAP

go/nbs/s3_table_reader.go
```diff
@@ -16,14 +16,16 @@ import (
 const (
 	s3RangePrefix   = "bytes"
 	s3ReadAmpThresh = uint64(5)
-	s3BlockSize     = 0
+	s3BlockSize     = (1 << 20) * 5  // 8MiB
+	s3MaxReadSize   = (1 << 20) * 20 // 32MiB
```
rafael-atticlabs: done

go/nbs/store.go
```diff
@@ -163,7 +165,7 @@ func (nbs *NomsBlockStore) GetMany(hashes []hash.Hash) []chunks.Chunk {
 		tables = nbs.tables
 
 		if nbs.mt != nil {
-			remaining = nbs.mt.getMany(reqs)
+			remaining = nbs.mt.getMany(reqs, &sync.WaitGroup{})
```
rafael-atticlabs: fair. will wait on it. it'll be noop. and it'll go away soon with channel getMany

go/nbs/table_reader.go
rafael-atticlabs: will open bug

Issue #2982, opened 2016-12-23
NBS: Remove reverse-order hack in httpBatchStore


cmasone-attic:
In order to allow value writing to send chunks to the server in write-order, I had to take out the code that sorted chunks-to-be-written by ref-height before sending them to the server. Unfortunately, Pull() relies on functionality like this to allow it to process and SchedulePut() chunks in a top-down fashion. There's now a hack that causes httpBatchStore.Flush() to send chunks in reverse-insert-order.

Stop this. Do a good thing instead :-)
@FYI rafael-atticlabs

rafael-atticlabs:
I actually think that this "hack" is a decent half-step towards "the right thing". Sending in reverse order is convenient because it's an in-order scan.

If we could modify sync so that each height level is written in "chunk" order, then just sending chunks in reverse insertion order might actually create *optimal* locality (i.e. not just parent-child, but sibling).

A (later) step three might be to allow "writes" to take place in this order (i.e. top-down, but in correct "chunk" order), so as to avoid the temporary buffering step).

rafael-atticlabs:
Blocked on https://github.com/attic-labs/noms/issues/3178

PR #2983, opened 2016-12-23
Modify httpBatchStore so that writing values maintains some locality

### LINKED ISSUE #2968: (NBS) *Prevent* noms sync from actively *hurting* chunk locality
https://github.com/attic-labs/noms/issues/2880 focused on offering guarantees about good chunk locality (related chunks being stored near each other in NBS), but it's now clear that two mechanisms in noms are actively working *against* this goal: (1) writing over http and (2) sync.

While we work on a design for retaining and enforcing locality throughout noms (including over http & sync), I think there's some low hanging fruit which would be compatible with the existing wire protocol, and that is to just start with *preventing* writing over http from destroying locality.

If we can do this, at least we will accomplish the common case of writing to a remote server the first time having decent locality.

Right now writing to a remote server involves temporarily storing chunks in a ldb instances and then shipping them all over on commit. The trouble is that what the code currently does is store them in ldb ordered by ascending ref height and then ascending hash. This has the effect of essentially *shuffling* all chunks of a given height on their way over to the server.

So here's my idea: *just for the case of writing to a remote noms server*, simply buffer out chunks in the order they were written (e.g. to a file), and then send them up to the server in exactly that order. The in-process validation will already retain the one required wire invariant: that no chunk shall be written before the server has seen all referenced chunks.

Implementation idea for this: essentially use a local nbs store rather than ldb. As we are writing keep a in-memory list of hashes written. On commit, just read them out in that order (which will be ideal read-order for nbs) as we ship them over to the server.

If this works, I think something slightly more complicated, but similar could work for the case of sync.

cc @cmasone-attic

cmasone-attic:
NBS benefits from related chunks being near one another. Initially,
let's use write-order as a proxy for "related".

This patch contains a pretty heinous hack to allow sync to continue
putting chunks into httpBatchStore top-down without breaking
server-side validation. Work to fix this is tracked in #2982

This patch fixes #2968, at least for now

### DIFF
--- a/go/datas/http_batch_store.go
+++ b/go/datas/http_batch_store.go
@@ -21,6 +21,7 @@ import (
 	"github.com/attic-labs/noms/go/constants"
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
+	"github.com/attic-labs/noms/go/nbs"
 	"github.com/attic-labs/noms/go/types"
 	"github.com/attic-labs/noms/go/util/verbose"
 	"github.com/golang/snappy"
@@ -41,12 +42,13 @@ type httpBatchStore struct {
 	getQueue      chan chunks.ReadRequest
 	hasQueue      chan chunks.ReadRequest
 	writeQueue    chan writeRequest
-	flushChan     chan struct{}
+	flushChan     chan nbs.EnumerationOrder
 	finishedChan  chan struct{}
 	rateLimit     chan struct{}
 	requestWg     *sync.WaitGroup
 	workerWg      *sync.WaitGroup
-	unwrittenPuts *orderedChunkCache
+	unwrittenPuts *nbs.NomsBlockCache
+	cacheMu       *sync.RWMutex
 }
 
 func newHTTPBatchStore(baseURL, auth string) *httpBatchStore {
@@ -62,12 +64,13 @@ func newHTTPBatchStore(baseURL, auth string) *httpBatchStore {
 		getQueue:      make(chan chunks.ReadRequest, readBufferSize),
 		hasQueue:      make(chan chunks.ReadRequest, readBufferSize),
 		writeQueue:    make(chan writeRequest, writeBufferSize),
-		flushChan:     make(chan struct{}),
+		flushChan:     make(chan nbs.EnumerationOrder),
 		finishedChan:  make(chan struct{}),
 		rateLimit:     make(chan struct{}, httpChunkSinkConcurrency),
 		requestWg:     &sync.WaitGroup{},
 		workerWg:      &sync.WaitGroup{},
-		unwrittenPuts: newOrderedChunkCache(),
+		unwrittenPuts: nbs.NewCache(),
+		cacheMu:       &sync.RWMutex{},
 	}
 	buffSink.batchGetRequests()
 	buffSink.batchHasRequests()
@@ -80,7 +83,6 @@ type httpDoer interface {
 }
 
 type writeRequest struct {
-	hash      hash.Hash
 	hints     types.Hints
 	justHints bool
 }
@@ -96,14 +98,17 @@ func makeHTTPClient(requestLimit int) *http.Client {
 }
 
 func (bhcs *httpBatchStore) Flush() {
-	bhcs.flushChan <- struct{}{}
+	bhcs.flushInternal(nbs.ReverseOrder)
+}
+
+func (bhcs *httpBatchStore) flushInternal(order nbs.EnumerationOrder) {
+	bhcs.flushChan <- order
 	bhcs.requestWg.Wait()
 	return
 }
 
 func (bhcs *httpBatchStore) Close() (e error) {
 	close(bhcs.finishedChan)
-	bhcs.unwrittenPuts.Destroy()
 	bhcs.requestWg.Wait()
 	bhcs.workerWg.Wait()
 
@@ -112,11 +117,20 @@ func (bhcs *httpBatchStore) Close() (e error) {
 	close(bhcs.hasQueue)
 	close(bhcs.writeQueue)
 	close(bhcs.rateLimit)
+
+	bhcs.cacheMu.Lock()
+	defer bhcs.cacheMu.Unlock()
+	bhcs.unwrittenPuts.Destroy()
 	return
 }
 
 func (bhcs *httpBatchStore) Get(h hash.Hash) chunks.Chunk {
-	if pending := bhcs.unwrittenPuts.Get(h); !pending.IsEmpty() {
+	checkCache := func(h hash.Hash) chunks.Chunk {
+		bhcs.cacheMu.RLock()
+		defer bhcs.cacheMu.RUnlock()
+		return bhcs.unwrittenPuts.Get(h)
+	}
+	if pending := checkCache(h); !pending.IsEmpty() {
 		return pending
 	}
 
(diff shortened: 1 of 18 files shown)


cmasone-attic:
PTAL

Now that I've gone all the way through the patch, I'm not sure that having NomsBlockCache be distinct from NomsBlockStore makes a lot of sense. On one hand, it's a nice place to put the code that creates and deletes the temp directory that the cache data is in, and it means that `ExtractChunks` is not part of the NomsBlockStore public API. On the other hand...it's a pretty thin wrapper, and the guts of `extractChunks` really needs to be inside NomsBlockStore so it can get at `nbs.tables` and the mutex that protects it.

I'm trying a little bit of further cleanup, but I broke something while doing so. I'll continue trying to chase that down, but please go ahead and review

go/datas/http_batch_store.go
```diff
@@ -313,16 +332,21 @@ func resBodyReader(res *http.Response) (reader io.ReadCloser) {
...
+		defer bhcs.cacheMu.RUnlock()
+		return bhcs.unwrittenPuts.Insert(c)
+	}
+	if !tryCache(c) {
 		return
 	}
 
 	bhcs.requestWg.Add(1)
-	bhcs.writeQueue <- writeRequest{c.Hash(), hints, false}
+	bhcs.writeQueue <- writeRequest{hints, false}
```
rafael-atticlabs: I'm wondering whether the machinery around the write queue (the channel, the persistent goroutine that are servicing the write-queue and the flush queue) are just unneeded complexity now. this stuff is a hold over from the before time when we didn't have any validation and thus were writing in larger batches, but we haven't done that in a while.

i think it's possible now to just have `SchedulePut` Put into the nbs (not care about whether the chunk was contained), and add its hints to the hint set, all inside the mutex, and have `Flush` just send everything over.

i'm probably forgetting about some critical thing, right?

go/datas/http_batch_store.go
```diff
@@ -448,8 +470,7 @@ func (bhcs *httpBatchStore) Root() hash.Hash {
 // UpdateRoot flushes outstanding writes to the backing ChunkStore before updating its Root, because it's almost certainly the case that the caller wants to point that root at some recently-Put Chunk.
 func (bhcs *httpBatchStore) UpdateRoot(current, last hash.Hash) bool {
 	// POST http://<host>/root?current=<ref>&last=<ref>. Response will be 200 on success, 409 if current is outdated.
-	bhcs.Flush()
-
+	bhcs.flushInternal(nbs.InsertOrder)
```
rafael-atticlabs: So is it the case that the sync user always sends over chunks by calling Flush and the simple write user always calls UpdateRoot()?

If so, I guess I'm ok with this, but it is a bit mysterious. If so, let's comment it somewhere and maybe put more detail in the bug you opened that the specific problem is that the API of http_batch_store is now kind of wonky.

As an alternative, maybe BatchStore gets a temporary method called "MarkFlushReverseOrder", or something that only sync calls before it starts writing top-down?

ghost (commented):
BTU.

In general, I think this is good. If you want to address some of these issues subsequently, I'm ok with that.

I think there's the opportunity to simplify things here quite a bit.

Unless I'm missing something, I don't think the Insert-return-whether inserted semantics are needed any more since we only ever send one write batch.

Also, it's a minor bummer that local sync (i.e. sync) still uses the ordered_put_cache.

I guess my temptation would be to go ahead and just *replace* the ordered_put_cache implementation with what is here called the NomsBlockCache, which just uses the normal NBS apis (the extract functions would need to become public).

go/datas/http_batch_store.go
cmasone-attic: This is the thing I was working on changing after I sent this patch, because I also don't like that this relies on making `noms sync` only thing that calls BatchStore.Flush(). This patch made that so, but there's no guarantee at all that it would remain that way.

Something mysteriously broke when I changed things that way, so I'll need to chase that down

go/nbs/cache.go
```diff
@@ -0,0 +1,62 @@
...
+	return &NomsBlockCache{store, dir}
+}
+
+// NomsBlockCache holds Chunks, allowing them to be retrieved by hash or enumerated in hash order.
+type NomsBlockCache struct {
+	chunks *NomsBlockStore
+	dbDir  string
+}
+
+// Insert stores c in the cache. If c is successfully added to the cache,
```
cmasone-attic: you're right. Maybe I don't actually need these semantics if I can get rid of all the goroutine machinery

go/nbs/mem_table.go
```diff
@@ -79,6 +79,20 @@ func (mt *memTable) getMany(reqs []getRecord, wg *sync.WaitGroup) (remaining boo
 	return
 }
 
+func (mt *memTable) extract(order EnumerationOrder, chunks chan<- extractRecord) {
+	sort.Sort(hasRecordByOrder(mt.order)) // ensure "insertion" order for extraction
```
cmasone-attic: I think it will be in insertion order, but we don't document that as a requirement, so I was programming defensively. I can just require that that be an invariant

go/nbs/table_reader.go
```diff
@@ -424,3 +424,34 @@ func (tr tableReader) calcReads(reqs []getRecord, blockSize, maxReadSize, ampThr
...
+		copy(hashes[ordinal][addrPrefixSize:], tr.suffixes[li:li+addrSuffixSize])
+	}
+	chunkLen := tr.offsets[tr.chunkCount-1] + uint64(tr.lengths[tr.chunkCount-1])
+	buff := make([]byte, chunkLen)
+	n, err := tr.r.ReadAt(buff, int64(tr.offsets[0]))
+	d.Chk.NoError(err)
+	d.Chk.True(uint64(n) == chunkLen)
+
+	sendChunk := func(i uint32) {
+		localOffset := tr.offsets[i] - tr.offsets[0]
```
cmasone-attic: Opened https://github.com/attic-labs/noms/issues/2986

go/nbs/table_set.go
```diff
@@ -96,6 +96,19 @@ func (ts tableSet) Prepend(mt *memTable) tableSet {
...
 
+func (ts tableSet) extract(order EnumerationOrder, chunks chan<- extractRecord) {
+	// Since new tables are _prepended_ to a tableSet, extracting chunks in ReverseOrder requires iterating ts.chunkSources from front to back, while doing insertOrder requires iterating back to front.
+	if order == ReverseOrder {
+		for _, cs := range ts.chunkSources {
+			cs.extract(order, chunks)
+		}
+		return
+	}
+	for i := len(ts.chunkSources) - 1; i >= 0; i-- {
```
cmasone-attic: so would I!

Go, however, apparently doesn't. Increment and decrement aren't expressions in Go (statements? whatever the right term is. You can't use them inline like that :-/)

go/datas/http_batch_store.go
cmasone-attic: No, I think this will work. As long as  `AddHints()` uses that mutex also, I think it will work.

go/nbs/table_set.go
rafael-atticlabs: :-)

ghost (approved):
Nice. LGTM

cmasone-attic:
Actually, can you look again? I've squashed so that the second commit includes all the changes since yesterday, which includes getting rid of the old batch-put machinery, that `flushInternal()` hack, and some other things.

rafael-atticlabs:
Still LG

Issue #2972, opened 2016-12-21
Marshal: Rename Noms struct generated from a Go struct.


arv:
Strawman is to use special field in the struct with a tag:

```go
type S struct {
  IgnoreTheName `noms:"@S2"`
}
```

arv:
@kalman

kalman:
As discussed, using an interface actually makes more sense, and it fits in nicely with the idiom of interfaces we're introducing in https://github.com/attic-labs/noms/issues/2970

marshal package:
```go
type MarshalOptions struct {
  StructName string
}

type MarshalOptionsGetter interface {
  GetMarshalOptions() MarshalOoptions
}
```

my package:
```go
type S struct {
  Foo string
}

func (S s) GetMarshalOptions() MarshalOptions {
  return MarshalOptions {StructName: "S2"}
}
```

kalman:
I've found one issue with this: it means that you can't cache the type of any struct which implements `MarshalOptionsGetter`. If you *were* to contrive an annotation based system for this, as suggested by OP, we could.

I don't think it's so bad, though.

arv:
It is true that we cannot precompute the noms struct type, unless our semantics is that `GetMarshalOptions` is only called once and that that function does not depend on any instance.

kalman:
@aboodman 

One reasonable way I can imagine doing this, with annotations, would be something like:
```go
type S struct {
  Abc string
  Xyz int

  SetsTheName `noms-global:"name=MyStruct`
}
```
(modulo naming)

aboodman:
Right I thought that was the null hypothesis. The annotations are less
typing, but less flexible. I'm not sure which is overall better.

kalman:
I realised we can punt on this decision because https://github.com/attic-labs/noms/issues/2970 provides a relatively straightforward workaround:
```go
type S struct {
  Abc string
}

func (s S) MarshalNoms() (types.Value, error) {
  return types.NewStruct("MyCustomName", types.StructData{
    "abc": types.String(s.Abc)
  }), err
}
```

There are a number of problems with that: needing to generate canonical struct field names, needing to update it when fields are added, the inconvenience of marshaling more complex field types.

But at least for now it's enough to unblock people.

PR #2984, opened 2016-12-23
Implement MarshalOptions to change marshaled struct name

kalman:
Previously, marshaling would only create Noms structs with the same
(title-ized) name as the Go struct. Now, structs can implement the
MarshalOptionsGetter interface to provide their own name, or even omit
the name altogether.

### DIFF
--- a/go/marshal/encode.go
+++ b/go/marshal/encode.go
@@ -70,8 +70,9 @@ import (
 //   //  omitted from the object if its value is empty, as defined above.
 //   Field int `noms:",omitempty"
 //
-// The name of the Noms struct is the name of the Go struct where the first
-// character is changed to upper case.
+// The name of the Noms struct, by default, is the name of the Go struct where
+// the first character is changed to upper case. This can be customized by
+// implementing MarshalOptionsGetter.
 //
 // Anonymous struct fields are currently not supported.
 //
@@ -110,6 +111,42 @@ func MustMarshal(v interface{}) types.Value {
 	d.Chk.NoError(err)
 	return r
 }
+
+// MarshalOptions customize how a type is marshalled.
+//
+// For example, to specify a name for a struct when it's marshalled:
+//
+//  type MyStruct struct {
+//    Foo string
+//  }
+//
+//  func (ms MyStruct) GetMarshalOptions() (opts marshal.MarshalOptions) {
+//    opts.Name = "SomeOtherName"
+//    return
+//  }
+//
+//  v, err := marshal.Marshal(MyStruct{"abc"})
+//  // v == types.NewStruct("SomeOtherName", types.StructData{
+//  //   "foo": types.String("abc"),
+//  // })
+type MarshalOptions struct {
+	// EmptyName sets the name of the Noms Struct to an empty string, regardless
+	// of the value of Name. Simply having Name as empty is not sufficient, since
+	// it defaults to the struct type's name.
+	EmptyName bool
+
+	// Name sets the name of the Noms struct when it's marshaled. If this is the
+	// empty string, defaults to the name of the Go struct unless EmptyName is
+	// true. If EmptyName is true, Name is ignored.
+	Name string
+}
+
+// MarshalOptionsGetter is the interface for types to provide MarshalOptions.
+type MarshalOptionsGetter interface {
+	GetMarshalOptions() (opts MarshalOptions)
+}
+
+var optionsGetterType = reflect.TypeOf((*MarshalOptionsGetter)(nil)).Elem()
 
 // UnsupportedTypeError is returned by encode when attempting to encode a type
 // that isn't supported.
@@ -223,30 +260,12 @@ func structEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc
 
 	parentStructTypes = append(parentStructTypes, t)
 	fields, structType, originalFieldIndex := typeFields(t, parentStructTypes)
-	if structType != nil {
-		e = func(v reflect.Value) types.Value {
-			values := make([]types.Value, len(fields))
-			for i, f := range fields {
-				values[i] = f.encoder(v.Field(f.index))
-			}
-			return types.NewStructWithType(structType, values)
-		}
-	} else if originalFieldIndex == nil {
-		// Slower path: cannot precompute the Noms type since there are Noms collections,
-		// but at least there are a set number of fields
-		name := strings.Title(t.Name())
-		e = func(v reflect.Value) types.Value {
-			data := make(types.StructData, len(fields))
-			for _, f := range fields {
-				fv := v.Field(f.index)
-				if !fv.IsValid() || f.omitEmpty && isEmptyValue(fv) {
-					continue
-				}
-				data[f.name] = f.encoder(fv)
-			}
-			return types.NewStruct(name, data)
-		}
-	} else {
+
+	// It won't be possible to use the pre-computed struct type if it implements
+	// MarshalOptionsGetter, because the struct name can change at runtime.
+	isOptsGetter := t.Implements(optionsGetterType)
+
+	if originalFieldIndex != nil {
 		// Slowest path - we are extending some other struct. We need to start with the
 		// type of that struct and extend.
 		e = func(v reflect.Value) types.Value {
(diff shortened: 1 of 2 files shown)


go/marshal/encode.go
```diff
@@ -223,18 +260,29 @@ func structEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc
 
 	parentStructTypes = append(parentStructTypes, t)
 	fields, structType, originalFieldIndex := typeFields(t, parentStructTypes)
-	if structType != nil {
+
+	// It won't be possible to use the pre-computed struct type if it implements
+	// MarshalOptionsGetter, because the struct name can change at runtime.
+	isOptsGetter := t.Implements(optionsGetterType)
```
kalman: This code appears to have changed a lot, but it hasn't. The reason I moved around the conditional blocks: it's easier to follow now with the extra check for `isOptsGetter` (IMO).

go/marshal/encode.go
```diff
@@ -244,28 +292,37 @@ func structEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc
 				}
 				data[f.name] = f.encoder(fv)
 			}
-			return types.NewStruct(name, data)
+			return types.NewStruct(getTypeName(v), data)
```
kalman: This "Slower path" is the only code that changed (and that, barely).

go/marshal/encode.go
```diff
@@ -223,18 +260,29 @@ func structEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc
...
 	parentStructTypes = append(parentStructTypes, t)
 	fields, structType, originalFieldIndex := typeFields(t, parentStructTypes)
-	if structType != nil {
+
+	// It won't be possible to use the pre-computed struct type if it implements
+	// MarshalOptionsGetter, because the struct name can change at runtime.
+	isOptsGetter := t.Implements(optionsGetterType)
+
+	if originalFieldIndex != nil {
+		// Slowest path - we are extending some other struct. We need to start with the
```
kalman: I honestly don't really understand this code, or whether this "original" thing is affected by the marshal options. It seems to me like it shouldn't (?).

go/marshal/encode_test.go
```diff
@@ -694,6 +694,62 @@ func TestEncodeOriginal(t *testing.T) {
...
+
+	s := MarshalOptionsTestT{
+		Abc: 42,
+		opts: MarshalOptions{
+			Name: "",
+		},
+	}
+	v, err := Marshal(s)
+	assert.NoError(err)
+	assert.True(types.NewStruct("MarshalOptionsTestT", types.StructData{
```
arv: I find this API strange. I would want `Name: ""` to result on a Noms struct without a name (`""`).

If we want a way to use the Go name I think that should be done with an extra option and not with the empty string.

go/marshal/encode.go
```diff
@@ -111,6 +112,42 @@ func MustMarshal(v interface{}) types.Value {
...
+// MarshalOptions customize how a type is marshalled.
+//
+// For example, to specify a name for a struct when it's marshalled:
+//
+//  type MyStruct struct {
+//    Foo string
+//  }
+//
+//  func (ms MyStruct) GetMarshalOptions() (opts marshal.MarshalOptions) {
+//    opts.Name = "SomeOtherName"
```
arv: Following seems better to me (not that it matters though)

```go
func (ms MyStruct) GetMarshalOptions() marshal.MarshalOptions {
  return marshal.MarshalOptions{Name: "SomeOtherName"}
}
```

go/marshal/encode.go
```diff
@@ -111,6 +112,42 @@ func MustMarshal(v interface{}) types.Value {
...
+	// true. If EmptyName is true, Name is ignored.
+	Name string
+}
+
+// MarshalOptionsGetter is the interface for types to provide MarshalOptions.
+type MarshalOptionsGetter interface {
+	GetMarshalOptions() (opts MarshalOptions)
+}
+
+var optionsGetterType = reflect.TypeOf((*MarshalOptionsGetter)(nil)).Elem()
```
arv: marshalOptionsGetterType

go/marshal/encode.go
arv: The problem is that if the orignal struct is named "A" and the options wants it to be named "B" the name will stay as "A" because with original we start with the original `types.Struct`.

go/marshal/encode.go
```diff
@@ -223,18 +260,29 @@ func structEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc
...
+	isOptsGetter := t.Implements(optionsGetterType)
+
+	if originalFieldIndex != nil {
+		// Slowest path - we are extending some other struct. We need to start with the
+		// type of that struct and extend.
 		e = func(v reflect.Value) types.Value {
-			values := make([]types.Value, len(fields))
-			for i, f := range fields {
-				values[i] = f.encoder(v.Field(f.index))
+			ret := v.FieldByIndex(originalFieldIndex).Interface().(types.Struct)
```
arv: @aboodman Just fixed an NPE here

go/marshal/encode.go
```diff
@@ -223,18 +260,29 @@ func structEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc
...
+			return ret
 		}
-	} else if originalFieldIndex == nil {
-		// Slower path: cannot precompute the Noms type since there are Noms collections,
-		// but at least there are a set number of fields
-		name := strings.Title(t.Name())
+	} else if structType == nil || isOptsGetter {
+		// Slower path: cannot precompute the Noms type since there are Noms
+		// collections, or behavior might change at marshal-time via
+		// MarshalOptions, but at least there are a set number of fields.
```
arv: Another option might be to cache this based on name too since I don't think it is common to change the name at runtime.

Maybe a TODO is sufficient since we do not yet know if this will be a performance critical.

go/marshal/encode.go
```diff
@@ -244,28 +292,37 @@ func structEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc
...
+			return types.NewStructWithType(structType, values)
 		}
 	}
 
 	encoderCache.set(t, e)
 	return e
 }
 
+func getTypeName(v reflect.Value) string {
+	if optsGetter, ok := v.Interface().(MarshalOptionsGetter); ok {
```
arv: Use t.Implements instead?

aboodman:
having to implement this interface seems unfortunately heavyweight. Why did we not do the annotations on a special field?

go/marshal/encode.go
kalman: Yeah I know you can do that. The pattern I said:
```go
opts.Name = "SomeOtherName"
```
is nice because it implies not changing other options (especially if you want to somehow use a placeholder value for the name). Also you don't need to type `marshal.MarshalOptions` multiple times :-)

go/marshal/encode.go
kalman: I agree it's uncommon (and probably incorrect) to change the name at runtime. However I'm afraid it'll cause hard to track down bugs. Better to fail in a predictable way.

go/marshal/encode.go
kalman: Well I still need to cast it to the interface.

go/marshal/encode_test.go
kalman: It's hard with default Go values for structs. It would mean that any type that implements the MarshalOptionsGetter, for any reason, would change its behavior.

kalman:
> having to implement this interface seems unfortunately heavyweight. Why did we not do the annotations on a special field?

We couldn't come up with a non-bizarrely-arbitrarily way to do this. Interfaces is the idiom that Go provides for customising the encoding of types, so it would be nice to get that to work.

To put it another way: an interface-based API is more laborious to use, but it's idiomatic, and documented at the language level. The annotation one is basically us inventing our own thing and that's confusing (I'm already going back-and-forth with the documentation prose to figure out how to use the simpler annotations).

I'm happy to keep discussing a good way to do this though, https://github.com/attic-labs/noms/issues/2972 is probably a more permanent place to have that discussion.

aboodman:
What about embedding the MarshalOptions struct? Marshal can test whether it
is present unambiguously.

kalman:
You still need to give the options a value. If you had some deeply nested
Go structure that you wanted to custom-marshal, you'd need to both (a) dig
into there, and (b) dig into *every* place that you wanted to change it
(including lists, maps, etc).

aboodman:
Ah yes, you're right.

PR #2991, opened 2016-12-24
add noms show --raw

aboodman:
@rafael-atticlabs fyi

### DIFF
--- a/cmd/noms/noms_show.go
+++ b/cmd/noms/noms_show.go
@@ -5,7 +5,9 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


aboodman:
It only shows the raw top level chunk which is exactly what I want. It's
useful for debugging the serialization.

arv:
That is also what I expected/want. Can we have a test for that?

aboodman:
Hm, It sorta already is testing that because it decodes the raw chunk. Assuming the decoder is working correctly (and it is tested elsewhere) it should decode chunks with random extra data in them.

But I did add some tests for more complex values.

PR #2992, opened 2016-12-24
Add `noms root` command for getting and setting root of entire db

aboodman:
Whichever gets to it first.

### DIFF
--- a/cmd/noms/noms_root.go
+++ b/cmd/noms/noms_root.go
@@ -1,0 +1,80 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/attic-labs/noms/cmd/util"
+	"github.com/attic-labs/noms/go/config"
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/hash"
+	flag "github.com/juju/gnuflag"
+)
+
+var nomsRoot = &util.Command{
+	Run:       runRoot,
+	UsageLine: "root <db-spec>",
+	Short:     "Get or set the current root hash of the entire database",
+	Long:      "See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the database argument.",
+	Flags:     setupRootFlags,
+	Nargs:     1,
+}
+
+var updateRoot = ""
+
+func setupRootFlags() *flag.FlagSet {
+	flagSet := flag.NewFlagSet("root", flag.ExitOnError)
+	flagSet.StringVar(&updateRoot, "update", "", "Replaces the entire database with the one with the given hash")
+	return flagSet
+}
+
+func runRoot(args []string) int {
+	if len(args) < 1 {
+		fmt.Fprintln(os.Stderr, "Not enough arguments")
+		return 0
+	}
+
+	cfg := config.NewResolver()
+	db, err := cfg.GetDatabase(args[0])
+	d.CheckErrorNoUsage(err)
+	defer db.Close()
+
+	currRoot := db.ValidatingBatchStore().Root()
+
+	if updateRoot == "" {
+		fmt.Println(currRoot)
+		return 1
+	}
+
+	fmt.Println(`WARNING
+
+This operation replaces the entire database with the instance having the given
+hash. The old database becomes eligible for GC.
+
+ANYTHING NOT SAVED WILL BE LOST
+
+Continue?
+`)
+	fmt.Scanln()
+
+	if updateRoot[0] == '#' {
+		updateRoot = updateRoot[1:]
+	}
+	h, ok := hash.MaybeParse(updateRoot)
+	if !ok {
+		fmt.Fprintf(os.Stderr, "Error: Hash %s does not exist in database\n", h.String())
+		return 1
+	}
+	ok = db.ValidatingBatchStore().UpdateRoot(h, currRoot)
+	if !ok {
+		fmt.Fprintln(os.Stderr, "Optimistic concurrency failure")
+		return 1
+	}
+
+	fmt.Printf("Success. Previous root was: %s\n", currRoot)
+	return 0
+}
--- a/cmd/noms/noms_root.go
+++ b/cmd/noms/noms_root.go
@@ -64,6 +64,7 @@ func runRoot(args []string) int {
 
 	db, err := cfg.GetDatabase(args[0])
 	d.CheckErrorNoUsage(err)
+	defer db.Close()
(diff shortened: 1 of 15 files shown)


go/datas/database.go
```diff
@@ -94,8 +94,13 @@ type Database interface {
 	// Regardless, Datasets() is updated to match backing storage upon return.
 	FastForward(ds Dataset, newHeadRef types.Ref) (Dataset, error)
 
+	// ValidatingBatchStore returns the BatchStore used to read and write
+	// groups of values to the database efficiently. This interface is a low-
+	// level detail of the database that should infrequently be needed by
+	// clients.
+	ValidatingBatchStore() types.BatchStore
```
cmasone-attic: This is very, very, very much intended to remain internal API. Kind of the only reason @rafael-atticlabs and I were OK with having this method on Database is that it was package-internal and only used by `Pull()`. There are some subtle and potentially surprising side effects of calling this method, and I think that exposing it outside the package is asking for long-term pain.

Instead of `cfg.GetDatabase()`, maybe you could use `cfg.GetChunkStore()` and just get/set the Root that way. I know we're talking about "getting rid of" ChunkStore, but that just means we'll replace it with BatchStore throughout the code base.

aboodman: Yup, guessed this wasn't s good approach but figured a patch was best way to start convo.

The http-client db doesn't have a chunkstore right? How can I do it in that case?

I thought about just exposing Root() and UpdateRoot()...

cmasone-attic:
Ugh. Yeah, you're right that there's no ChunkStore in the HTTP case. All you actually want is a RootTracker, and both BatchStore and ChunkStore satisfy that. Since we have a medium term goal to get rid of ChunkStore in favor of BatchStore, perhaps the Right Thing is to add `GetRootTracker` to `config.Resolver`. Then you can use that for this tool, and we can circle back when we do the ChunkStore -> BatchStore conversion and harmonize things at that point.

rafael-atticlabs:
What's the use case for this?

aboodman:
I needed it initially to debug the huge chunk bug. But generally I've found
it very useful, along with 'noms show --raw' (separate pr) to debug
problems at the serialization level.

It can also be used for a cheap and easy way to checkpoint/backup/restore
an entire db.

aboodman:
Actually `config` already has `GetChunkStore` so I think I can just use that.

cmasone-attic:
There's no ChunkStore in the HTTP case, though, as you pointed out earlier

aboodman:
oh bother

aboodman:
So ... I still don't get how you want me to get RootTracker for the HTTP case. RootTracker is implemented by httpBatchStore, which is private. Do you want me to expose it somewhere?

cmasone-attic:
...yeah, it's probably most sensible to do something like
```golang
func NewHTTPBatchStore(baseURL, auth string) types.BatchStore {
  return newHTTPBatchStore(baseURL, auth)
}
```

That way the `httpBatchStore` type can stay private.

aboodman:
@cmasone-attic ptal

cmd/noms/noms_root.go
```diff
@@ -0,0 +1,86 @@
...
+	if n != 1 || strings.ToLower(input) != "y" {
+		return 0
+	}
+
+	if updateRoot[0] == '#' {
+		updateRoot = updateRoot[1:]
+	}
+	h, ok := hash.MaybeParse(updateRoot)
+	if !ok {
+		fmt.Fprintf(os.Stderr, "Error: Hash %s does not exist in database\n", h.String())
```
cmasone-attic: `MaybeParse()` doesn't check whether the hash exists, it just determines whether the string parses and decodes to a byteslice of the right length.

If you're not going to validate anything about the object the provided hash points to, then change this error message to be something like "Hash %s does not parse".

If you _are_ going to validate whether this hash points to an existing object, you should probably also validate that that object has the appropriate type. For that you probably need a `Database`

aboodman:
In the case where this is a remote server, the remote server should do that
validation right?

cmasone-attic:
Yes, if you're talking to a remote DB then the server does that validation

aboodman:
I sorta feel like current behavior is correct then. If you decide to fuck up your own database, that's on you boss. It does print out the old root so you can revert back to it.

aboodman:
Hm, on second thought, I agree with you. Will add this validation.

aboodman:
PTAL

aboodman:
@arv, please review change to map.go

go/util/clienttest/client_test_suite.go
```diff
@@ -62,6 +64,7 @@ func (suite *ClientTestSuite) MustRun(m func(), args []string) (stdout string, s
 // If m()  panics the panic is caught, and returned with recoveredError
 // If m() calls exit.Exit() m() will panic and return ExitError with recoveredError
 func (suite *ClientTestSuite) Run(m func(), args []string) (stdout string, stderr string, recoveredErr interface{}) {
+	fmt.Println("running", args)
```
cmasone-attic: dbg output

cmd/noms/noms_root.go
```diff
@@ -0,0 +1,110 @@
...
+		return 1
+	}
+
+	fmt.Printf("Success. Previous root was: %s\n", currRoot)
+	return 0
+}
+
+func validate(r types.Value) bool {
+	if !types.IsSubtype(types.MakeMapType(types.StringType, types.ValueType), r.Type()) {
+		fmt.Fprintf(os.Stderr, "Root of database must be Map<string, Value>, but you specified: %s\n", r.Type().Describe())
```
cmasone-attic: It's actually `Map<String, Ref<Value>>`

cmd/noms/noms_root.go
```diff
@@ -0,0 +1,110 @@
...
+	if updateRoot[0] == '#' {
+		updateRoot = updateRoot[1:]
+	}
+	h, ok := hash.MaybeParse(updateRoot)
+	if !ok {
+		fmt.Fprintf(os.Stderr, "Invalid hash: %s\n", h.String())
+		return 1
+	}
+
+	db, err := cfg.GetDatabase(args[0])
```
cmasone-attic: you never close this DB

aboodman:
all fixed, thanks

go/types/map.go
```diff
@@ -247,6 +247,18 @@ func (m Map) Iter(cb mapIterCallback) {
 	})
 }
 
+// Any returns true if cb() return true for any of the items in the map.
+func (m Map) Any(cb func(k, v Value) bool) (yep bool) {
+	m.Iter(func(k, v Value) bool {
```
arv: This might be cleaner with `(stop bool)`.

aboodman:
jenkins: test this

aboodman:
@arv how could there be a js impl? it's a command line feature.

aboodman:
OH, of map.any. No.

PR #2997, opened 2016-12-31
Add new spec for nbs-aws

cmasone-attic:
The new spec is a URI, akin to what we use for HTTP It allows the
specification of a DynamoDB table, an S3 bucket, a database ID, and a
dataset ID: aws://table-name:bucket-name/database::dataset

The bucket name is optional and, if not provided, Noms will use a
ChunkStore implementation backed only by DynamoDB.

### DIFF
--- a/go/spec/spec.go
+++ b/go/spec/spec.go
@@ -17,6 +17,8 @@ import (
 	"github.com/attic-labs/noms/go/datas"
 	"github.com/attic-labs/noms/go/nbs"
 	"github.com/attic-labs/noms/go/types"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


cmasone-attic:
PTAL

go/spec/spec.go
```diff
@@ -247,6 +250,14 @@ func (sp Spec) createDatabase() datas.Database {
 	switch sp.Protocol {
 	case "http", "https":
 		return datas.NewRemoteDatabase(sp.Href(), sp.Options.Authorization)
+	case "aws":
+		u, _ := url.Parse(sp.Href())
+		parts := strings.SplitN(u.Host, ":", 2) // [table] [, bucket]?
+		sess := session.Must(session.NewSession(aws.NewConfig().WithRegion("us-west-2")))
```
willhite: Hmmm, I wonder if the region should be part of the spec, too. Once we open source this, people might want to run it in a different region.

cmasone-attic: Issue filed to think about this. It might be tough to add another optional thing, because I think the syntax might become ambiguous. We can talk about it when we're all togehter later this week. https://github.com/attic-labs/noms/issues/2999

PR #2998, opened 2017-01-01
Js sequence refactor

### LINKED ISSUE #2904: Port Go sequence refactor to JS
This refactor was done to Go: 068eb0cb71c11e3f163d1fcb76e5a49c3d01efb7
We need to port it to JS.

When that's done we should also implement Set.at and Map.at in JS.

@rafael-atticlabs

rafael-atticlabs:
Fixes #2904 and lays some ground work for lazy sequence decoding.

In particular, removes `IndexedSequence` and `OrderedSequence`. What remain are `MetaSequence` and each of the leaf sequences which directly inherit from `Sequence`.

### DIFF
--- a/js/noms/src/sequence.js
+++ b/js/noms/src/sequence.js
@@ -16,6 +16,11 @@ import type Value from './value.js'; // eslint-disable-line no-unused-vars
 import Hash from './hash.js';
 import {compare, equals} from './compare.js';
 
+// Sequence<T> is the base class of all prolly-tree nodes. It represents a sequence of "values"
+// which are grouped together and defined by the output of a rolling hash value as a result of
+// chunking. In this context, |T| need not be a `Value` type. In the case of non-leaf prolly-tree
+// levels, it will be a |MetaTuple|, and in the case of a Map leaf sequence, it will be a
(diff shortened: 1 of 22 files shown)


rafael-atticlabs:
Also, It doesn't look to me like we need `Indexed/OrderedSequenceCursor/Iterator` any-more either, but I'm leaving that for a subsequent patch.

rafael-atticlabs:
(Final note): I removed `List.toJS()` which didn't appear to have an usages.

js/noms/src/sequence.js
```diff
@@ -11,6 +11,10 @@ import type {AsyncIteratorResult} from './async-iterator.js';
 import Ref from './ref.js';
 import type {Type} from './type.js';
 import {ValueBase} from './value.js';
+import type {EqualsFn} from './edit-distance.js';
+import type Value from './value.js'; // eslint-disable-line no-unused-vars
+import Hash from './hash.js';
+import {compare, equals} from './compare.js';
 
 export default class Sequence<T> {
```
kalman: Could you document this now + what `T` is?

PR #3001, opened 2017-01-03
Enable noms serve to serve nbs-aws stores

cmasone-attic:
There's no reason that this shouldn't work, so make it do so


cmasone-attic:
PTAL

PR #3002, opened 2017-01-03
[NBS] Implement Streaming GetMany

### LINKED ISSUE #2980: (NBS) fix race in table_reader
https://github.com/attic-labs/noms/commit/d8d8c6c7e14058aafffc05964725578cacac3f31 created a race where one a reader of rec.found could collide with a writer of rec.data. It's possible this would be fine, but the behavior is undefined, given slice storage, so it'd be better to avoid it.

### LINKED ISSUE #2981: (NBS) Allow `GetMany()` to stream chunks back over a channel
That way the server wouldn't need to start sending response bytes until all chunks are fetched.

rafael-atticlabs:
Adds the ability to stream individual chunks requested via GetMany() back to caller.

Removes readAmpThresh and maxReadSize. Lowers the S3ReadBlockSize to 512k.

Fixes #2981
Fixes #2980

### DIFF
--- a/go/nbs/mem_table.go
+++ b/go/nbs/mem_table.go
@@ -7,6 +7,9 @@ package nbs
 import (
 	"sort"
 	"sync"
+
+	"github.com/attic-labs/noms/go/chunks"
+	"github.com/attic-labs/noms/go/hash"
 )
 
 type memTable struct {
@@ -69,11 +72,12 @@ func (mt *memTable) get(h addr) []byte {
 	return mt.chunks[h]
 }
 
-func (mt *memTable) getMany(reqs []getRecord, wg *sync.WaitGroup) (remaining bool) {
-	for i, r := range reqs {
+func (mt *memTable) getMany(reqs []getRecord, foundChunks chan *chunks.Chunk, wg *sync.WaitGroup) (remaining bool) {
+	for _, r := range reqs {
 		data := mt.chunks[*r.a]
 		if data != nil {
-			reqs[i].data = data
+			c := chunks.NewChunkWithHash(hash.Hash(*r.a), data)
+			foundChunks <- &c
 		} else {
 			remaining = true
 		}
--- a/go/nbs/store.go
+++ b/go/nbs/store.go
@@ -28,7 +28,7 @@ const (
 	// StorageVersion is the version of the on-disk Noms Chunks Store data format.
 	StorageVersion = "0"
 
-	defaultMemTableSize uint64 = 512 * 1 << 20 // 512MB
+	defaultMemTableSize uint64 = (1 << 20) * 128 // 128MB
 	defaultAWSReadLimit        = 1024
 
 	InsertOrder EnumerationOrder = iota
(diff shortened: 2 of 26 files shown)


go/nbs/store.go
```diff
@@ -183,41 +183,27 @@ func (nbs *NomsBlockStore) GetMany(hashes []hash.Hash) []chunks.Chunk {
...
 }
 
-func toGetRecords(hashes []hash.Hash) []getRecord {
+func toGetRecords(hashes hash.HashSet) []getRecord {
 	reqs := make([]getRecord, len(hashes))
-	for i, h := range hashes {
+	for h, _ := range hashes {
 		a := addr(h)
-		reqs[i] = getRecord{
+		reqs = append(reqs, getRecord{
```
cmasone-attic: Since you're appending here, instead of setting elements directly by index, I believe you need to set the capacity of `reqs` to 0 above. Otherwise, you'll wind up with a slice containing `len(hashes)` empty records, and then all the stuff you care about appended to that.

go/nbs/table_reader.go
```diff
@@ -389,36 +378,32 @@ func (tr tableReader) calcReads(reqs []getRecord, blockSize, maxReadSize, ampThr
...
-		readEnd := rec.offset + uint64(tr.lengths[rec.ordinal]) // implicitly include the first chunk
+	for i := 0; i < len(offsetRecords); {
+		rec := offsetRecords[i]
+		length := tr.lengths[rec.ordinal]
 
-		// As we scan forward, for each location/length, we'll include it in the current read if
-		// the total number of bytes we'll read contains fewer than X bytes we don't care about.
-		readAmp := uint64(0)
+		if !readStarted {
+			readStarted = true
```
cmasone-attic: I think maybe that, if we increment `reads` every time we set `readStarted` to true, that's equivalent to what we're doing now, and it lets us get rid of the extra check after the loop, which feels a bit clunky.

go/nbs/mem_table.go
```diff
@@ -69,11 +72,12 @@ func (mt *memTable) get(h addr) []byte {
 	return mt.chunks[h]
 }
 
-func (mt *memTable) getMany(reqs []getRecord, wg *sync.WaitGroup) (remaining bool) {
-	for i, r := range reqs {
+func (mt *memTable) getMany(reqs []getRecord, foundChunks chan *chunks.Chunk, wg *sync.WaitGroup) (remaining bool) {
```
rafael-atticlabs: Yeah. I see what you are saying. I think I'm going to open a bug for this so we can think about it. As you point out, fixing it would involve basically making an internal notion of chunk, which seems to suggest

-It's reasonable for a store to depend on Chunk
-It's probably better to move the other stores away from the `chunks` package.

For instance, we could move the chunk specific stuff together with `hash` (since it's pretty annoying that that name collides with the stdlib) and put all ChunkStore implementations in `store` or something.

rafael-atticlabs:
jenkins: test this

Issue #2989, opened 2016-12-23
Need @at annotation for javascript


aboodman:
For splore!

kalman:
Done!

PR #3003, opened 2017-01-03
Refactor JS path code and implement @type

kalman:
This brings it much closer to the Go implementation, and eventually
implementing the @at annotation.

### DIFF
--- a/js/noms/src/path-test.js
+++ b/js/noms/src/path-test.js
@@ -23,10 +23,10 @@ function hashIdx(v: Value): string {
   return `[#${getHash(v).toString()}]`;
 }
 
-async function assertResolvesTo(expect: Value | null, ref: Value, str: string) {
+async function assertResolvesTo(expect: Value | null, target: Value, str: string) {
   const j = s => JSON.stringify(s);
   const p = Path.parse(str);
-  const actual = await p.resolve(ref);
+  const actual = await p.resolve(target);
   if (expect === null) {
     assert.isTrue(actual === null, `Expected null, but got ${j(actual)}`);
   } else if (actual === null) {
@@ -292,14 +292,14 @@ suite('Path', () => {
   });
 
   test('type annotation', async () => {
-    const m = new Map([
+    const mkv = [
       ['string', 'foo'],
       ['bool', false],
       ['number', 42],
       ['List<number|string>', new List([42, 'foo'])],
       ['Map<bool, bool>', new Map([[true, false]])],
-    ]);
-
+    ];
+    const m = new Map(mkv);
     const s = newStruct('', {
       str: 'foo',
       num: 42,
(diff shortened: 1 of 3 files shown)


kalman:
I'm working towards https://github.com/attic-labs/noms/issues/2989 but there needs to be a little bit more work until I can support it. Luckily raf coincidentally has a PR out that does this little work.

kalman:
Where "little" is actually an enormously painful refactor.

kalman:
Compare to https://github.com/attic-labs/noms/blob/master/go/types/path.go

kalman:
adding @arv not sure if aa is back yet

js/noms/src/path.js
```diff
@@ -200,13 +233,35 @@ function parsePathIndex(str: string): [indexType | null, Hash | null, string] {
   throw new SyntaxError(`Invalid index: ${idxStr}`);
 }
 
-function getAnnotation(str: string): [string /* ann */, string /* rem */] {
+function getAnnotation(str: string)
+    : [string /* ann */, boolean /* hasArg */, string /* arg */, string /* rem */] {
```
arv: Maybe return `{ann: string, hasArg: boolean, arg: string, rem: string}` instead

js/noms/src/path.js
```diff
@@ -382,6 +414,20 @@ export class HashIndexPath {
   }
 }
 
+/**
+ * TypeAnnotation is a PathPart annotation to resolve to the type of the value
+ * it's resolved in.
+ */
+class TypeAnnotation {
+  resolve(v: Value): Promise<Value | null> {
+    return Promise.resolve(getTypeOfValue(v));
```
arv: This cannot resolve to null. Also, maybe tighten this a bit? `Promise<Type<TypeDesc>>`

js/noms/src/path-test.js
```diff
@@ -277,8 +281,44 @@ suite('Path', () => {
...
     t('.foo[42]@soup', 'Unsupported annotation: @soup');
+    t('.foo@key', 'Cannot use @key annotation on: .foo');
+    t('.foo@key()', '@key annotation does not support arguments');
+    t('.foo@key(42)', '@key annotation does not support arguments');
+    t('.foo@type()', '@type annotation does not support arguments');
+    t('.foo@type(42)', '@type annotation does not support arguments');
+  });
+
+  test('type annotation', async () => {
+    const m = new Map([
```
kalman: It's used further to test a type annotation, but I've pulled out the key-values to make this nicer.

js/noms/src/path-test.js
```diff
@@ -277,8 +281,44 @@ suite('Path', () => {
...
+
+    const s = newStruct('', {
+      str: 'foo',
+      num: 42,
+    });
+
+    const tests = [];
+
+    await m.forEach((v: string, k: Value) => {
+      tests.push(assertResolvesTo(getTypeOfValue(v), m, `["${k}"]@type`));
```
kalman: I actually messed the type annotations up here. `v` is a `Value`, `k` is a `string`. It happened because the k/v arguments are the confusing way around, but the little refactor cleared things up.

Sad that Flow didn't figure this out... or maybe it *did* figure it out, the type annotations were wrong, and it just ignored them.

js/noms/src/path.js
kalman: Good idea.

js/noms/src/path.js
kalman: It was like this because TypeAnnotation needs to implement the Part interface. But it seems happy with the tighter type, so cool.

PR #3007, opened 2017-01-03
TestHandleGetRefs

rafael-atticlabs:
The order of returned chunks is no longer guaranteed.

### DIFF
--- a/go/datas/remote_database_handlers_test.go
+++ b/go/datas/remote_database_handlers_test.go
@@ -158,17 +158,14 @@ func TestBuildWriteValueRequest(t *testing.T) {
 	assert.Equal(len(hints), count)
 
 	outChunkChan := make(chan interface{}, 16)
-	go deserializeToChan(gr, outChunkChan)
+	chunks.DeserializeToChan(gr, outChunkChan)
+	close(outChunkChan)
+
 	for c := range outChunkChan {
 		assert.Equal(chnx[0].Hash(), c.(*chunks.Chunk).Hash())
 		chnx = chnx[1:]
 	}
 	assert.Empty(chnx)
-}
-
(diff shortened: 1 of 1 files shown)


go/datas/remote_database_handlers_test.go
```diff
@@ -158,19 +158,16 @@ func TestBuildWriteValueRequest(t *testing.T) {
 	assert.Equal(len(hints), count)
 
 	outChunkChan := make(chan interface{}, 16)
-	go deserializeToChan(gr, outChunkChan)
+	chunks.DeserializeToChan(gr, outChunkChan)
```
cmasone-attic: Won't this hang if the code being tested tries to put more than 16 items into `outChunkChan`? Obviously it _shouldn't_ do that, but it might. It feels a bit gross to rely on the chunk buffer to avoid deadlocking.

I guess my more realistic concern is that someone expands the test to write more than 16 chunks, and you could mitigate that by making the chunk buffer be the same size as `chnx`

rafael-atticlabs: done

go/datas/remote_database_handlers_test.go
```diff
@@ -227,13 +224,20 @@ func TestHandleGetRefs(t *testing.T) {
 	)
 
 	if assert.Equal(http.StatusOK, w.Code, "Handler error:\n%s", string(w.Body.Bytes())) {
-		chunkChan := make(chan interface{})
-		go deserializeToChan(w.Body, chunkChan)
+		chunkChan := make(chan interface{}, 2)
```
cmasone-attic: THis actually has the same problem as above

cmasone-attic:
lgtm

Issue #3006, opened 2017-01-03
(NBS) Investigate http timeouts while writing to S3/NBS


rafael-atticlabs:
HttpBatchStore presently sets a response-header timeout of 120s which seems to be getting hit by the downloader. This may be due partially to the NBS memtable size being set to 512MB and write concurrency to 5 because it means that the server will buffer up to 2.5GB of data, allowing the client's idle timer to start.

However, it's still a mystery why it would take > 120s to write 2.5GB to S3.

rafael-atticlabs:
Fixed by https://github.com/attic-labs/noms/commit/e025471eab95c4cbaf8fe5704834e07a33d4843a

PR #3008, opened 2017-01-03
Address write timeout

rafael-atticlabs:
Relevant to: https://github.com/attic-labs/noms/issues/3006.

Ups the response-header timeout to 4m, and adds some additional log output which hopefully will shed some light on why some writes are taking so long.

### DIFF
--- a/go/nbs/s3_table_persister.go
+++ b/go/nbs/s3_table_persister.go
@@ -8,9 +8,11 @@ import (
 	"bytes"
 	"sort"
 	"sync"
+	"time"
 
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/util/sizecache"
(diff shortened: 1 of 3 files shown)


go/nbs/s3_table_persister.go
```diff
@@ -65,6 +68,7 @@ func (s3p s3TablePersister) Compact(mt *memTable, haver chunkReader) chunkSource
 		d.Chk.NoError(err)
 		s3tr := &s3TableReader{s3: s3p.s3, bucket: s3p.bucket, h: name}
 
+		verbose.Log("Compacted table of %d Kb in %s", len(data), time.Since(t1))
```
cmasone-attic: Shouldn't it be `len(data)/1024`?

rafael-atticlabs: why yes, it should.

PR #3012, opened 2017-01-04
Add the marshal.Marshaler interface

kalman:
This lets types provide their own Noms value marshaling.

### DIFF
--- a/go/marshal/encode.go
+++ b/go/marshal/encode.go
@@ -8,6 +8,7 @@
 package marshal
 
 import (
+	"fmt"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


kalman:
There were some subtle things here, if you can think of any other reasonable test cases, LMK.

marshal.Unmarshaler is almost ready, it was a little trickier.

go/marshal/encode.go
```diff
@@ -89,12 +93,14 @@ import (
 func Marshal(v interface{}) (nomsValue types.Value, err error) {
 	defer func() {
 		if r := recover(); r != nil {
-			switch r.(type) {
+			switch r := r.(type) {
```
arv: oh go!

PR #3013, opened 2017-01-04
fix bug in http_batch_store storing of hints

rafael-atticlabs:
TBR=@cmasone-attic


cmasone-attic:
well...that'd do it. LGTM

PR #3016, opened 2017-01-04
ValueStore writes chunks only when referenced

cmasone-attic:
The old strategy for writing values was to recursively encode them,
putting the resulting chunks into a BatchStore from the bottom up as
they were generated. The BatchStore implementation was responsible for
handling concurrency, so chunks from different Values would be
interleaved if the there were multiple calls to WriteValue happening
at the same time.

The new strategy tries to keep chunks from the same 'level' of a
graph together by caching chunks as they're encoded and only writing
them once they're referenced by some other value. When a collection
is written, the graph representing it is encoded recursively, and
chunks are generated bottom-up. The new strategy should, in practice,
mean that the children of a given parent node in this graph will be
cached until that parent gets written, and then they'll get written
all at once.

### DIFF
--- a/go/types/value_store.go
+++ b/go/types/value_store.go
@@ -13,35 +13,45 @@ import (
 	"github.com/attic-labs/noms/go/util/sizecache"
 )
 
-// ValueReader is an interface that knows how to read Noms Values, e.g. datas/Database. Required to avoid import cycle between this package and the package that implements Value reading.
+// ValueReader is an interface that knows how to read Noms Values, e.g.
+// datas/Database. Required to avoid import cycle between this package and the
+// package that implements Value reading.
 type ValueReader interface {
 	ReadValue(h hash.Hash) Value
 }
 
-// ValueWriter is an interface that knows how to write Noms Values, e.g. datas/Database. Required to avoid import cycle between this package and the package that implements Value writing.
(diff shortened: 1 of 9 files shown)


cmasone-attic:
OK, PTAL

go/types/value_store.go
```diff
@@ -103,32 +144,53 @@ func (lvs *ValueStore) ReadValue(h hash.Hash) Value {
...
 		return r
 	}
+
+	// TODO: It _really_ feels like there should be some refactoring that allows us to only have to walk the refs of |v| once, but I'm hesitant to undertake that refactor right now.
 	hints := lvs.chunkHintsFromCache(v)
-	lvs.bs.SchedulePut(c, height, hints)
-	lvs.set(hash, (*presentChunk)(v.Type()))
-	lvs.valueCache.Drop(hash)
+
+	lvs.pendingMu.Lock()
```
rafael-atticlabs: maybe do this block inside a function so that unlock is guaranteed?

ghost (approved):
LGTM w/ nit.

\o/

JS next?

go/types/value_store.go
cmasone-attic: DOne

PR #3017, opened 2017-01-04
Remember hints for written chunks

rafael-atticlabs:
Prior to this patch, ValueStore only kept a record of where referenced `Refs` originally reside in from chunks read from the server. This ignored the case of a client doing subsequent commits with the same ValueStore (for example, writing multiple states of a map). This was resulting in the server being forced to load a ton of chunks to validate.

### DIFF
--- a/go/types/value_store_test.go
+++ b/go/types/value_store_test.go
@@ -30,7 +30,7 @@ func TestCheckChunksInCache(t *testing.T) {
 
 	b := NewEmptyBlob()
 	cs.Put(EncodeValue(b, nil))
-	cvs.set(b.Hash(), hintedChunk{b.Type(), b.Hash()})
+	cvs.set(b.Hash(), hintedChunk{b.Type(), b.Hash()}, false)
 
 	bref := NewRef(b)
 	assert.NotPanics(func() { cvs.chunkHintsFromCache(bref) })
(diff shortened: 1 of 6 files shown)


cmasone-attic:
Can you add a test that validates this behavior?

js/noms/src/value-store.js
```diff
@@ -213,19 +216,34 @@ class HashCacheEntry {
...
-    this._cache.set(hash.toString(), entry);
+  add(hash: Hash, entry: HashCacheEntry, toPending: boolean) {
+    if (toPending) {
+      this._pending.set(hash.toString(), entry);
+    } else {
+      this._cache.set(hash.toString(), entry);
+    }
   }
 
+  mergePendingHints() {
```
cmasone-attic: Fix whitespace in and after this function

go/types/value_store_test.go
```diff
@@ -30,12 +30,38 @@ func TestCheckChunksInCache(t *testing.T) {
...
+	cvs := newLocalValueStore(cs)
+
+	l := NewList()
+	r := NewRef(l)
+	i := 0
+	for r.Height() == 1 {
+		l = l.Append(Number(i))
+		r = NewRef(l)
+		i++
+	}
```
cmasone-attic: This is cool. It took me a second to figure out what's going on, but now I got it and it's clever

PR #3018, opened 2017-01-04
Few cleanups to type simplification, plus enable it in various jobs

aboodman:


go/types/simplify.go
```diff
@@ -35,8 +35,6 @@ import (
 // Anytime any of the above cases generates a union as output, the same process
 // is applied to that union recursively.
 func makeSimplifiedUnion(in ...*Type) *Type {
-	d.Chk.True(len(in) > 0)
```
aboodman: it turns out we create empty unions. whoops.

go/types/type_cache.go
```diff
@@ -539,11 +539,12 @@ func MakeStructType(name string, fieldNames []string, fieldTypes []*Type) *Type
 }
 
 func MakeUnionType(elemTypes ...*Type) *Type {
-	staticTypeCache.Lock()
```
aboodman: Putting the lock here was too coarse because sometimes makeSimplifiedUnion tries to acquire it again re-entrantly.

PR #3023, opened 2017-01-05
Hack `noms sync` to be able to support syncing from arbitrary paths

aboodman:
Before it could only sync from hashes. This is not a good long-term
approach because it assumes caller can write to source db, which
might not be true.

### DIFF
--- a/cmd/noms/noms_sync.go
+++ b/cmd/noms/noms_sync.go
@@ -81,7 +81,19 @@ func runSync(args []string) int {
 		lastProgressCh <- last
 	}()
 
-	sourceRef := types.NewRef(sourceObj)
+	var sourceRef types.Ref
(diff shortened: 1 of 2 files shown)


aboodman:
cc @cmasone-attic

rafael-atticlabs:
Deferring to @cmasone-attic. He's now the expert on what's safe to do with Flush. This is definitely smelly as it's reading & writing from multiple abstraction layers, but it may be fine for now.

cmasone-attic:
Yeah, exposing Flush like this makes potential for weirdness. I think that Pull() can be modified to handle this properly. What would be the priority of getting that done?

aboodman:
I think it would be good to get to it in the next week or so. It would be
really valuable to be able to sync again during development and we
basically can't do that now.

PR #3025, opened 2017-01-05
Write sequence read-ahead cache to correct index

kalman:


kalman:
CC @ehalpern - it doesn't look like Go has this problem, but... maybe?

js/noms/src/sequence.js
```diff
@@ -126,10 +126,11 @@ export class SequenceCursor<T, S: Sequence<any>> {
   getChildSequence(): Promise<?S> {
     if (this.readAhead && this.sequence.isMeta && !this.childSeqs) {
       // Only readAhead when enabled, for meta sequences.
-      this.childSeqs = [];
+      const childSeqs = [];
       for (let i = this.idx; i < this.sequence.length; i++) {
-        this.childSeqs.push(this.sequence.getChildSequence(i));
+        childSeqs[i] = this.sequence.getChildSequence(i);
```
kalman: Writing to `[i]` is fine because:
```javascript
a = [];
a[1] = 42;
console.log(a); // [undefined, 42]
```

js/noms/src/sequence.js
```diff
@@ -126,10 +126,11 @@ export class SequenceCursor<T, S: Sequence<any>> {
   getChildSequence(): Promise<?S> {
     if (this.readAhead && this.sequence.isMeta && !this.childSeqs) {
       // Only readAhead when enabled, for meta sequences.
-      this.childSeqs = [];
+      const childSeqs = [];
```
kalman: Need this temporary variable because this change makes flow sad without it ("childSeqs may be null or undefined"). I don't know why it didn't have this warning before.

kalman:
The reason I noticed this was because I was observing behavior similar to this, in some cases:
```javascript
m = someNomsMap();
m.has(42); // true
iter = m.iteratorAt(42);
m.next(); // 43!?
```
I tried to figure out why https://github.com/attic-labs/noms/blob/master/js/noms/src/map-test.js#L568 doesn't fail, didn't get there by the end of the day. Any idea?

kalman:
Ah, I think it doesn't fail because the test only has meta-nodes with 2 elements. This means that the read-ahead cache, iterating from `1`, gets:
```
sequence: ['a', 'b']
read-ahead: [read-ahead for 'b', undefined]
```
so when it tries to read `[1]`, the next iterator position - which should be 'b' - it gets `undefined` which falls back to reading it.

If the test had a meta-node of size 3, then we'd get if iterating from `1`:
```
sequence: ['a', 'b', 'c']
read-ahead: [read-ahead for 'b', read-ahead for 'c', undefined]
```
then when it tries to `[1]` it does actually get 'c' not 'b'.

Might be worthwhile seeing if we can change that test... but it'll take a bit of effort to update the rest of the tests. I can look into it tomorrow.

kalman:
Well I made the test change and it still passes, so ¯\_(ツ)_/¯

Issue #3019, opened 2017-01-04
Server should use `GetMany` to load hints


rafael-atticlabs:
For NBS/S3, in particular, this will likely dramatically reduce the latency of beginning to read chunks off the wire for a `writeValue` operation

rafael-atticlabs:
Looking at the go samples running on aws last night, it was apparent that go need the general cursor-based read ahead that js got (in read-ahead mode, preload all children of a meta node).

I prototyped this and saw some improvement, but since go is multi-threaded, our batching strategy is now in our way in the sense that for every logical "ReadMany", we end up sending 6 total requests rather than 1.

@cmasone-attic, Consider implementing `ValueStore.ReadManyValues` which would kind of pull into the client the ability to signal a desire to load multiple values/chunks at once. This would be useful for both this bug as well as general go read-ahead.

PR #3026, opened 2017-01-05
Make server use GetMany to load hinted chunks

cmasone-attic:
Now that we have GetMany, the server can use it directly to let
the chunk-fetching layer figure out the best way to batch up
requests. A small refactor allows ValidatingBatchingSink to directly
update the hint cache instead of relying on logic inside ReadValue
to do it. I made that change because ReadValue now also does a bunch
of other things around caching read values and checking a cache of
'pending Puts' that will never have anything in it when used from
the server.

Toward issue #3019

### DIFF
--- a/go/chunks/dynamo_store.go
+++ b/go/chunks/dynamo_store.go
@@ -122,7 +122,9 @@ func (s *DynamoStore) Get(h hash.Hash) Chunk {
 func (s *DynamoStore) GetMany(hashes hash.HashSet, foundChunks chan *Chunk) {
 	for h, _ := range hashes {
 		c := s.Get(h)
-		foundChunks <- &c
+		if !c.IsEmpty() {
+			foundChunks <- &c
+		}
 	}
 	return
 }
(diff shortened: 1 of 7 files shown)


cmasone-attic:
PTAL

cmasone-attic:
I took this approach rather than creating `ReadManyValues()` because ReadValue() currently caches the Values it reads, which this code doesn't really want, but which consumes memory and cycles.

cmasone-attic:
Ok, PTAL now

go/chunks/dynamo_store.go
```diff
@@ -122,7 +122,9 @@ func (s *DynamoStore) Get(h hash.Hash) Chunk {
 func (s *DynamoStore) GetMany(hashes hash.HashSet, foundChunks chan *Chunk) {
 	for h, _ := range hashes {
 		c := s.Get(h)
-		foundChunks <- &c
+		if !c.IsEmpty() {
```
rafael-atticlabs: ah. yeah. that would be needed.

ghost (approved):
LGTM

PR #3028, opened 2017-01-05
NBS: Create directory for store if it doesn't exist

### LINKED ISSUE #2909: Create NBS directory if absent
Right now if you try to open a NBS store in a directory that doesn't exist, it'll crash. Seems like we should mirror LDB's behavior here. They have an ErrorIfAbsent flag, but it defaults to false, so I think we can just silently create for now and see if the use case arises for ErrorIfAbsent.

cmasone-attic:
Fixes #2909

### DIFF
--- a/go/nbs/block_store_test.go
+++ b/go/nbs/block_store_test.go
@@ -9,6 +9,7 @@ import (
 	"crypto/rand"
 	"io/ioutil"
 	"os"
+	"path/filepath"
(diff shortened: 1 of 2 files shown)


cmasone-attic:
PTAL

cmasone-attic:
jenkins: test this

ghost (approved):
LGTM

PR #3029, opened 2017-01-05
Type accretion for struct fields should use equal

### LINKED ISSUE #2993: Arguably incorrect type accretion
```
	s := types.NewStruct("", types.StructData{
		"foo": types.NewStruct("", types.StructData{
			"bar": types.Number(42),
		}),
	})
	fmt.Println(s.Type().Describe())
	s = s.Set("foo", s.Get("foo").(types.Struct).Set("baz", types.Number(43)))
	fmt.Println(s.Type().Describe())
```

The two lines printed are the same. I expected the second to have `baz`.

Is there a reason this works this way, or it should the check here have been for type equality not subtype?

arv:
We used to use isSubtype but we should just create a more specific type
when we can.

Fixes #2993

### DIFF
--- a/go/marshal/encode_test.go
+++ b/go/marshal/encode_test.go
@@ -686,15 +686,20 @@ func TestEncodeOriginal(t *testing.T) {
 	assert.True(MustMarshal(s).Equals(orig.Set("foo", types.Number(43))))
 
 	// Field type of base are preserved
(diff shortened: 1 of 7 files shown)


aboodman:
lgtm

arv:
jenkins: test this

PR #3032, opened 2017-01-05
Fix sync bug when all chunks are already present in the sink

### LINKED ISSUE #3101: Can we remove `Spec.Spec` ?
I'm not clear why it exists and it prevents people from using the parsed spec struct like a spec, because if they change any of the value of the struct, that field will be out of date.

It doesn't look like it is used for anything.

cmasone-attic:
In the edge case where all source chunks are already present in the
sink AND the sink dataset does not yet exist, the httpBatchStore code
was sending chunks in reverse order.  This patch ensures that,
regardless of how few chunks are sent, any operation that sends chunks
to the server also resets chunk write order.

Fixes #3101


cmasone-attic:
PTAL

also, FYI @aboodman

PR #3035, opened 2017-01-06
Introduce noms version 8. Use it to guard type simplification.

aboodman:


aboodman:
Main change is in version.go (factored out from constants.go). Most of the remaining changes are fallout.

One notable exception to above is that I had to dork around with util/command.go and friends to get this flag to show up in all the noms subcommands without having to modify them all individually.

aboodman:
Working on js... will add to this patch.

js/noms/src/version.js
```diff
@@ -4,4 +4,24 @@
 
 // @flow
 
-export default '7';
+const stable = '7';
+const next = '8';
+let useNext = false;
+
+export default {
```
arv: This is an anti pattern. Export all the functions instead. You might still want to `export default '7';`

arv:
LGTM

PR #3036, opened 2017-01-06
Add ValueStore.ReadManyValues()

### LINKED ISSUE #3019: Server should use `GetMany` to load hints

cmasone-attic:
The more code can use GetMany(), the better performance gets on top of
NBS. To this end, add a call to ValueStore that allows code to read
many values concurrently. This can be used e.g. by read-ahead code
that's navigating prolly trees to increase performance.

Fixes #3019

### DIFF
--- a/go/chunks/chunk_store.go
+++ b/go/chunks/chunk_store.go
@@ -19,8 +19,9 @@ type ChunkStore interface {
 	RootTracker
 }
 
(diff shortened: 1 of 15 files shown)


cmasone-attic:
PTAL

ghost (approved):
nice. LGTM

PR #3037, opened 2017-01-07
Use an env variable instead of flag to switch on vnext.

aboodman:
I'm not usually a fan of environment variables but there are just too many different programs involved, the flag is a major pita.

### DIFF
--- a/cmd/noms/noms.go
+++ b/cmd/noms/noms.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/attic-labs/noms/cmd/util"
-	"github.com/attic-labs/noms/go/version"
 	flag "github.com/juju/gnuflag"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/noms/noms.go
```diff
@@ -68,12 +67,12 @@ func main() {
 	for _, cmd := range commands {
 		if cmd.Name() == args[0] {
 			flags := cmd.Flags()
-			version.RegisterVersionFlags(*flags)
```
aboodman: The command line utilities and programs were reverted back to how they were before my prior change.

aboodman:
cc @rafael-atticlabs

rafael-atticlabs:
heh. LGTM. Just submitted this: https://github.com/attic-labs/noms/issues/3040.

aboodman:
Gonna do this a different way.

PR #3039, opened 2017-01-08
Implement marshal.Unmarshaler interface

kalman:


go/marshal/decode.go
```diff
@@ -305,6 +337,17 @@ func nomsValueDecoder(v types.Value, rv reflect.Value) {
 	rv.Set(reflect.ValueOf(v))
 }
 
+func marshalerDecoder(t reflect.Type) decoderFunc {
+	return func(v types.Value, rv reflect.Value) {
```
kalman: Does this need to go in decoderCache? For that matter, should marshalerEncoder have used the encoderCache?

arv: I don't think needs to go in the cache. There is no computation that can be done ahead of time here.

PR #3042, opened 2017-01-09
Make DeserializeToChan return an error, use it everywhere

cmasone-attic:
Chunk deserialization can run into errors sometimes if, e.g. the
client hangs up during a writeValue request. The old error strategy
worked by throwing a "catchable" error and recovering. That's OK if
you've only got one goroutine, but since the writeValue handler starts
so many goroutines, architecting the code to deal with error handling
by panic/recover is dicey.

Instead, make DeserializeToChan return an error in the more common
failure cases and handle it by passing it over a channel and raising
it from a central place.

### DIFF
--- a/go/datas/http_batch_store.go
+++ b/go/datas/http_batch_store.go
@@ -272,40 +272,16 @@ func (bhcs *httpBatchStore) getRefs(hashes hash.HashSet, batch chunks.ReadBatch)
 		d.Panic("Unexpected response: %s", http.StatusText(res.StatusCode))
 	}
 
-	rl := make(chan struct{}, 16)
-	chunks.Deserialize(reader, &readBatchChunkSink{&batch, &sync.RWMutex{}}, rl)
-}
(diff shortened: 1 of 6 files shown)


cmasone-attic:
PTAL

go/datas/http_batch_store.go
```diff
@@ -272,40 +272,16 @@ func (bhcs *httpBatchStore) getRefs(hashes hash.HashSet, batch chunks.ReadBatch)
...
-		or.Satisfy(&c)
-	}
-	rb.mu.RUnlock()
-
-	rb.mu.Lock()
-	defer rb.mu.Unlock()
-	delete(*(rb.batch), c.Hash())
-}
+	chunkChan := make(chan interface{}, 16)
+	go func() { defer close(chunkChan); chunks.Deserialize(reader, chunkChan) }()
```
rafael-atticlabs: go riddance, readBatchChunkSink!

ghost (approved):
LGTM

PR #3045, opened 2017-01-10
Enable v8 by default

aboodman:
Third attempt at landing this. Now we enable it by default except with an env var that confirms.


go/constants/version.go
```diff
@@ -5,7 +5,29 @@
...
 package constants
 
+import (
+	"fmt"
+	"os"
+)
+
 // TODO: generate this from some central thing with go generate, so that JS and Go can be easily kept in sync
-const NomsVersion = "7"
+const NomsVersion = "7.1"
```
aboodman: I decided to do this so that next stable version could be `8`.

arv:
The failure looks it is due to this PR

go/constants/version.go
arv: Seems OK.

aboodman:
What MD files? I updated the ones I could think of.

aboodman:
jenkins: test this

aboodman:
jenkins: test this

aboodman:
@arv tests pass now.

arv:
> What MD files? I updated the ones I could think of.

I meant that you could include the changes to the docs in this PR but if you want to do that in a different PR that is fine with me.

aboodman:
Oh I actually changed the md files already using the github UI. Let me know
if you see one that still needs fixing.

PR #3046, opened 2017-01-10
Go read-ahead multiple meta-1 levels

### LINKED ISSUE #2915: Un-written p-tree mutations will be ignored by iteration
Looking at the read-ahead code, it occurs to me (and it's my fault for missing this in the review), that the read-ahead doesn't check for the presence of "uncommitted" sequences (sequences which are referenced directly from a meta-tuple).

I haven't tried this, but I suspect if you mutate a prolly tree that was read *out of* a `ValueStore` and then iterate over it, the process will probably crash.

cc @ehalpern

rafael-atticlabs:
Prior to this patch, go both & js employee different read-ahead tactics, with the same result: reading ahead about one meta-level (~150 leaf sequences == ~512k) at a time.

JS cursors with `readAhead` set would preload all children at their current position and forward. Go iter() would spin up a sequence cursor to speed ahead of the current consumed position.

This patch adds:

-The approach taken by JS to go of preloading tree children
-Slightly modified mechanism of advancing ahead of the consumed position.
-Usage of recently landed ReadManyValues() which avoids alot of wire requests for a single chunk when many chunks are needed

Using both mechanisms, go iteration is able to read forward N meta-levels (currently set to 30 or ~16MB).

On my MBP:

Prior this patch
read blob from ldb: 55 MB/s
read blob from ldb over http: 30 MB/s
read blob from nbs: 80 MB/s
read blob from nbs over http: 34 MB/s

After this patch
read blob from ldb: 98 MB/s
read blob from ldb over http: 70 MB/s
read blob from nbs: 315 MB/s
read blob from nbs over http: 170 MB/s

Fixes #2915


go/types/blob.go
```diff
@@ -30,10 +30,9 @@ func NewEmptyBlob() Blob {
 }
 
 // BUG 155 - Should provide Write... Maybe even have Blob implement ReadWriteSeeker
-func (b Blob) Reader() io.ReadSeeker {
-	iter := newSequenceIterator(b.seq, 0)
-	return &BlobReader{b.seq, iter, nil, 0}
-
+func (b Blob) Reader() *BlobReader {
```
rafael-atticlabs: Note that blob.go was (a) returned to its state prior to commit of original go read ahead and (b) added `BlobReader.Copy`

go/types/blob.go
```diff
@@ -128,16 +127,53 @@ func (cbr *BlobReader) Read(p []byte) (n int, err error) {
...
+	}
+	if err == io.EOF && cbr.cursor.idx < cbr.cursor.seq.seqLen() {
 		cbr.currentReader = nil
 		err = nil
 	}
 
 	return
 }
 
+func (cbr BlobReader) Copy(w io.Writer) (n int64) {
```
rafael-atticlabs: This is the new code in Blob.go

rafael-atticlabs:
@arv, please see minor js refactor.

PR #3048, opened 2017-01-10
compactingChunkStore.count() must block on compaction

### LINKED ISSUE #3044: (NBS) Race in compactingChunkSource.count() can corrupt NBS manifest
compactingChunkSource.count() should never return before the compaction is completed. The reason is that it won't know its ultimate size until compaction finishes. In particular it may return a value > 0 when the resulting table will have 0 chunks and thus shouldn't be included.

This was the bug I noticed before and ran across again tonight. The race occurs when a table is compacted which has all duplicate chunks and UpdateRoot() is invoked before compaction finishes.

The check for chunkCount > 0 passes because it returns the size of the memtable. But then the call to hash() waits for compaction which results in an empty table which then get unioned into the manifests table set. After that the manifest can never be opened because there is no table with name `0000000`.

As an aside, I noticed that when an empty table is compacted we actually write the file and remove it. We should probably avoid writing it in the first place.

cmasone-attic:
Compaction not only persists the contents of a memTable, it also
filters out duplicate chunks. This means that calling count() on a
compactingChunkStore before and after compaction completes could lead
to different results. In the case where the memTable contains only
duplicate chunks, this is Very Bad becuase it leads to an non-existent
table winding up in the NBS manifest.

Fixes #3044

### DIFF
--- a/go/nbs/compacting_chunk_source.go
+++ b/go/nbs/compacting_chunk_source.go
@@ -76,9 +76,9 @@ func (ccs *compactingChunkSource) close() error {
 }
 
 func (ccs *compactingChunkSource) count() uint32 {
-	cr := ccs.getReader()
-	d.Chk.True(cr != nil)
-	return cr.count()
+	ccs.wg.Wait()
+	d.Chk.True(ccs.cs != nil)
+	return ccs.cs.count()
 }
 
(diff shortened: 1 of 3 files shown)


cmasone-attic:
PTAL

ghost (approved):
lgtm

PR #3060, opened 2017-01-11
Fix bug in read ahead for sequence.iter

rafael-atticlabs:
The issue is that iter() using readAheadLeafCursors is tied to whether a cursor is created with the `readAhead` flag. However, readAhead flags can *never* be set of "leaf" cursors.

Check the flag on parent cursor instead (if there is no parent, its fine since readAhead wont do anything).

TBR @ehalpern


rafael-atticlabs:
jenkins: test this

ehalpern:
LGTM

PR #3061, opened 2017-01-11
Work around environment variable issue.

arv:
We are using babel-plugin-transform-inline-environment-variables which
replaces `process.env.FOO` with the value of the `FOO` environment
variable at compile time.

However, due to our pipeline we end with something like:

```js
var NAME = 'NOMS_VERSION_NEXT';
process.env[NAME]
```

which does not get replaced in development mode. If it was a const the
transformer could replace it but var bindings can change.


arv:
TBR

PR #3062, opened 2017-01-11
Only request remaining hashes in httpBatchStore.GetMany()

cmasone-attic:
In httpBatchStore.GetMany(), we check our unwritten
puts to see if any of the requested chunks already
exist locally. If any do, we're _supposed_ to remove
their hashes from the set slated to be requested from
the server. That logic was borked.

Fixes https://github.com/attic-labs/attic/issues/503

### DIFF
--- a/go/datas/http_batch_store.go
+++ b/go/datas/http_batch_store.go
@@ -163,7 +163,7 @@ func (bhcs *httpBatchStore) GetMany(hashes hash.HashSet, foundChunks chan *chunk
 	wg := &sync.WaitGroup{}
 	wg.Add(len(remaining))
 	bhcs.requestWg.Add(1)
-	bhcs.getQueue <- chunks.NewGetManyRequest(hashes, wg, foundChunks)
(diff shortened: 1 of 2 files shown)


cmasone-attic:
PTAL

go/datas/http_batch_store.go
```diff
@@ -163,7 +163,7 @@ func (bhcs *httpBatchStore) GetMany(hashes hash.HashSet, foundChunks chan *chunk
 	wg := &sync.WaitGroup{}
 	wg.Add(len(remaining))
 	bhcs.requestWg.Add(1)
-	bhcs.getQueue <- chunks.NewGetManyRequest(hashes, wg, foundChunks)
+	bhcs.getQueue <- chunks.NewGetManyRequest(remaining, wg, foundChunks)
```
rafael-atticlabs: don't  we need a len(remaining) == 0 check here as well?

cmasone-attic: It'd work correctly without doing that, but it's wasteful.

PR #3064, opened 2017-01-12
Increase Chunk locality when writing Values through ValueStore

### LINKED ISSUE #3051: (NBS) order chunk cousins together
Right now, child chunks of a given parent are getting put together in nbs. That means that we can effectively read data in ~512k chunks. The next step would be ordering chunk cousins together (children of parent's siblings). Doing this effectively would allow us to read in multiple MB chunks.

cmasone-attic:
Readahead + NBS benefit greatly when "related" Chunks are close to
each other. The current code did a good job of writing siblings in the
Chunk graph next to each other, but "cousins" (that is, children whose
parents are siblings) might wind up spread quite far apart.  This
patch makes WriteValue hold onto novel Chunks until it sees a
_grandparent_ come through the pipeline. All of that Chunk's queued
grandchildren will be Put at that time.
    
Additionally, ValueStore.Flush() now takes a Hash and flushes all
Chunks that are reachable from the Chunk with that Hash, as opposed
to simply flushing all Chunks to the BatchStore. This means that
there's now no supported way to write orphaned Chunks/Values to a
Database.
    
Fixes #3051

### DIFF
--- a/go/types/list_test.go
+++ b/go/types/list_test.go
@@ -958,17 +958,17 @@ func TestListDiffLargeWithSameMiddle(t *testing.T) {
 	vs1 := newLocalValueStore(cs1)
 	nums1 := generateNumbersAsValues(4000)
 	l1 := NewList(nums1...)
-	ref1 := vs1.WriteValue(l1).TargetHash()
-	vs1.Flush()
-	refList1 := vs1.ReadValue(ref1).(List)
+	hash1 := vs1.WriteValue(l1).TargetHash()
+	vs1.Flush(hash1)
+	refList1 := vs1.ReadValue(hash1).(List)
 
 	cs2 := chunks.NewTestStore()
 	vs2 := newLocalValueStore(cs2)
 	nums2 := generateNumbersAsValuesFromToBy(5, 3550, 1)
 	l2 := NewList(nums2...)
-	ref2 := vs2.WriteValue(l2).TargetHash()
-	vs2.Flush()
-	refList2 := vs2.ReadValue(ref2).(List)
+	hash2 := vs2.WriteValue(l2).TargetHash()
+	vs2.Flush(hash2)
+	refList2 := vs2.ReadValue(hash2).(List)
 
 	// diff lists without value store
 	diff1 := accumulateDiffSplices(l2, l1)
--- a/go/types/map_test.go
+++ b/go/types/map_test.go
@@ -190,8 +190,12 @@ func newMapTestSuite(size uint, expectRefStr string, expectChunkCount int, expec
 				idx := uint64(0)
 				l2.Iter(func(key, value Value) (stop bool) {
 					entry := elems.entries[idx]
-					if !key.Equals(entry.key) || !value.Equals(entry.value) {
-						fmt.Printf("%s != %s or %s != %s", EncodedValue(key), EncodedValue(entry.key), EncodedValue(value), EncodedValue(entry.value))
+					if !key.Equals(entry.key) {
+						fmt.Printf("%d: %s (%s)\n!=\n%s (%s)\n", idx, EncodedValueWithTags(key), key.Hash(), EncodedValueWithTags(entry.key), entry.key.Hash())
+						stop = true
+					}
+					if !value.Equals(entry.value) {
+						fmt.Printf("%s (%s) !=\n%s (%s)\n", EncodedValueWithTags(value), value.Hash(), EncodedValueWithTags(entry.value), entry.value.Hash())
 						stop = true
 					}
 					idx++
--- a/go/types/map_test.go
+++ b/go/types/map_test.go
@@ -6,6 +6,7 @@ package types
 
 import (
 	"bytes"
+	"fmt"
 	"math/rand"
 	"sort"
 	"sync"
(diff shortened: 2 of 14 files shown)


cmasone-attic:
PTAL!

go/types/value_store.go
```diff
@@ -229,35 +242,115 @@ func (lvs *ValueStore) WriteValue(v Value) Ref {
...
-		lvs.pendingPuts[h] = pendingChunk{c, height, hints}
-		v.WalkRefs(func(reachable Ref) {
-			if pc, present := lvs.pendingPuts[reachable.TargetHash()]; present {
+	lvs.setHintsForReachable(v, h, true)
+	lvs.set(h, (*presentChunk)(v.Type()), false)
+	lvs.valueCache.Drop(h)
+	return r
+}
+
+func (lvs *ValueStore) putGrandchildren(v Value, c chunks.Chunk, height uint64, hints Hints) {
```
rafael-atticlabs: suggestion this method be called `bufferChunk.

suggested comment:

`bufferChunk` enqueues `c` (which is the serialization of `v`) within this `ValueStores`. Buffered chunks are flushed progressively to the underlying BatchStore in a way which attempts to locate children and grandchildren sequentially together. The following invariants are retained:

1) For any given chunk currently in the buffer, only direct children of the chunk may also be presently buffered (any grandchildren will have been flushed).

2) The total data occupied by buffered chunks does not exceed `lvs.pendingPutMax'

go/types/value_store.go
```diff
@@ -229,35 +242,115 @@ func (lvs *ValueStore) WriteValue(v Value) Ref {
...
 		})
-	}()
+		return
+	}
 
-	lvs.setHintsForReachable(v, v.Hash(), true)
-	lvs.set(h, (*presentChunk)(v.Type()), false)
-	lvs.valueCache.Drop(h)
-	return r
+	if height > 1 {
```
rafael-atticlabs: // Enforce invariant (1)

go/types/value_store.go
```diff
@@ -229,35 +242,115 @@ func (lvs *ValueStore) WriteValue(v Value) Ref {
...
+			}
+
+			if _, present := lvs.pendingParents[childHash]; present {
+				lvs.pendingPutSize -= uint64(putChildren(childHash))
+				delete(lvs.pendingParents, childHash)
+			}
+		})
+	}
+
+	for lvs.pendingPutSize > lvs.pendingPutMax {
```
rafael-atticlabs: // Enforce invariant (2)

go/types/value_store.go
```diff
@@ -229,35 +242,115 @@ func (lvs *ValueStore) WriteValue(v Value) Ref {
...
+				lvs.pendingPutSize -= put(reachable.TargetHash(), pc)
+			}
+		})
+		delete(lvs.pendingParents, root) // If not present, this is idempotent
+		lvs.pendingPutSize -= put(root, pc)
+
+		// Merge in pending hints
+		lvs.cacheMu.Lock()
+		defer lvs.cacheMu.Unlock()
+		for h, entry := range lvs.pendingHints {
```
rafael-atticlabs: this doesn't look right. it seems like the thing we should be doing here is flushing all hints *except* those for chunks which will remaining in `pendingPuts` after this call.

i would have expected this to be

```
if _, present := lvs.pendingPuts[h]; !present {
  lvs,hintCache[h] = entry
  delete(lvs.pendingHints, h)
} else {
  // no else.
}
```

go/types/list_test.go
```diff
@@ -958,17 +958,17 @@ func TestListDiffLargeWithSameMiddle(t *testing.T) {
 	vs1 := newLocalValueStore(cs1)
 	nums1 := generateNumbersAsValues(4000)
 	l1 := NewList(nums1...)
-	ref1 := vs1.WriteValue(l1).TargetHash()
-	vs1.Flush()
-	refList1 := vs1.ReadValue(ref1).(List)
+	hash1 := vs1.WriteValue(l1).TargetHash()
+	vs1.Flush(hash1)
```
rafael-atticlabs: can you remind me why we need `Flush` for these test cases to work? It seems like reading a given value out of a ValueStore should always work if you've just written it (i.e. without flushing). I must be forgetting something....

rafael-atticlabs: Is it because we are asserting the number of reads/writes below?

go/types/map_test.go
```diff
@@ -181,12 +182,22 @@ func newMapTestSuite(size uint, expectRefStr string, expectChunkCount int, expec
 			expectPrependChunkDiff: expectPrependChunkDiff,
 			expectAppendChunkDiff:  expectAppendChunkDiff,
 			validate: func(v2 Collection) bool {
+				if v2.Len() != uint64(elems.entries.Len()) {
```
rafael-atticlabs: what happened here? why did this test need to change?

ghost (commented):
BTU

go/types/map_test.go
cmasone-attic: So...This test fails on ToT, but very infrequently. I've added some additional logging code that shows exactly what's failing, and it's really confusing. As far as I can tell, two Structs with the same Type, contents and Hash fail the equality check. I have no idea how that's possible, though I assume a deeply insidious race :-/

ghost (approved):
LGTM

PR #3067, opened 2017-01-12
Parallelize walk.Walk

rafael-atticlabs:
This re-implements walk.Walk to traverse "down" prolly trees, loading entire subsequent levels with large batches of ReadManyValues.

### DIFF
--- a/go/types/collection.go
+++ b/go/types/collection.go
@@ -8,5 +8,6 @@ type Collection interface {
 	Value
 	Len() uint64
 	Empty() bool
+	IsLeaf() bool
 	sequence() sequence
 }
--- a/go/walk/walk.go
+++ b/go/walk/walk.go
@@ -6,46 +6,84 @@
 package walk
 
 import (
-	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
 	"github.com/attic-labs/noms/go/types"
 )
 
 type SkipValueCallback func(v types.Value) bool
 
+// WalkValues loads prolly trees progressively by walking down the tree. We don't wants to invoke
+// the value callback on internal sub-trees (which are valid values) because they are not logical
+// values in the graph
+type valueRec struct {
+	v  types.Value
+	cb bool
+}
+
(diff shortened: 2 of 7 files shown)


go/types/collection.go
```diff
@@ -8,5 +8,6 @@ type Collection interface {
 	Value
 	Len() uint64
 	Empty() bool
+	IsLeaf() bool
```
cmasone-attic: Sort of unfortunate, but meh

go/walk/walk.go
```diff
@@ -6,46 +6,79 @@
...
 
 import (
-	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
 	"github.com/attic-labs/noms/go/types"
 )
 
 type SkipValueCallback func(v types.Value) bool
 
+// WalkValues loads prolly trees progressively by walking down the tree. We don't walk to invoke
```
cmasone-attic: You mean "We don't _want_ to invoke...", right?

go/walk/walk.go
```diff
@@ -6,46 +6,79 @@
...
-		v := vr.ReadValue(target)
-		if v == nil {
-			d.Chk.Fail("Attempt to visit absent ref:%s", target.String())
-			return
+
+		hs := hash.HashSet{}
+		for h, _ := range refs {
+			if _, ok := visited[h]; !ok {
+				hs[h] = struct{}{}
+				visited[h] = struct{}{}
```
cmasone-attic: HashSet has an Insert method you should use

go/types/collection.go
rafael-atticlabs: yeah, the other option was moving this into `types`. I figured you'd prefer this.

go/walk/walk.go
rafael-atticlabs: yes. thank you.

go/walk/walk.go
rafael-atticlabs: done

Issue #3068, opened 2017-01-12
Implement parallel JS walk


rafael-atticlabs:
JS: https://github.com/attic-labs/noms/commit/3b3835492c1e91af78008082afd77a1753488997

PR #3070, opened 2017-01-12
(JS) Implement parallel walk io

rafael-atticlabs:
Fixes: #3068

### DIFF
--- a/js/noms/package.json
+++ b/js/noms/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@attic/noms",
   "license": "Apache-2.0",
-  "version": "67.1.0",
+  "version": "68.0.0",
   "description": "Noms JS SDK",
   "repository": "https://github.com/attic-labs/noms/tree/master/js/noms",
   "main": "dist/commonjs/noms.js",
--- a/js/noms/src/value.js
+++ b/js/noms/src/value.js
@@ -11,7 +11,7 @@ import type {Type} from './type.js';
 import type Ref from './ref.js';
 import type {ValueReader} from './value-store.js';
 
-export type ValueCallback = (v: Value) => ?void | Promise<?void>;
+export type ValueCallback = (v: Value) => void | Promise<void>;
 
 /**
  * ValueBase is the base class for non primitive Noms values.
--- a/js/noms/src/value.js
+++ b/js/noms/src/value.js
@@ -9,8 +9,9 @@ import type {primitive} from './primitives.js';
 import {ensureHash} from './get-hash.js';
 import type {Type} from './type.js';
 import type Ref from './ref.js';
-import type {WalkCallback} from './walk.js';
 import type {ValueReader} from './value-store.js';
+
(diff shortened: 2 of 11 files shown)


js/noms/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@attic/noms",
   "license": "Apache-2.0",
-  "version": "67.1.0",
+  "version": "68.0.0",
```
rafael-atticlabs: Note I bumped the major version since the semantics changed on ValueBase.walkValues

rafael-atticlabs:
jenkins: test this

js/noms/src/walk.js
```diff
@@ -18,24 +24,78 @@ export type WalkCallback = (v: Value) => ?boolean | Promise<?boolean>;
  * this node's children.
  *
  * If `cb` returns undefined or `Promise.resolve()`, the default is to continue recursing (`false`).
```
arv: This comment is not correct any more.

js/noms/src/walk.js
```diff
@@ -18,24 +24,78 @@ export type WalkCallback = (v: Value) => ?boolean | Promise<?boolean>;
...
+ *
  */
+
+interface ValueRec {
```
arv: Use type here? Also, what is cb?

```js
type ValueRec = {
  value: Value,
  done: boolean,
};
```

arv: cb -> needsCallback?

rafael-atticlabs: done

js/noms/src/walk.js
rafael-atticlabs: good catch. now it is (switched to existing semantics).

js/noms/src/value.js
```diff
@@ -9,9 +9,10 @@ import type {primitive} from './primitives.js';
 import {ensureHash} from './get-hash.js';
 import type {Type} from './type.js';
 import type Ref from './ref.js';
-import type {WalkCallback} from './walk.js';
 import type {ValueReader} from './value-store.js';
 
+export type ValueCallback = (v: Value) => ?void | Promise<?void>;
```
arv: This type is strange. `?T` means `null | void | T` so `?void` is the same as `null | void | void` Do we want to allow `null` or not?

js/noms/src/walk.js
```diff
@@ -18,24 +24,84 @@ export type WalkCallback = (v: Value) => ?boolean | Promise<?boolean>;
...
-    case 'string':
-      return;
-  }
+  while (values.length > 0 || refs.length > 0) {
+    // Visit all values located *within* loaded chunks.
+    while (values.length > 0) {
+      const rec = values.pop();
+      if (rec.needsCallback) {
+        let skip = cb(rec.v);
+        if (skip && skip !== true) {
```
arv: `skip` is either undefined or null so `skip !== true` is always true.

PR #3073, opened 2017-01-13
Add DateTime util struct

### LINKED ISSUE #2970: Marshal: Provide a way to marshal time.Time
Idea is to do add Marshaler and Unmarshaler interface

```go
type Marshaler interface {
        MarshalNoms() (types.Value, error)
}

type Unmarshaler interface {
        UnmarshalNoms(types.Value) error
}
```

Similar to how encoding/json does this:

https://golang.org/pkg/encoding/json/#Marshaler
https://golang.org/pkg/encoding/json/#Unmarshaler

arv:
This allows us to consistently marshal back and forth between time.Time
and a Noms struct.

Fixes #2970

### DIFF
--- a/go/util/datetime/date_time.go
+++ b/go/util/datetime/date_time.go
@@ -1,0 +1,50 @@
+// Copyright 2017 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Package datetime implements marshalling of Go DateTime values into Noms structs
+// with type DateTimeType.
+package datetime
+
+import (
+	"math"
+	"time"
+
+	"github.com/attic-labs/noms/go/marshal"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// DateTime is an alias for time.Time that allows us to marshal date time to
+// Noms.
+type DateTime time.Time
+
+// DateTimeType is the Noms type used to represent date time objects in Noms.
+var DateTimeType = types.MakeStructTypeFromFields("DateTime", types.FieldMap{
+	"secSinceEpoch": types.NumberType,
+})
+
+// MarshalNoms makes DateTime implement marshal.Marshaler and it makes
+// DateTime marshal into a Noms struct with type DateTimeType.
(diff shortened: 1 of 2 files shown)


arv:
Should this be in /go/util/datetime? is there a better place?

I still think that use rfc 3339 might have been better since that is human readable.

go/util/datetime/date_time.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/attic-labs/noms/go/marshal"
+	"github.com/attic-labs/noms/go/types"
+)
+
+// DateTime is an alias for time.Time that allows us to marshal date time to
+// Noms.
+type DateTime time.Time
+
+// DateTimeType is the Noms type used to represent date time objects in Noms.
```
aboodman: Might want to indicate somewhere that it is valid for secSinceEpoch to be fractional.

aboodman:
lgtm

aboodman:
Is rfc 3339 completely unambiguous or are there multiple allowed serializations? Because if the latter I think that we should use seconds.

kalman:
Multiple allowed serializations, most notably based on timezone, but I think you can also leave out trailing 0s or something.

One idea would be to enforce the timezone is always `Z` which is the special character they use to denote UTC.

kalman:
The same program in Go:
```go
package main

import (
	"fmt"
	"time"
)

func main() {
	fmt.Println(time.Now().UTC().Format(time.RFC3339))
}
```
outputs `2017-01-14T00:10:39Z`

Javascript:
```javascript
new Date().toISOString()
```
outputs `2017-01-14T00:10:53.258Z`

It looks like we could modify Go's `time.RFC3339` to include milliseconds then it would be... ok?

arv:
I'm uploading  an rfc3339 pr too

aboodman:
The problem is that nothing anywhere enforces it. We don't have anyplace in Noms to put such policies. I think it would be better to just use seconds since then there's nothing to enforce.

We can modify the HRS to print it out in RFC3339. Same w/ Splore.

Note that other databases always use some kind of timestamp for their time representation, presumably for same reason.

kalman:
SGTM

arv:
@aboodman convinced me against rfc3339

PR #3074, opened 2017-01-14
Walk avoids blobs

rafael-atticlabs:
Recursive walk now avoids loading child chunks of multi-level blobs.

Fixes: #3069

Note, I had been resisting moving Walk into types, but this patch pushed me over the edge. That's why `IsLeaf` disappeared from `Collection`. `Walk` just needs internal knowledge from types to do its job correctly and efficiently.

I recommend reviewing this patch as individual commits.


arv:
lg minus failure... let me know when it works again.

PR #3081, opened 2017-01-18
Ref count HTTP batch stores in the spec code

kalman:
This prevents the same Database being created more than once.

### DIFF
--- a/js/noms/src/spec.js
+++ b/js/noms/src/spec.js
@@ -17,7 +17,7 @@ import RefCountingBatchStore from './ref-counting-batch-store.js';
 
 // Cache of database URL -> Database object. It's reasonable for clients to
 // open a spec without worrying about whether the Database is already open.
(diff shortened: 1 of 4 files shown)


js/noms/src/spec.js
```diff
@@ -6,12 +6,26 @@
...
 import Dataset, {datasetRe} from './dataset.js';
 import Database from './database.js';
 import HttpBatchStore, {DEFAULT_MAX_READS} from './http-batch-store.js';
 import MemoryStore from './memory-store.js';
 import type Value from './value.js';
+import RefCountingBatchStore from './ref-counting-batch-store.js';
+
+// Cache of database URL -> Database object. It's reasonable for clients to
+// open a spec without worrying about whether the Database is already open.
+const batchStores: {[key: string]: RefCountingBatchStore} = {};
```
arv: Object.create(null)

PR #3092, opened 2017-01-19
Switch to Yarn

arv:
brew install yarn


arv:
Succes... Ready for review

samples/js/counter/.eslintrc.js
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = {
+  rules: {
+    // Allow console
+    'no-console': 0,
+    // Used to distinguish between user errors and exceptions.
+    'no-throw-literal': 0,
+  },
+  extends: 'noms',
+};
```
ehalpern: It's unfortunate that .babelrc, .eslintrc.js and .flowconfig have to be repeated everywhere. Is there any benefit to sym-linking to single defaults or is the additional complexity of sharing not worth it?

arv: We can go back to symlinking. I initially thought it was silly to symlink `.babelrc` and `.eslintrc.js` since they are pretty much just a single line (like a symlink 🙂) anyway.

arv:
@ehalpern Do you like  this better?

PR #3093, opened 2017-01-19
Make database root a Map<String, Ref<Value>>

### LINKED ISSUE #2869: The root chunk is huge
Currently the root map encodes the type of every dataset. Clients must download this chunk even though nobody ever really uses it.

I think that the thing to do here is to just change Noms to have this root map have the type `Map<string, Ref<Value>>`.

cmasone-attic:
Since the head of each dataset can have an arbitrarily complex
type, type accretion leads the Datasets map at the root of the
DB to become very large. This type info isn't really very useful
at that level either. So, get rid of it by making this map be
from String -> Ref<Value>.

Fixes #2869


cmasone-attic:
PTAL. I've included @arv because he helped me out with JS type stuff yesterday. @rafael-atticlabs, when I spoke to @aboodman about this patch yesterday he suggested I have you review it because you're in a better position to decide (with me) if this is the right approach here.

js/noms/src/database.js
```diff
@@ -63,7 +65,30 @@ export default class Database {
    * id in the above Datasets Map.
    */
   getDataset(id: string): Dataset {
```
arv: How about something along the lines of:

```js
getDataset(id: string): Dataset {
  const getRef = async () => {
    const sets = await this.datasets();
    const valRef = sets.get(id);
    ...
  };
  return new Dataset(this, id, getRef());
}
```

all those then callbacks makes the code harder to follow.

cmasone-attic: Done

go/datas/remote_database_handlers.go
```diff
@@ -380,14 +380,14 @@ func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, cs c
...
-	// Ensure that proposed new Root is a Map and, if it has anything in it, that it's <String, <Ref<Commit>>
+	// Ensure that proposed new Root is a Map and, if it has anything in it, that it's <String, <Ref<Value>>
 	v := types.DecodeValue(c, nil)
 	if v.Type().Kind() != types.MapKind {
 		d.Panic("Root of a Database must be a Map")
 	}
 	m := v.(types.Map)
-	if !m.Empty() && !isMapOfStringToRefOfCommit(m) {
-		d.Panic("Root of a Database must be a Map<String, Ref<Commit>>, not %s", m.Type().Describe())
+	if !m.Empty() && !isMapOfStringToRefOfValue(m) {
```
rafael-atticlabs: This is a bummer and kind of gets at the question of what invariants are we trying to enforce. To be precise, this no longer offers assurance that the entire database hasn't be corrupted by setting the root to be a `Map<String, Ref<Garbage>>`.

If want to retain the same level of integrity, I think the obvious thing to do is to ALSO diff the old map to the new map and assert that all *new* map `Ref<Values>` are  actually`Ref<Commit>`.

@aboodman Thoughts?

go/types/value_store.go
```diff
@@ -353,8 +353,8 @@ func (lvs *ValueStore) checkChunksInCache(v Value, readValues bool) Hints {
 		}
 
 		targetType := getTargetType(reachable)
-		if !entry.Type().Equals(targetType) {
-			d.Panic("Value to write contains ref %s, which points to a value of a different type: %+v != %+v", reachable.TargetHash(), entry.Type(), targetType)
+		if entry.Type().Kind() != ValueKind && targetType.Kind() != ValueKind && !entry.Type().Equals(targetType) {
```
rafael-atticlabs: I think we need the old behavior here. Which is that if the type is more specific than the information we have, we're forced to load the value and check its actual type.

cmasone-attic: I'm not sure that we ever actually did that. There was a comment in the JS saying we should, but looking at the code I just changed, I don't think that was actually happening.

go/datas/remote_database_handlers.go
aboodman: Agree with raf. I think the server should protect itself against clients replacing map values with garbage.

go/types/value_store.go
cmasone-attic: done

go/datas/remote_database_handlers.go
```diff
@@ -374,20 +374,31 @@ func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, cs c
...
+			d.Panic("Can't UpdateRoot from a non-present Chunk")
+		}
+
+		datasets = lastVal.(types.Map)
+	}
+
 	// Ensure that proposed new Root is a Map and, if it has anything in it, that it's <String, <Ref<Commit>>
-	v := types.DecodeValue(c, nil)
-	if v.Type().Kind() != types.MapKind {
+	if proposed.Type().Kind() != types.MapKind {
```
rafael-atticlabs: `if m, ok := proposed.(types.Map); ok`

we don't have wrappers anymore =-)

cmasone-attic:
@arv, can you take another look at JS. I made the change you asked for

arv:
Still LGTM

PR #3094, opened 2017-01-20
Retry NomsBlockStore.Flush() with exponential backoff

### LINKED ISSUE #3089: Commits to NBS fail if root got updated by someone else
I have two processes:
p1 opens a database with an NBS store
p2 opens a database with an NBS store and commits
p1 tries to commit and fails because the root has been updated

this is preventing me from using NBS in the testing as I would use an HTTP store in production.

cmasone-attic:
Fixes #3089

### DIFF
--- a/go/nbs/store.go
+++ b/go/nbs/store.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"sort"
 	"sync"
+	"time"
 
 	"github.com/attic-labs/noms/go/chunks"
 	"github.com/attic-labs/noms/go/constants"
@@ -17,6 +18,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/jpillora/backoff"
 )
 
 // The root of a Noms Chunk Store is stored in a 'manifest', along with the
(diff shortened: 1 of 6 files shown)


go/nbs/store.go
```diff
@@ -313,6 +315,13 @@ func (nbs *NomsBlockStore) AddHints(hints types.Hints) {
 }
 
 func (nbs *NomsBlockStore) Flush() {
-	success := nbs.UpdateRoot(nbs.root, nbs.root)
-	d.Chk.True(success)
+	b := &backoff.Backoff{
+		Min:    50 * time.Millisecond,
+		Max:    10 * time.Second,
```
cmasone-attic: I have no real basis for this maximum. I set the minimum at 50ms because I vaguely remembered that being the same order of magnitude as a dynamoDB write. Please feel free to suggest other numbers.

cmasone-attic:
PTAL

go/nbs/store.go
rafael-atticlabs: Think a lower min might make sense. Seems like we're mainly concerned with two cases:

1) Someone updated behind your back a long time ago, but isn't about to try to update any time soon. In this case *any* min will suffice, since we'll succeed on the next UpdateRoot

2) There's someone else whose about to try the same thing and we want to avoid a three-stoges routine of going exactly back & forth. I suspect that 10s of microseconds would accomplish this as long as we keep jitter.

The only downside I know of to picking too low is that we get a few first back & forths with another Flusher -- which means read load on the dynamo table.

So to counter your from air proposal, here's another: 128 microseconds.

PR #3104, opened 2017-01-23
Introduce walk.WalkDifferentStructs

rafael-atticlabs:
This patch adds a static function which can walk graphs looking for (and diffing) two structs. It uses type information to avoid traversing sub-values which can't contain structs. It also uses a similar approach as sync to avoid visiting common sub-chunk-graphs.

### DIFF
--- a/go/types/walk.go
+++ b/go/types/walk.go
@@ -109,14 +109,14 @@ func mightContainStructs(t *Type) (mightHaveStructs bool) {
 	return
 }
 
-// Efficiently returns the set of |removed| (reachable from |last|, but not |current|) and |added|
-// (reachable from |current| but not |last|) types.Structs. Internally avoids visiting chunks which
-// are common to both value graphs.
+// WalkDifferentStructs Efficiently returns the set of |removed| (reachable from |last|, but not
+// |current|) and |added| (reachable from |current| but not |last|) types.Structs. Internally
+// avoids visiting chunks which are common to both value graphs.
 func WalkDifferentStructs(last, current Value, vr ValueReader) (added, removed map[hash.Hash]Struct) {
 	added, removed = map[hash.Hash]Struct{}, map[hash.Hash]Struct{}
 
 	oldRefs, currentRefs := RefByHeight{}, RefByHeight{}
-	oldValues, currentValues := []Value{}, []Value{}
+	oldValues, currentValues := ValueSlice{}, ValueSlice{}
 	if last != nil {
 		if current != nil && last.Equals(current) {
 			return // same values
@@ -128,7 +128,7 @@ func WalkDifferentStructs(last, current Value, vr ValueReader) (added, removed m
 		currentValues = append(currentValues, current)
 	}
 
-	walkLocalValues := func(values *[]Value, structs map[hash.Hash]Struct, refs *RefByHeight) {
+	walkLocalValues := func(values *ValueSlice, structs map[hash.Hash]Struct, refs *RefByHeight) {
 		for len(*values) > 0 {
 			v := (*values)[len(*values)-1]
 			*values = (*values)[:len(*values)-1]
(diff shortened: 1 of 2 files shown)


go/types/walk.go
```diff
@@ -91,3 +95,156 @@ func WalkValues(target Value, vr ValueReader, cb SkipValueCallback) {
 		}
 	}
 }
+
+func mightContainStructs(t *Type) (mightHaveStructs bool) {
```
rafael-atticlabs: @aboodman Please look at this.

go/types/walk_test.go
```diff
@@ -249,3 +250,118 @@ func (suite *WalkTestSuite) SetupTest() {
...
+
+	l1 = bl.Append(s2, s4)
+	l2 = bl.Append(s1, s2, s3)
+
+	suite.AssertDiffs(l1, l2, []Value{s1, s3}, []Value{s4})
+	suite.AssertDiffs(l1, l1, []Value{}, []Value{})
+	suite.AssertDiffs(l2, l1, []Value{s4}, []Value{s1, s3})
+
+	// Big, committed collections of structs
+	l1 = suite.vs.ReadValue(suite.vs.WriteValue(l1).TargetHash()).(List)
```
cmasone-attic: I think you mean "flushed" rather than "committed"? Given the grandchild strategy, I think that we won't even have flushed any parts of `l1` or `l2`. They're only of height 2, right?

If you actually are testing the case where chunks of these prolly trees have gotten written out to the batch store layer, then you'll want to do something like 
```golang
l1Hash := suite.vs.WriteValue(l1).TargetHash()
suite.vs.Flush(l1Hash)
l1 = suite.vs.ReadValue(l1Hash).(List)
```

go/types/walk.go
```diff
@@ -91,3 +95,156 @@ func WalkValues(target Value, vr ValueReader, cb SkipValueCallback) {
...
+	}
+
+	t.WalkValues(func(v Value) {
+		mightHaveStructs = mightHaveStructs || mightContainStructs(v.(*Type))
+	})
+
+	return
+}
+
+// Efficiently returns the set of |removed| (reachable from |last|, but not |current|) and |added|
```
rafael-atticlabs: done

go/types/walk.go
```diff
@@ -91,3 +95,156 @@ func WalkValues(target Value, vr ValueReader, cb SkipValueCallback) {
...
+// (reachable from |current| but not |last|) types.Structs. Internally avoids visiting chunks which
+// are common to both value graphs.
+func WalkDifferentStructs(last, current Value, vr ValueReader) (added, removed map[hash.Hash]Struct) {
+	added, removed = map[hash.Hash]Struct{}, map[hash.Hash]Struct{}
+
+	oldRefs, currentRefs := RefByHeight{}, RefByHeight{}
+	oldValues, currentValues := []Value{}, []Value{}
```
rafael-atticlabs: done

go/types/walk.go
```diff
@@ -91,3 +95,156 @@ func WalkValues(target Value, vr ValueReader, cb SkipValueCallback) {
...
+		// Iterate over all "loaded values". Record structs and refs of unloaded values.
+		walkLocalValues(&oldValues, removed, &oldRefs)
+		walkLocalValues(&currentValues, added, &currentRefs)
+
+		if len(oldRefs) == 0 && len(currentRefs) == 0 {
+			continue
+		}
+
+		sort.Sort(oldRefs)
+		sort.Sort(currentRefs)
```
rafael-atticlabs: done

go/types/walk_test.go
rafael-atticlabs: fixed. sadly there are all three things here.

1) Uncommitted sequences are those which have yet to be handed to a ValueStore
2) Unflushed are those which haven't be handed to a BatchStore
3) Neither are those which are now in ChunkStore

rafael-atticlabs: but fixed

PR #3105, opened 2017-01-23
Fix warning of unmet peer dependency

mikegray:


js/webpack-config/package.json
```diff
@@ -6,6 +6,7 @@
   "version": "2.3.0",
   "main": "index.js",
   "dependencies": {
+    "babel-core": "^6.0.0",
```
arv: not 6.0.0 I assume? We should just have gotten the latest one if you did:

`yarn add babel-core`

mikegray:
PTAL

PR #3106, opened 2017-01-23
Switch Map/Set Diff to use top-down by default

cmasone-attic:
The only thing that wants what we used to call the "best" diff
algorithm is the command-line tools. Non-interactive programs all want
the algorithm that finishes up fastest, which is top-down.

Fixes https://github.com/attic-labs/attic/issues/627

### DIFF
--- a/go/types/map.go
+++ b/go/types/map.go
@@ -58,15 +58,28 @@ func NewStreamingMap(vrw ValueReadWriter, kvs <-chan Value) <-chan Map {
 	return outChan
 }
 
-// Computes the diff from |last| to |m| using "best" algorithm, which balances returning results early vs completing quickly.
+// Diff computes the diff from |last| to |m| using the top-down algorithm,
+// which completes as fast as possible while taking longer to return early
+// results than left-to-right.
 func (m Map) Diff(last Map, changes chan<- ValueChanged, closeChan <-chan struct{}) {
+	if m.Equals(last) {
+		return
+	}
+	orderedSequenceDiffTopDown(last.seq, m.seq, changes, closeChan)
+}
+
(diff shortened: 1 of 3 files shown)


go/types/map.go
```diff
@@ -58,15 +58,28 @@ func NewStreamingMap(vrw ValueReadWriter, kvs <-chan Value) <-chan Map {
...
 func (m Map) Diff(last Map, changes chan<- ValueChanged, closeChan <-chan struct{}) {
 	if m.Equals(last) {
 		return
 	}
+	orderedSequenceDiffTopDown(last.seq, m.seq, changes, closeChan)
+}
+
+// DiffHybrid computes the diff from |last| to |m| using a hybrid algorithm
+// which balances returning results early vs completing quickly, if possible.
+func (m Map) DiffHybrid(last Map, changes chan<- ValueChanged, closeChan <-chan struct{}) {
```
cmasone-attic: At first, I tried to avoid hanging a method for this on `Map` and `Set`, but the problem is that the function signature for List Diffs is just different enough that you can't easily make a function that takes two Collections and runs diff on them. What I tried looked like

```golang
func DiffCollections(last, current Collection, changes interface{}, stopChan <-chan struct{}) {
   ...
}
```

and I had to do a bunch of type-asserting on `changes` depending on what kinds of Collection you'd passed in...AND whether you'd passed in a `chan<-` or just a `chan`. And, since List doesn't even have a top-down, left-right, or best anyway it all felt kinda pointless :-/

cmasone-attic:
PTAL

rafael-atticlabs:
LGTM

PR #3109, opened 2017-01-24
NBS UpdateRoot() returns false when last != nbs.Root()

cmasone-attic:
This code initially panicked in this case, because there didn't used
to be a reasonable way that a caller might wind up trying to update
away from a Root that didn't match NomsBlockStore's internal
bookkeeping. Now, given the new Flush() behavior there is. So, just
return false and allow the caller to take appropriate action.

Towards #3089


cmasone-attic:
PTAL

There's detailed explanation of what Dan's seeing at https://github.com/attic-labs/noms/issues/3089#issuecomment-274892962

rafael-atticlabs:
LGTM

PR #3111, opened 2017-01-24
Fix cache/value store caching

rafael-atticlabs:
* Fixes a long standing bug in which the RemoteBatchStore is accidentally caching all chunks
* ValueStore's value cache now stores `Promise<?Value>` so that concurrent `readValues` of the same value can share a single decoding
* Removes the debug-only chunk hash check which keeps tripping up perf investigations

### DIFF
--- a/js/noms/src/value-store.js
+++ b/js/noms/src/value-store.js
@@ -197,7 +197,6 @@ class SizeCache<T> {
     let entry = this._cache.get(key);
     if (entry) {
       this._cache.delete(key);
-      entry.size.then(() => this.expire());
     }
 
     entry = new CacheEntry(size, value);
--- a/js/noms/src/value-store.js
+++ b/js/noms/src/value-store.js
@@ -64,29 +64,28 @@ export default class ValueStore {
       resolveSize = resolveFn;
     }));
 
-    const valueP = Promise.resolve().then(() => {
-      const pc = this._pendingPuts.get(hash.toString());
-      const chunkP = pc ? Promise.resolve(pc.c) : this._bs.get(hash);
-      return chunkP.then(chunk => {
-        resolveSize(chunk.data.byteLength);
+    const pc = this._pendingPuts.get(hash.toString());
+    const chunkP = pc ? Promise.resolve(pc.c) : this._bs.get(hash);
 
-        if (chunk.isEmpty()) {
-          this._knownHashes.addIfNotPresent(hash, new HashCacheEntry(false));
-          return null;
-        }
+    const valueP = chunkP.then(chunk => {
+      resolveSize(chunk.data.byteLength);
 
-        const v = decodeValue(chunk, this);
-        this._knownHashes.cacheChunks(v, hash, false);
-        // hash is trivially a hint for v, so consider putting that in the cache.
-        // If we got to v by reading some higher-level chunk, this entry gets dropped on
-        // the floor because r already has a hint in the cache. If we later read some other
-        // chunk that references v, cacheChunks will overwrite this with a hint pointing to that
-        // chunk. If we don't do this, top-level Values that get read but not written -- such as
-        // the existing Head of a Database upon a Commit -- can be erroneously left out during a
-        // pull.
-        this._knownHashes.addIfNotPresent(hash, new HashCacheEntry(true, getTypeOfValue(v), hash));
-        return v;
-      });
+      if (chunk.isEmpty()) {
+        this._knownHashes.addIfNotPresent(hash, new HashCacheEntry(false));
+        return null;
+      }
+
+      const v = decodeValue(chunk, this);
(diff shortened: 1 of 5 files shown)


js/noms/src/value-store.js
```diff
@@ -193,24 +201,34 @@ class SizeCache<T> {
...
-  add(hash: Hash, size: number, value: ?T): void {
+  add(hash: Hash, size: Promise<number>, value: T): void {
     const key = hash.toString();
-    if (this._cache.has(key)) {
+    let entry = this._cache.get(key);
+    if (entry) {
       this._cache.delete(key);
-    } else {
-      this._size += size;
+      this._size -= entry.size;
```
kalman: I think this is problematic if the promise hasn't resolved yet? `entry.size` will be 0, but when it does resolve it'll add its resolvedSize?

js/noms/src/value-store.js
```diff
@@ -53,36 +53,44 @@ export default class ValueStore {
   }
 
   // TODO: This should return Promise<?Value>
-  async readValue(hash: Hash): Promise<any> {
+  readValue(hash: Hash): Promise<any> {
     const entry = this._valueCache.entry(hash);
```
kalman: Seems like this should be using `this._valueCache.get`, since the entry isn't used. Then again maybe `Cache.get` isn't actually used anywhere and you should delete that.

js/noms/src/value-store.js
```diff
@@ -143,15 +151,15 @@ class PendingChunk {
 interface Cache<T> {  // eslint-disable-line no-undef
   entry(hash: Hash): ?CacheEntry<T>;  // eslint-disable-line no-undef
   get(hash: Hash): ?T;  // eslint-disable-line no-undef
-  add(hash: Hash, size: number, value: T): void;  // eslint-disable-line no-undef
+  add(hash: Hash, size: Promise<number>, value: T): void;  // eslint-disable-line no-undef
```
kalman: I'm observing that `Cache` is only ever used with `T` as a `Promise<?Value>`

and making size a Promise makes it look weird that value isn't also a Promise.

It's probably not worth going down this route, but you could delete the generics on this and CacheEntry and make it `add(Hash, Promise<number>, Promise<Value>)` etc.

js/noms/src/value-store.js
rafael-atticlabs: removed get

js/noms/src/value-store.js
```diff
@@ -53,36 +53,44 @@ export default class ValueStore {
...
-    }
+    let resolveSize;
+    const sizeP: Promise<number> = new Promise((resolveFn => {
+      resolveSize = resolveFn;
+    }));
+
+    const valueP = Promise.resolve().then(() => {
+      const pc = this._pendingPuts.get(hash.toString());
+      const chunkP = pc ? Promise.resolve(pc.c) : this._bs.get(hash);
+      return chunkP.then(chunk => {
```
rafael-atticlabs: nice.

js/noms/src/value-store.js
rafael-atticlabs: yeah. i think i'm ok with this for now.

js/noms/src/value-store.js
rafael-atticlabs: fixed. now it's totally FUNCTIONAL. expire() now waits until it has all the sizes and the boots people out.

js/noms/src/value-store.js
```diff
@@ -193,24 +201,34 @@ class SizeCache<T> {
...
 
-    if (this._size > this._maxSize) {
-      for (const [key, {size}] of this._cache) {
-        if (this._size <= this._maxSize) {
-          break;
+    entry = new CacheEntry(0, value);
+    this._cache.set(key, entry);
+
+    size.then(resolvedSize => {
+      notNull(entry).size = resolvedSize;
```
rafael-atticlabs: it was, but now its gone

js/noms/src/value-store.js
```diff
@@ -188,47 +192,51 @@ class SizeCache<T> {
...
-      for (const [key, {size}] of this._cache) {
-        if (this._size <= this._maxSize) {
-          break;
-        }
-        this._cache.delete(key);
-        this._size -= size;
-      }
+    entry = new CacheEntry(size, value);
+    this._cache.set(key, entry);
+    entry.size.then(() => this.expire());
```
kalman: I don't understand this line... it means you might call `this.expire()` twice on cache entries, each time they're added, then potentially a 2nd time if they're replaced? I guess it's idempotent, but it seems unnecessary?

rafael-atticlabs: right you are. fixed.

rafael-atticlabs: thanks for such a close review on this patch =-)

PR #3118, opened 2017-01-26
Modify frag tool to evaluate using the grandchild strategy

cmasone-attic:
Also, allow frag tool to be pointed at any path in a NBS-backed
Noms DB.

### DIFF
--- a/go/nbs/frag/main.go
+++ b/go/nbs/frag/main.go
@@ -6,127 +6,162 @@ package main
 
 import (
 	"fmt"
-	"log"
 	"os"
+	"strings"
 	"sync"
 
+	"github.com/attic-labs/noms/go/config"
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/datas"
 	"github.com/attic-labs/noms/go/hash"
 	"github.com/attic-labs/noms/go/nbs"
+	"github.com/attic-labs/noms/go/spec"
 	"github.com/attic-labs/noms/go/types"
 	"github.com/attic-labs/noms/go/util/profile"
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/dustin/go-humanize"
 	flag "github.com/juju/gnuflag"
-)
-
-var (
-	dir    = flag.String("dir", "", "Write to an NBS store in the given directory")
-	table  = flag.String("table", "", "Write to an NBS store in AWS, using this table")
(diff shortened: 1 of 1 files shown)


cmasone-attic:
PTAL

rafael-atticlabs:
Sorry for the delay. It looks to me like this is effectively sampling every *other* depth level. I get why this is tricky to implement. Here's a suggestion of how to make it simpler:

```
childrenQueue := [][]Value{[]Value{startValue}) // a queue of []Value. each []Value is sequence of siblings of a given chunk.

visited := hash.HashSet{}

// Iterate over the queue of sibling-lists
for len(childrenQueue) > 0 {
  children := popFirst(childrenQueue) // select one

  // build list of all grandkids
  allGrandkidHashes := hash.HashSet{}
  for _, child := range children {
    visited.Insert(child.Hash())
    allGrandkidHashes.InsertAll(getChildren(child))
  }
  
  // sample on allGrandkidHashes with CalcReads, etc...

  // fetch all non-leaf grandkids which haven't been visited
  grandKids := db.ReadManyValues(onlyRefHeight2OrMore(allGrandkidHashes))

  // for each child, build the direct set of its children and enqueue to childrenQueue   
  for _, child := range children {
    directGrandKids := selectValues(grandKids, getChildren(child))
    pushBack(childrenQueue, directGrandKids)
  }
}
```

We talked about this in the iteration of frag, but now that ReadManyValues exists, I think it's even easier and more direct to implement concurrency using ReadManyValues/GetMany rather than goroutines.

cmasone-attic:
I think the problem is that we think about this in a fundamentally different way :-/ I don't understand how your proposal works, so -- to me -- it's not simpler. Also, I'm trying to parallelize work and it doesn't look like your solution does that.

Since you seem to have an implementation in your head that makes sense to you, though, perhaps you should just take over this work instead of me taking a fourth and fifth pass at it.

cmasone-attic:
I looked back over my code this morning, and wrote a separate program walk the chunk graph and sanity-check that I'm visiting all the nodes, and I'm pretty sure that my solution calculates reads for every level of the graph -- not just every other level. The structure of the two approaches is basically the same. I have a channel of sibling-groups, while you have this queue. I pull a set of siblings off the channel, build a list of all grandkids, do CalcReads and whatever, and then for each child I build the set of its children and put that on the channel.

Additionally, I've looked at your code more and I think there are a couple of reasons that it's more compact than mine. Obviously, it's more of an outline than code, so it elides logic in a bunch of places. Also, I think it doesn't do one of the things you asked me to do, which is report numbers for both the case where we visit each node once as well as the case where the whole graph is traversed regardless of whether nodes are being visited more than once.

PR #3128, opened 2017-01-30
Fix print_diff when map is keyed by non-primitive

willhite:


go/diff/print_diff.go
```diff
@@ -72,9 +72,17 @@ func PrintDiff(w io.Writer, v1, v2 types.Value, leftRight bool) (err error) {
...
 
-		switch parentEl.(type) {
+		switch parent := parentEl.(type) {
 		case types.Map:
-			key = lastPart.(types.IndexPath).Index
+			if indexPath, ok := lastPart.(types.IndexPath); ok {
+				key = indexPath.Index
+			} else if hip, ok := lastPart.(types.HashIndexPath); ok {
+				hip1 := hip
+				hip1.IntoKey = true
```
kalman: Cute

kalman: Consider explaining why this works, took me a bit of time to remember why

willhite: Added a comment.

PR #3140, opened 2017-02-01
Pass fetch options through to updateRoot

kalman:


js/noms/src/http-batch-store.js
```diff
@@ -94,14 +98,16 @@ export class Delegate {
...
+  _updateRootOptions: FetchOptions;
   _body: ArrayBuffer;
 
   constructor(rpc: RpcStrings, fetchOptions: FetchOptions) {
     this._rpc = rpc;
     this._readBatchOptions = mergeOptions(readBatchOptions, fetchOptions);
     this._writeBatchOptions = mergeOptions(writeBatchOptions, fetchOptions);
-    this._rootOptions = fetchOptions;
+    this._getRootOptions = fetchOptions;
+    this._updateRootOptions = mergeOptions(updateRootOptions, fetchOptions);
```
arv: Any reason not to share the options for write and updateRoot? Seems unlikely that they would be different.

kalman: Shrug, just following what was already there, where names matched up with the RPCs.

Issue #3142, opened 2017-02-02
NBS: Ensure Compact handles too many novel tables


cmasone-attic:
It's possible that a client could have more novel tables to add to a manifest than we're willing to allow to be present in a manifest. In this case, compacting only tables that are present upstream (the current strategy) won't be sufficient.

Issue #3141, opened 2017-02-01
NBS: Ensure Compact handles optimistic lock failures optimally


cmasone-attic:
The initial plan for Compact is to have it block UpdateRoot, compact its novel tables together with whatever tables exist in the manifest, build a new manifest containing the compacted table (and dropping the compactees), and attempt to write it.

if that write hits an optimistic lock failure, the code will just start over with the new manifest and try again. In the common case, whatever process beat us in the race to update the manifest will ALSO have done a compaction, so on the second try the code will be able to just add its novel tables to the manifest and continue -- no re-compaction needed.

There is an edge case, though, where the client has more than one novel table to add and the additional tables are what pushes the manifest over the size threshold. Another client with fewer tables could conceivably update that manifest without needing to compact, so the guy-with-many-tables would wind up compacting multiple times.

The optimization we can do here is keeping track of which tables were compacted and, if those tables are present in the manifest the client discovers when the optimistic lock fails, it can just drop them, add in the big compacted table, and move forward.

FYI @rafael-atticlabs

Issue #3130, opened 2017-01-30
NBS compaction can blow up memory


cmasone-attic:
the first pass at compaction is going to just take a bunch of the smallest tables and merge them into one. Since `tableWriter` wants one big buffer to write into, this process may explode RAM.

rafael-atticlabs:
this is now fixed.

PR #3143, opened 2017-02-02
First pass at compaction

### LINKED ISSUE #2944: All NBS stores should share the index cache
Thankfully, NBS supports concurrent file access, so it's not wrong (like it was with LDB) to have multiple views on the same store open, but it's wasteful as each view has heated up a bunch of the same memory

### LINKED ISSUE #3132: NBS: compaction
Some databases in prod have already gotten so many commits that the manifest is too big to fit in DynamoDB.

Whoops.

We need some kind of compaction to get unblocked here

cmasone-attic:
The first cut at compaction blocks UpdateRoot() while it compacts n/2
tables down into a single, large table (where n == number of tables
named in the NBS manifest). It then attempts to update the manifest
with one referencing the compacted table, the novel tables from the
client, and the remaining upstream tables that were not compacted.

If the update fails, probably due to an optimistic lock failure, the
client drops the compacted table it just created, pulls in the tables
from the newly-discovered upstream manifest, and tries again.

Known flaws:
- may explode RAM (#3130)
- doesn't handle novel tables > max tables (#3142)
- may handle optimistic-lock-failures suboptimally (#3141)

Fixes #3132

Also, fixes #2944 because doing so simplifies some code.

### DIFF
--- a/go/nbs/block_store_test.go
+++ b/go/nbs/block_store_test.go
@@ -198,9 +198,9 @@ func (suite *BlockStoreSuite) TestChunkStoreFlushOptimisticLockFail() {
 }
 
 func (suite *BlockStoreSuite) TestCompactOnUpdateRoot() {
-	testMax := 5
+	testMaxTables := 5
 	mm := fileManifest{suite.dir}
-	smallTableStore := newNomsBlockStore(mm, newFSTableSet(suite.dir, nil), 2, testMax)
+	smallTableStore := newNomsBlockStore(mm, newFSTableSet(suite.dir, nil), 2, testMaxTables)
 	inputs := [][]byte{[]byte("ab"), []byte("cd"), []byte("ef"), []byte("gh"), []byte("ij"), []byte("kl")}
 	chunx := make([]chunks.Chunk, len(inputs))
 	for i, data := range inputs {
@@ -208,22 +208,22 @@ func (suite *BlockStoreSuite) TestCompactOnUpdateRoot() {
 	}
 
 	root := smallTableStore.Root()
-	suite.NoError(smallTableStore.PutMany(chunx[:testMax]))
+	suite.NoError(smallTableStore.PutMany(chunx[:testMaxTables]))
 	suite.True(smallTableStore.UpdateRoot(chunx[0].Hash(), root)) // Commit write
 
 	exists, _, mRoot, specs := mm.ParseIfExists(nil)
 	suite.True(exists)
 	suite.Equal(chunx[0].Hash(), mRoot)
-	suite.Len(specs, testMax)
+	suite.Len(specs, testMaxTables)
 
 	root = smallTableStore.Root()
-	suite.NoError(smallTableStore.PutMany(chunx[testMax:]))
-	suite.True(smallTableStore.UpdateRoot(chunx[testMax].Hash(), root)) // Should compact
+	suite.NoError(smallTableStore.PutMany(chunx[testMaxTables:]))
+	suite.True(smallTableStore.UpdateRoot(chunx[testMaxTables].Hash(), root)) // Should compact
 
 	exists, _, mRoot, specs = mm.ParseIfExists(nil)
 	suite.True(exists)
-	suite.Equal(chunx[testMax].Hash(), mRoot)
-	suite.Len(specs, testMax)
+	suite.Equal(chunx[testMaxTables].Hash(), mRoot)
+	suite.Len(specs, testMaxTables)
 }
 
 func assertInputInStore(input []byte, h hash.Hash, s chunks.ChunkStore, assert *assert.Assertions) {
--- a/go/nbs/block_store_test.go
+++ b/go/nbs/block_store_test.go
@@ -197,6 +197,35 @@ func (suite *BlockStoreSuite) TestChunkStoreFlushOptimisticLockFail() {
 	suite.True(suite.store.UpdateRoot(c2.Hash(), suite.store.Root()))
 }
 
+func (suite *BlockStoreSuite) TestCompactOnUpdateRoot() {
+	testMax := 5
+	mm := fileManifest{suite.dir}
+	smallTableStore := newNomsBlockStore(mm, newFSTableSet(suite.dir, nil), 2, testMax)
+	inputs := [][]byte{[]byte("ab"), []byte("cd"), []byte("ef"), []byte("gh"), []byte("ij"), []byte("kl")}
+	chunx := make([]chunks.Chunk, len(inputs))
+	for i, data := range inputs {
+		chunx[i] = chunks.NewChunk(data)
+	}
+
+	root := smallTableStore.Root()
+	suite.NoError(smallTableStore.PutMany(chunx[:testMax]))
+	suite.True(smallTableStore.UpdateRoot(chunx[0].Hash(), root)) // Commit write
+
+	exists, _, mRoot, specs := mm.ParseIfExists(nil)
+	suite.True(exists)
+	suite.Equal(chunx[0].Hash(), mRoot)
+	suite.Len(specs, testMax)
+
+	root = smallTableStore.Root()
+	suite.NoError(smallTableStore.PutMany(chunx[testMax:]))
+	suite.True(smallTableStore.UpdateRoot(chunx[testMax].Hash(), root)) // Should compact
+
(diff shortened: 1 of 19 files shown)


cmasone-attic:
PTAL

go/nbs/block_store_test.go
```diff
@@ -197,6 +197,35 @@ func (suite *BlockStoreSuite) TestChunkStoreFlushOptimisticLockFail() {
 	suite.True(suite.store.UpdateRoot(c2.Hash(), suite.store.Root()))
 }
 
+func (suite *BlockStoreSuite) TestCompactOnUpdateRoot() {
+	testMax := 5
```
rafael-atticlabs: nit: `testMaxTables`

go/nbs/table_set.go
```diff
@@ -5,165 +5,281 @@
...
+	sort.Sort(chunkSourcesByDescendingCount(ns.upstream))
+
+	max := func(a, b int) int {
+		if a > b {
+			return a
+		}
+		return b
+	}
+
+	partition := len(ns.upstream) - max(2, len(ns.upstream)/2)
```
rafael-atticlabs: I think the fanciness here in order to avoid allocating a new slice buffer isn't worth it. can you just do the naive thing here that's easy to read. i.e.

```
tablesToCompact := ...
compactedTable := ...
upstream := 
```

go/nbs/table_set.go
```diff
@@ -5,165 +5,281 @@
...
+func (ts tableSet) Compact() (ns tableSet, compactees chunkSources) {
+	ns = tableSet{
+		novel:    make(chunkSources, len(ts.novel)),
+		upstream: make(chunkSources, len(ts.upstream), len(ts.upstream)+1),
+		p:        ts.p,
+		rl:       ts.rl,
+	}
+	copy(ns.novel, ts.novel)
+	copy(ns.upstream, ts.upstream)
+	sort.Sort(chunkSourcesByDescendingCount(ns.upstream))
```
rafael-atticlabs: Does it bother you that we go out of our way to *prepend* a new table to the front of the list, but this appears to not only compact the N smallest tables but also reorder the remaining tables by size.

go/nbs/table_set.go
```diff
@@ -5,165 +5,281 @@
...
+	flattened.upstream = append(flattened.upstream, ts.novel...)
+	flattened.upstream = append(flattened.upstream, ts.upstream...)
+	return
+}
+
+// Merge returns a new tableSet holding the novel tables managed by |ts| and
+// those specified by |specs|. Tables in |ts.upstream| that are not referenced
+// by |specs| are returned in |dropped| so that the caller can close() them
+// appropriately.
+func (ts tableSet) Merge(specs []tableSpec) (merged tableSet, dropped chunkSources) {
```
rafael-atticlabs: `Rebase`

ghost (commented):
In general, lg. Some small stuff...

go/nbs/block_store_test.go
cmasone-attic: dun

go/nbs/file_table_persister.go
```diff
@@ -15,30 +15,36 @@ import (
...
-			defer checkClose(temp)
-			io.Copy(temp, bytes.NewReader(data))
-			tempName = temp.Name()
-		}
-		return tempName, name, chunkCount
-	}()
+	return ftp.toReader(mt.write(haver))
+}
+
+func (ftp fsTablePersister) toReader(name addr, data []byte, chunkCount uint32) chunkSource {
```
cmasone-attic: doen

go/nbs/store.go
```diff
@@ -290,13 +304,24 @@ func (nbs *NomsBlockStore) UpdateRoot(current, last hash.Hash) bool {
 		nbs.mt = nil
 	}
 
-	actual, tableNames := nbs.mm.Update(nbs.tables.ToSpecs(), nbs.root, current, nil)
+	candidate := nbs.tables
+	var compactees chunkSources
+	if candidate.Size() > nbs.maxTables {
```
cmasone-attic: No, `Size()` is just returning the number of tables, which is unaffected by compacting a memTable (do we need a different term for this now?). A `tableSet` is immutable; nothing changes the number of tables in a set after it's constructed.

go/nbs/store.go
```diff
@@ -290,13 +304,24 @@ func (nbs *NomsBlockStore) UpdateRoot(current, last hash.Hash) bool {
...
+		candidate, compactees = candidate.Compact() // Compact() must only compact upstream tables
```

go/nbs/store.go
```diff
@@ -305,7 +330,10 @@ func (nbs *NomsBlockStore) Version() string {
 	return nbs.nomsVersion
 }
 
-func (nbs *NomsBlockStore) Close() (err error) {
+// Close will block and compact nbs.tables if it contains more entries than
```
cmasone-attic: Done

go/nbs/table.go
```diff
@@ -205,6 +205,7 @@ type chunkReader interface {
 	get(h addr) []byte
 	getMany(reqs []getRecord, foundChunks chan *chunks.Chunk, wg *sync.WaitGroup) bool
 	count() uint32
+	data() uint64
```
cmasone-attic: done

go/nbs/table_persister.go
```diff
@@ -4,7 +4,94 @@
...
+	if chunkCount == 0 {
+		return
+	}
+
+	maxSize := maxTableSize(uint64(chunkCount), totalData)
+	buff := make([]byte, maxSize) // This can blow up RAM (BUG 3130)
+	tw := newTableWriter(buff)
+
+	// Use "channel of channels" ordered-concurrency pattern so that chunks from a given table stay together, preserving whatever locality was present in that table.
+	chunkChans := make(chan chan extractRecord)
```
cmasone-attic: I may be understanding the question wrong, but yes, it will load all input tables concurrently (modulo the rate-limiter channel `rl`). We talked about this the other day, and when I told you that it took a while to do a compaction, you pointed this out as an optimization opportinity.

I used the "channel of channels" pattern that I picked up from your readahead to ensure that the chunks from the different tables don't get all mixed up together. That's what the comment above is referring to.

go/nbs/table_persister.go
```diff
@@ -4,7 +4,94 @@
...
+				defer func() { close(c); wg.Done(); <-rl }()
+				rl <- struct{}{}
+				s.extract(InsertOrder, c)
+			}(src, chunks)
+			chunkChans <- chunks
+		}
+		wg.Wait()
+	}()
+
+	for chunks := range chunkChans {
```
cmasone-attic: done

go/nbs/table_reader.go
```diff
@@ -19,6 +19,7 @@ import (
 
 type tableIndex struct {
 	chunkCount        uint32
+	totalData         uint64
```
cmasone-attic: This is `tableIndex`

go/nbs/table_set.go
cmasone-attic: Well, the Prepending thing was sort of an opportunistic enhancement. I don't know if it actually matters in practice, so it seemed not worth it to maintain it when compacting. It seems like locality within a table is far-and-away more important, and we don't really know what future work we're going to do to maintain that. Given that, it seemed better to just be simple here.

go/nbs/table_set.go

go/nbs/table_set.go

go/nbs/store.go
rafael-atticlabs: I mean because it could turn out that a memory table that is compacting will end up having 0 chunks (all dups), in which case our compactingChunkSource table shouldn't have been counted.

IOW, it seems like this point should synchronize with any memtable compaction that's taking place before continuing.

go/nbs/table_persister.go
rafael-atticlabs: cool. i only asked because i got the impression that you weren't going to do that in this patch.

go/nbs/table_reader.go
rafael-atticlabs: sorry. table writer.

go/nbs/table_set.go
rafael-atticlabs: ok

ghost (approved):
LGTM

PR #3145, opened 2017-02-05
Jest

arv:
This changes to use Jest. Benefits of jest is parallel tests and jest can run the minimal set of tests that are affected by a change (by looking at dependencies).

The main work was to disentangle our cyclic dependencies. To do this I had to remove some runtime assertions in encode value as well as expose the values of a struct without going through a struct mirror.


arv:
PTAL

This is now ready for review. I've split up the PR into logical commits that can be reviewed one after the other.

@rafael-atticlabs FYI

js/noms/src/indexed-sequence.js
```diff
@@ -17,6 +17,8 @@ export async function newCursorAtIndex(sequence: Sequence<any>, idx: number,
   while (sequence) {
     cursor = new IndexedSequenceCursor(cursor, sequence, 0, readAhead);
     idx -= cursor.advanceToOffset(idx);
+    // TODO: The Sequence type needs to be cleaned up.
+    // $FlowIssue: xxx
```
kalman: (there is an issue for this somewhere, but whatevs)

js/noms/src/jest.js
```diff
@@ -0,0 +1,21 @@
...
+
+// @flow
+
+declare var afterAll;
+declare var afterEach;
+declare var beforeAll;
+declare var beforeEach;
+declare var describe;
+declare var it;
+
```
kalman: maybe put a comment in here

js/noms/src/http-batch-store-test.js
```diff
@@ -9,11 +9,13 @@ import {assert} from 'chai';
 import Hash from './hash.js';
 import {emptyHash} from './hash.js';
 import HttpBatchStore from './http-batch-store.js';
-import mock from 'mock-require';
+import {invariant} from './assert.js';
+
+declare var jest;
```
kalman: Is it possible to make `jest` be default-exported from jest.js?

(obviously it would be fake, in a sense, but better for documentation)

arv: Good idea. I think that would actually work

PR #3147, opened 2017-02-06
Funnel extract failure back to main goroutine

cmasone-attic:
Panics on background goroutines take down the server. This patch
hacks in a mechanism to pipe failures during NBS tableReader.extract
back to the main goroutine so the server doesn't die on this failure
and I can diagnose it.


cmasone-attic:
TBR

PR #3153, opened 2017-02-07
Better naming for graphql fields

rafael-atticlabs:
TBR @arv  Sorry, was going to this in the last patch, but forgot. Just name changes. I think this is better than a sea of "value".


go/ngql/README.md
```diff
@@ -19,23 +19,23 @@ List:
 ```
 type FooList {
   size: Float!
-  values: [Foo!]!
+  elements: [Foo!]!
```
arv: +1

PR #3157, opened 2017-02-07
Graphql empty collections and types

rafael-atticlabs:
We need to fully map the noms types into graphql even though the mapping isn't useful, otherwise the endpoint may die when attempting to create a schema for types that aren't mapped.

Note the reference the proposed *real* solution for implementation once we're convinced this will all work now.

For now, Blobs & Types just get returned as a string representation of their hash value.

Also, a few more type naming cleanups


rafael-atticlabs:
TBR @arv

Issue #3156, opened 2017-02-07
NBS: Some edge case leads snappy to allocate output slice


cmasone-attic:
This is the source of the "corruption" in #3148 

For some large chunks, `snappy.Encode` decides to allocate some new output storage instead of using the buffer we pass it. We were assuming that snappy DID use the buffer we passed it, so this resulted in big blocks of 0's in the table file and chunks being left out (!!!). One such table is `767sh0tj3daui5rh6tj9nb21iu1it2df`, which github won't let me attach :-/

Even using what's in that broken table, I can't craft a test case that reproduces the problem. Unfortunately, the problematic chunks aren't in there.

I'm going to modify the code to detect this case and manually copy the compressed chunk data into the table, but also ship the uncompressed data off to storage somewhere for analysis.

cmasone-attic:
We don't understand why yet, but we're still seeing some of this. Working around for now

cmasone-attic:
We figured it out!

When trying to calculate the maximum possible table size for a compacted table, we were using the _compressed_ lengths of chunks in the tables being compacted -- because that's the only information we had! We realized, though, that we could repurpose a currently-unused field in the NBS table footer format. So, now NBS will store the total _uncompressed_ length of all stored chunks (as opposed to the total _compressed_ length, which it used to store) in the table footer. This allows compaction code to correctly estimate how much space it'll take to store all those chunks in a freshly compacted table with a bunch of other chunk data.

PR #3158, opened 2017-02-07
Gather more info about Bug 3156

cmasone-attic:
There's some case that causes chunks that compress to more than about
55k (we think these are quite big, chunks that are many hundreds of K
in size) not to wind up correctly inserted into tables. It looks like
the snappy library believes the buffer we've allocated may not be
large enough, so it allocates its own space and this screws us up.

This patch changes two things:
1) The CRC in the NBS format is now the CRC of the _compressed_ data
2) Such chunks will be manually copied into the table, so they won't
   be missing anymore

Also, when the code detects a case where the snappy library decided
to allocate its own storage, it saves the uncompressed data off to
the side, so that it can be pushed to durable storage. Such chunks
are stored on disk or in S3 named like "<chunk-hash>-errata", and
logging is dumped out so we can figure out which tables were supposed
to contain these chunks.

Towards #3156

### DIFF
--- a/go/nbs/file_table_persister.go
+++ b/go/nbs/file_table_persister.go
@@ -20,7 +20,17 @@ type fsTablePersister struct {
 }
 
 func (ftp fsTablePersister) Compact(mt *memTable, haver chunkReader) chunkSource {
-	return ftp.persistTable(mt.write(haver))
+	name, data, count, errata := mt.write(haver)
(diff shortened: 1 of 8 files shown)


cmasone-attic:
PTAL. I don't have any unit tests for the "errata" pathway yet

ghost (approved):
LGTM

PR #3160, opened 2017-02-08
Union of scalar and empty struct

rafael-atticlabs:
Add support for unions containing scalar values by "boxing" those scalars in that context. Also add a `hash` field to Struct so that empty structs have at least one field

### DIFF
--- a/go/ngql/query_test.go
+++ b/go/ngql/query_test.go
@@ -260,7 +260,7 @@ func (suite *QueryGraphQLSuite) TestListOfUnionOfScalars() {
 		types.Bool(true),
 	)
 
-	suite.assertQueryResult(list, "{root{elements{... on BooleanValue{scalarValue} ... on StringValue{stringValue: scalarValue} ... on NumberValue{numberValue: scalarValue}}}}", `{"data":{"root":{"elements":[{"scalarValue":28},{"stringValue":"bar"},{"booleanValue":true}]}}}`)
+	suite.assertQueryResult(list, "{root{elements{... on BooleanValue{b: scalarValue} ... on StringValue{s: scalarValue} ... on NumberValue{n: scalarValue}}}}", `{"data":{"root":{"elements":[{"n":28},{"s":"bar"},{"b":true}]}}}`)
 }
 
 func (suite *QueryGraphQLSuite) TestCyclicStructs() {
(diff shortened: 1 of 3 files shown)


go/ngql/query_test.go
```diff
@@ -237,6 +243,32 @@ func (suite *QueryGraphQLSuite) TestListOfUnionOfStructs() {
...
 
+func (suite *QueryGraphQLSuite) TestListOfUnionOfStructsConflictingFieldTypes() {
+	list := types.NewList(
+		types.NewStruct("Foo", types.StructData{
+			"a": types.Number(28),
+		}),
+		types.NewStruct("Bar", types.StructData{
+			"a": types.String("bar"),
+		}),
+		types.NewStruct("Baz", types.StructData{
```
arv: Another problem... What if the names are the same but the fields are different? We will generate two GraphQL types that have the same name 😢

go/ngql/types.go
```diff
@@ -32,9 +37,29 @@ type getFieldFn func(v interface{}, fieldName string, ctx context.Context) types
 // returning one or more *noms* values whose presence is indicated by the provided arguments.
 type getSubvaluesFn func(v types.Value, args map[string]interface{}) (interface{}, error)
 
+// GraphQL requires all memberTypes in a Union to be Structs, so when a noms union contains a
+// scalar, we represent it in that context as a "boxed" value. E.g.
+// Boolean! =>
+// type BooleanValue {
+//   scalarValue: Boolean!
```
arv: Another option would be to do:

```
type BooleanValue {
  boolean: Boolean!
}

type NumberValue {
  number: Number!
}

type StringValue {
  string: String!
}
```

not sure if it would make simpler or more complicated to use?

go/ngql/query_test.go
rafael-atticlabs: The GraphQL spec says that all fields of a struct are always optional. Thus, I think the thing to do will be to define the struct as the union of all the fields that it might have -- which as chance has it it what the type simplification kind of does. For this usage, we'd have to union not just field instances within a collection, but all usages of that Struct reachable from a given value.

rafael-atticlabs: Ok to just open a bug for that?

arv: sure

PR #3164, opened 2017-02-09
Always print Bug 3156 logging

cmasone-attic:
Apparently, there's some issue running demo-server with --verbose
in prod, so we don't do it. This means that the logging info I
added isn't showing up. Change the logging code to use fmt.Fprintf()

Also, add unit test for errata functionality.

Towards #3156


cmasone-attic:
TBR

PR #3165, opened 2017-02-09
NBS: Calculate maxTableSize precisely

### LINKED ISSUE #3156: NBS: Some edge case leads snappy to allocate output slice

cmasone-attic:
Though Raf and I can't figure out how, it's clear that the method we
initially used for calculating the max amount of space for
snappy-compressed chunk data was incorrect. That's the root cause of
of all the chunks to be written and summing the snappy.MaxEncodedLen()
for each.

Fixes #3156

### DIFF
--- a/go/nbs/compacting_chunk_source.go
+++ b/go/nbs/compacting_chunk_source.go
@@ -81,10 +81,10 @@ func (ccs *compactingChunkSource) count() uint32 {
 	return ccs.cs.count()
 }
 
-func (ccs *compactingChunkSource) byteLen() uint64 {
+func (ccs *compactingChunkSource) lens() []uint32 {
 	ccs.wg.Wait()
 	d.Chk.True(ccs.cs != nil)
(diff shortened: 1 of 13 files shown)


cmasone-attic:
PTAL

ghost (approved):
LGTM

PR #3166, opened 2017-02-09
ensure types have a name

rafael-atticlabs:
TBR @arv 

More hackery to get around not translating noms types yet.


go/ngql/types.go
```diff
@@ -393,7 +397,7 @@ func getTypeName(nomsType *types.Type) string {
 		return strings.Join(names, "Or")
 
 	default:
-		panic("type name not implemented")
+		panic(fmt.Sprintf("%d: (getTypeName) not reached", nomsType.Kind()))
```
arv: I thought we had a NomsKind to string function?

PR #3168, opened 2017-02-10
Work around issue #3156

cmasone-attic:
While I dig back in to figure out how to get the debugging
info we need, work around this problem to unblock others.


cmasone-attic:
FYI @mikegray @willhite 

TBR

PR #3169, opened 2017-02-10
Add missing content type header to writeValue/updateRoot

arv:
When we do a writeValue or updateRoot we do a POST with a request body
of some binary data. Some clients (RN) do not set the content-type
header which then leads to failure on our server.

### DIFF
--- a/js/noms/src/http-batch-store.js
+++ b/js/noms/src/http-batch-store.js
@@ -52,6 +52,9 @@ const readBatchOptions = {
 
 const writeBatchOptions = {
   method: 'POST',
+  headers: {
+    'Content-Type': 'application/octet-stream',
+  },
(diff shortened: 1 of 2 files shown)


js/noms/src/http-batch-store.js
```diff
@@ -52,11 +52,12 @@ const readBatchOptions = {
...
   method: 'POST',
+  headers: {
+    'Content-Type': 'application/octet-stream',
+  },
 };
 
-const updateRootOptions = {
-  method: 'POST',
-};
+const updateRootOptions = writeBatchOptions;
```
kalman: `updateRoot` has no post form data, so it's odd to put a content type on it. Not a big deal I suppose.

PR #3170, opened 2017-02-10
NBS: Store total uncompressed data size in NBS tables

### LINKED ISSUE #3156: NBS: Some edge case leads snappy to allocate output slice

cmasone-attic:
BUG 3156 is caused by the compaction code trying to estimate the
maximum possible table size for chunk data pulled from a bunch of
existing tables. The problem was that we only had _compressed_ data
lengths for the chunks in existing tables, so we were drastically
underestimating the worst-case space that we might need during
compaction.

The fix is to have tables store the total number of _uncompressed_
bytes that were inserted, so that the compaction code can use this to
get the right estimate when putting together a bunch of tables.

Fixes #3156

### DIFF
--- a/go/nbs/table_writer.go
+++ b/go/nbs/table_writer.go
@@ -9,7 +9,6 @@ import (
 	"encoding/binary"
 	"fmt"
 	"hash"
-	"os"
 	"sort"
 
 	"github.com/attic-labs/noms/go/d"
@@ -18,11 +17,11 @@ import (
 
 // tableWriter encodes a collection of byte stream chunks into a nbs table. NOT goroutine safe.
 type tableWriter struct {
-	buff              []byte
-	pos               uint64
-	totalPhysicalData uint64
-	prefixes          prefixIndexSlice // TODO: This is in danger of exploding memory
-	blockHash         hash.Hash
+	buff                  []byte
+	pos                   uint64
+	totalUncompressedData uint64
+	prefixes              prefixIndexSlice // TODO: This is in danger of exploding memory
+	blockHash             hash.Hash
 
 	snapper snappyEncoder
 }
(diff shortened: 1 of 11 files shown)


cmasone-attic:
PTAL

cmasone-attic:
TBR

go/nbs/table_writer.go
```diff
@@ -37,14 +36,12 @@ func (r realSnappyEncoder) Encode(dst, src []byte) []byte {
...
-	for _, length := range lengths {
-		max += prefixTupleSize + lengthSize + addrSuffixSize + checksumSize // Metadata
-		d.Chk.True(int64(length) < maxInt)
-		max += uint64(snappy.MaxEncodedLen(int(length)))
-	}
-	return
+func maxTableSize(numChunks, totalData uint64) uint64 {
+	avgChunkSize := totalData / numChunks
+	d.Chk.True(avgChunkSize < maxChunkSize)
+	maxSnappySize := snappy.MaxEncodedLen(int(avgChunkSize))
```
rafael-atticlabs: I think this equivalent to maxEncodedLen(totalData)

ghost (commented):
LGTM

PR #3185, opened 2017-02-11
Fix staging folder to correspond to build folder

zcstarr:


zcstarr:
@rafael-atticlabs  PTAL

PR #3188, opened 2017-02-13
Fix graphql generated css

zcstarr:


zcstarr:
Fixes the 404 with the graphql css @arv PTAL

PR #3189, opened 2017-02-14
NBS: tableSet.Flatten() must exclude empty tables

cmasone-attic:
I missed this in the compaction patch :-/ I caught it in another
test when the code panic'd while trying to write a manifest with
an empty table in it. So at least it got caught there?


rafael-atticlabs:
lgtm

Issue #3127, opened 2017-01-28
Drop support for LevelDBChunkStore (and make NBS the default ChunkStore for file system storage)


rafael-atticlabs:
Blocked on:

 * https://github.com/attic-labs/attic/issues/896

There are generally two big reasons to do this:

1) NBS has and (will continue to have) far superior performance for many reasons, but primarily because LDB is incurring (huge amounts) of IO to ensure fast in-order key scans which NOMs never performs.

2) Implementing important features up the stack (like correct data validation, batched writing and more) will be dramatically easier if we change the semantics of the ChunkStore interface to more closely match the behavior of NBS (i.e. transactions, fast has-checks). Doing the work to make LevelDB support these semantics would further degrade its perf and be time consuming.

PR #3190, opened 2017-02-14
Remove usages of LevelDBStore

cmasone-attic:
NBS is stable enough that we've made it the default store for command
line tools, and the go-to store for tests that require temporary, but
persistent, storage.

We intend to remove support for LevelDB-backed chunk storage
completely ASAP. This patch removes all usage of LevelDBStore from
noms.git, but doesn't remove LevelDBStore _just_ yet as there are
still some dependencies on it elsewhere.

Toward #3127


cmasone-attic:
The meaningful changes are pretty much in the `spec` package. I changed the default behavior in the case where the user provides no protocol. I think I hit all the spots, and all the tests pass, so this should be good.

I haven't removed the actual impl yet, as I said, but I hope to be able to do that soon

PTAL

PR #3193, opened 2017-02-15
Remove LevelDBStore

### LINKED ISSUE #3127: Drop support for LevelDBChunkStore (and make NBS the default ChunkStore for file system storage)

rafael-atticlabs:
Fixes #3127 

TBR @cmasone-attic


cmasone-attic:
LGTM

PR #3194, opened 2017-02-15
Fix GraphQL to use unique type names

### LINKED ISSUE #3161: (GraphQL) Structs with same name but different fields will result in conflicting schema types
GraphQL treats all fields as optional, so the value->schema process probably probably "simplify" all shapes of a given struct name to a common "union" shape.

arv:
The names of GraphQL types needs to be globally unique. We therefore
name struct types as NAME_HASH (where hash is the first 6 chars of
the noms hash).

Also, make sure that we always map the noms type to the same GraphQL
type, even if the boxedIfScalar is true.

Fixes #3161

### DIFF
--- a/go/ngql/query.go
+++ b/go/ngql/query.go
@@ -47,7 +47,8 @@ func constructQueryType(rootValue types.Value, tm *typeMap) *graphql.Object {
 		}})
 }
 
-// Query takes |rootValue|, builds a GraphQL scheme from rootValue.Type() and executes |query| against it, encoding the result to |w|.
+// Query takes |rootValue|, builds a GraphQL scheme from rootValue.Type() and
+// executes |query| against it, encoding the result to |w|.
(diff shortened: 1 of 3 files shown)


ghost (approved):
LGTM. 

FWIW, I think this is clearly the right first approach. The other one would be to "simplify" multiple shapes of the same struct name into a single GraphQL type.

That would result in a bit less ugliness, but it has the following downsides:

-Implementing it correctly is harder (might require two passes over the type hierarchy).
-The GraphQL spec is clear that *all object properties are always optional*, so it's correct, but it's not clear there's a clean way to implement it with graphql-go. (I.e. I don't see any obvious way for a resolver to say "this property isn't here right now").

We'll have to see how often in practice we need to deal with selecting properties off of types within a union. It's kind of a shame that GraphQL doesn't support parameterized types, but maybe that's a bad idea for other reasons.

PR #3195, opened 2017-02-16
Log when s3TableReader hits the rate limiter

cmasone-attic:
We're getting some TCP hangups when trying to talk to S3.
We have an S3 reading rate limiter that's supposed to
prevent issues like this, so the question is whether that's
set too high. Rather than just turning the knob and seeing
if things are OK, this patch adds some logging to verify
that we're actually hitting the rate limiter.


cmasone-attic:
PTAL

ghost (approved):
lgtm

PR #3197, opened 2017-02-16
Revert "Drop s3 read rate limit to 768"

cmasone-attic:
Reverts attic-labs/noms#3196

Based on further logging, I'm not sure this will help. So, revert for the time being.


cmasone-attic:
TBR

PR #3199, opened 2017-02-16
Revert "Log when s3TableReader hits the rate limiter"

cmasone-attic:
Reverts attic-labs/noms#3195

There doesn't seem to be any correlation between hitting the rate limit and the connection resets, so shut off this log spam


cmasone-attic:
TBR

PR #3200, opened 2017-02-16
Propagate not found as GQL errors

kalman:


kalman:
@arv added you as reviewer but never mind (unless you feel comfortable). however, FYI.

rafael-atticlabs:
lgtm

PR #3201, opened 2017-02-17
Fix NPE in GraphQL

arv:
When we have a cyclic type we ended up getting the non complete type out of the map.

Use graphql.FieldsThunk which is designed to allow recursive types.

### DIFF
--- a/go/ngql/query_test.go
+++ b/go/ngql/query_test.go
@@ -303,6 +303,30 @@ func (suite *QueryGraphQLSuite) TestCyclicStructs() {
 	suite.assertQueryResult(s1, "{root{a b{elements{a}}}}", `{"data":{"root":{"a":"aaa","b":{"elements":[{"a":"bbb"}]}}}}`)
 }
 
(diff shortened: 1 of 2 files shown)


arv:
I need to update the comment about how we use NonNull to support cycles...

arv:
Updated comment and removed hack.

go/ngql/query_test.go
```diff
@@ -280,7 +280,6 @@ func (suite *QueryGraphQLSuite) TestListOfUnionOfScalars() {
 
 	suite.assertQueryResult(list, "{root{elements{... on BooleanValue{b: scalarValue} ... on StringValue{s: scalarValue} ... on NumberValue{n: scalarValue}}}}", `{"data":{"root":{"elements":[{"n":28},{"s":"bar"},{"b":true}]}}}`)
 }
-
```
kalman: add back

kalman (approved):
lgtm

PR #3212, opened 2017-02-22
Share s3, dynamodb clients

cmasone-attic:
These objects manage their own pools of HTTP connections and
other resources, so it's generally best to share them
process-wide if you can.

Fixes #3027


cmasone-attic:
Just a quick patch to see if this helps any. PTAL

ghost (approved):
LGTM

Issue #3204, opened 2017-02-21
Implement `unionTypes(a, b *types.Type) *types.Type`


rafael-atticlabs:
This is a variant of the algorithm we have now for simplifying types. And when we add optional struct fields, it will *become* the algorithm for simplifying types.

The difference between union and simplify is that when simplify encounters two structs and a field is present in only one, it drops it. Union must keep it.

ehalpern:
@aboodman: Given @rafael-atticlabs's suggestion that this ultimately replace makeSimplifiedUnion, do you still have reservations about implementing this in noms rather than attic?

aboodman:
I think it might be easier to iterate at first with this in attic closer to
the calling code. But not a huge deal.

PR #3214, opened 2017-02-22
Implement `unionTypes(a, b *types.Type) *types.Type` (#3204)

ehalpern:
Implements full unions building off of makeSimplifiedUnion.
The difference between simplified unions and full unions are
in how structs are handled. In the former, we produce the intersection
of input structs to ensure the simplified type is a supertype of the input
types. In the later, we produce the union of the input structs which leads
to a type that is a subtype of all inputs.

fixes: #3204

### DIFF
--- a/go/types/simplify.go
+++ b/go/types/simplify.go
@@ -1,10 +1,8 @@
 package types
 
-import (
-	"github.com/attic-labs/noms/go/d"
-)
+import "github.com/attic-labs/noms/go/d"
 
-// makeSimplifiedUnion returns a type that is a supertype of all the input types, but is much
+// makeSupertype returns a type that is a supertype of all the input types but is much
 // smaller and less complex than a straight union of all those types would be.
 //
 // The resulting type is guaranteed to:
@@ -29,12 +27,12 @@ import (
 // - The map group is collapsed like so:
 //     {Map<K1,V1>|Map<K2,V2>...} -> Map<K1|K2,V1|V2>
 // - Each struct group is collapsed like so:
-//     {struct{foo:number,bar:string}, struct{bar:bool, baz:blob}} ->
+//     {struct{foo:number,bar:string}, struct{bar:blob, baz:bool}} ->
 //       struct{bar:string|blob}
 //
 // Anytime any of the above cases generates a union as output, the same process
 // is applied to that union recursively.
-func makeSimplifiedUnion(in ...*Type) *Type {
+func makeSupertype(in ...*Type) *Type {
 	ts := make(typeset, len(in))
 	for _, t := range in {
 		// De-cycle so that we handle cycles explicitly below. Otherwise, we would implicitly crawl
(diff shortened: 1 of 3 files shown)


go/types/type_cache.go
```diff
@@ -537,6 +537,10 @@ func MakeUnionType(elemTypes ...*Type) *Type {
 	return makeSimplifiedUnion(elemTypes...)
 }
 
+func MakeFullUnionType(elemTypes ...*Type) *Type {
```
aboodman: I feel like the terminology is getting confused here. Let's ask @arv in meatspace what the right terms are.

This thing you're adding now feels more like a _union_ of the input types than what I had. The thing I had might be called `ExtractCommonSupertype()` ?

go/types/union.go
```diff
@@ -43,8 +41,32 @@ func makeSimplifiedUnion(in ...*Type) *Type {
...
-	return makeSimplifiedUnionImpl(ts)
+	// makeUnionImpl de-cycles internally.
+	return makeUnionImpl(ts, true)
+}
+
+// makeUnion returns a type that is the union of all the input types.
+//
+// The resulting type is similar to the type produced by makeSimplifiedUnion
+// with the one exception: Each named Struct will contain the union of all fields in
+// the inputs as opposed to just the common fields. In effect this means the resulting
```
aboodman: It's not true that the resulting type is a subtype:

`makeUnion(List<A>, List<B>) => List<A|B>`.

`List<A|B>` is a supertype of `List<A>` and `List<B>`, not a subtype.

rafael-atticlabs:
I'm going to completely defer to @aboodman on this since he's the simplification expert.

go/types/type_cache.go
ehalpern: Agreed. I was reluctant to mess with the existing terminology cause I was trying to minimize the change.

go/types/union.go
ehalpern: Good catch. I meant to review that claim before the checkin.

ehalpern:
PTAL @aboodman 

- the new type maker is now called  MakeMergedType
- renamed makeSimplifiedUnion to makeSupertype
- deferred renaming MakeUnionType (which delegates to makeSupertype)

go/types/simplify.go
```diff
@@ -43,8 +41,32 @@ func makeSimplifiedUnion(in ...*Type) *Type {
 		ts[t] = struct{}{}
 	}
 
-	// makeSimplifiedUnionImpl de-cycles internally.
-	return makeSimplifiedUnionImpl(ts)
+	// Impl de-cycles internally.
+	return makeSimplifiedTypeImpl(ts, false)
+}
+
+// makeMergedType returns a the type that results from merging all input types.
```
aboodman: typo: `a the`

aboodman:
Please just call this makeSimplifiedType and makeSimplifiedType2. They are both trying to do the same thing, just slightly different algorithms. And I think makeSimplifiedType is the best name I've seen so far for this concept.

aboodman:
And the `2` will go away soon.

ehalpern:
Done. PTAL @aboodman

rafael-atticlabs:
\o/

Issue #3210, opened 2017-02-21
GraphQL: Cannot quer field elements in empty *


arv:
```json
{
  "data": null,
  "errors": [
    {
      "message": "Cannot query field \"elements\" on type \"EmptyMap\".",
      "locations": [
        {
          "line": 1,
          "column": 150
        }
      ]
    }
  ]
}
```

This makes things annoying...

rafael-atticlabs:
To capture some meatspace discussion:

noms lacks schemas. this makes grapqhl usage difficult

A schema largely serves the purpose of defining a contract between writers and readers of data. One can imagine access control features which enforce this, but the central problem is just have a place to describe the contract.

Prior to accretive types, noms containers (Set, Map, List) described the type of things they *can* contain. This described a schema in the same sense that SQL table descriptions do. Now noms types describe the things the *do* contain. This means that there is no longer any kind of first class contract between readers and writers.

A client directly using noms can generally tolerate this ambiguity by navigating a graph of data incrementally and potentially introspection what exists at each level.

However, GraphQL as spec'd has a hard time with this. Queries are written against a static schema and the design assumes this is relatively unchanging. The current noms grapql implementation dynamically creates a schema based on the type of the data it discovers. This makes it hard to write queries that will work with varying actual states of the data.

This arises for example when collections of data *may* be empty. If a `Map` generally contains `Struct Foo`, but happens at a given moment to be empty, the generated schema will describe the type as `EmptyMap` (having no element types) and then the Queries

```
{
  root {
    ... on EmptyMap {
      size
    }
    ... on StringToFooMap {
      size
      elements {
        key
        value {
          baz
        }
      }
    }
  }
}
```

is *never* valid. So clients have no way to tolerate the absence of `data at a given time.`

A variant of this is the way types are simplified. Because `Struct Foo { a b }` and `Struct Foo{ a }` are simplified to `Struct Foo { a }` within a container, it's impossible to ever query for `b` within `Foo`. Optional struct fields would address this.

rafael-atticlabs:
cc @aboodman

rafael-atticlabs:
We've identified two kinds of solutions to this:

1) Create a convention such that datasets can optionally express a schema. This might take the form of a field in `::ds.meta.schema: types.Type`. If that is present, the `/graphql/` endpoint will use it as the basis for a generated schema, rather than the type of the value actually present. When optional fields are supported, it will additionally enforce that type is a subtype of the expressed schema.

[...]

PR #3218, opened 2017-02-22
NGQL: Add support for meta.schema

### LINKED ISSUE #3213: GraphQL: use a meta field to define schema
Suggestion:

- If the `meta` field of a `Commit` has a `schema` field of type `Type` we use that type to generate the GraphQL schema.

Future work would be to have the commit method validate that the commit value is a subtype of the type declared in the schema.

arv:
If the value is a commit with a meta.schema of type Type, use that to
compute the GraphQL schema.

Fixes #3213

### DIFF
--- a/go/ngql/query_test.go
+++ b/go/ngql/query_test.go
@@ -359,3 +359,44 @@ func (suite *QueryGraphQLSuite) TestError() {
 	suite.Equal(buff.String(), `{"data":null,"errors":[{"message":"Some error string","locations":null}]}
 `)
 }
+
+func (suite *QueryGraphQLSuite) TestCustomSchema() {
+	val := types.NewStruct("Value", types.StructData{
+		"x": types.Number(42),
+		"s": types.String("hi"),
+	})
+	commit := makeCommit(val, types.EmptyStruct)
+	suite.assertQueryResult(commit, "{root{value{x}}}", `{"data":{"root":{"value":{"x":42}}}}`)
+	suite.assertQueryResult(commit, "{root{value{s}}}", `{"data":{"root":{"value":{"s":"hi"}}}}`)
+
+	schema := types.MakeStructTypeFromFields("Commit", types.FieldMap{
+		"value": types.MakeStructTypeFromFields("Value", types.FieldMap{
+			"x": types.NumberType,
+			"s": types.StringType,
+		}),
+		"meta": types.MakeStructTypeFromFields("Meta", types.FieldMap{
+			"schema": types.TypeType,
+		}),
+	})
+
+	commit = makeCommit(val, makeMetaStructWithSchema(schema))
+	suite.assertQueryResult(commit, "{root{value{x}}}", `{"data":{"root":{"value":{"x":42}}}}`)
+	suite.assertQueryResult(commit, "{root{value{s}}}", `{"data":{"root":{"value":{"s":"hi"}}}}`)
+
+	// no s field. Gets caught in validation
+	schema = types.MakeStructTypeFromFields("Commit", types.FieldMap{
+		"value": types.MakeStructTypeFromFields("Value", types.FieldMap{
+			"x": types.NumberType,
+		}),
+		"meta": types.MakeStructTypeFromFields("Meta", types.FieldMap{
+			"schema": types.TypeType,
+		}),
+	})
+
+	commit = makeCommit(val, makeMetaStructWithSchema(schema))
+	suite.assertQueryResult(commit, "{root{value{x}}}", `{"data":{"root":{"value":{"x":42}}}}`)
+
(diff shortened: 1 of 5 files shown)


arv:
@rafael-atticlabs FYI

go/ngql/query_test.go
```diff
@@ -359,3 +359,44 @@ func (suite *QueryGraphQLSuite) TestError() {
...
+		}),
+		"meta": types.MakeStructTypeFromFields("Meta", types.FieldMap{
+			"schema": types.TypeType,
+		}),
+	})
+
+	commit = makeCommit(val, makeMetaStructWithSchema(schema))
+	suite.assertQueryResult(commit, "{root{value{x}}}", `{"data":{"root":{"value":{"x":42}}}}`)
+
+	// Should give an error message because we didn't declare s in the schema.
```
kalman: It's about to not give an error...

kalman:
There is also a compile error.

arv:
PTAL

go/ngql/query_test.go
```diff
@@ -359,3 +359,45 @@ func (suite *QueryGraphQLSuite) TestError() {
...
+	})
+
+	commit = makeCommit(val, makeMetaStructWithSchema(schema))
+	suite.assertQueryResult(commit, "{root{value{x}}}", `{"data":{"root":{"value":{"x":42}}}}`)
+
+	// s is not part of the schema but is part of the data.
+	suite.assertQueryResult(commit, "{root{value{s}}}", `{"data":{"root":{"value":{}}}}`)
+	// b is part of the schema but not part of the data.
+	suite.assertQueryResult(commit, "{root{value{b}}}", `{"data":{"root":{"value":{"b":null}}}}`)
+}
```
kalman: Did you want to test the `... on FooBar` behavior?

arv: I did... One more commit coming

go/ngql/types.go
```diff
@@ -203,10 +203,17 @@ func structToGQLObject(nomsType *types.Type, tm *typeMap) *graphql.Object {
 			structDesc.IterFields(func(name string, nomsFieldType *types.Type) {
 				fieldType := nomsTypeToGraphQLType(nomsFieldType, false, tm)
 
+				if nonNull, ok := fieldType.(*graphql.NonNull); ok {
```
kalman: Wait - did you deliberately not update the schema generation code? That still produces not-nullable types.

I mean, this seems fine (until we have optional types), just pointing it out.

arv: Did you want all types to be nullable? It is only struct fields that need this hack.

aboodman:
This change has a lot of implications all over the place. I don't think we should jump into it so rapidly. Was the other solution from https://github.com/attic-labs/noms/issues/3210 not doable:

> Turn off run-time query validation. The main benefit of graphql is design-time validation. Run-error query validation isn't super helpful and in the case is especially troublesome. Or as @kalman put it, it should be the concern of some higher abstraction to give errors, not the query system. To wit, graphql actually assumes that all properties are optional, so there's already tolerated uncertainly about what data exists, it just presently enforces that you must declare the set of things with might exist.

To be specific what I'm worried about here is that if GQL only really works against datasets that have this schema thing, then the practice of developing against Noms dramatically changes. Now all the sudden everything needs to have a schema. I'm not even sure I understand what all that would mean ... we are a distributed system with different parties committing to the same dataset. Who has responsibility for setting the schema?

aboodman:
Also, are you allowed to go beyond the schema (expando?). If not that makes the system super rigid, especially for distributed multiparty usecases. But if so, then the schema doesn't really fix the GQL issue.

aboodman:
I see now that the other fix already landed. So then my comment is just -- delay this?

rafael-atticlabs:
I think that something like *doesn't* mean that GQL only works in practice with datasets that have a schema, but I agree that assuming @kalman's patch to shut off run-time query validation works, we can wait on this.

aboodman:
Independent of this patch I would still be very reluctant to add the notion
of schemas in noms 8. We already have so much going on and that is a very
core feature that needs a lot of consideration.

aboodman:
I think the core question is where the schema goes, who has permission to
update it, and how subtyping is allowed (if at all). This also interacts
closely with access control which hasn't been thought through.

arv:
I agree. Let's hold off on this and see how far we can get with GraphQL without a schema.

The mismatch comes from GraphQL wanting you to define a schema that you expect to operate on and not a schema that reflects the actual data you have. When the data doesn't match the schema you are supposed to get an error.

This PR was one way to provide a way to tell GraphQL what our expected schema is. Another option we talked about was to hand write the schema and have the client pass it.

PR #3220, opened 2017-02-23
Add marshal.MustUnmarshal() function

willhite:


go/marshal/decode.go
```diff
@@ -85,6 +86,11 @@ func Unmarshal(v types.Value, out interface{}) (err error) {
 	return
 }
 
+func MustUnmarshal(v types.Value, out interface{}) {
```
arv: GoDoc

go/marshal/decode.go
```diff
@@ -85,6 +86,11 @@ func Unmarshal(v types.Value, out interface{}) (err error) {
...
+	err := Unmarshal(v, out)
+	d.Chk.NoError(err)
```
arv: Also, `d.PanicIfError`

PR #3223, opened 2017-02-23
Implement `TypesIntersect(a, b *types.Type) bool`

### LINKED ISSUE #3203: Implement `ContainCommonSupertype(a, b *types.Type) bool`
`ContainCommonSupertype(a, b *types.Type) bool` is defined recursively for `a` & `b`:

* if either type is `Value` => return true
* if they are of different kind => return false
* else
  * if a & b are structs then, return true iff they are the same name and share a field name and the respective types of the field ContainCommonSupertype
  * if a & b are sets or list, return true iff there exists some element type in both where one ContainCommonSupertype the other
  * if a & b are maps, return true iff there exists a common key type and some pair of respective value types ContainCommonSupertype.
  * else, return true.

ehalpern:
fixes: #3203

### DIFF
--- a/go/types/intersect.go
+++ b/go/types/intersect.go
@@ -11,10 +11,10 @@ import "github.com/attic-labs/noms/go/d"
 //    - if either type is Union, return true iff at least one variant of |a| intersects with one variant of |b|
 //    - if |a| & |b| are not the same kind, return false
 //    - else
-//      - if both are structs, return true iff they have the same name, share a field name and the type
-//        of that field intersects
+//      - if both are structs, return true iff their names are equal or one name is "", they share a field name
+//        and the type of that field intersects
 //      - if both are refs, sets or lists, return true iff the element type intersects
-//      - if both are maps, return true iff they have a key and value type that intersect
+//      - if both are maps, return true iff they have a key with the same type and value types that intersect
 //      - else return true
 func TypesIntersect(a, b *Type) bool {
 	// Avoid cycles internally.
@@ -25,11 +25,8 @@ func typesIntersectImpl(a, b *Type) bool {
 	if a.Kind() == ValueKind || b.Kind() == ValueKind {
 		return true
 	}
-	if a.Kind() == UnionKind {
-		return unionIntersects(a, b)
-	}
-	if b.Kind() == UnionKind {
-		return unionIntersects(b, a)
+	if a.Kind() == UnionKind || b.Kind() == UnionKind {
+		return unionsIntersect(a, b)
 	}
 	if a.Kind() != b.Kind() {
 		return false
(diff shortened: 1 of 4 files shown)


ehalpern:
fyi: @aboodman. @rafael-atticlabs, let me know if you'd like someone else to review.

go/types/intersect.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return false
+}
+
+func containersIntersect(kind NomsKind, a, b *Type) bool {
+	d.Chk.True(kind == a.Desc.Kind() && kind == b.Desc.Kind())
+	return typesIntersectImpl(a.Desc.(CompoundDesc).ElemTypes[0], b.Desc.(CompoundDesc).ElemTypes[0])
+}
+
+// TODO: consider requiring that both maps share a key type
```
rafael-atticlabs: agree. let's go ahead and do that.

go/types/intersect.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func structsIntersect(a, b *Type) bool {
+	d.Chk.True(StructKind == a.Kind() && StructKind == b.Kind())
+	aDesc := a.Desc.(StructDesc)
+	bDesc := b.Desc.(StructDesc)
+	if aDesc.Name != bDesc.Name {
+		return false
+	}
+	for _, f := range aDesc.fields {
```
rafael-atticlabs: nit: let's avoid the map lookup here. desc.Fields is always sorted. do a parallel scan through both. i.e

```
for i, j := 0, 0; i < len(aDesc.fields) && j < len(bDesc.fields) {
  aName, bName := aDesc.fields[i].name, bDesc.fields[j].name;
  if aName < bName {
    i++
  } else if bName < aName {
    j++
  } else if typesIntersectImp(aDesc.fields[i].t, bDesc.fields[j].t) {
    return true
  }
  return false
}
```

go/types/intersect_test.go
```diff
@@ -0,0 +1,88 @@
+package types
+
+import (
+	"testing"
+
+	"github.com/attic-labs/testify/assert"
+)
+
+func TestIntersectImpl(t *testing.T) {
```
rafael-atticlabs: add some tests which cover hit & miss cases for containers with union parameters. e.g. List<A|B|C> & List<C|D>.

Also maybe one "recursive" struct case, like {a: List<B>} & {a: List<A|B>}

ghost (commented):
btu

ehalpern:
PTAL @rafael-atticlabs

arv:
How is this different from IsSubtype?

rafael-atticlabs:
@arv 
```
TypesIntersect(Struct { a: Number, b: String }, Struct { a: Number, c: Bool }) == true
```

ghost (approved):
LGTM. nice.

arv:
Can we use another word to describe this. Type intersection is already a well defined term in type theory.

aboodman:
Note that this doesn't return a new type... it just returns whether the two
types do intersect or not. Is this not type intersection from theory?

arv:
You can always get the intersection of two different types but there might not be any values that fit in the type intersection.

```
type T = number & string;
const n: T = ???;
```

There is no value that can be a number and a string....

```
type T2 = {b: string} & {a: number};
const n2: T2 = {a: 42, b: 'a'}
```

Given this it is not clear what it means that `{a: number, b: string} & {b: string, c: boolean}` returns `true`? Maybe the question we are asking is if there is a common field in two types?

arv:
I think the terminology needs some tweaking. The question we are asking is the type may help get us towards the other type. In other words if there is a common (non trivial) super type 

containsCommonSuperType
mayContribute{To,Towards}

rafael-atticlabs:
Per meatspace discussion with @arv, let's change the name of this to be `ContainCommonSupertype(a, b *types.Type) bool`

aboodman:
I like that name a lot!

On Fri, Feb 24, 2017 at 10:18 AM, Rafael Weinstein <notifications@github.com

aboodman:
Except for it it's not grammatical. Should be ContainsCommonSupertype
(plural)

ehalpern:
@aboodman: Sounds good, but I think ContainCommonSupertype is correct since it refers to two arguments (i.e. a and b contain common supertypes)

rafael-atticlabs:
there are two things you are asking the question about. e.g. "a and b contain a common supertype".

ehalpern:
Done

aboodman:
You are right! My mistake.

PR #3224, opened 2017-02-24
Update nbs/README.md

aboodman:
I did some marketing for you. Let me know if you think it is OK (and especially accurate).


aboodman:
Eh, on second thought we can always fix it asynchronously (still tell me what you think)

rafael-atticlabs:
lgtm

PR #3225, opened 2017-02-24
Introduce Map.Iterator() and friends

aboodman:


go/types/map_iterator.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"github.com/attic-labs/noms/go/d"
+)
+
+// MapIterator can efficiently iterate through a Noms Map.
+type MapIterator struct {
+	cursor    *sequenceCursor
+	LastKey   Value
+	LastValue Value
```
kalman: Can we remove LastKey/LastValue until we need them? They were for implementing the fancy SetIterator compositions, but we're not composing MapIterators.

aboodman: We need them right now. That's why I implemented this.

PR #3228, opened 2017-02-24
Add Set.IteratorAt and Set.IteratorFrom

arv:
These two were missing for Set

### DIFF
--- a/go/types/set.go
+++ b/go/types/set.go
@@ -217,7 +217,21 @@ func (s Set) IterAll(cb setIterAllCallback) {
 }
 
 func (s Set) Iterator() SetIterator {
-	return &setIterator{s: s, cursor: nil}
+	return s.IteratorAt(0)
+}
+
+func (s Set) IteratorAt(idx uint64) SetIterator {
+	return &setIterator{
+		cursor: newCursorAtIndex(s.seq, idx, false),
+		s:      s,
+	}
+}
+
+func (s Set) IteratorFrom(val Value) SetIterator {
+	return &setIterator{
+		cursor: newCursorAtValue(s.seq, val, false, false, false),
+		s:      s,
+	}
 }
 
 func (s Set) elemType() *Type {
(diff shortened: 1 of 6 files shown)


arv:
I need this for GraphQL...

go/types/set.go
```diff
@@ -217,7 +217,21 @@ func (s Set) IterAll(cb setIterAllCallback) {
...
+func (s Set) IteratorAt(idx uint64) SetIterator {
+	return &setIterator{
+		cursor: newCursorAtIndex(s.seq, idx, false),
+		s:      s,
+	}
+}
+
+func (s Set) IteratorFrom(val Value) SetIterator {
+	return &setIterator{
+		cursor: newCursorAtValue(s.seq, val, false, false, false),
```
kalman: We should enable readahead for the iterators. I missed it in the map iterator review too, could you fix those?

go/types/set_iterator.go
```diff
@@ -32,48 +32,42 @@ type SetIterator interface {
...
+	cursor       *sequenceCursor
+	currentValue Value
 }
 
 func (si *setIterator) Next() Value {
 	if si.cursor == nil {
-		si.cursor = newCursorAt(si.s.seq, emptyKey, false, false, false)
-	} else {
-		si.cursor.advance()
+		d.Panic("Cannot use a nil SetIterator")
```
kalman: Could you remove this condition and the panic, I don't think it's useful. Also in map iterator. I made that comment but Aaron didn't see it I suppose.

go/types/set_iterator.go
```diff
@@ -32,48 +32,42 @@ type SetIterator interface {
...
+		si.cursor.advance()
+	} else {
+		si.currentValue = nil
 	}
-	return nil
+	return si.currentValue
 }
 
 func (si *setIterator) SkipTo(v Value) Value {
 	d.Chk.NotNil(v, "setIterator.SkipTo() called with nil value")
```
kalman: `d.PanicIfTrue(v == nil)`

arv:
Thanks, will fix'em...

PR #3229, opened 2017-02-25
GraphQL map set at

arv:


arv:
First commit is https://github.com/attic-labs/noms/pull/3228

go/ngql/types.go
```diff
@@ -257,63 +257,80 @@ func getListValues(v types.Value, args map[string]interface{}) (interface{}, err
...
 	if c, ok := args[countKey].(int); ok {
 		count = uint64(c)
 	}
 
 	// Clamp ranges
+	if idx < 0 {
+		idx = 0
+	}
+	if idx > s.Len() {
+		idx = s.Len()
```
kalman: You can immediately return an empty set in this case. Is this stuff happy with a nil interface result (obviously if not just a single element value).

You can also immediately return in the count <= 0 case.

Like
```go
if count <= 0 || idx >= s.Len() {
  // return nil, nil ?
  return []interface{}{}, nil
}
```

go/ngql/types.go
```diff
@@ -257,63 +257,80 @@ func getListValues(v types.Value, args map[string]interface{}) (interface{}, err
...
+	if idx > m.Len() {
+		idx = m.Len()
+	}
 	if count < 0 {
 		count = 0
 	}
-	if count > m.Len() {
-		count = m.Len()
+	if idx+count > m.Len() {
+		count = m.Len() - idx
```
kalman: Same comments. All of this looks awfully similar to the set code can you share code?

PR #3231, opened 2017-02-27
Add additional logging in s3_table_reader.go

cmasone-attic:
This is to help create a support case for the S3
connection-reset-by-peer issue we're seeing on prod.


cmasone-attic:
PTAL

FYI @rafael-atticlabs

PR #3233, opened 2017-02-28
GraphQL: Add support for key and through

### LINKED ISSUE #3227: Implement ranged queries for collections via GQL
For lists/sets/maps:

`elements(at:)` => returns single element
`elements(at:, count:)` => returns [at, at + count)

For sets/maps

`elements(key:)` => returns single element
`elements(key:, count:)` => returns count elements starting at key
`elements(key:, through:)` => returns elements [key, through]

arv:
If key is provided as an argument to a Map/Set elements that defines
where the collection should start iterating from.

If through is provided the iteration will end after visiting through
(or if the key is larger than through)

If both key and at are present, at is ignored.

If key is present but count and through are missing then we use a count
of 1.

Closes #3227

### DIFF
--- a/go/ngql/query.go
+++ b/go/ngql/query.go
@@ -26,6 +26,7 @@ const (
 	sizeKey        = "size"
 	targetHashKey  = "targetHash"
 	targetValueKey = "targetValue"
+	throughKey     = "through"
 	tmKey          = "tm"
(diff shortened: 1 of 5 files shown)


ghost (approved):
LGTM. really nice.

Issue #3222, opened 2017-02-23
Stop implicitly creating directories for nbs


aboodman:
A long time ago, we decided that commands like:

`noms serve /foo/bar` should implicitly create a directory at `/foo/bar`. This was a bad idea. We are constantly accidentally creating directories because of typos.

We should remove this. Creating a database for a non-existing file path should be the same as creating a database for a non-existent http server. It should result in a nice error being printed to the console.

More generally, anytime you use `config.NewResolver()` or `spec.Spec` with a non-existing file path, it should return an error, not implicitly create it.

PR #3235, opened 2017-02-28
Remove implicit directory creation when creating a local db

jesseditson:
- remove Mkdir from NewLocalStore and NewLocalStoreFactory
- add specific error messages for directory does not exist and path is not a directory
- add tests for missing directory and path not directory

fixes: #3222

### DIFF
--- a/cmd/noms/noms_ds_test.go
+++ b/cmd/noms/noms_ds_test.go
@@ -6,7 +6,6 @@ package main
 
 import (
 	"testing"
-	"os"
 
 	"github.com/attic-labs/noms/go/datas"
 	"github.com/attic-labs/noms/go/nbs"
@@ -26,25 +25,21 @@ type nomsDsTestSuite struct {
 
 func (s *nomsDsTestSuite) TestEmptyNomsDs() {
 	dir := s.DBDir
-	dbName := dir+"/name"
 
-	os.MkdirAll(dbName, 0777)
-	cs := nbs.NewLocalStore(dbName, clienttest.DefaultMemTableSize)
+	cs := nbs.NewLocalStore(dir, clienttest.DefaultMemTableSize)
 	ds := datas.NewDatabase(cs)
 
 	ds.Close()
 
-	dbSpec := spec.CreateDatabaseSpecString("nbs", dbName)
+	dbSpec := spec.CreateDatabaseSpecString("nbs", dir)
 	rtnVal, _ := s.MustRun(main, []string{"ds", dbSpec})
 	s.Equal("", rtnVal)
 }
 
 func (s *nomsDsTestSuite) TestNomsDs() {
 	dir := s.DBDir
-	dbName := dir+"/name"
 
-	os.MkdirAll(dbName, 0777)
-	cs := nbs.NewLocalStore(dbName, clienttest.DefaultMemTableSize)
+	cs := nbs.NewLocalStore(dir, clienttest.DefaultMemTableSize)
 	db := datas.NewDatabase(cs)
 
 	id := "testdataset"
@@ -60,9 +55,9 @@ func (s *nomsDsTestSuite) TestNomsDs() {
 	err = db.Close()
 	s.NoError(err)
 
-	dbSpec := spec.CreateDatabaseSpecString("nbs", dbName)
-	datasetName := spec.CreateValueSpecString("nbs", dbName, id)
-	dataset2Name := spec.CreateValueSpecString("nbs", dbName, id2)
+	dbSpec := spec.CreateDatabaseSpecString("nbs", dir)
+	datasetName := spec.CreateValueSpecString("nbs", dir, id)
+	dataset2Name := spec.CreateValueSpecString("nbs", dir, id2)
 
 	// both datasets show up
 	rtnVal, _ := s.MustRun(main, []string{"ds", dbSpec})
--- a/cmd/noms/noms_ds_test.go
+++ b/cmd/noms/noms_ds_test.go
@@ -6,6 +6,7 @@ package main
 
 import (
 	"testing"
+	"os"
 
 	"github.com/attic-labs/noms/go/datas"
 	"github.com/attic-labs/noms/go/nbs"
@@ -25,21 +26,25 @@ type nomsDsTestSuite struct {
 
 func (s *nomsDsTestSuite) TestEmptyNomsDs() {
 	dir := s.DBDir
+	dbName := dir+"/name"
 
-	cs := nbs.NewLocalStore(dir+"/name", clienttest.DefaultMemTableSize)
+	os.MkdirAll(dbName, 0777)
+	cs := nbs.NewLocalStore(dbName, clienttest.DefaultMemTableSize)
 	ds := datas.NewDatabase(cs)
 
 	ds.Close()
 
-	dbSpec := spec.CreateDatabaseSpecString("nbs", dir+"/name")
+	dbSpec := spec.CreateDatabaseSpecString("nbs", dbName)
 	rtnVal, _ := s.MustRun(main, []string{"ds", dbSpec})
 	s.Equal("", rtnVal)
 }
 
 func (s *nomsDsTestSuite) TestNomsDs() {
 	dir := s.DBDir
+	dbName := dir+"/name"
 
-	cs := nbs.NewLocalStore(dir+"/name", clienttest.DefaultMemTableSize)
+	os.MkdirAll(dbName, 0777)
+	cs := nbs.NewLocalStore(dbName, clienttest.DefaultMemTableSize)
 	db := datas.NewDatabase(cs)
 
 	id := "testdataset"
@@ -55,9 +60,9 @@ func (s *nomsDsTestSuite) TestNomsDs() {
 	err = db.Close()
 	s.NoError(err)
 
-	dbSpec := spec.CreateDatabaseSpecString("nbs", dir+"/name")
-	datasetName := spec.CreateValueSpecString("nbs", dir+"/name", id)
-	dataset2Name := spec.CreateValueSpecString("nbs", dir+"/name", id2)
+	dbSpec := spec.CreateDatabaseSpecString("nbs", dbName)
+	datasetName := spec.CreateValueSpecString("nbs", dbName, id)
+	dataset2Name := spec.CreateValueSpecString("nbs", dbName, id2)
 
 	// both datasets show up
 	rtnVal, _ := s.MustRun(main, []string{"ds", dbSpec})
--- a/cmd/noms/noms_sync_test.go
+++ b/cmd/noms/noms_sync_test.go
@@ -6,7 +6,6 @@ package main
 
 import (
 	"os"
-	"path"
 	"testing"
 
 	"github.com/attic-labs/noms/go/datas"
@@ -34,8 +33,7 @@ func (s *nomsSyncTestSuite) TestSyncValidation() {
 	source1.Database().Close()
 	sourceSpecMissingHashSymbol := spec.CreateValueSpecString("nbs", s.DBDir, source1HeadRef.String())
 
-	db2dir := path.Join(s.TempDir, "db2")
-	sinkDatasetSpec := spec.CreateValueSpecString("nbs", db2dir, "dest")
+	sinkDatasetSpec := spec.CreateValueSpecString("nbs", s.DBDir2, "dest")
 
 	defer func() {
 		err := recover()
@@ -46,8 +44,7 @@ func (s *nomsSyncTestSuite) TestSyncValidation() {
 }
 
 func (s *nomsSyncTestSuite) TestSync() {
-	db2dir := path.Join(s.TempDir, "db2")
-	defer s.NoError(os.RemoveAll(db2dir))
+	defer s.NoError(os.RemoveAll(s.DBDir2))
 
 	sourceDB := datas.NewDatabase(nbs.NewLocalStore(s.DBDir, clienttest.DefaultMemTableSize))
 	source1 := sourceDB.GetDataset("src")
@@ -59,11 +56,11 @@ func (s *nomsSyncTestSuite) TestSync() {
 	sourceDB.Close()
 
 	sourceSpec := spec.CreateValueSpecString("nbs", s.DBDir, "#"+source1HeadRef.String())
-	sinkDatasetSpec := spec.CreateValueSpecString("nbs", db2dir, "dest")
+	sinkDatasetSpec := spec.CreateValueSpecString("nbs", s.DBDir2, "dest")
 	sout, _ := s.MustRun(main, []string{"sync", sourceSpec, sinkDatasetSpec})
 
 	s.Regexp("Created", sout)
-	db := datas.NewDatabase(nbs.NewLocalStore(db2dir, clienttest.DefaultMemTableSize))
+	db := datas.NewDatabase(nbs.NewLocalStore(s.DBDir2, clienttest.DefaultMemTableSize))
 	dest := db.GetDataset("dest")
 	s.True(types.Number(42).Equals(dest.HeadValue()))
 	db.Close()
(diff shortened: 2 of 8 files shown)


jesseditson:
Seeing the same tests (`TestScalars`) fail on master, so assuming this PR doesn't break them. However, I believe this approach may be the wrong one after digging in to the use case a bit - we don't want to prevent this mkdir, but rather the resolver one. Closing.

aboodman:
Did you mean to close this?

aboodman:
Oh duh, sorry didn't read your comment carefully enough.

jesseditson:
@aboodman Yeah, I either failed to build correctly or failed to fix the issue at the right place in the stack, figured I'd close it until later when I can verify the fix! Let me know if you'd prefer it to stay open and/or if you'd like to verify on your side! Go is new to me, so mostly working through the build process - the tests and code were easy enough to read though!

jesseditson:
Heh - was just missing `go install`, this is working as expected. Reopening, ready for review.

aboodman:
To @cmasone-attic for review.

When I patch this into my client and do `noms ds foo`, noms prints a stack trace. Ideally, I would like it to print a nice error message instead with no stack trace.

@cmasone-attic if you can recommend an easy way to do this, let's do it in this patch. Otherwise, this is already a fine improvement.

aboodman:
(thanks @jesseditson!)

aboodman:
Oh, actually this is causing TestNomsCommitDuplicate to fail. You can replicate this by doing:

```
cd <nomsroot>/cmd/noms
go test ./...
```

You should also test the entire go tree:

```
cd <nomsroot>/go
go test ./...
```

jesseditson:
@aboodman Ah cool! Couldn't see the failures on my end, but had a failure locally on both master and this branch, so just assumed it was the same.

Makes sense about the stack trace! I'll move it to a normal `panic`

aboodman:
I added you to be a collaborator of this project just now. I think that when you accept, you'll be able to see the errors from the bots.

Looking at this code, I think that the right thing to do is that the various noms command should be "catching" this sort of error using `d.TryCatch` (https://github.com/attic-labs/noms/blob/master/go/d/try.go#L76).

But that would require updating lots of call sites, so I think we should do that in a different patch.

jesseditson:
Looks like jenkins is hungry for more permissions, here's what it looks like on my end:


RE: Try/Catch - is the motivation there is to avoid embedding the user-facing message that deep? If the message is ok where it is, it seems like just using `panic` directly might be a simpler path, but obviously I have very little context.

aboodman:
Hm, I tried pushing some more buttons. What does jenkins say now?

Re: try/catch. panic() won't help -- I may be misunderstanding, but what
I'm trying to do is prevent commands like `noms ds <not-a-dir>` from
dumping a stack.

Switching d.PanicIfError to panic won't help because panic is what's
causing the stack trace. In order to stop it, we need to either (a) not
panic in the first place (e.g., return the error up the call stack
manually), or else (b) catch the panic high up in the call stack (like in
the noms command) and print out a friendly error instead.

d.TryCatch and d.PanicIfError are designed to be used together to
facilitate solution (b). But the code in the command is not using
d.TryCatch so it's not working with your patch.

Put another way, your patch is a step toward fixing this, but another step
is required: all the noms commands need to wrap their calls to Noms in
d.TryCatch and print out a friendly error message in the Catch function.
But this second step is a lot more work, so I think we should save it for a
separate patch.

jesseditson:
Ah, I must be misunderstanding the purpose of `d.Wrap` - I assumed that was there purely to add the trace, and that `panic` would just print and exit. I'm sure the real reason differs.
Totally makes sense now!

Wondering if Jenkins is not letting me see it because this commit was before I had access? I'll push an update in a sec and see if I can see the test results.

jesseditson:
Still seeing the two failing tests locally (`TestScalars`, `TestQueryGraphQL`), but this latest update should get `cmd/noms` test back to where they should be!

jesseditson:
Unfortunately still can't see test results, so LMK how that goes!

arv:
How are you running the tests? 

You need to make sure that you do not run the tests under vendor.

jesseditson:
I've been running `go test ./...` - but I also tried `go test $(go list ./... | grep -v /vendor/)`, with the same results!

jesseditson:
(this is from inside `go` - tests inside `cmd/noms` pass happily now)

aboodman:
Are you sure that `TestScalars` and `TestQueryGraphQL` fail without your change? They pass locally for me, and they seem to be passing on the bots.

If they are failing locally, can you please show an example of one failing?

jesseditson:
I'll verify!

cmd/noms/noms_sync_test.go
```diff
@@ -63,6 +65,7 @@ func (s *nomsSyncTestSuite) TestSync() {
 	sout, _ := s.MustRun(main, []string{"sync", sourceSpec, sinkDatasetSpec})
 
 	s.Regexp("Created", sout)
+	os.Mkdir(db2dir, 0777)
```
aboodman: Instead of replicating this everywhere, how about modifying `ClientTestSuite::SetupSuite()` to create the directory? So that it will already be created when every test tries to use it.

jesseditson:
Still seeing failures on master :(

<details>
  <summary>Here's the trace:</summary>

  ```
# (prompt)  ~/go/src/github.com/attic-labs/noms/go [master*]
$ go test $(go list ./... | grep -v /vendor/)
ok  	github.com/attic-labs/noms/go/chunks	0.026s
ok  	github.com/attic-labs/noms/go/config	0.040s
?   	github.com/attic-labs/noms/go/constants	[no test files]
ok  	github.com/attic-labs/noms/go/d	0.021s
ok  	github.com/attic-labs/noms/go/datas	0.893s
ok  	github.com/attic-labs/noms/go/diff	0.088s
ok  	github.com/attic-labs/noms/go/hash	0.016s
ok  	github.com/attic-labs/noms/go/marshal	0.034s
ok  	github.com/attic-labs/noms/go/merge	0.059s
ok  	github.com/attic-labs/noms/go/migration	0.096s
ok  	github.com/attic-labs/noms/go/nbs	2.039s
?   	github.com/attic-labs/noms/go/nbs/benchmarks	[no test files]
?   	github.com/attic-labs/noms/go/nbs/benchmarks/cachedrop	[no test files]
?   	github.com/attic-labs/noms/go/nbs/benchmarks/chunker	[no test files]
?   	github.com/attic-labs/noms/go/nbs/benchmarks/gen	[no test files]
?   	github.com/attic-labs/noms/go/nbs/frag	[no test files]
?   	github.com/attic-labs/noms/go/nbs/test	[no test files]
--- FAIL: TestScalars (0.00s)
        Error Trace:    query_test.go:35
			query_test.go:45
	Error:  	Not equal: "{\"data\":{\"root\":2.147483648e+09}}\n" (expected)
			        != "{\"data\":{\"root\":2147483648}}\n" (actual)
		
        Error Trace:    query_test.go:35
			query_test.go:46
	Error:  	Not equal: "{\"data\":{\"root\":-2.147483648e+09}}\n" (expected)
			        != "{\"data\":{\"root\":-2147483648}}\n" (actual)
		
        Error Trace:    query_test.go:35
			query_test.go:48
	Error:  	Not equal: "{\"data\":{\"root\":1e-08}}\n" (expected)
			        != "{\"data\":{\"root\":1e-8}}\n" (actual)
		
--- FAIL: TestQueryGraphQL (0.06s)
FAIL
FAIL	github.com/attic-labs/noms/go/ngql	0.092s
?   	github.com/attic-labs/noms/go/perf/codec-perf-rig	[no test files]
?   	github.com/attic-labs/noms/go/perf/hash-perf-rig	[no test files]
ok  	github.com/attic-labs/noms/go/perf/suite	3.098s
ok  	github.com/attic-labs/noms/go/spec	0.110s
ok  	github.com/attic-labs/noms/go/types	12.730s
ok  	github.com/attic-labs/noms/go/types/perf	0.016s
?   	github.com/attic-labs/noms/go/util/clienttest	[no test files]
ok  	github.com/attic-labs/noms/go/util/datetime	0.022s
?   	github.com/attic-labs/noms/go/util/exit	[no test files]
ok  	github.com/attic-labs/noms/go/util/functions	0.015s
?   	github.com/attic-labs/noms/go/util/integrationtest	[no test files]
ok  	github.com/attic-labs/noms/go/util/jsontonoms	0.026s
?   	github.com/attic-labs/noms/go/util/math	[no test files]
?   	github.com/attic-labs/noms/go/util/outputpager	[no test files]
?   	github.com/attic-labs/noms/go/util/profile	[no test files]
?   	github.com/attic-labs/noms/go/util/progressreader	[no test files]
ok  	github.com/attic-labs/noms/go/util/random	0.025s
ok  	github.com/attic-labs/noms/go/util/sizecache	0.022s
?   	github.com/attic-labs/noms/go/util/status	[no test files]
?   	github.com/attic-labs/noms/go/util/test	[no test files]
?   	github.com/attic-labs/noms/go/util/verbose	[no test files]
ok  	github.com/attic-labs/noms/go/util/writers	0.022s
  ```

</details>

aboodman:
I can replicate those failures in Go 1.8. Looks like something changed between Go 1.7 and 1.8.

cmasone-attic:
`TestCSVImporterFromBlob` now fails thusly:
```
panic(0xb03f00, 0xc4204d8d40)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
testing.tRunner.func1(0xc42008a180)
	/usr/local/go/src/testing/testing.go:579 +0x25d
panic(0xb03f00, 0xc4204d8d40)
	/usr/local/go/src/runtime/panic.go:458 +0x243
github.com/attic-labs/noms/go/d.PanicIfError(0xf14260, 0xc4204e0e40)
	/var/workspace/src/github.com/attic-labs/noms/go/d/try.go:44 +0x74
github.com/attic-labs/noms/go/nbs.newLocalStore(0xc42016d340, 0x19, 0x800000, 0xc420028a00, 0x80, 0xc42003a818)
	/var/workspace/src/github.com/attic-labs/noms/go/nbs/store.go:145 +0x68
github.com/attic-labs/noms/go/nbs.NewLocalStore(0xc42016d340, 0x19, 0x800000, 0xc4204e0e20)
	/var/workspace/src/github.com/attic-labs/noms/go/nbs/store.go:140 +0x70
github.com/attic-labs/noms/samples/go/csv/csv-import.(*testSuite).TestCSVImporterFromBlob.func1.1(0xad7960, 0xc4204e0e20)
	/var/workspace/src/github.com/attic-labs/noms/samples/go/csv/csv-import/importer_test.go:127 +0x40
github.com/attic-labs/noms/samples/go/csv/csv-import.(*testSuite).TestCSVImporterFromBlob.func1(0xb7e6ba, 0x6)
	/var/workspace/src/github.com/attic-labs/noms/samples/go/csv/csv-import/importer_test.go:131 +0xc7
github.com/attic-labs/noms/samples/go/csv/csv-import.(*testSuite).TestCSVImporterFromBlob(0xc4200175c0)
	/var/workspace/src/github.com/attic-labs/noms/samples/go/csv/csv-import/importer_test.go:151 +0x5c
reflect.Value.call(0xc42022b6e0, 0xc420028840, 0x13, 0xb7da73, 0x4, 0xc42044df40, 0x1, 0x1, 0x5b63f8, 0xb792a0, ...)
	/usr/local/go/src/reflect/value.go:434 +0x5c8
reflect.Value.Call(0xc42022b6e0, 0xc420028840, 0x13, 0xc42044df40, 0x1, 0x1, 0xc42002ac00, 0x0, 0xf8)
	/usr/local/go/src/reflect/value.go:302 +0xa4
github.com/attic-labs/noms/vendor/github.com/attic-labs/testify/suite.Run.func2(0xc42008a180)
	/var/workspace/src/github.com/attic-labs/noms/vendor/github.com/attic-labs/testify/suite/suite.go:95 +0x1cb
testing.tRunner(0xc42008a180, 0xc420331ab0)
	/usr/local/go/src/testing/testing.go:610 +0x81
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:646 +0x2ec
```

cmd/noms/noms_ds_test.go
```diff
@@ -25,21 +26,25 @@ type nomsDsTestSuite struct {
 
 func (s *nomsDsTestSuite) TestEmptyNomsDs() {
 	dir := s.DBDir
+	dbName := dir+"/name"
```
cmasone-attic: Presumably `s.DBDir` is already an empty directory, no?

Can't you just use that, which should already exist? If you change it, and the test fails then feel free to keep it like this -- I'm not going to ask you to debug that kind of thing :-)

cmd/noms/noms_ds_test.go
```diff
@@ -25,21 +26,25 @@ type nomsDsTestSuite struct {
...
 
-	dbSpec := spec.CreateDatabaseSpecString("nbs", dir+"/name")
+	dbSpec := spec.CreateDatabaseSpecString("nbs", dbName)
 	rtnVal, _ := s.MustRun(main, []string{"ds", dbSpec})
 	s.Equal("", rtnVal)
 }
 
 func (s *nomsDsTestSuite) TestNomsDs() {
 	dir := s.DBDir
+	dbName := dir+"/name"
```
cmasone-attic: Same Q as above. Unclear why the (admittedly already-existing) code makes a new directory. Do we need that?

go/nbs/store.go
```diff
@@ -92,8 +93,21 @@ type LocalStoreFactory struct {
 	maxTables  int
 }
 
+func CheckDir(dir string) error {
+	stat, err := os.Stat(dir)
+	if os.IsNotExist(err) {
+		return fmt.Errorf("Directory does not exist: %s", dir)
```
cmasone-attic: Presumably, `err` is already an error that says `dir` doesn't exist...why not just return that?

go/util/clienttest/client_test_suite.go
```diff
@@ -43,6 +43,9 @@ func (suite *ClientTestSuite) SetupSuite() {
 	suite.out = stdOutput
 	suite.err = errOutput
 	exit.Exit = mockExit
+
+	os.Mkdir(suite.DBDir, 0777)
+	os.Mkdir(path.Join(suite.TempDir, "db2"), 0777)
```
cmasone-attic: What's this second directory for? Is it that a number of the tests make a directory called "db2" and then try to use it? If so, that's kind of an unstated dependency...the test code only works because it "knows" that the harness creates this directory. If many of the tests depend on this directory existing in order to be correct, then the path to this dir should be exposed via some member on ClientTestSuite, and the tests should use that -- as we did with `suite.DBDir`

arv:
So Go 1.8 changed the JSON output. I'll update the tests to not care about the exact output...

cmd/noms/noms_ds_test.go
jesseditson: Works for me!

go/nbs/store.go
jesseditson: This is mainly because I felt that the original message (`Error:	stat foo: no such file or directory`) wasn't very user friendly (files are not allowed, and `stat` could be confusing without context) - that being said, users are likely devs and the original error can be tested using `os.IsNotExist` when this gets wrapped in a try/catch down the road (as per @aboodman's earlier comments). In light of that, I think it's better to return this error so that some display logic can detect what kind of error this is.

go/util/clienttest/client_test_suite.go
jesseditson: Makes sense to me!

jesseditson:
@cmasone-attic feedback addressed - also rebased against master! Not sure which tests are failing in CI, let me know if any of those are things I can address!

cmasone-attic:
The failure remains the one I indicated above, with the stack trace above. it's in `samples/go/csv/csv-import/importer_test.go`

jesseditson:
Ah, I understand now! Thought you were responding to Aaron up there. Fixed up that test. Also realized why the recommended test command omits vendor - I had been running go test in the subdirs, not the root. I'll run all the suites when testing from now on!

aboodman:
screen shot 2017-03-01 at 1 28 19 pm

aboodman:
Thanks @jesseditson !

jesseditson:
My pleasure! Sorry about the rocky start, but sure was fun getting my hands dirty with some golang. What an enjoyable language.

PR #3237, opened 2017-03-01
GraphQL: Add support for keys to Maps

arv:
This adds a `keys` argument to Map types. This allows you to get
multiple entries in one line:

```graphql
{
  root {
    elements(keys: ["a", "c"]) {
      key
      value
    }
  }
}
```

Outputs:

```json
{
  "data": {
    "root": {
      "elements": [
        {"key": "a", "value": 1},
        {"key": "c", "value": 3}
      ]
    }
  }
}
```

If `keys` is present all other args are ignored.

Issue #3227


arv:
@aboodman PTAL at the changes to MapIterator

@rafael-atticlabs PTAL at the GraphQL specifics

arv:
+ @kalman Since @rafael-atticlabs needs some rest...

go/ngql/types.go
```diff
@@ -330,9 +341,32 @@ func getArgValue(arg interface{}, nomsType *types.Type) (types.Value, error) {
...
+}
+
+func getCollectionArgs(col nomsCollection, args map[string]interface{}, factory iteratorFactory) (iter interface{}, nomsKey, nomsThrough types.Value, count uint64, singleExactMatch bool, err error) {
+	typ := col.Type()
+	len := col.Len()
 	nomsKeyType := typ.Desc.(types.CompoundDesc).ElemTypes[0]
 
+	if keys, ok := args[keysKey]; ok {
+		var nomsValue types.Value
+		nomsValue, err = getArgValue(keys, types.MakeListType(nomsKeyType))
```
kalman: It's kinda weird to be constructing this temporary noms list just to use it to iterate through map keys. Can you just use a `[]string`? I know it involves changing a bit of code.

go/ngql/types.go
```diff
@@ -330,9 +341,32 @@ func getArgValue(arg interface{}, nomsType *types.Type) (types.Value, error) {
 	return nil, fmt.Errorf("%s is not a subtype of %s", nomsVal.Type().Describe(), nomsType.Describe())
 }
 
-func getCollectionArgs(typ *types.Type, len uint64, args map[string]interface{}, factory iteratorFactory) (iter interface{}, nomsKey, nomsThrough types.Value, count uint64, singleExactMatch bool, err error) {
+type nomsCollection interface {
+	types.Value
+	Len() uint64
+}
+
+func getCollectionArgs(col nomsCollection, args map[string]interface{}, factory iteratorFactory) (iter interface{}, nomsKey, nomsThrough types.Value, count uint64, singleExactMatch bool, err error) {
```
kalman: I think "using keys prevents using any other arguments" makes sense to prevent key/count, but as a general rule, it seems wrong?

go/types/map_iterator.go
```diff
@@ -4,22 +4,27 @@
...
+	Next() (k, v Value)
+}
+
+// mapIterator can efficiently iterate through a Noms Map.
+type mapIterator struct {
 	cursor       *sequenceCursor
-	CurrentKey   Value
-	CurrentValue Value
+	currentKey   Value
+	currentValue Value
```
arv: I'll update attic... This is a better model and it is more consistent with LisIterator and SetIterator.

This was only added last week so I think changing this should be safe.

go/ngql/types.go
arv: We could allow them. It just does not seem useful.

go/ngql/types.go
arv: Sure. Should be simple enough.

go/types/map_iterator.go
kalman: Have a look how they're used, also CC @aboodman

go/ngql/types.go
arv: I made the change but I don't really like it. I understand that creating a types.List and an Iterator for that is not cheap but having to duplicate this code just to optimize this case seems bad. I'll upload my changes and you can look.

go/types/map_iterator.go
arv: i fixed the usages in Attic 😛

go/ngql/types.go
```diff
@@ -671,3 +678,21 @@ func maybeGetScalar(v types.Value) interface{} {
...
+	idx  int
+}
+
+func (it *mapIteratorForKeys) Next() (k, v types.Value) {
+	if it.idx >= len(it.keys) {
+		return
+	}
+	k = it.keys[it.idx]
+	it.idx++
+	if k != nil {
```
kalman: why bother checking for nil? doesn't seem possible/advisable to be nil.

arv: Old code... removed

PR #3242, opened 2017-03-02
Close output channels in streaming collection builders

cmasone-attic:
Since these functions return a receive-only channel, the
caller _can't_ close the channel it pulls the constructed
collection off of. Since the builder function can easily
know when it's safe to close the channel, it seems fine
to just have NewStreamingWhatever() do the close.


cmasone-attic:
PTAL. I believe we've been leaking channels

PR #3244, opened 2017-03-02
Fix NBS LocalStoreFactory

cmasone-attic:
While it's reasonable to take out the automatic directory
creation for paths that the user types, I think we want the
factory to always be able to use the paths that it authors
underneath the directory the user passed in.


cmasone-attic:
PTAL

PR #3246, opened 2017-03-03
Support other endpoints and extra args in GraphiQL

kalman:


kalman:
@rafael-atticlabs I don't know who owns this

samples/js/graphiql/src/main.js
```diff
@@ -93,27 +117,28 @@ class Prompt extends React.Component<void, PromptProps, void> {
...
-  const url = `${params.db}/graphql/`;
+  const url = `${params.db}/${params.endpoint}`;
 
   const headers = new Headers();
   headers.append('Content-Type', 'application/x-www-form-urlencoded; charset=utf-8');
   if (params.auth) {
     headers.append('Authorization', `Bearer ${params.auth}`);
   }
 
+  const extra = params.extra ? ('&' + params.extra) : '';
```
arv: encodeURIComponent here?

kalman: Not what I want to do, the purpose of this is to have non-encoded URI components so you can go `db=foo&something=bar`, which wouldn't be right if it were URI encoded.

Issue #3236, opened 2017-02-28
GraphQL is pretty slow


arv:
The GraphQL endpoint is surprisingly slow (order of 5-10s) for a pretty simple query (8kb result)

Things we could do to potentially speed things up:

 - [ ] When we are only getting one element from a collection, do not use an iterator with read ahead
 - [ ] Parallelize the resolvers
 - [ ] Instrument/benchmark https://github.com/graphql-go/graphql

rafael-atticlabs:
Just to be clear, it's not generally slow, a specific query you created seems to taking longer than one might expect. Can you include that query here for posterity?

arv:
```graphql
{
  root {
    value {
      byTag {
        elements {
          key
          value {
            elements(count: 1) {
              value {
                elements(count: 1) {
                  cover {
                    images {
                      elements {
                        key {
                          height
                          width
                        }
                        value {
                          blobRef {
                            targetHash
                          }
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

arv:
We also compute the schema from scratch on every request. We should probably cache the graphql types between requests...

rafael-atticlabs:
It occurred to me this morning that we could move in the direction of optimal value resolution without having to hack graphql-go.

When we fetch data from collections, we'll always know exactly the start->end range we need. If we add methods to collection which allow directly fetching a range of values, we can always be assured to load exactly the sequences we need. It also has the benefit of not needing to open a cursor first.

We used to have something like this in js (`List.toJS`). I don't think it'd be super hard to implement in go, and I suspect it'd be 60-80% of the benefit of the "right" strategy.

arv:
It will help with the `count:1` case.  It will also help with the paging case.

rafael-atticlabs:
Thoughts about this:

My hypothesis is that the slowness we see (for instance populating the filter ui with data) is primarily latency waiting on lots of individual reads (sequentially) from s3.

I reviewed the graphql-go code yesterday and resolution of fields is entirely serial. So I'd suspect that resolving the data we need is sub-optimal in several directions:

1) There's no parallelizing in descending the request graph
2) In many cases, we are fetching fewer or more sequences than we'd eventually need (i.e. not reading ahead enough or too much)

In some sense, I think the "optimal" solution would look like something which starts from the root chunk, enumerates exactly the set of chunks which are needed at the next level, fetches them in parallel and then keep iterating out like that.

[...]

PR #3249, opened 2017-03-03
GraphQL: Do not do read ahead when count is 1

arv:
When doing a query like

```graphql
values(count:1)
```

We used to do a read ahead. This removes that read ahead.

Towards #3236

### DIFF
--- a/go/ngql/types.go
+++ b/go/ngql/types.go
@@ -348,7 +348,7 @@ func getCollectionArgs(col types.Collection, args map[string]interface{}, factor
 		}
 		iter = factory.IteratorAt(uint64(idx))
 	} else if count == 1 && !singleExactMatch {
-		// no  key, no at, no through, but a count:1
+		// no key, no at, no through, but a count:1
 		iter = factory.First()
 	} else {
 		iter = factory.IteratorAt(0)
--- a/go/ngql/types.go
+++ b/go/ngql/types.go
@@ -261,6 +261,9 @@ func getSetElements(v types.Value, args map[string]interface{}) (interface{}, er
 		IteratorAt: func(at uint64) interface{} {
 			return s.IteratorAt(at)
 		},
+		First: func() interface{} {
+			return &setFirstIterator{s: s}
+		},
 	})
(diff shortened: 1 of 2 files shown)


go/ngql/types.go
```diff
@@ -334,28 +343,17 @@ func getCollectionArgs(col types.Collection, args map[string]interface{}, factor
...
 		}
 		iter = factory.IteratorAt(uint64(idx))
+	} else if count == 1 && !singleExactMatch {
+		// no  key, no at, no through, but a count:1
+		iter = factory.First()
 	} else {
 		iter = factory.IteratorAt(0)
 	}
 
-	if err != nil {
```
arv: This could never happen

go/ngql/types.go
```diff
@@ -334,28 +343,17 @@ func getCollectionArgs(col types.Collection, args map[string]interface{}, factor
...
-		return
-	}
-	if iter == nil {
```
arv: This could never happen

go/ngql/types.go
```diff
@@ -748,3 +750,23 @@ func (it *mapIteratorForKeys) Next() (k, v types.Value) {
 	it.idx++
 	return
 }
+
+type setFirstIterator struct {
+	s types.Set
+}
+
+func (it *setFirstIterator) Next() types.Value {
+	return it.s.First()
```
arv: I initially kept track of done but turns out that we never hit that since we already handle the end using count.

go/ngql/types.go
```diff
@@ -334,28 +343,17 @@ func getCollectionArgs(col types.Collection, args map[string]interface{}, factor
 		if idx < 0 {
 			idx = 0
 		} else if uint64(idx) > length {
-			// count is already 0
+			count = 0
 			return
 		}
 		iter = factory.IteratorAt(uint64(idx))
+	} else if count == 1 && !singleExactMatch {
+		// no  key, no at, no through, but a count:1
```
kalman: has extra space

PR #3250, opened 2017-03-03
Make some ngql methods public

kalman:


go/ngql/query.go
```diff
@@ -37,29 +37,33 @@ const (
...
+	rootType := NomsTypeToGraphQLType(rootNomsType, false, tm)
 
 	return graphql.NewObject(graphql.ObjectConfig{
 		Name: rootQueryKey,
 		Fields: graphql.Fields{
 			rootKey: &graphql.Field{
 				Type: rootType,
 				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
-					return maybeGetScalar(rootValue), nil
+					return MaybeGetScalar(rootValue), nil
```
kalman: Not super happy about exposing `MaybeGetScalar`, but I need to. If you can suggest a better name then great

kalman:
CC @rafael-atticlabs

go/ngql/query.go
arv: Maybe{Wrap,Box}Scalar ?

arv: Sorry... that is not what it does

arv: Can't think of a better name...

PR #3251, opened 2017-03-03
Add TypeMarshaler interface

arv:
This allows go structs to participate in the MarshalType flow even
when they implement the Marshaler interface or if they have an original
tag.

### DIFF
--- a/go/marshal/encode_type.go
+++ b/go/marshal/encode_type.go
@@ -86,14 +86,14 @@ func encodeType(t reflect.Type, parentStructTypes []reflect.Type, tags nomsTags,
 	}
 
 	if t.Implements(marshalerInterface) {
+		// There is no way to determine the noms type now. For Marshal it can be
+		// different each time MarshalNoms is called and is handled further up the
+		// stack.
 		if options.ReportErrors {
 			err := fmt.Errorf("Cannot marshal type which implements %s, perhaps implement %s for %s", marshalerInterface, typeMarshalerInterface, t)
 			panic(&marshalNomsError{err})
 		}
 
-		// There is no way to determine the noms type now. For Marshal it can be
-		// different each time MarshalNoms is called. This is handled further up the
-		// stack.
 		return nil
 	}
 
--- a/go/marshal/encode_type.go
+++ b/go/marshal/encode_type.go
@@ -38,6 +38,7 @@ func MarshalType(v interface{}) (nt *types.Type, err error) {
 	rv := reflect.ValueOf(v)
 	nt = encodeType(rv.Type(), nil, nomsTags{}, encodeTypeOptions{
 		IgnoreOmitEmpty: true,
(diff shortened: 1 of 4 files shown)


go/marshal/encode_type.go
```diff
@@ -54,13 +55,36 @@ func MustMarshalType(v interface{}) *types.Type {
...
+		if typ == nil {
+			panic(fmt.Errorf("nil result from %s.MarshalNomsType", t.String()))
+		}
+		return typ
+
+	}
+
 	if t.Implements(marshalerInterface) {
 		// There is no way to determine the noms type now. For Marshal it can be
 		// different each time MarshalNoms is called. This is handled further up the
```
kalman: Could we actually generate an error here which says along the lines of `"Cannot marshal type which implements marshal.Marshaler, perhaps implement marshal.TypeMarshaler for $typename"`?

go/marshal/encode_type.go
```diff
@@ -54,13 +55,36 @@ func MustMarshalType(v interface{}) *types.Type {
...
+		v := reflect.New(t)
+		typ, err := v.Interface().(TypeMarshaler).MarshalNomsType()
+		if err != nil {
+			panic(&marshalNomsError{err})
+		}
+		if typ == nil {
+			panic(fmt.Errorf("nil result from %s.MarshalNomsType", t.String()))
+		}
+		return typ
+
```
kalman: extra blank line

go/marshal/encode_type.go
arv: Not at the moment. This is used by other parts of the system and they depend on nil. I could refactor to have those parts check for an error instead

arv: I'm doing this... turns out to be simple.

go/marshal/encode_type.go
```diff
@@ -68,24 +69,28 @@ var typeOfTypesType = reflect.TypeOf((*types.Type)(nil))
...
-
 	}
 
 	if t.Implements(marshalerInterface) {
+		if options.ReportErrors {
+			err := fmt.Errorf("Cannot marshal type which implements %s, perhaps implement %s for %s", marshalerInterface, typeMarshalerInterface, t)
+			panic(&marshalNomsError{err})
+		}
+
 		// There is no way to determine the noms type now. For Marshal it can be
```
kalman: Move comment up

go/marshal/encode_type.go
```diff
@@ -108,6 +113,12 @@ func encodeType(t reflect.Type, parentStructTypes []reflect.Type, tags nomsTags,
 		case "String":
 			return types.StringType
 		}
+
+		if options.ReportErrors {
+			err := fmt.Errorf("Cannot marshal type %s, it requires type parameters", t)
```
kalman: I don't know what this error means

PR #3261, opened 2017-03-08
Make GraphQL struct fields nullable

arv:
This is needed because we do not have optional fields in noms structs.
Therefore the noms struct type has the field but the the noms struct may
not have it.

### DIFF
--- a/go/ngql/types.go
+++ b/go/ngql/types.go
@@ -69,7 +69,8 @@ func isScalar(nomsType *types.Type) bool {
 	}
 }
 
-// Note: Always returns a graphql.NonNull() as the outer type.
+// NomsTypeToGraphQLType creates a GraphQL type from a Noms type that knows how
+// to resolve the Noms values.
 func NomsTypeToGraphQLType(nomsType *types.Type, boxedIfScalar bool, tm *typeMap) graphql.Type {
 	key := typeMapKey{nomsType.Hash(), boxedIfScalar && isScalar(nomsType)}
 	gqlType, ok := (*tm)[key]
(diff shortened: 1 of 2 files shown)


go/ngql/types.go
```diff
@@ -196,13 +196,15 @@ func structToGQLObject(nomsType *types.Type, tm *typeMap) *graphql.Object {
 			}
 
 			structDesc.IterFields(func(name string, nomsFieldType *types.Type) {
-				fieldType := NomsTypeToGraphQLType(nomsFieldType, false, tm)
+				fieldType := unpackNonNullType(NomsTypeToGraphQLType(nomsFieldType, false, tm))
```
kalman: This `unpackNonNullType` thing doesn't feel right. It makes me wonder if we should do a sweep and *not* mark every type as non-null in the first place, then push that decision into the containers or wherever the fields are stored.

I.e. I'm looking at `NomsTypeToGraphQLType` which wraps everything in `graphql.NewNonNull`. What if:
* We didn't do that.
* We add the non-nullability into the list/set/map keys/values and ref.

Perhaps the problem is that graphql conflates optionality with nullability, but oh well.

arv:
PTAL

go/ngql/types.go
```diff
@@ -176,7 +169,7 @@ func unionToGQLUnion(nomsType *types.Type, tm *typeMap) *graphql.Union {
 			key := typeMapKey{nomsType.Hash(), isScalar}
 			memberType := (*tm)[key]
 			// Member types cannot be non-null and must be struct (graphl.Object)
```
kalman: Comment can just be "Member types must be struct" if it's even useful.

go/ngql/types.go
```diff
@@ -565,8 +560,8 @@ func listAndSetToGraphQLObject(nomsType *types.Type, tm *typeMap) *graphql.Objec
 	nomsValueType := nomsType.Desc.(types.CompoundDesc).ElemTypes[0]
 	var listType, valueType graphql.Type
 	if !isEmptyNomsUnion(nomsValueType) {
-		listType = NomsTypeToGraphQLType(nomsValueType, false, tm)
-		valueType = unpackNonNullType(listType)
```
kalman: Is it possible to remove `unpackNonNullType` function entirely?

PR #3263, opened 2017-03-09
Noms type to GraphQL input types

arv:
We already had a way to generate GraphQL output types from Noms types
but no way to do input types. Input types are used in arguments
position.

We currently do not do any conversion from the JSON like data in
argument position. We leave it to the users to determine if they want
to convert the data into noms data structures or go to Go structures
or operate directly on the JSON like data.

### DIFF
--- a/go/ngql/query.go
+++ b/go/ngql/query.go
@@ -58,10 +58,13 @@ func NewContext(vr types.ValueReader, tm *typeMap) context.Context {
 // Query takes |rootValue|, builds a GraphQL scheme from rootValue.Type() and
 // executes |query| against it, encoding the result to |w|.
 func Query(rootValue types.Value, query string, vr types.ValueReader, w io.Writer) {
+	schemaConfig := graphql.SchemaConfig{}
 	tm := NewTypeMap()
+	queryWithSchemaConfig(rootValue, query, schemaConfig, vr, tm, w)
+}
 
-	queryObj := constructQueryType(rootValue, tm)
-	schemaConfig := graphql.SchemaConfig{Query: queryObj}
+func queryWithSchemaConfig(rootValue types.Value, query string, schemaConfig graphql.SchemaConfig, vr types.ValueReader, tm *typeMap, w io.Writer) {
+	schemaConfig.Query = constructQueryType(rootValue, tm)
 	schema, _ := graphql.NewSchema(schemaConfig)
 	ctx := NewContext(vr, tm)
 
@@ -73,6 +76,7 @@ func Query(rootValue types.Value, query string, vr types.ValueReader, w io.Write
 
 	err := json.NewEncoder(w).Encode(r)
 	d.PanicIfError(err)
+
 }
 
 // Error writes an error as a GraphQL error to a writer.
--- a/go/ngql/query.go
+++ b/go/ngql/query.go
@@ -76,7 +76,6 @@ func queryWithSchemaConfig(rootValue types.Value, query string, schemaConfig gra
 
 	err := json.NewEncoder(w).Encode(r)
 	d.PanicIfError(err)
-
 }
 
 // Error writes an error as a GraphQL error to a writer.
--- a/go/ngql/query_test.go
+++ b/go/ngql/query_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/attic-labs/graphql"
 	"github.com/attic-labs/noms/go/chunks"
 	"github.com/attic-labs/noms/go/types"
+	"github.com/attic-labs/testify/assert"
 	"github.com/attic-labs/testify/suite"
 )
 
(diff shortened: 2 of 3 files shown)


go/ngql/query.go
```diff
@@ -73,6 +76,7 @@ func Query(rootValue types.Value, query string, vr types.ValueReader, w io.Write
 
 	err := json.NewEncoder(w).Encode(r)
 	d.PanicIfError(err)
+
```
kalman: Added \n

go/ngql/query_test.go
```diff
@@ -834,3 +834,105 @@ func (suite *QueryGraphQLSuite) TestStructWithMissingField() {
...
+							Type: inType,
+						},
+					},
+					Resolve: resolver,
+				},
+			},
+		}),
+	}
+	queryWithSchemaConfig(root, query, schemaConfig, suite.vs, tm, buf)
+	suite.JSONEq(expected+"\n", string(buf.Bytes()))
```
kalman: Is the `+"\n"` needed with `suite.JSONEq`?

go/ngql/types.go
```diff
@@ -22,12 +22,20 @@ type typeMap map[typeMapKey]graphql.Type
...
 
 func NewTypeMap() *typeMap {
 	return &typeMap{}
 }
 
+type graphQLTypeMode bool
+
+const (
+	inputMode  = graphQLTypeMode(true)
+	outputMode = graphQLTypeMode(false)
```
kalman: More idiomatic:
```go
const (
  inputMode  = iota
  outputMode = iota
)
```
and `typeMode` as an `int`.

This would also be a logical place to explain the difference between input and output types, as you explained to me.

go/ngql/types.go
```diff
@@ -559,8 +689,10 @@ func argsWithSize() graphql.Fields {
 func listAndSetToGraphQLObject(nomsType *types.Type, tm *typeMap) *graphql.Object {
 	nomsValueType := nomsType.Desc.(types.CompoundDesc).ElemTypes[0]
 	var listType, valueType graphql.Type
+	var keyInputType graphql.Input
 	if !isEmptyNomsUnion(nomsValueType) {
 		valueType = NomsTypeToGraphQLType(nomsValueType, false, tm)
+		keyInputType = NomsTypeToGraphQLInputType(nomsValueType, tm)
```
kalman: As discussed: perhaps add test for non-scalar key type.

go/ngql/query_test.go
arv: No, old code survived the transition...

arv:
PTAL

This reworks how key/through/keys converts from GraphQL input type to noms value.

go/ngql/types.go
```diff
@@ -476,6 +578,22 @@ func mapEntryToGraphQLObject(keyType, valueType graphql.Type, nomsKeyType, nomsV
 	}))
 }
 
+func mapEntryToGraphQLInputObject(keyType, valueType graphql.Input, nomsKeyType, nomsValueType *types.Type, tm *typeMap) graphql.Input {
+	return graphql.NewInputObject(graphql.InputObjectConfig{
+		Name: fmt.Sprintf("%s%sEntryInput", getTypeName(nomsKeyType), getTypeName(nomsValueType)),
+		Fields: graphql.InputObjectConfigFieldMapThunk(func() graphql.InputObjectConfigFieldMap {
+			return graphql.InputObjectConfigFieldMap{
+				keyKey: &graphql.InputObjectFieldConfig{
+					Type: keyType,
```
kalman: should these be NotNull?

go/ngql/types.go
```diff
@@ -207,12 +273,68 @@ func structToGQLObject(nomsType *types.Type, tm *typeMap) *graphql.Object {
...
+
+	entryType := mapEntryToGraphQLInputObject(keyType, valueType, nomsKeyType, nomsValueType, tm)
+	return graphql.NewList(entryType)
+}
+
+func structToGQLInputObject(nomsType *types.Type, tm *typeMap) graphql.Input {
+	// GraphQL input types cannot contain cycles.
+	unresolved := types.ToUnresolvedType(nomsType)
+	if unresolved.HasUnresolvedCycle() {
+		return nil
```
kalman: Is it possible to set errors anywhere for these, or do we just silently fail?

go/ngql/types.go
```diff
@@ -730,6 +848,52 @@ func MaybeGetScalar(v types.Value) interface{} {
 	return v
 }
 
+func argToNomsValue(arg interface{}, nomsType *types.Type) types.Value {
+	switch nomsType.Kind() {
+	case types.BoolKind:
+		return types.Bool(arg.(bool))
+	case types.NumberKind:
+		if i, ok := arg.(int); ok {
+			return types.Number(float64(i))
```
kalman: I'm not sure you need the extra `float64` cast, since number types can be cast to each other, and types.Number is just a number type.

go/ngql/types.go
arv: Yes.

go/ngql/types.go
arv: yeah, not needed.

go/ngql/types.go
arv: Not sure where to report these. At the moment if we end up with nil for the input type we do not add the arguments (key, keys, through). Unknown errors are errors in GrpahQL (we still have that validation on) but it might be confusing.

Failing when run into these is not an option because we run into these when we automatically create the schema based on the data in the dataset.

One option might be to return an error from NomsTypeToGraphQLInputType but ignore that in the auto schema generation... I'll give it a try..

arv:
Made it return `(graphql.Input, error)`. A bit more boilerplate code but it simplified handling of the error cases a bit.

PR #3266, opened 2017-03-13
Add network connection logging

cmasone-attic:
Run the `ss` tool in the background when issuing an S3 request. Only print its output when reading data from S3 fails, though. Also retry reads that fail in this way

### DIFF
--- a/go/nbs/s3_table_reader.go
+++ b/go/nbs/s3_table_reader.go
@@ -7,6 +7,8 @@ package nbs
 import (
 	"fmt"
 	"io"
+	"os"
+	"os/exec"
 
 	"github.com/attic-labs/noms/go/d"
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


cmasone-attic:
PTAL

go/nbs/s3_table_reader.go
```diff
@@ -48,7 +50,12 @@ func newS3TableReader(s3 s3svc, bucket string, h addr, chunkCount uint32, indexC
 		size := indexSize(chunkCount) + footerSize
 		buff := make([]byte, size)
 
-		n, err := source.readRange(buff, fmt.Sprintf("%s=-%d", s3RangePrefix, size))
+		rngHdr := fmt.Sprintf("%s=-%d", s3RangePrefix, size)
+		n, err := source.readRange(buff, rngHdr)
+		for rErr, ok := err.(readError); ok; rErr, ok = err.(readError) {
```
zcstarr: will this retry terminate at some point ?

cmasone-attic:
removed retries. PTAL

PR #3267, opened 2017-03-14
Refactor ngql to use an object with some state

arv:
And introduce a way to provide a naming function so that we can
create names that are stable

### DIFF
--- a/go/ngql/query.go
+++ b/go/ngql/query.go
@@ -29,7 +29,6 @@ const (
 	targetHashKey  = "targetHash"
 	targetValueKey = "targetValue"
 	throughKey     = "through"
-	tmKey          = "tm"
 	valueKey       = "value"
 	valuesKey      = "values"
 	vrKey          = "vr"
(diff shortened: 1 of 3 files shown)


go/ngql/query.go
```diff
@@ -29,7 +29,6 @@ const (
 	targetHashKey  = "targetHash"
 	targetValueKey = "targetValue"
 	throughKey     = "through"
-	tmKey          = "tm"
```
kalman: 👍

go/ngql/query.go
```diff
@@ -38,8 +37,15 @@ const (
 // NewRootQueryObject creates a "root" query object that can be used to
 // traverse the value tree of rootValue.
 func NewRootQueryObject(rootValue types.Value, tm *TypeMap) *graphql.Object {
+	tc := &TypeConverter{tm, DefaultNameFunc}
```
kalman: This doesn't need to be a pointer.

go/ngql/query.go
```diff
@@ -56,21 +62,27 @@ func NewRootQueryObject(rootValue types.Value, tm *TypeMap) *graphql.Object {
 // NewContext creates a new context.Context with the extra data added to it
 // that is required by ngql.
 func NewContext(vr types.ValueReader, tm *TypeMap) context.Context {
-	return context.WithValue(context.WithValue(context.Background(), vrKey, vr), tmKey, tm)
+	return context.WithValue(context.Background(), vrKey, vr)
```
kalman: It seems like we should be able to get rid of Context entirely and put everything on TypeConverter? Where "everything" is the ValueReader at this point.

go/ngql/types.go
```diff
@@ -13,18 +13,55 @@ import (
...
+	return tc.nomsTypeToGraphQLInputType(nomsType)
+}
+
 // TypeMap is used as a cache in NomsTypeToGraphQLType and
 // NomsTypeToGraphQLInputType.
 type TypeMap map[typeMapKey]graphql.Type
 
 type typeMapKey struct {
-	h             hash.Hash
+	name          string // name must be unique anyway...
```
kalman: What does this comment mean (not sure what the "anyway..." means)

arv: It used to have the hash there to ensure uniqueness but the name has to unique anyway. I removed the comment.

go/ngql/query.go
arv: As the API is now it is a bit hard. Once I've removed the old API I can make that change.

Issue #3255, opened 2017-03-07
Fix s3 'connection reset by peer' failures


cmasone-attic:
3 pronged plan of attack:

1) update aws sdk (done)
2) add a VPC endpoint for a more stable connection to S3
3) log request ID (#3206) for use in support followup if needed

rafael-atticlabs:
@cmasone-attic is this bug actionable at all? should be keep it open?

cmasone-attic:
I think at this point we've worked around it and were pretty much out of options anyway :-/

PR #3269, opened 2017-03-14
Land simpler network connection logging for #3255

cmasone-attic:
Trying to run `ss` in a background process on every request wound up
opening too many FDs. This is because, in order to read stderr and
stdout from the child process, the server process needed to open a
couple of pipes. For every single one. Unless they exited VERY
fast, that means that there'd be many of these child processes alive
at the same time, each requiring 2 FDs -- and this is in addition to
the incoming network connections the server is handling and all its
connections to S3.

Whoops.

This patch only spins up `ss` in the event that a request has already
failed during read. The downside is that we're getting a snapshot from
_after_ the failure. The upside is that it's likely to actually
function! The hope is that the situation that lead to the read failing
persists long enough that we can see it even when sampling just after
the failure. Maybe not, but hopefully it's a start on figuring out

### DIFF
--- a/go/nbs/s3_table_reader.go
+++ b/go/nbs/s3_table_reader.go
@@ -7,6 +7,8 @@ package nbs
 import (
 	"fmt"
 	"io"
+	"os"
+	"os/exec"
 
 	"github.com/attic-labs/noms/go/d"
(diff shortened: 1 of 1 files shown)


cmasone-attic:
PTAL

PR #3272, opened 2017-03-15
NGQL: Handle null better in input data

arv:
Struct fields are not allowed to be null or missing.

Lists/Sets may not contain null.

### DIFF
--- a/go/ngql/query_test.go
+++ b/go/ngql/query_test.go
@@ -1185,6 +1185,97 @@ func (suite *QueryGraphQLSuite) TestVariables() {
 			},
 		},
 	})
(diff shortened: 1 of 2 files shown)


go/ngql/query_test.go
```diff
@@ -1185,6 +1185,97 @@ func (suite *QueryGraphQLSuite) TestVariables() {
 			},
 		},
 	})
+	test(m2, `{"data":null,"errors":[{"message":"Variable \"$ks\" got invalid value [{}].\nIn element #1: In field \"n\": Expected \"String!\", found null.","locations":[{"line":1,"column":12}]}]}`,
+		q,
+		map[string]interface{}{
+			"ks": []interface{}{
```
kalman: I think you can leave out this `[]interface{}`, Go can figure out the type?

arv: There are cases where the type name is not needed. This case is needed though.

PR #3274, opened 2017-03-16
Add retries to S3 reading

cmasone-attic:
We hit the point of diminishing returns with #3255.
Just do retries with an exponential backoff, logging
failures so we can at least go back in logs and see
how often we see this failure in future, and if we
get storms of retries or not.


cmasone-attic:
PTAL

cmasone-attic:
jenkins: test this

Issue #1466, opened 2016-05-10
Parser for human readable serialization


arv:
We want a parser for the human readable serialization so that it can be used for dumping data.
#1153

arv:
The current format is ambiguous.

- There is an overlap between struct types and struct values.
- There is an overlap between empty maps, sets and structs
- There is an overlap between structs and maps.

Here is a diff that tests the ambiguous cases:

```diff
diff --git a/go/types/encode_human_readable_test.go b/go/types/encode_human_readable_test.go
index 6c222c9..ef0dd0e 100644
--- a/go/types/encode_human_readable_test.go
+++ b/go/types/encode_human_readable_test.go
@@ -423,3 +423,24 @@ func TestEncodedValueMaxLines(t *testing.T) {
        WriteEncodedValueMaxLines(&buf, l1, 5)
        assert.Equal(expected, buf.String())
 }
+
+func TestConflicts(t *testing.T) {
+       s := NewStruct("struct", StructData{})
+       assertWriteHRSEqual(t, "struct {}", s)
+       typ := MakeStructTypeFromFields("", FieldMap{})
+       assertWriteHRSEqual(t, "struct {}", typ)
+
+       s2 := NewStruct("", StructData{})
+       assertWriteHRSEqual(t, " {}", s2)
+       set := NewSet()
+       assertWriteHRSEqual(t, "{}", set)
+       m := NewMap()
+       assertWriteHRSEqual(t, "{}", m)
+
+       s3 := NewStruct("", StructData{
+               "Number": NumberType,
+       })
+       assertWriteHRSEqual(t, " {\n  Number: Number,\n}", s3)
+       m2 := NewMap(NumberType, NumberType)
+       assertWriteHRSEqual(t, "{\n  Number: Number,\n}", m2)
+}
```

arv:
A possible solution could be:

- Disallow a struct to be named `struct`, `map` or `set`
- Prefix the value serialization of map and set with `map` and `set`.

Then the test above would be:

```diff
diff --git a/go/types/encode_human_readable_test.go b/go/types/encode_human_readable_test.go
index 6c222c9..aa978ab 100644
--- a/go/types/encode_human_readable_test.go
+++ b/go/types/encode_human_readable_test.go
@@ -423,3 +423,24 @@ func TestEncodedValueMaxLines(t *testing.T) {
        WriteEncodedValueMaxLines(&buf, l1, 5)
        assert.Equal(expected, buf.String())
 }
+
[...]

PR #3277, opened 2017-03-17
Add parser for Noms Types

arv:
At this point it only parses the type syntax.

Towards #1466

### DIFF
--- a/go/nomdl/lexer.go
+++ b/go/nomdl/lexer.go
@@ -9,43 +9,49 @@ import (
 	"text/scanner"
 )
 
-type token struct {
-	Pos  scanner.Position
-	Type rune
-	Text string
-}
-
-func (t token) String() string {
-	return t.Text
-}
-
 type lexer struct {
 	scanner   *scanner.Scanner
-	nextToken token
+	peekToken rune
 }
 
-func new(s *scanner.Scanner) *lexer {
-	rv := &lexer{scanner: s}
-	rv.next()
-	return rv
+func newLexer(s *scanner.Scanner) *lexer {
+	return &lexer{scanner: s}
 }
 
-func (lex *lexer) next() token {
-	nextToken := lex.nextToken
-	r := lex.scanner.Scan()
-	lex.nextToken = token{
-		Pos:  lex.scanner.Pos(),
-		Text: lex.scanner.TokenText(),
-		Type: r,
+func (lex *lexer) next() rune {
+	if lex.peekToken != 0 {
+		tok := lex.peekToken
+		lex.peekToken = 0
+		return tok
 	}
-	return nextToken
+
+	return lex.scanner.Scan()
 }
 
-func (lex *lexer) peek() token {
-	return lex.nextToken
+func (lex *lexer) peek() rune {
+	if lex.peekToken != 0 {
+		return lex.peekToken
+	}
+	tok := lex.scanner.Scan()
+	lex.peekToken = tok
+	return tok
 }
 
-func (lex *lexer) eat(expected rune) token {
+func (lex *lexer) pos() scanner.Position {
+	if lex.peekToken != 0 {
+		panic("Cannot use pos after peek")
+	}
+	return lex.scanner.Pos()
+}
+
+func (lex *lexer) tokenText() string {
+	if lex.peekToken != 0 {
+		panic("Cannot use tokenText after peek")
+	}
+	return lex.scanner.TokenText()
+}
+
+func (lex *lexer) eat(expected rune) rune {
 	tok := lex.next()
 	lex.check(expected, tok)
 	return tok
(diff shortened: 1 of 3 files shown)


go/nomdl/lexer.go
```diff
@@ -0,0 +1,106 @@
+// Copyright 2017 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
```
aboodman: Is there supposed to be a blank line here?

cmasone-attic:
@arv you included a nomdl binary in the patch :-)

arv:
@cmasone-attic I'll remove it in the next commit

go/nomdl/lexer.go
```diff
@@ -0,0 +1,106 @@
...
+func (t token) String() string {
+	return t.Text
+}
+
+type lexer struct {
+	scanner   *scanner.Scanner
+	nextToken token
+}
+
+func new(s *scanner.Scanner) *lexer {
```
cmasone-attic: This package has `new()` and `New()`, which seems suboptimal. Also...isn't `new` a reserved word in Go?

go/nomdl/lexer.go
```diff
@@ -0,0 +1,106 @@
...
+	Text string
+}
+
+func (t token) String() string {
+	return t.Text
+}
+
+type lexer struct {
+	scanner   *scanner.Scanner
+	nextToken token
```
cmasone-attic: On my first readthrough, it seems like `nextToken` is often redundant with state that's stored inside `scanner`. Can you provide some comments explaining what you needed it for? Up here is fine, or wherever it is in this file that you take advantage of copying the state out of scanner and storing it here.

go/nomdl/parser.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+	}
+	return types.MakeStructTypeFromFields(name, fields)
+}
+
+func (p *Parser) parseSingleElemType(allowEmptyUnion bool) *types.Type {
+	p.lex.eat('<')
+	if allowEmptyUnion && p.lex.eatIf('>') {
+		return types.MakeUnionType()
+	}
```
cmasone-attic: Interesting...this `allowEmptyUnion` means that `List<>` and `Set<>` are valid? What do those mean in practical Noms?

go/nomdl/parser.go
```diff
@@ -0,0 +1,224 @@
...
+		return nil
+	}
+	p.lex.eat('>')
+	return types.MakeCycleType(uint32(s))
+}
+
+func (p *Parser) parseMapType() *types.Type {
+	var keyType, valueType *types.Type
+	p.lex.eat('<')
+	if p.lex.eatIf('>') {
```
cmasone-attic: And same for this...what does `Map<>` mean?

go/nomdl/parser_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/attic-labs/noms/go/types"
+	"github.com/attic-labs/testify/suite"
+)
+
+type ParserSuite struct {
```
cmasone-attic: You might consider https://blog.golang.org/subtests, which I've started using instead of testify suites. The benefit is mainly that you can actually run individual subtests via `go test`

go/nomdl/parser.go
```diff
@@ -0,0 +1,224 @@
...
+//   `Map` `<` (Type `,` Type)? `>`
+//
+// RefType :
+//   `Set` `<` Type `>`
+//
+// SetType :
+//   `Set` `<` Type? `>`
+//
+// StructType :
+//   `struct` StructName? `{` StructField? | (StructField `,`)* `}`
```
cmasone-attic: nit of all nits: I think that this expression implies that, if you have multiple fields in your struct definition, a comma is required after each one. Your tests indicate that the terminal comma is optional, however. Which is nicer, so I'd advocate for making the comment line up with the nicer code behavior :-)

go/nomdl/parser_test.go
```diff
@@ -0,0 +1,157 @@
...
+        }`, types.MakeStructTypeFromFields("S", types.FieldMap{
+		"x": types.NumberType,
+		"y": types.StringType,
+	}))
+
+	suite.assertParseError(`struct S {
+                x: Number
+                y: String
+        }`, `Unexpected token Ident, expected ",", example:3:18`)
+}
```
cmasone-attic: I think that, from the code and the spec above, nested struct types are supported, right?

```
struct S {
  t: struct { n: Number }
}
```

It'd be good to have a test case for complicated stuff like that, and the same with structs definitions nested inside collection definitions and stuff

go/nomdl/parser.go
arv: Good catch.

go/nomdl/parser.go
arv: They are used for empty collections.

```go
types.NewList().Type() == MustParseType("List<>")
```

go/nomdl/lexer.go
arv: Apparently I could override new. I renamed it to newLexer.

go/nomdl/parser_test.go
arv: This didn't need a suite at all. At some point I used initParser but that is no longer used

arv:
@cmasone-attic PTAL

go/nomdl/parser.go
```diff
@@ -0,0 +1,231 @@
...
+//
+// SetType :
+//   `Set` `<` Type? `>`
+//
+// StructType :
+//   `struct` StructName? `{` StructFields? `}`
+//
+// StructFields :
+//   StructField
+//   StructField `,` (StructField `,`)*
```
cmasone-attic: Not to be a huge jerk, but doesn't this disallow `struct S { x: Bool, y: Bool }`? Meaning that it requires a terminal comma if there are multiple fields? Would this express the right thing:
```
StructFields:
  StructField
  StructField `,` StructFields?
```

cmasone-attic:
LGTM w/comment nit

go/nomdl/parser.go
arv: You are right. This still didn't end up correct.

Your suggestion should work.

PR #3278, opened 2017-03-17
Remove ChunkStore backpressure mechanism

### LINKED ISSUE #3234: Remove backpressure
We never actually used this and, with NBS, we no longer really expect to return this from the server.

Baby step toward #3178 

Deprecates #2640, #1387, and #1276

cmasone-attic:
This was something that evolved from the way that Dynamo stores
data, and a way to allow clients to make incremental write
progress. We never actually made the clients handle it
properly, though, and so much has changed since we wrote it
that it's only going to be in the way of building something
better.

Fixes #3234

### DIFF
--- a/go/chunks/chunk_store.go
+++ b/go/chunks/chunk_store.go
@@ -5,7 +5,6 @@
 package chunks
 
 import (
-	"fmt"
 	"io"
 
 	"github.com/attic-labs/noms/go/hash"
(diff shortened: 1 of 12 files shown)


cmasone-attic:
PTAL. It turns out that the new testify handles nil pointers to custom error types differently, for some reason. This caused a bunch of tests that checked the return value of `PutMany()` to fail. So, since Raf and I agreed we want to get rid of `BackpressureError` anyway, I just did that :-)

PTAL

fyi @rafael-atticlabs

rafael-atticlabs:
\o/

PR #3279, opened 2017-03-17
Roll to latest testify

cmasone-attic:


cmasone-attic:
TBR

I built and tested locally, but I'm going to let Jenkins cycle before landing this.

PR #3282, opened 2017-03-20
Test for the correct kind of error in s3_table_reader.go

cmasone-attic:
A connection reset doesn't get caught by io.ReadFull and converted to
an io.ErrUnexpectedEOF like I thought.  So, I need to try to type
assert the error I get to *net.OpError and then use that to see if it
was a connection reset.

Workaround for #3255


cmasone-attic:
PTAL. I'm hopeful that this is actually the right kind of error to check for :-)

Issue #3284, opened 2017-03-21
"original" in struct prevents type marshaling


kalman:
Consider:
```go
	type S struct {
		Foo int          `noms:",omitempty"`
		Bar types.Struct `noms:",original"`
	}
```
This is intentionally an error: https://github.com/attic-labs/noms/blob/master/go/marshal/encode_type_test.go#L355 because it doesn't make sense for "original" to be part of the type.

but it's actually pretty annoying as an error, I'd rather it just skip and let me marshal the type.

aboodman:
Seems reasonable.

arv:
Treating `original` as skip (`-`) when doing marshal type makes sense to me.

PR #3285, opened 2017-03-21
Skip "original" field when marshaling type

kalman:
Instead of failing. Fixes https://github.com/attic-labs/noms/issues/3284


go/marshal/encode_type.go
```diff
@@ -20,8 +20,9 @@ import (
 // The rules for MarshalType is the same as for Marshal, except for omitempty
 // is ignored since that cannot be determined statically.
 //
-// If a Go struct contains a noms tag with original an error is returned since
-// the Noms type depends on the original Noms value which is not available.
+// If a Go struct contains a noms tag with original an error the field is
```
arv: grammar

PR #3287, opened 2017-03-22
Optional fields

### LINKED ISSUE #2327: Optional fields in structs
Our lack of optional fields quickly leads to combinatorial explosion of types. If we have data like:

```
List(
  Struct({a:int}),
  Struct({b:int}),
  Struct({a:int, b:int}),
)
```

... we infer the type as:

```
List<
  Struct{a:int} |
  Struct{b:int} |
  Struct{a:int, b:int}
>
```

I think we should instead infer (conceptually):

```
List<Struct{a:?int, b:?int}>
```

I'm still not sure exactly how optional should work in Noms though. It could be a special concept (optional), or there could be a new type `Nil` that just means no value is acceptable (so you'd actually see: `List<Struct{a:int|nil, b:int|nil}>`.

Not sure what the tradeoff is.

arv:
This adds optional fields to structs

Note that is not valid for a struct value to have optional fields, only struct types may have optional fields.

Most of the commits can be reviewed in isolation.

Fixes #2327


arv:
I plan to update the API to use rest params but that can be done in a followup commit/pr

Once this is settled there is some refactoring that can be done to unify the code for how we handle fields throughout types/

arv:
I'm assigning to a bunch of people:

@kalman Look at marshal
@cmasone-attic Look at nomdl
@rafael-atticlabs Look at the rest

kalman:
marshal stuff lgtm

arv:
jenkins: test this

go/ngql/query_test.go
```diff
@@ -307,6 +331,8 @@ func (suite *QueryGraphQLSuite) TestListOfUnionOfScalars() {
 	suite.assertQueryResult(list, "{root{values{... on BooleanValue{b: scalarValue} ... on StringValue{s: scalarValue} ... on NumberValue{n: scalarValue}}}}", `{"data":{"root":{"values":[{"n":28},{"s":"bar"},{"b":true}]}}}`)
 }
 
+// TODO: Add tests for List<struct S { x?: Number }> but I need to do type simplification first!
```
aboodman: Comment out of date?

go/types/simplify.go
```diff
@@ -191,68 +167,51 @@ func simplifyMaps(ts typeset, merge bool) *Type {
...
+	// times a field name is present is less that then number of types we
+	// are simplifying then the field must be optional.
+	// If we see an optional field we do not increment the count for it and
+	// it will be treated as optional in the end.
+	type fieldTypeInfo struct {
+		count   int
+		typeset typeset
 	}
-	commonFields := map[string]typeset{}
+	commonFields := map[string]fieldTypeInfo{}
```
aboodman: `commonFields` is no longer a good name for this since it's no longer only the fields that the structs have in common. It's more like `fieldCounts`.

aboodman: Or `fieldInfos`

go/types/simplify.go
```diff
@@ -191,68 +167,51 @@ func simplifyMaps(ts typeset, merge bool) *Type {
...
-			if !ok {
-				ts = typeset{}
+			fti.typeset.Add(t)
+			if !optional {
+				fti.count++
 			}
-			ts.Add(f.t)
-			unionFields[f.name] = ts
-		}
+			commonFields[name] = fti
```
aboodman: You don't need to add this back to the map in the case where it was already there

go/nomdl/parser.go
```diff
@@ -100,7 +100,7 @@ func (p *Parser) ensureAtEnd() {
 //   Ident
 //
 // StructField :
-//   StructFieldName `:` Type
+//   StructFieldName `?`? `:` Type
```
cmasone-attic: "Yo, dawg, I put a question mark by your question mark so that your optional marker can be optional"

go/marshal/encode.go
```diff
@@ -447,18 +448,21 @@ func typeFields(t reflect.Type, parentStructTypes []reflect.Type, options encode
 			index:     i,
 			nomsType:  nt,
 			omitEmpty: tags.omitEmpty,
+			optional:  tags.omitEmpty,
```
kalman: I thought about this a bit and I'm not sure that re-using the omitEmpty tag for optional is the right thing to do. Better to have a clean break and anybody that wants to migrate from the omitEmpty semantics to the optional semantics can do so themselves.

arv: Let's talk about it today. It is clear that omitempty in MarshalType MUST imply optional because the field is not always there.

PR #3290, opened 2017-03-24
NGQL: Add work around for using a Set as a List

arv:
If the GraphQL schema expects a List we also allow a Set.


rafael-atticlabs:
Just curious: how did this come up?

kalman:
Some importer was generating the photo labels as a `set<string>` instead of a `list<string>`, but we were expecting a `list<string>` because type marshalling generated that from the definition of a Photo (Go struct).

PR #3292, opened 2017-03-24
One more level of error unwrapping in s3_table_reader.go

cmasone-attic:
My mistake was assuming that, when a network read error happens, the
Err field of the resulting net.OpError would directly contain a
unix.Errno. What actually happens is that there's an *os.SyscallError
in there, which is the thing that's actually wrapped around the
unix.Errno. Hence, my `IsConnReset()` function was always returning
false.

This patch has been tested in the field, so I'm confident it will
work to mitigate #3255 in practice.


cmasone-attic:
PTAL

PR #3296, opened 2017-03-28
Fix v7 hyperlink in README.md

samherrmann:
The hyperlink points to an import statement on the master branch. By using the v7 tag in the URL, the URL will point the the intended location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Noms is currently under heavy development. Attic Labs plans to use it as an inte
 
 ### Data Format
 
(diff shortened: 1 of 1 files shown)


aboodman:
Thanks for the report, that is a good point. In order to accept your patch, we need you to complete our contributor agreement. See: https://github.com/attic-labs/noms/blob/master/CONTRIBUTING.md#contributing-code.

samherrmann:
@aboodman I have signed the contributor agreement.

aboodman:
Thanks very much!

PR #3298, opened 2017-03-28
Add missing .babelrc and .flowconfig

kalman:


kalman:
TBR

These used to be in js/noms but that was deleted.

samples/js/graphiql/package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "license": "Apache-2.0",
   "scripts": {
-    "preinstall": "cp ../../../js/noms/.babelrc ../.eslintrc.js ../../../js/noms/.flowconfig .",
+    "preinstall": "cp ../.babelrc ../.eslintrc.js ../.flowconfig .",
```
arv: Why not keep these in samples/js/graphiql?

kalman: Just being consistent with where .flowconfig is.

arv: Then we should move it. At this point we only have one js sample.

kalman: Yeah then we wouldn't need the preinstall.

PR #3302, opened 2017-03-29
Correctly create custom http.Transport in http_batch_store.go

cmasone-attic:
For the longest time, there's been a 'go vet' warning telling us that
when we build a custom http.Client in http_batch_store.go, we were
copying a Mutex down in there. That seems like not what you want to do.
Turns out we should be creating our own http.Transport object and
sharing that, instead of copying the default one every time we build
a new httpBatchStore.


cmasone-attic:
PTAL

ghost (approved):
LGTM

PR #3303, opened 2017-03-29
Resurrect MakeUnionType2 as MakeUnionTypeIntersectStructs

arv:
This is needed by attic and it uses internal APIs so it has to be
exported 😢

A better path forward would be to fully implement intersection types
but that is a large change for an edge case feature.

### DIFF
--- a/go/types/simplify.go
+++ b/go/types/simplify.go
@@ -32,17 +32,17 @@ import "github.com/attic-labs/noms/go/d"
 //
 // Anytime any of the above cases generates a union as output, the same process
 // is applied to that union recursively.
-func makeSimplifiedType(in ...*Type) *Type {
+func makeSimplifiedType(intersectStructs bool, in ...*Type) *Type {
 	ts := make(typeset, len(in))
 	for _, t := range in {
 		// De-cycle so that we handle cycles explicitly below. Otherwise, we would implicitly crawl
 		// cycles and recurse forever.
-		t := ToUnresolvedType(t)
-		ts[t] = struct{}{}
+		ut := ToUnresolvedType(t)
+		ts[ut] = struct{}{}
 	}
 
 	// Impl de-cycles internally.
-	return makeSimplifiedTypeImpl(ts)
+	return makeSimplifiedTypeImpl(ts, intersectStructs)
 }
 
 // typeset is a helper that aggregates the unique set of input types for this algorithm, flattening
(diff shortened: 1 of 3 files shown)


go/types/simplify_test.go
```diff
@@ -16,141 +16,143 @@ import (
...
-		{simplifyMaps,
-			[]*Type{MakeMapType(StringType, MakeStructTypeFromFields("A", FieldMap{"foo": StringType})),
-				MakeMapType(StringType, MakeStructTypeFromFields("A", FieldMap{"foo": StringType, "bar": BoolType})),
-			},
-			MakeMapType(StringType, MakeStructType2("A",
-				StructField{"foo", StringType, false},
-				StructField{"bar", BoolType, true},
-			))},
-	}
+	for _, intersectStruct := range []bool{false, true} {
```
arv: These tests are just wrapped in the loop.

go/types/simplify_test.go
```diff
@@ -16,141 +16,143 @@ import (
...
+			{structSimplifier(""),
+				[]*Type{MakeStructTypeFromFields("", FieldMap{"foo": BoolType}),
+					MakeStructTypeFromFields("", FieldMap{"foo": StringType})},
+				MakeStructTypeFromFields("", FieldMap{"foo": MakeUnionType(BoolType, StringType)})},
+			// struct{foo:Bool}|struct{foo:Bool,bar:Number} -> struct{foo:Bool,bar?:Number}
+			{structSimplifier(""),
+				[]*Type{MakeStructTypeFromFields("", FieldMap{"foo": BoolType}),
+					MakeStructTypeFromFields("", FieldMap{"foo": BoolType, "bar": NumberType})},
+				MakeStructType2("",
+					StructField{"bar", NumberType, !intersectStruct},
```
arv: and any place where the expected result type had an optional field it is made required if we are intersecting structs.

go/types/simplify.go
```diff
@@ -208,8 +212,8 @@ func simplifyStructs(expectedName string, ts typeset) *Type {
 	for name, fti := range allFields {
 		fields = append(fields, StructField{
 			Name:     name,
-			Type:     makeSimplifiedTypeImpl(fti.typeset),
-			Optional: fti.count < count,
+			Type:     makeSimplifiedTypeImpl(fti.typeset, intersectStructs),
+			Optional: !(intersectStructs && fti.anyRequired) && fti.count < count,
```
ehalpern: While a little more verbose, this would be slightly easier to read if anyRequired were instead named anyNonOptional. This would make the relationship to 'optional' immediately obvious.

arv: Good idea. Thanks

PR #3305, opened 2017-03-29
Make marshal.MustFoo not go via defer-recover

kalman:


kalman:
The panic-recover thing makes the stack trace harder to read.

go/marshal/decode.go
```diff
@@ -89,8 +88,13 @@ func Unmarshal(v types.Value, out interface{}) (err error) {
 // Unmarshals a Noms value into a Go value using the same rules as Unmarshal().
 // Panics on failure.
 func MustUnmarshal(v types.Value, out interface{}) {
-	err := Unmarshal(v, out)
-	d.PanicIfError(err)
+	rv := reflect.ValueOf(out)
```
arv: Can you do the refactoring for this case too to prevent code duplication?

kalman: I tried that at first and it wasn't a lot more readable. Let me know what you think of it now. Much more de-duplication would make it worse IMO.

arv:
2640880 LGTM

PR #3307, opened 2017-03-30
Make sure we flatten unions before calling simplify types

arv:
We ran into an issue where we passed non flatten types to simplify types
which lead to us having two `struct Commit` in a union.


arv:
TBR

Landing when bot is happy

arv:
hmm... does not seem ready yet

arv:
PTAL

PR #3310, opened 2017-03-30
Fix union ordering between two cycle types

arv:
We were not enforcing the order between two cycle types. In other words
the following two should be equal:

```
Cycle<0> | Cycle<1> = Cycle<1> | Cycle<0>
```


arv:
PTAL -- this is as good as I can make it without just listing all the cases that should panic.

PR #3315, opened 2017-03-30
Replace and collect

arv:
This replaces all `struct N { ... }` with `struct N {}` in a type
and returns a the types it replaces for future use

### DIFF
--- a/go/types/simplify.go
+++ b/go/types/simplify.go
@@ -236,3 +236,25 @@ func (tc *TypeCache) simplifyStructs(expectedName string, ts typeset, intersectS
 
 	return tc.makeStructType(expectedName, fields)
 }
+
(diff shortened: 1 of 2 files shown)


arv:
Only last commit

go/types/simplify.go
```diff
@@ -236,3 +236,25 @@ func (tc *TypeCache) simplifyStructs(expectedName string, ts typeset, intersectS
 
 	return tc.makeStructType(expectedName, fields)
 }
+
+func replaceAndCollectStructTypes(tc *TypeCache, t *Type) (*Type, map[string]map[*Type]struct{}) {
+	collected := map[string]map[*Type]struct{}{}
```
arv: I decided to use *Type instead of hash.Hash here because I do not want to invoke writeValue at this level. I was thinking it would be fine to have duplicates at this level.

Issue #3313, opened 2017-03-30
Unskip skipped union order tests


arv:
https://github.com/attic-labs/noms/pull/3312
https://github.com/attic-labs/noms/commit/90b62269ccf5beef1c906e1dfbd291d31d77207e#diff-8f43baefa378ee15c4c1cc3dae5ee293R72

arv:
Done

PR #3318, opened 2017-03-31
Implement Type Simplification v2

### LINKED ISSUE #3184: Implement type simplification v2
Expand on the [initial design](https://github.com/attic-labs/noms/issues/2995) and "union" structs of the same name if one is reachable from another. replace "descendant" with `Cycle`. This explains how cycles are auto detected.

IOW: 

```
Struct Foo {
  a: String
  b: Struct Foo {
    a: String
  }
}
```

=>

```
Struct Foo {
  a: String
  b?: Cycle<1>
}
```

rafael-atticlabs:
Fixes #3184


go/types/simplify.go
```diff
@@ -199,37 +199,64 @@ func (tc *TypeCache) simplifyStructs(expectedName string, ts typeset, intersectS
...
+		for i, et := range elemTypes {
+			et2, c := removeAndCollectStructFields(tc, et, seen, pendingStructs)
+			newElemTypes[i] = et2
+			changed = changed || c
+		}
+		if !changed {
+			return t, false
+		}
+
+		return newType(CompoundDesc{t.Kind(), newElemTypes}, tc.nextTypeId()), true
```
arv: I'm not sure why we don't want to intern this?

go/types/simplify.go
```diff
@@ -199,37 +199,64 @@ func (tc *TypeCache) simplifyStructs(expectedName string, ts typeset, intersectS
...
+			pendingStructs[name] = pending
+		}
+		seen[t] = newStruct
+
+		newFields := make(structFields, len(desc.fields))
+		changed := false
+		for i, f := range desc.fields {
+			newType, c := removeAndCollectStructFields(tc, f.Type, seen, pendingStructs)
+			newFields[i] = StructField{Name: f.Name, Type: newType, Optional: f.Optional}
+			changed = changed || c
```
arv: It doesn't look like we are using this changed. We should probably use it below to determine if we want to use the old desc.fields instead of newFields

aboodman:
I'm OK with this approach. Defer to @arv for review.

arv:
I see you removed both the skips from https://github.com/attic-labs/noms/issues/3313

@rafael-atticlabs I added a commit to this PR... (by pushing to rafael-atticlabs/typeSimplification2)

go/types/simplify.go
arv: I fixed it but I'm not sure it is worth it here... Maybe it would be better to remove the change out param from removeAndCollectStructFields.

arv:
I'm landing this. There is some followup work that needs to be done but I'll do that in different PRs

Issue #3325, opened 2017-03-31
`Make*Type` must return simplified types


rafael-atticlabs:

arv:
We had to do a temporary workaround here:

https://github.com/attic-labs/noms/commit/1c00c80f8d98559f356d5f87fa11aabfaf579ab9#diff-8b1dd2dd8dd8119e01566fc4f37de621R464

PR #3331, opened 2017-03-31
Simplify List, Map, Set and Ref types too

arv:
Towards #3325

### DIFF
--- a/go/types/assert_type_test.go
+++ b/go/types/assert_type_test.go
@@ -334,17 +334,17 @@ func TestAssertTypeCycleUnion(tt *testing.T) {
 	//   },
 	// }
 
-	tb := MakeStructType("",
+	tb := MakeStructType("A",
 		StructField{
 			"b",
-			MakeStructType("", StructField{"c", MakeCycleType(1), false}),
+			MakeStructType("B", StructField{"c", MakeCycleType(1), false}),
 			false,
 		},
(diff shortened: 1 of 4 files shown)


go/types/assert_type_test.go
```diff
@@ -334,17 +334,17 @@ func TestAssertTypeCycleUnion(tt *testing.T) {
 	//   },
 	// }
 
-	tb := MakeStructType("",
+	tb := MakeStructType("A",
```
arv: I updated these as I was simplifying structs...

go/types/value_store.go
```diff
@@ -458,8 +458,8 @@ func (lvs *ValueStore) checkChunksInCache(v Value, readValues bool) Hints {
 			lvs.set(targetHash, entry, false)
 		}
 		// At this point, entry should have the most specific type info possible. Unless it matches targetType, or targetType is 'Value', bail.
-		if !(targetType.TargetKind() == ValueKind || entry.Type().Equals(targetType)) {
-			d.Panic("Value to write contains ref %s, which points to a value of a different type: %s != %s", reachable.TargetHash(), targetType.Describe(), entry.Type().Describe())
+		if !IsSubtype(targetType, entry.Type()) {
```
arv: I'm not sure this is the way we want to do this?

Issue #3324, opened 2017-03-31
`TypeOf(Value)` should derive type from the value


rafael-atticlabs:
`TypeOf(Value)` will always return a simplified type

PR #3333, opened 2017-04-01
Update struct in memory representation

arv:
To contain struct name and field names.

This is in preparation for removing types from the value.

Towards #3324

### DIFF
--- a/go/types/struct.go
+++ b/go/types/struct.go
@@ -16,46 +16,58 @@ import (
 )
 
 var EmptyStructType = MakeStructType("")
-var EmptyStruct = Struct{ValueSlice{}, EmptyStructType, &hash.Hash{}}
+var EmptyStruct = Struct{"", structValueFields{}, EmptyStructType, &hash.Hash{}}
 
 type StructData map[string]Value
 
+type structValueField struct {
+	name  string
+	value Value
+}
+
+type structValueFields []structValueField
+
+func (fs structValueFields) Len() int           { return len(fs) }
+func (fs structValueFields) Swap(i, j int)      { fs[i], fs[j] = fs[j], fs[i] }
+func (fs structValueFields) Less(i, j int) bool { return fs[i].name < fs[j].name }
+
 type Struct struct {
-	values []Value
+	name   string
+	fields structValueFields
 	t      *Type
 	h      *hash.Hash
 }
 
 func NewStruct(name string, data StructData) Struct {
-	fieldNames := make(sort.StringSlice, len(data))
+	valueFields := make(structValueFields, len(data))
+	typeFields := make(structFields, len(data))
 	i := 0
-	for fn := range data {
-		fieldNames[i] = fn
+	for name, value := range data {
+		typeFields[i] = StructField{name, value.Type(), false}
+		valueFields[i] = structValueField{name, value}
 		i++
 	}
 
-	sort.Sort(fieldNames)
-	fields := make(structFields, len(data))
-	values := make(ValueSlice, len(data))
-	for i, name := range fieldNames {
-		fields[i] = StructField{name, data[name].Type(), false}
-		values[i] = data[name]
-	}
+	sort.Sort(typeFields)
+	sort.Sort(valueFields)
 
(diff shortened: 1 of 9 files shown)


vendor/github.com/fatih/structs/structs.go
```diff
@@ -65,7 +65,7 @@ func New(s interface{}) *Struct {
 func (s *Struct) Map() map[string]interface{} {
 	out := make(map[string]interface{})
 
-	fields := s.structFields()
+	fields := s.structTypeFields()
```
rafael-atticlabs: maybe this was an accidental change?

go/types/struct.go
```diff
@@ -16,46 +16,58 @@ import (
...
+type structValueFields []structValueField
+
+func (fs structValueFields) Len() int           { return len(fs) }
+func (fs structValueFields) Swap(i, j int)      { fs[i], fs[j] = fs[j], fs[i] }
+func (fs structValueFields) Less(i, j int) bool { return fs[i].name < fs[j].name }
+
 type Struct struct {
-	values []Value
+	name   string
+	fields structValueFields
```
rafael-atticlabs: I think it'd be nice to keep values as a `ValueSlice` and just add `name String` and `fieldNames []string`. Having values just be a value slice was motivated by wanting a fast path to just hand in a slice of values and the constructor wouldn't need to do any more allocation.

In the new world, I think maybe the "fast-path" constructor should hand in name, fieldNames and valueSlice. the constructor would need to validate that len(fieldNames) == len(valueSlice) and that fieldNames are sorted and valid idents.

go/types/struct.go
```diff
@@ -106,66 +118,93 @@ func (s Struct) desc() StructDesc {
...
-	_, i := s.desc().findField(n)
+	f, i := s.findField(n)
 	if i == -1 {
 		return nil, false
 	}
-	return s.values[i], true
+	return f.value, true
+}
+
+func (s Struct) findField(name string) (*structValueField, int) {
```
rafael-atticlabs: I suggest adding `searchField` which `findField` calls. i.e.

```
func (s Struct) searchField(name string) int {
	return sort.Search(len(s.fieldNames), func(i int) bool { return s.fieldNames[i] >= name })
}

func (s Struct) findField(name string) (idx int, found bool) {
	idx := s.searchField(name)
	if idx == len(s.fields) || s.fieldNames[idx] != name {
		return -1, false
	}
	return idx, true
}
```

that was the mutation functions (set, in particular) below can use `searchField` to find the position that the field goes and use `copy` to efficiently create new fieldName and values slices, without needing to sort. i.e. something like:

```
func (s Struct) Set(n string, v Value) Struct {
	i := s.searchField(n)
	var fieldNames []string
        var values ValueSlice

	if i != len(s.fieldNames) && s.fieldNames[i] == n {
		fieldNames = s.fieldNames
		values = make(ValueSlice, len(s.values))
		copy(values, s.values)
		values[i] = v
	} else {
		fieldNames = make([]string, len(s.fieldNames)+1)
		values = make(ValueSlice, len(s.values)+1)
		copy(values[:i], s.values[:i])
		values[i] = v
		copy(values[i+1:], s.values[i:])
	}

	return Struct{s.name, fieldNames, values, &hash.Hash{}}
}
```
```
func (s Struct) Delete(n string) Struct {
	idx := s.findField(n)
	if idx == -1 {
		return s
	}

	fieldNames := make(string, len(s.values)-1)
	copy(fieldNames[:i], s.fieldNames[:i])
	copy(fieldNames[i-1:], s.fieldNames[:i])

	values := make(string, len(s.values)-1)
	copy(values[:i], s.values[:i])
	copy(values[i:], s.values[i+1:])

	return Struct{s.name, fieldNames, values, &hash.Hash{}}
}
```

go/types/struct.go
arv: Not sure I follow? You would need to allocate the `[]string` for the names. Are you imagining that the encoder would inter that too?

arv:
I made some changes but I didn't change to use two slices. Waiting on response before doing that.

PR #3336, opened 2017-04-03
Fix dynamoManifest to properly handle an empty set of tables

cmasone-attic:
This a) doesn't make semantic sense, and b) doesn't
work with dynamoDB manifests.


cmasone-attic:
PTAL

cmasone-attic:
This is the right fix, now, and makes dynamodbManifest properly handle the case where the root is updated, but there are still no tables in the DB. PTAL

cmasone-attic:
It does. I've now added a test for that case to file_manifest_test.go

PR #3337, opened 2017-04-03
Remove types.Value Type() in favor of types.TypeOf()

### LINKED ISSUE #3324: `TypeOf(Value)` should derive type from the value

arv:
BREAKING CHANGE

This removes the `Type()` method from the `types.Value` interface.
Instead use the `types.TypeOf(v types.Value) bool` function.

Fixes #3324


arv:
This does not do any optimizations. It only replaces `val.Type()` with `types.TypeOf(val)`. Later PRs will update the code to compute the type from the value as well as skipping asking for the type when we do not need it any more.

Issue #3335, opened 2017-04-03
Marshal go slice as set


ehalpern:
Also mentioned in #2376.

PR #3339, opened 2017-04-04
Go slice to Noms Set mashalling

ehalpern:
fixes: #3335

### DIFF
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -38,16 +38,16 @@ import (
 // into corresponding Go array elements. If the Go map was nil a new map is
 // created if any value is set.
 //
-// To unmarshal Noms sets, it depends on the presence of a `noms:",set"` tag:
-//  - Without (default), Unmarshal decodes into corresponding Go list elements.
-//  - With, Unmarshal decodes into Go map keys corresponding to the set values,
-//    with values struct{}{}. Map values must have struct{} type.
+// To unmarshal a Noms set into a Go map, the field must be tagged with `noms:",set"`,
+// and it must have a type of map[<value-type>]struct{}. Unmarshal decodes into
+// Go map keys corresponding to the set values and assigns each key a value of struct{}{}.
 //
 // When unmarshalling onto interface{} the following rules are used:
 //  - types.Bool -> bool
 //  - types.List -> []T, where T is determined recursively using the same rules.
-//  - types.Set -> depends on `noms:",set"` annotation: without, same as
-//    types.List, with, same as types.Map.
+//  - types.Set -> depends on `noms:",set"` annotation and field type:
+//    - without the annotation, same as types.List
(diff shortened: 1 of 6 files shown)


go/marshal/encode.go
```diff
@@ -489,6 +493,33 @@ func listEncoder(t reflect.Type, parentStructTypes []reflect.Type) encoderFunc {
...
+
+	var elemEncoder encoderFunc
+	// lock e until encoder(s) are initialized
+	var init sync.RWMutex
+	init.Lock()
+	defer init.Unlock()
+	e = func(v reflect.Value) types.Value {
+		init.RLock()
+		defer init.RUnlock()
+		values := make([]types.Value, v.Len(), v.Len())
```
arv: `values := make([]types.Value, v.Len())` is semantically the same

go/marshal/decode_test.go
```diff
@@ -764,6 +767,7 @@ func TestDecodeSet(t *testing.T) {
 		"c": types.NewSet(),
 		"d": types.NewMap(),
 		"e": types.NewSet(),
+		"g": types.NewSet(),
```
arv: Keep this in alphabetical order?

go/marshal/decode.go
```diff
@@ -38,16 +38,16 @@ import (
 // into corresponding Go array elements. If the Go map was nil a new map is
 // created if any value is set.
 //
-// To unmarshal Noms sets, it depends on the presence of a `noms:",set"` tag:
-//  - Without (default), Unmarshal decodes into corresponding Go list elements.
-//  - With, Unmarshal decodes into Go map keys corresponding to the set values,
-//    with values struct{}{}. Map values must have struct{} type.
+// To unmarshal a Noms set into a Go map, the field must be tagged with `noms:",set"`,
+// and it must have a type of map[<value-type>]struct{}. Unmarshal decodes into
+// Go map keys corresponding to the set values and assigns each key a value of struct{}{}.
```
arv: Can we add code to unmarshal a noms set into a go slice/array too?

arv: ignore this comment.

go/marshal/encode.go
ehalpern: Good catch. I picked this up from copied code. Fixed both instances.

go/marshal/decode_test.go
ehalpern: Turns out this was just wrong but the value is ignored in the test.

go/marshal/decode.go
ehalpern: As discussed, the code already handled unmarshalling into slices and arrays regardless of annotation. It would be nice to produce an error if a slice/array is annotated as a set but the noms types is a list, but I don't think I'm going to do that in this patch.

Issue #3180, opened 2017-02-11
Change completeness validation to be lazy


rafael-atticlabs:
Remove the requirement that data be written bottom up and instead track the set of unresolved refs. On commit, check that all unresolved refs are contained.

Issue #3123, opened 2017-01-27
Removing the *hinting* system


rafael-atticlabs:
Towards #3122, the noms wire protocol uses a system of hints to reduce the IO load on a server of validating that chunks written to the server contain valid references.

The "hints" are references to other chunks which the client knows to contain the same references (and already be committed to the server). Thus, if the server can find these references within contained chunks which have already been validated, it can assume they are correct.

Generally, this works relatively well, but has some problems:

1) It's complex. The code within ValueStore which manages it has gotten more & more complex as noms has evolved. As is, it's pretty hard to reason about and we have comparatively low confidence that it's behavior is correct. Thankfully, incorrect behavior only results in greater IO on the server, not corruption.

2) The RAM required to maintain this hinting information isn't bounded and long-lived processes which do multiple commits continue to accumulate hinting information with no clear way to know how to expire some of it.

2) We had to relax the server to not *require* all hints necessary to write, and thus it's possible that a lazy client could bypass all hints and force the server to do a large amount of IO. The only obvious recourse here is to create heuristics around how much IO is too much, but those would be fairly arbitrary.

3) We currently have a *write once at the end* write model for noms because writing in batches and committing at some point in the future doesn't have an obvious way to fit into this model. As a result, we always write locally to disk first. In practice, this means that alot of jobs are forced needlessly to write to disk, and it means that we can parallelize job work and shipping data to the server.

rafael-atticlabs:
cc @cmasone-attic @aboodman

rafael-atticlabs:
I have two ideas for this. Both approaches are trying to find a cheaper way to validate that a ref is contained and of the claimed type.

Both ideas share step (1):

-depend NBS's relative cheep has-checking ability to ensure that a referenced chunk is present

ChunkStore already requires `Has(hash hash.Hash)` -- just use it, knowing that NBS can do this fast. It disadvantages some ChunkStore impls, but I think NBS's advantages are large and it seems silly to not leverage them.

Idea (1):

Add `CheckType(chunk hash.Hash, typeHashPrefix Uint(16|32|64))` to `ChunkStore`. NBS would implement this by storing it in the table index in ordinal order (probably in parallel with suffix index).

This solution is kind brute force, but works and I think we could be fine with 4 or ever 2 bytes of typeHashPrefix.

Idea (2):

Right now a chunks' address is the hash of its serialization. I.e. a `Ref` is logically encoded as something like this:

[...]

PR #3340, opened 2017-04-04
Rip out hinting and make validation lazy

### LINKED ISSUE #3178: Remove "hinting" system
Split out as a near-term step for https://github.com/attic-labs/noms/issues/3123.

cmasone-attic:
The hinting mechanism used to assist in server-side validation
of values served us well, but now it's in the way of building a
more suitable validation strategy. Tear it out and go without
validation for a hot minute until #3180 gets done.

Fixes #3178


cmasone-attic:
PTAL. The interesting stuff is in `value_store.go`, `http_batch_store.go` and `remote_database_handlers.go`, most likely

cmasone-attic:
Ok! This stack of patches now implements lazy server-side ref validation as well as ripping out hinting.

The things we've lost as a result of this are client-side ref checking, and (I believe) any checking that the refs encountered during writing point to values of the type they claim to point to. I think that our plan to fix this in the fullness of time is #3123, but that we can tolerate not having it right now, yeah?

PTAL. I intend to continue on to take `refHeight` out of `SchedulePut()` and also clear up #2982, but this PR is growing fairly large so I'd like to take those on separately.

rafael-atticlabs:
LGTM, modulo:

In addition to the comments inline:

-NomsVersion needs to be bumped since this is a wire protocol change

-I'm not sure if this is what you meant by lacking client-side validation, but this change makes it so that there is no validation using `localDatabase` (i.e. writing directly to a file NBS). I'm ok with landing this patch without that (largely because BR is only writing over HTTP right now), but we should open a bug and fix it in the next patch.

Some high-level thoughts about code structure and layering. I'm ok, addressing some of these in following on commits:

-I think ValidatingBatchStore is increasingly a strange abstraction. It's tempting at this point to encapsulate the concern of checking completeness into a smaller utility which is used by someone else. I.e.

```
type CompletenessChecker ***

func (cc) AddRef(h *hash.Hash)
func (cc) AddNovel(h *hash.Hash)
func (cc) IsComplete (fn hasManyFunc) bool
```

Regardless of how its encapsulated, I think it might be more efficient to: 

-collect all of the ref & novel hashes
-store them as a `[]*hash.Hash`
-on `IsComplete`, sort each, subtract `novel` from `refs` and then call the `hasManyFunc` on the remainder

Reasons for this:
-There's not a huge benefit removing a novel hash from refs at the time its added, and some other novel chunk will likely add it back
-storing pointers to hashes will burn less ram, which should generally be safe because someone else is already keeping the hashes alive
-its less work just to sort both slices when you need to answer `IsComplete` than maintaining hashes
-it results in correctly computing the set of unresolved refs thus minimizing the work that `hasMany` needs to do.

go/datas/database_test.go
```diff
@@ -98,7 +98,8 @@ func (suite *DatabaseSuite) TestReadWriteCachePersists() {
 	suite.NoError(err)
 }
 
-func (suite *DatabaseSuite) TestWriteRefToNonexistentValue() {
+// TODO: Re-enable as part of BUG 3180
+func (suite *DatabaseSuite) SkipTestWriteRefToNonexistentValue() {
```
cmasone-attic: Yes, right

cmasone-attic: Actually, it needed to be tweaked a little too, as the failure now won't happen until a BatchStore Flush or a Database Commit()

go/datas/serialize_hints.go
```diff
@@ -11,17 +11,8 @@ import (
 
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
-	"github.com/attic-labs/noms/go/types"
 )
 
-func serializeHints(w io.Writer, hints types.Hints) {
```
cmasone-attic: yeppers

go/datas/serialize_hints_test.go
```diff
@@ -14,17 +14,16 @@ import (
 
 func TestHintRoundTrip(t *testing.T) {
```
cmasone-attic: whole file is gone now

cmasone-attic:
Totally forgot about NomsVersion, thanks. 

As for the local thing, yeah...but is it as simple as just making a `LocalDatabase` be backed by a `localBatchStore` that does this new validation, instead of the `BatchStoreAdapter` we're using right now? I think it is, so I should be able to do that right now.

ValidatingBatchingStore has always been sort of weird, and I agree its just getting weirder. Now that its functionality is less tied to the innards of ValueStore, I think it makes a lot of sense to make it some other kind of utility; though it does still need access to a ValueReader, I believe it only needs its public API. It does two things, now: validating, and also batching up chunks to be put so that we don't have to keep them all in memory until we're done, but also don't have to write them one at a time. Is your argument that this should be split out, which I agree with, or that it's not really doing a lot for us?

There are definitely some good ideas in your CompletenessChecker plan, especially just keeping hash pointers. I debated not bothering to keep `novel` up to date while enqueuing chunks, but figured that it was unlikely to be a huge performance hit. Also, I believe that my solution already correctly computes the set of unresolved refs, minimizing the work that hasMany needs to do. Is that not so?

Doing all this work with `HashSet` seemed more readable to me than using slices and sorting them and doing clever things to diff them. If it turns out the map lookups are a performance hit, I'm happy to change it, but it seemed like it wouldn't be a very big deal. Also, `HasMany()` returns a `HashSet`, so I'm already going to be dealing with at least one of those. Are you also asking for that to be different? I'm fine changing that, too, but since a part of your argument relies on the simplicity of using slices to answer `IsComplete`, I think it's important to be precise.

cmasone-attic:
Ok, PTAL. I realized that I'd done the work to restore validation checking in the LocalDatabase case on the branch where I'd also done #2982, so I needed to pull that patch in as well. It's not that much code, I don't think.

rafael-atticlabs:
I'm still lgtm, but now the problem is that when we are writing locally, we're now *staging* into a seperate nbs and the on commit, reading it all back out and pushing over to the real place. unless, I'm misunderstanding. if so, i'm totally fine landing and then refactoring from here.

cmasone-attic:
Yes, writes are being staged locally and then being read out and rewritten. Easiest way to fix that is probably to duplicate logic, unless we want to rip apart ValidatingBatchingSink right now. Which is an option

rafael-atticlabs:
Up to you. I'm fine landing and opening bug to fix.

cmasone-attic:
#3343 will track work to fix the local-staging stuff

PR #3342, opened 2017-04-06
Add datetime.IsZero() and DateTimeNow() functions

willhite:
These changes were necessary to add DateTime structs to Commit structs in attic.

### DIFF
--- a/go/util/datetime/date_time.go
+++ b/go/util/datetime/date_time.go
@@ -56,3 +56,32 @@ func (dt *DateTime) UnmarshalNoms(v types.Value) error {
 	*dt = DateTime(time.Unix(int64(s), int64(frac*1e9)))
 	return nil
 }
+
+// DateTimeNow returns a DateTime struct representing the current time. This
+// struct is created using seconds only. Since nanoseconds aren't used, it's
+// value is unchanged when Unmarshalling/Marshalling:
+//   dt1 := DateTimeNow()
+//   dt2 := DateTime{}
+//   marshal.MustUnmarshal(marshal.MustMarshal(dt1), &dt2)
+//   dt1 == dt2 // Will be true
+func DateTimeNow() DateTime {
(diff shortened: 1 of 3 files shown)


go/util/datetime/date_time.go
```diff
@@ -56,3 +56,32 @@ func (dt *DateTime) UnmarshalNoms(v types.Value) error {
...
 }
+
+// DateTimeNow returns a DateTime struct representing the current time. This
+// struct is created using seconds only. Since nanoseconds aren't used, it's
+// value is unchanged when Unmarshalling/Marshalling:
+//   dt1 := DateTimeNow()
+//   dt2 := DateTime{}
+//   marshal.MustUnmarshal(marshal.MustMarshal(dt1), &dt2)
+//   dt1 == dt2 // Will be true
+func DateTimeNow() DateTime {
```
arv: This should just be Now

go/util/datetime/date_time.go
```diff
@@ -56,3 +56,32 @@ func (dt *DateTime) UnmarshalNoms(v types.Value) error {
...
+	return DateTime(time.Unix(time.Now().Unix(), 0))
```
arv: I don't understand why nanoseconds are not used?

aboodman:
I'll be in soon but it seems like there is some problem marshalling zero
dates. Maybe it is possible to fix that instead?

aboodman:
Was reading docs for time. Time on bus. I think problem is that
DateTime.Marshal uses UnixNano which cannot represent the zero time,
resulting in the zero time (and many other times) not roundtripping through
noms.

This is easy fix to Marshal though, to just get the date components
separately.

aboodman:
Also not sure we should be marshalling zero dates in the first place.

arv:
@aboodman PTAL

PR #3351, opened 2017-04-07
NewStreamingBlob: print out err details from reader

aboodman:
We're getting an error here but can't see what it is in the logs.


aboodman:
Separately, I think NewStreamingBlob should return an error, not panic.

aboodman:
Thanks. The errors I'm seeing are actually UnexpectedEOF. I assume this is
because the server is hanging up on us but I wonder why that is happening.

PR #3352, opened 2017-04-07
Introduce @target annotation for paths

### LINKED ISSUE #2172: Do paths automatically traverse refs?
And should they?

e.g. `foo.bar[42].monkey`

What if foo is a struct and bar is `Ref<List>`. What if bar is `List<Ref<struct>>` ?

===

Update, after more experience with paths and stuff, I think it sounds great to go with what we were converging on below:

- Yes, we automatically traverse refs
- You can do `@ref` on any path to return its ref
- Since all Noms value (can) have refs, this should always return a value
- But since not all Noms values are stored in the database independently you won't necessarily be able to take the resulting `targetHash` and use it for anything

aboodman:
Fixes #2172

### DIFF
--- a/cmd/noms/noms_log.go
+++ b/cmd/noms/noms_log.go
@@ -189,7 +189,7 @@ func printCommit(node LogNode, path types.Path, w io.Writer, db datas.Database) 
 		}
 
 		if showValue {
-			_, err = writeCommitLines(node, path, maxLines, lineno, w)
+			_, err = writeCommitLines(node, path, maxLines, lineno, w, db)
 		} else {
 			_, err = writeDiffLines(node, path, db, maxLines, lineno, w)
 		}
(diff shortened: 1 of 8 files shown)


go/types/path.go
```diff
@@ -421,14 +429,30 @@ Switch:
...
 	return "@type"
 }
 
+// TargetAnnotation is a PathPart annotation to resolve to the targetValue of the Ref it is resolved on.
+type TargetAnnotation struct {
+}
+
+func (ann TargetAnnotation) Resolve(v Value, vr ValueReader) Value {
+	if r, ok := v.(Ref); ok {
+		return r.TargetValue(vr)
```
kalman: It would be helpful to throw a useful error if `vr` is nil because it seems like an easy gotcha.

samples/go/poke/main.go
```diff
@@ -58,7 +58,7 @@ func poke() (win bool) {
 		return
 	}
 
-	oldVal := inPath.Resolve(inRoot)
```
kalman: There is definitely a database right above here.

cmd/noms/noms_log.go
```diff
@@ -283,7 +283,7 @@ func writeCommitLines(node LogNode, path types.Path, maxLines, lineno int, w io.
 	}
 	mlw := &writers.MaxLineWriter{Dest: w, MaxLines: uint32(maxLines), NumLines: uint32(lineno)}
 	pw := &writers.PrefixWriter{Dest: mlw, PrefixFunc: genPrefix, NeedsPrefix: true, NumLines: uint32(lineno)}
-	v := path.Resolve(node.commit)
```
aboodman: done

samples/go/poke/main.go
aboodman: Fair.

aboodman: Wait, I'm confused -- I put a db in the call one line below?

cmd/noms/noms_log.go
```diff
@@ -325,8 +325,8 @@ func writeDiffLines(node LogNode, path types.Path, db datas.Database, maxLines,
 	parentCommit := parent.(types.Ref).TargetValue(db).(types.Struct)
 	var old, neu types.Value
 	functions.All(
-		func() { old = path.Resolve(parentCommit) },
-		func() { neu = path.Resolve(node.commit) },
+		func() { old = path.Resolve(parentCommit, nil) },
+		func() { neu = path.Resolve(node.commit, nil) },
```
kalman: shouldn't be too hard to pass it in here either

samples/go/nomdex/nomdex_update.go
```diff
@@ -171,7 +171,7 @@ func addElementsToGraphBuilder(gb *types.GraphBuilder, db datas.Database, rootOb
 		typeCacheMutex.Unlock()
 		if !ok || hasPath {
 			pathResolved := false
-			tv := relPath.Resolve(v)
+			tv := relPath.Resolve(v, nil)
```
kalman: I see a db above here

PR #3353, opened 2017-04-07
Fix NBS optimistic locking

### LINKED ISSUE #3349: NBS optimistic locking in broken
We use the root hash to optimistically lock...but, we allow flush() which alters the set of tables without changing the root. so multiple writers who only flush can stomp on each other.

cmasone-attic:
Introduce a "lock" hash into NBS manifests to address the bad
interaction between Flush() and optimistic locking. Our original
design didn't include Flush(), which changes the set of tables without
updating the root. Thus... an optimistic locking strategy predicated
on checking the currently-persisted root hash is not robust to
interleaved Flush() calls from multiple clients.

Fixes #3349

### DIFF
--- a/go/nbs/dynamo_manifest.go
+++ b/go/nbs/dynamo_manifest.go
@@ -40,14 +40,13 @@ type ddbsvc interface {
 type dynamoManifest struct {
 	table, db string
 	ddbsvc    ddbsvc
-	lock      addr
 }
 
 func newDynamoManifest(table, namespace string, ddb ddbsvc) manifest {
-	return &dynamoManifest{table: table, db: namespace, ddbsvc: ddb}
+	return dynamoManifest{table: table, db: namespace, ddbsvc: ddb}
 }
 
-func (dm *dynamoManifest) LoadIfExists(readHook func()) (exists bool, vers string, root hash.Hash, tableSpecs []tableSpec) {
+func (dm dynamoManifest) ParseIfExists(readHook func()) (exists bool, vers string, lock addr, root hash.Hash, tableSpecs []tableSpec) {
 	result, err := dm.ddbsvc.GetItem(&dynamodb.GetItemInput{
 		ConsistentRead: aws.Bool(true), // This doubles the cost :-(
 		TableName:      aws.String(dm.table),
@@ -66,7 +65,7 @@ func (dm *dynamoManifest) LoadIfExists(readHook func()) (exists bool, vers strin
 		exists = true
 		vers = *result.Item[versAttr].S
 		root = hash.New(result.Item[rootAttr].B)
-		copy(dm.lock[:], result.Item[lockAttr].B)
+		copy(lock[:], result.Item[lockAttr].B)
 		if hasSpecs {
 			tableSpecs = parseSpecs(strings.Split(*result.Item[tableSpecsAttr].S, ":"))
 		}
@@ -89,10 +88,7 @@ func validateManifest(item map[string]*dynamodb.AttributeValue) (valid, hasSpecs
 	return false, false
 }
 
-func (dm *dynamoManifest) Update(specs []tableSpec, root, newRoot hash.Hash, writeHook func()) (actual hash.Hash, tableSpecs []tableSpec, err error) {
-	tableSpecs = specs
-
-	lock := generateLockHash(newRoot, specs)
+func (dm dynamoManifest) Update(lastLock, newLock addr, specs []tableSpec, newRoot hash.Hash, writeHook func()) (lock addr, actual hash.Hash, tableSpecs []tableSpec) {
 	putArgs := dynamodb.PutItemInput{
 		TableName: aws.String(dm.table),
 		Item: map[string]*dynamodb.AttributeValue{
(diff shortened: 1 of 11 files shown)


go/nbs/dynamo_manifest.go
```diff
@@ -39,13 +40,14 @@ type ddbsvc interface {
 type dynamoManifest struct {
 	table, db string
 	ddbsvc    ddbsvc
+	lock      addr
```
cmasone-attic: This is the main thing I hemmed and hawwed about and am still not sure is the right thing. On one hand, the details of this 'lock' hash and how it's calculated and how to check it are totally the concern of the `manifest` implementations. On the other, making it their concern means both `manifest` implementations are now stateful, where before they were stateless, which was nice.

cmasone-attic:
PTAL

go/nbs/mmap_table_reader.go
```diff
@@ -67,6 +67,10 @@ func newMmapTableReader(dir string, h addr, chunkCount uint32, indexCache *index
 		buff, err = unix.Mmap(int(f.Fd()), aligned, int(fi.Size()-aligned), unix.PROT_READ, unix.MAP_SHARED)
 		d.PanicIfError(err)
 		index = parseTableIndex(buff[indexOffset-aligned:])
+
+		if indexCache != nil {
```
rafael-atticlabs: were we just not ever checking the indexCache for the file case?

go/nbs/dynamo_manifest.go
rafael-atticlabs: Yeah, I don't have a problem with this implementation, but abstractly, AWS share:

1-table file format
2-table naming
3-root hash encoding (20 byte hash).

And differ on 
1-Manifest encoding

The one difference makes sense because we need to optimistically lock on an attribute in Dynamo.

I think it makes sense that both implementations should be required to compute the same logical lock hash given the same root hash and included table hashes.

rafael-atticlabs: Seperately, I noticed when debugging last night, it appeared as though I watched repeated openings of the same manifest which, each time resulted in a different ordering of the table files. I wondered what that was about. Not because it's wrong, just surprising.

rafael-atticlabs: Ah. i see the issue you are getting at. The manifests because mutable, where as before they were immutable. Yeah, I think i'd be more comfortable if the lock hash was also owned by the nbs store.

go/nbs/dynamo_manifest.go
```diff
@@ -76,25 +79,28 @@ func validateManifest(item map[string]*dynamodb.AttributeValue) (valid, hasSpecs
...
+		return len(item) == 5, false
 	}
 	return false, false
 }
 
-func (dm dynamoManifest) Update(specs []tableSpec, root, newRoot hash.Hash, writeHook func()) (actual hash.Hash, tableSpecs []tableSpec) {
+func (dm *dynamoManifest) Update(specs []tableSpec, root, newRoot hash.Hash, writeHook func()) (actual hash.Hash, tableSpecs []tableSpec, err error) {
 	tableSpecs = specs
 
+	lock := generateLockHash(newRoot, specs)
```
rafael-atticlabs: nit: `newLock`

go/nbs/mmap_table_reader.go
cmasone-attic: we were not

go/nbs/dynamo_manifest.go
cmasone-attic: Hm. generally, `newFoo` is, in our code, a function that constructs a thing of type `Foo`. Since `Lock` isn't a type, here, that naming convention seems wrong to me.

rafael-atticlabs: ok. no biggy. "newRoot" is right there as well, that's why I suggested it.

go/nbs/dynamo_manifest.go
cmasone-attic: re: the different ordering, i've seen that too. I think it's because, in `tableSet.Rebase()`, which we do whenever we open a manifest, we do the following to remove dupes in the set of tables we want to open:
```golang
	// Create a list of tables to open so we can open them in parallel.
	tablesToOpen := map[addr]tableSpec{}
	for _, spec := range specs {
		if _, present := tablesToOpen[spec.name]; !present { // Filter out dups
			tablesToOpen[spec.name] = spec
		}
	}
```
and then we iterate that map. Do you think it's important to fix the ordering? I'm sure we could come up with a way

rafael-atticlabs: no. it's just a bit smelly. made it hard to debug, but i don't see a concrete problem. mainly because our ordering isn't significant. as long as it isn't causing false lock failures.

cmasone-attic:
ok, PTAL at this before I rebase.

go/nbs/dynamo_manifest.go
cmasone-attic: oh, the variable name! Thought you meant the function name and was quite confused

PR #3358, opened 2017-04-08
Rename 'lock' attribute in dynamo_manifest.go

cmasone-attic:
Turns out 'lock' is a reserved word for DynamoDB attributes,
so change it.


cmasone-attic:
TBR

PR #3360, opened 2017-04-08
correctly encode struct cycles

rafael-atticlabs:
ValueEncoder was encoding any repeated struct name as a cycle. It needs to only do this with non-empty names.

Sorry for not catching this in review, @arv. I blame the rats.


rafael-atticlabs:
Also, minor removal of outdated comments and dead code.

rafael-atticlabs:
TBR @arv

PR #3365, opened 2017-04-08
Zero-tolerance for unnamed struct cycles

rafael-atticlabs:
TBR @arv


rafael-atticlabs:
Also, clean up a bit of doc.

PR #3366, opened 2017-04-08
Remove to unresolved type

### LINKED ISSUE #3364: Remove `ToUnresolvedType`
AFAIC, `ToUnresolvedType` is only used by `ContainCommonSupertype` and I'm not convinced it's usage is correct anymore. I think generally, doing any computation of types should involve needing to handle cycles and unresolved cycles should only ever appear if the type really is unresolved.

rafael-atticlabs:
Fixes #3364


rafael-atticlabs:
TBR, @arv, @ehalpern

ehalpern:
LGTM

PR #3367, opened 2017-04-08
Types refactor

rafael-atticlabs:
Review this PR in patch order. 

1. Decoder no longer needed to remove struct cycles. That happened as we decoded
2. There were a bunch of tests that were asserting something about structs after removing cycles. I suspect these tests did something useful at some point, but I'm pretty they weren't any more. Apologies if I got this wrong. If so, assign bug to me to restore anything important.
3. Remove dead code
4. Refactor and add test.
5. Just moving code around (mostly type_cache -> make_type)
6. Remove (no longer neccesary) typeDepth counter in ValueDecoder 

TBR @arv


go/types/simplify_test.go
```diff
@@ -64,20 +64,6 @@ func TestSimplifyType(t *testing.T) {
 			act := simplifyType(in, intersectStructs)
 			assert.True(exp.Equals(act), "Expected: %s\nActual: %s", exp.Describe(), act.Describe())
 		}
-		testSame := func(t *Type) {
-			test(t, resolveCycleTypes(t))
-		}
-
-		testSame(BlobType)
```
arv: These were useful. It ensures that we don't do anything stupid.

go/types/type.go
```diff
@@ -111,41 +111,33 @@ func TypeOf(v Value) *Type {
 
 // HasStructCycles determines if the type contains any struct cycles.
 func HasStructCycles(t *Type) bool {
-	return hasStructCycles(t, map[string]int{})
+	return hasStructCycles(t, nil)
 }
 
-func hasStructCycles(t *Type, seenStructs map[string]int) bool {
+func hasStructCycles(t *Type, visited []*Type) bool {
```
arv: Not sure this is better? It comes down to higher constant factor O(1) vs lower constant factor O(n).

But lets keep the slice solution. It is simpler to understand.

go/types/simplify_test.go
rafael-atticlabs: I'm not sure I understand. They were tests that we didn't do random bad stuff to types?

arv:
No they were testing that the output of simplify was what we expect.

rafael-atticlabs:
Ah. I get it now. Sorry. I'll put them back (without resolveCycles) tomorrow.

Issue #2307, opened 2016-08-06
Arbitrary precision number format isn't quite right (I think)


rafael-atticlabs:
I'm worried that there's flexibility in how encode certain numbers. I think we want a fairly simple and correct way to deterministically represent normalized numbers.

For example, I think we current encode `256` as `256` (base) and `(0)` exponent. It would be better to encode as `1` (base) and `7` (exponent)

arv:
That is correct. We should change this next time we do a version change.

aboodman:
This can't be a p1 because it is a format change. Or if it is a p1, then we need a way to queue such changes and not land them immediately, or have them between runtime switches, or something

PR #3368, opened 2017-04-09
Encode all noms quantities as varint

### LINKED ISSUE #2119: Consider using varint for all quantities in the noms format
I had originally reasoned that chunk-level compression (snappy) would do a good enough job removing zeros from quantities (e.g. chunk height, number of elements following in a sequence, etc...), that there was no point to using var int.

Yesterday, I discovered a reason. For #617, I tracked down a bug which caused me to think about the fact that mostly those quantities will be 8 byte values and will tend to be mostly zeros. When we hash over the value bytes, it means that (a) these zeros will tend to skew the target chunk size down and also reduce entropy into the rolling hash.

rafael-atticlabs:
Fixes #2119 

Also, don't encode types within refs when passing noms values to the rolling value hasher. Until we can compress bytes into rvh, this will allow meta tuples to be a reasonable size.


rafael-atticlabs:
TBR @arv 

Sorry for the cowboy routine this weekend. I can nearly roll noms into attic now, but it's hitting the infinite chunking bug now, which I believe this patch will address.

arv:
Please remove test-data.bak/

go/types/value_encoder.go
```diff
@@ -26,8 +27,10 @@ func (w *valueEncoder) writeKind(kind NomsKind) {
 
 func (w *valueEncoder) writeRef(r Ref) {
 	w.writeHash(r.TargetHash())
-	w.writeType(r.TargetType(), map[string]*Type{})
-	w.writeUint64(r.Height())
+	if !w.forRollingHash {
```
arv: I don't understand how this is safe. Doesn't this lead to a different hash?

arv:
I'm even more concerned about #2307 now... Our uvarint format is not very good.

go/types/value_encoder.go
rafael-atticlabs: The rolling has encodes values and passes them to buzhash. The only sense in which this affects the hash is where the chunk boundary in drawn.

rafael-atticlabs:
This didn't affect how we encode numbers. This only affected counties within the encoding. Maybe I'm misunderstanding your concern.

arv:
I looked at the bug about number encoding and I realized that the issues were not as severe as I thought and like you said this changed does not use our number encoding.

Now that I had the number encoding in my head I just fixed it... https://github.com/attic-labs/noms/pull/3370

PR #3375, opened 2017-04-11
Add IsValueSubtypeOf and IsCommit

### LINKED ISSUE #3326: Implement `IsValueSubtypeOf(value, Type)`

### LINKED ISSUE #3348: Add datas.IsCommit
We currently have `datas.IsCommitType`. It would be good to have one that operates on the value directly.

It would check:
 - That we have a struct named "Commit"
 - That we have meta, parents and value fields (and no other) and that they are non optional.
 - meta is a struct
 - parents is `Set<Ref<Cycle<Commit>>>`.

Maybe this would fall out from https://github.com/attic-labs/noms/issues/3326 ?

arv:
This adds IsValueSubtypeOf which skips computing the type of the value.

Use IsValueSubtypeOf to implement IsCommit which checks if a value is a
commit.

Replace usages of IsSubtype(t, TypeOf(v)) with IsValueSubtypeOf(v, t).

Fixes #3326
Fixes #3348


go/types/subtype.go
```diff
@@ -126,3 +126,110 @@ func compoundSubtype(requiredType, concreteType *Type, parentStructTypes []*Type
...
+		return true
+	case UnionKind:
+		for _, et := range t.Desc.(CompoundDesc).ElemTypes {
+			if IsValueSubtypeOf(v, et) {
+				return true
+			}
+		}
+		return false
+	default:
+		if v.Kind() != t.TargetKind() {
```
rafael-atticlabs: do we want to put a panic here if t.TargetKind() is CycleKind? Seems like that's a programmer error rather than a false.

rafael-atticlabs: not totally sure.

go/types/subtype.go
```diff
@@ -126,3 +126,110 @@ func compoundSubtype(requiredType, concreteType *Type, parentStructTypes []*Type
...
+			return false
+		}
+	}
+
+	switch desc := t.Desc.(type) {
+	case StructDesc:
+		_, found := indexOfType(t, parentStructTypes)
```
rafael-atticlabs: hmm... this doesn't seem right. seems like you need to go as deep as the value is. see test suggestion below

go/types/subtype_test.go
```diff
@@ -437,3 +437,206 @@ func TestIsSubtypeOptionalFields(tt *testing.T) {
...
+	assertTrue(
+		NewStruct("Struct", StructData{"x": Bool(true)}),
+		MakeStructType("Struct", StructField{"y", BoolType, true}),
+	)
+	assertFalse(
+		NewStruct("Struct", StructData{"x": Bool(true)}),
+		MakeStructType("Struct", StructField{"x", StringType, true}),
+	)
+
+	assertTrue(
```
rafael-atticlabs: a bit worried the cycle case isn't detected correctly. what about the following case:

```
Type = Struct Node {
  value: Number
  children: List<Cycle<Node>>
}

Value = Struct Node {
  value: 1
  children: List(Struct Node {
    value: false
    children: List()
  }
}

File: go/datas/remote_database_handlers.go [hunk 4]
```diff
@@ -469,8 +469,8 @@ func assertMapOfStringToRefOfCommit(proposed, datasets types.Map, vr types.Value
 				d.Panic("Root of a Database must be a Map<String, Ref<Commit>>, but key %s maps to a %s", change.V.(types.String), types.TypeOf(val).Describe())
 			}
 			// TODO: Add IsCommit
-			if targetType := types.TypeOf(ref.TargetValue(vr)); !IsCommitType(targetType) {
-				d.Panic("Root of a Database must be a Map<String, Ref<Commit>>, not the ref at key %s points to a %s", change.V.(types.String), targetType.Describe())
+			if targetValue := ref.TargetValue(vr); !IsCommit(targetValue) {
```
rafael-atticlabs: remove comment above.

arv:
Done and done.

PR #3376, opened 2017-04-11
NBS: clean up chunk extract error reporting

### LINKED ISSUE #3148: NBS: Remove/clean up extract error reporting
We're currently hitting some kind of corrupt data when extracting during compaction, which we need to debug _now_. Once we're OK, clean up the error reporting and make it better.

cmasone-attic:
At one point, we had some hard-to-diagnose failures
decoding chunks. We tracked the problem down and fixed
it, but it's good to keep the error reporting. It's done
more cleanly and efficiently, now.

Fixes #3148


cmasone-attic:
I was cruising through this part of the code, and this seemed like a pretty easy cleanup. PTAL

cmasone-attic:
PTAL

Issue #3371, opened 2017-04-10
Assert failure in `ValueStore.bufferChunk`


rafael-atticlabs:
```
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: thumbnailer/run: stderr: panic: 
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: thumbnailer/run: stderr: #015                          #015#011Error Trace:#011value_store.go:256
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: #015#011            #011#015#011#011#011ref.go:78
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: #015#011Error:      #011Should be false
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: 
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: 
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: goroutine 138 [running]:
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: panic(0x908160, 0xc44b4b5f40)
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: #011/usr/local/go/src/runtime/panic.go:500 +0x1a1
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: github.com/attic-labs/attic/vendor/github.com/attic-labs/noms/go/d.panicker.Errorf(0xc42283a080, 0x79, 0x0, 0x0, 0x0)
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: #011/var/workspace/src/github.com/attic-labs/attic/vendor/github.com/attic-labs/noms/go/d/try.go:25 +0xa6
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: github.com/attic-labs/attic/vendor/github.com/attic-labs/noms/go/d.(*panicker).Errorf(0xc95f28, 0xc42283a080, 0x79, 0x0, 0x0, 0x0)
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: #011<autogenerated>:1 +0x7e
Apr  9 07:23:18 ip-172-31-31-70 attic[37836]: github.com/attic-labs/attic/vendor/github.com/attic-labs/testify/assert.Fail(0xc4f120, 0xc95f28, 0x9ce2b8, 0xf, 0x0, 0x0, 0x0, 0x35487c28e5d73897)
[...]

PR #3383, opened 2017-04-12
Don't read chunk during ValueStore.bufferChunk()

cmasone-attic:
The last patch did this in order to allow bad-behavers
to still have a chance of succeding if they write Values
top down. This ensures that they won't, and therefore
will run afoul of lazy completeness checking.

Follow on for #3371


cmasone-attic:
PTAL

PR #3386, opened 2017-04-13
Define DateTime using struct embedding rather than type renaming.

aboodman:
This results in us inheriting all the methods of DateTime automatically.

### DIFF
--- a/go/util/datetime/date_time.go
+++ b/go/util/datetime/date_time.go
@@ -14,8 +14,7 @@ import (
 	"github.com/attic-labs/noms/go/types"
 )
 
-// DateTime is an alias for time.Time that allows us to marshal date time to
-// Noms.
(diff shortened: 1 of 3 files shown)


aboodman:
jenkins: test this

go/util/datetime/date_time.go
```diff
@@ -16,7 +16,9 @@ import (
 
 // DateTime is an alias for time.Time that allows us to marshal date time to
```
arv: update comment?

aboodman:
Regarding your comment: it's just more convenient to not have to cast to time.Time to get access to these methods.

arv:
jenkins: test this

go/util/datetime/date_time.go
arv: ^^^

PR #3388, opened 2017-04-14
Remove DynamoStore

cmasone-attic:
We're no longer using this, and forthcoming changes to ChunkStore
mean that we'd have to do work to continue supporting it.


cmasone-attic:
PTAL

PR #3389, opened 2017-04-16
Remove warning

rafael-atticlabs:
This is happening so often it's filling up space on servers. Still need to understand how it's happening.

### DIFF
--- a/go/types/value_store.go
+++ b/go/types/value_store.go
@@ -5,8 +5,6 @@
 package types
 
 import (
-	"fmt"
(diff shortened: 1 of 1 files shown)


rafael-atticlabs:
TBR @cmasone-attic

aboodman:
One interesting thing that I saw --- this started happening on my local
machine, then I restarted the server and it went away. So it's not related
to the persistent state of the db ... Not sure if that's helpful.

PR #3391, opened 2017-04-17
make noop update root not do any work

rafael-atticlabs:


go/nbs/store.go
```diff
@@ -397,6 +397,10 @@ func (nbs *NomsBlockStore) updateManifest(current, last hash.Hash) error {
 		nbs.mt = nil
 	}
 
+	if current == last && len(nbs.tables.novel) == 0 {
```
cmasone-attic: oh, I thought you meant that we'd do this _after_ we go get the updated manifest. I think we have to do that, don't we?

PR #3394, opened 2017-04-17
Disable read ahead for iter/iterator prolly tree methods.

rafael-atticlabs:


go/merge/three_way_list.go
```diff
@@ -112,7 +112,7 @@ func merge(s1, s2 types.Splice) types.Splice {
 
 func apply(source, target types.List, offset uint64, s types.Splice) types.List {
 	toAdd := make(types.ValueSlice, s.SpAdded)
-	iter := source.IteratorAt(s.SpFrom)
+	iter := source.IteratorAt(s.SpFrom, false)
```
kalman: seems like these merge functions could go either way (true or false) but no readahead seems fine.

go/ngql/types.go
```diff
@@ -439,9 +439,9 @@ func getListElements(v types.Value, args map[string]interface{}) interface{} {
 	var iter listOrSetIterator
 	switch l := l.(type) {
 	case types.List:
-		iter = l.IteratorAt(uint64(idx))
+		iter = l.IteratorAt(uint64(idx), false)
 	case types.Set:
-		iter = l.IteratorAt(uint64(idx))
+		iter = l.IteratorAt(uint64(idx), false)
```
kalman: kinda icky, but maybe this should be:
```go
const readAheadThreshhold 100 // or whatever
iter = l.IteratorAt(uint64(idx), count >= readaheadThreshhold)
```
same elsewhere in ngql. It seems like a bummer to lose readahead entirely here.

arv:
This API is kind of icky. Can we have IterWithReadahead,  MapWithReadahead etc (making Iter and Map be the right default)

go/ngql/types.go
rafael-atticlabs: I don't think there's a good way to pick a constant. GQL is the easy case here since we know the exact begin/end point. I think the thing to do for GQL is to add API which allows you fetch ranges and then we can load those optimally.

rafael-atticlabs:
Thinking more about this change, I think I've concluded that read ahead should be the special case and that the right thing to do right now is simply not enable it unless in particular situations (like opening a blob reader, or IterAll). Then visit the places this hurts. For GQL, i think the situation is different. In the earlier patch, there were only two places that wanted read ahead enabled and they weren't that critical

PR #3399, opened 2017-04-18
Split decode vs completeness validation out of ValidatingBatchingSink

### LINKED ISSUE #3343: Split concerns out of ValidatingBatchingSink (decode vs completeness)
This API is weirdly named, now, and doesn't do any batching. Probably we should have some kind of `CompletenessChecker` that's divorced from the work of validating that chunks decode into values properly and extracting refs from Chunks.

cmasone-attic:
ValidatingBatchingSink stopped batching a while ago, and has generally
made less and less sense over time. Splitting out decoding from
validation allows for clearer code in the server-side writeValue
handler.

Additionally, since we intend to make new chunks that are Put into a
ChunkStore not persist until Flush() or UpdateRoot() as a part of
now the only CS implementation -- splitting out these concerns allows
localBatchStore to stop caching all new chunks as it goes.

Fixes #3343


cmasone-attic:
It turns out that I kinda need to land this either with or before the big Chunk/BatchStore patch. So I figured why not just finish it out, since I think it'll address the too-many-fds problem we're having, and it'll shrink the other patch a little.

PTAL

go/datas/local_batch_store.go
```diff
@@ -11,53 +11,40 @@ import (
...
 
-// SchedulePut simply calls Put on the underlying ChunkStore.
+// SchedulePut calls Put on the underlying ChunkStore and adds any refs in c
+// to a pool of unresolved refs which are validated against the underlying
+// ChunkStore during Flush() or UpdateRoot().
 func (lbs *localBatchStore) SchedulePut(c chunks.Chunk) {
 	lbs.once.Do(lbs.expectVersion)
-	lbs.unwrittenPuts.Insert(c)
+	lbs.cs.Put(c)
+	lbs.mu.Lock()
```
rafael-atticlabs: maybe put the mutex inside the completeness checker?

rafael-atticlabs: if you moved it outside because the server doesn't need to worry about concurrency, FWIW, I suspect that given the other work that's happening there, the mutex around each chunk isn't significant.

go/types/validating_decoder.go
```diff
@@ -0,0 +1,40 @@
...
+	Chunk *chunks.Chunk
+	Value *Value
+}
+
+// Decode decodes c and checks that the hash of the resulting value
+// matches c.Hash(). It returns a DecodedChunk holding both c and a pointer to
+// the decoded Value.
+func (vbs *ValidatingDecoder) Decode(c *chunks.Chunk) DecodedChunk {
+	h := c.Hash()
+	v := decodeFromBytesWithValidation(c.Data(), vbs.vs)
```
rafael-atticlabs: what happens here if you pass `nil` for the ValueReader?

cmasone-attic: I think it would probably be OK for the current usage of the API, but since this is API is visible outside the package I can't guarantee that all future callers will know what's safe to do with the returned Value and what's not. So, I erred on the side of safety and least-surprise.

go/datas/local_batch_store.go
cmasone-attic: yes, that's exactly why I did it. Discussed on slack and I'm going to leave it this way for now. My rationale is that I have one caller that wants locking and one that doesn't, so there's no clear indicator to me whether having the lock inside the completeness checker is the way to go. Hopefully future use cases will demonstrate what most callers want and then I'll make the change if it's warranted.

PR #3403, opened 2017-04-19
Merge BatchStore into ChunkStore

### LINKED ISSUE #2945: Refactor ChunkStore / BatchStore
It's been apparent for some time that what noms *really* depends on is `BatchStore` -- not ChunkStore. The key difference is durability expectations. ChunkStore should probably evolve to look like BatchStore, and we should just remove BatchStore if possible

cmasone-attic:
BatchStore is dead, long live ChunkStore! Merging these two required
some modification of the old ChunkStore contract to make it more
BatchStore-like in places, most specifically around Root(), Put() and
PutMany().

The first big change is that Root() now returns a cached value for the
root hash of the Store. This is how NBS worked already, so the more
interesting change here is the addition of Rebase(), which loads the
latest persistent root. Any chunks that appeared in backing storage
since the ChunkStore was opened (or last rebased) also become
visible.

UpdateRoot() has been replaced with Commit(), because UpdateRoot() was
ALREADY doing the work of persisting novel chunks as well as moving
the persisted root hash of the ChunkStore in both NBS and
httpBatchStore. This name, and the new contract (essentially Flush() +
UpdateRoot()), is a more accurate representation of what's going on.

As for Put(), the former contract for claimed to block until the chunk
was durable. That's no longer the case. Indeed, NBS was already not
fulfilling this contract. The new contract reflects this, asserting
that novel chunks aren't persisted until a Flush() or Commit() --
which has replaced UpdateRoot(). Novel chunks are immediately visible
to Get and Has calls, however.

In addition to this larger change, there are also some tweaks to
ValueStore and Database. ValueStore.Flush() no longer takes a hash,
and instead just persists any and all Chunks it has buffered since the
last time anyone called Flush(). Database.Close() used to have some
side effects where it persisted Chunks belonging to any Values the
caller had written -- that is no longer so. Values written to a
Database only become persistent upon a Commit-like operation (Commit,
CommitValue, FastForward, SetHead, or Delete).
```
/******** New ChunkStore interface ********/

type ChunkStore interface {
     ChunkSource
     RootTracker
}

// RootTracker allows querying and management of the root of an entire tree of
// references. The "root" is the single mutable variable in a ChunkStore. It
// can store any hash, but it is typically used by higher layers (such as
// Database) to store a hash to a value that represents the current state and
// entire history of a database.
type RootTracker interface {
     // Rebase brings this RootTracker into sync with the persistent storage's
     // current root.
     Rebase()

     // Root returns the currently cached root value.
     Root() hash.Hash

     // Commit atomically attempts to persist all novel Chunks and update the
     // persisted root hash from last to current. If last doesn't match the
     // root in persistent storage, returns false.
     // TODO: is last now redundant? Maybe this should just try to update from
     // the cached root to current?
     // TODO: Does having a separate RootTracker make sense anymore? BUG 3402
     Commit(current, last hash.Hash) bool
}

// ChunkSource is a place chunks live.
type ChunkSource interface {
     // Get the Chunk for the value of the hash in the store. If the hash is
     // absent from the store nil is returned.
     Get(h hash.Hash) Chunk

     // GetMany gets the Chunks with |hashes| from the store. On return,
     // |foundChunks| will have been fully sent all chunks which have been
     // found. Any non-present chunks will silently be ignored.
     GetMany(hashes hash.HashSet, foundChunks chan *Chunk)

     // Returns true iff the value at the address |h| is contained in the
     // source
     Has(h hash.Hash) bool

     // Returns a new HashSet containing any members of |hashes| that are
     // present in the source.
     HasMany(hashes hash.HashSet) (present hash.HashSet)

     // Put caches c in the ChunkSink. Upon return, c must be visible to
     // subsequent Get and Has calls, but must not be persistent until a call
     // to Flush(). Put may be called concurrently with other calls to Put(),
     // PutMany(), Get(), GetMany(), Has() and HasMany().
     Put(c Chunk)

     // PutMany caches chunks in the ChunkSink. Upon return, all members of
     // chunks must be visible to subsequent Get and Has calls, but must not be
     // persistent until a call to Flush(). PutMany may be called concurrently
     // with other calls to Put(), PutMany(), Get(), GetMany(), Has() and
     // HasMany().
     PutMany(chunks []Chunk)

     // Returns the NomsVersion with which this ChunkSource is compatible.
     Version() string

     // On return, any previously Put chunks must be durable. It is not safe to
     // call Flush() concurrently with Put() or PutMany().
     Flush()

     io.Closer
}
```

Fixes #2945


cmasone-attic:
ok, PTAL!

go/types/validating_decoder.go
```diff
@@ -14,9 +14,7 @@ type ValidatingDecoder struct {
 }
 
 func NewValidatingDecoder(cs chunks.ChunkStore) *ValidatingDecoder {
-	return &ValidatingDecoder{
-		newLocalValueStore(cs),
-	}
+	return &ValidatingDecoder{NewValueStore(cs)}
```
rafael-atticlabs: can we just add this concern into ValidatingChunkStore and give it a flag for whether it or not it should trust that chunks are encoded correctly? For the server, it'd mean that it could also just use a ValidatingChunkStore and also that each chunk would only need to be decoded once.

go/types/value_store.go
```diff
@@ -277,55 +295,64 @@ func (lvs *ValueStore) bufferChunk(v Value, c chunks.Chunk, height uint64) {
...
+			lvs.bufferedChunkSize -= uint64(len(chunk.Data()))
 		}
-		v.WalkRefs(func(reachable Ref) {
-			if pending, present := lvs.bufferedChunks[reachable.TargetHash()]; present {
-				lvs.bufferedChunkSize -= put(reachable.TargetHash(), pending)
+
+		for parent := range lvs.withBufferedChildren {
+			if pending, present := lvs.bufferedChunks[parent]; present {
+				v := DecodeValue(pending, lvs)
+				// TODO: Consider gathering up the hashes of the referenced, buffered chunks, doing a HasMany() and only writing chunks that aren't in the ChunkStore
```
rafael-atticlabs: nah, NBS is better at filtering those out on the other side.

go/datas/database_common.go
```diff
@@ -188,12 +193,15 @@ func (dbc *databaseCommon) getRootAndDatasets() (currentRootHash hash.Hash, curr
 	return
 }
 
-func (dbc *databaseCommon) tryUpdateRoot(currentDatasets types.Map, currentRootHash hash.Hash) (err error) {
-	// TODO: This Map will be orphaned if the UpdateRoot below fails
+func (dbc *databaseCommon) tryCommitChunks(currentDatasets types.Map, currentRootHash hash.Hash) (err error) {
```
cmasone-attic: There's already a method on databaseCommon called `doCommit()`, and to me `tryCommit` implies "I'll attempt a doCommit and return an error instead of panicing if bad things happen" or some such. This name is an attempt to indicate that this is trying an operation on the ChunkStore.

go/types/validating_decoder.go
cmasone-attic: ...Isn't that just re-creating `ValidatingBatchingSink`? Minus the batching part, but I don't think that was really the piece that bothered you so.

Where do you see the server decoding chunks multiple times? I'm pretty sure they're decoded just once.

go/types/value_store.go
cmasone-attic: done and done

go/chunks/chunk_store.go
```diff
@@ -34,11 +33,23 @@ type Factory interface {
 // Database) to store a hash to a value that represents the current state and
 // entire history of a database.
 type RootTracker interface {
+	// Rebase brings this RootTracker into sync with the persistent storage's
+	// current root.
+	Rebase()
+
+	// Root returns the currently cached root value.
```
cmasone-attic: done

go/chunks/chunk_store.go
```diff
@@ -57,19 +68,24 @@ type ChunkSource interface {
...
 	HasMany(hashes hash.HashSet) (present hash.HashSet)
 
-	// Returns the NomsVersion with which this ChunkSource is compatible.
-	Version() string
-}
-
-// ChunkSink is a place to put chunks.
-type ChunkSink interface {
-	// Put writes c into the ChunkSink, blocking until the operation is complete.
+	// Put caches c in the ChunkSink. Upon return, c must be visible to
```
cmasone-attic: Done

go/chunks/chunk_store.go
```diff
@@ -57,19 +68,24 @@ type ChunkSource interface {
...
+	// subsequent Get and Has calls, but must not be persistent until a call
+	// to Flush(). Put may be called concurrently with other calls to Put(),
```
cmasone-attic: Word.

Issue #3404, opened 2017-04-19
Rationalize Flush() semantics across ChunkStore, ValueStore, and Database


cmasone-attic:
Even after fixing #2945, there will still be a little oddness around `Flush` at the various levels of our Value and Chunk writing abstractions.

1) At the ChunkStore layer, Flush() is supposed to just be Commit() where last == current. I should ensure that's actually so and get rid of ChunkStore Flush() in favor of just calling Commit() appropriately.
2) Right now, ValueStore.Flush() calls Flush() on its underlying ChunkStore. For databaseCommon, this kind of sucks because it Flushes ValueStore right before calling Commit on the underlying ChunkStore, which leads to two ChunkStore.Commit() calls right in a row, which is lame. The problem is that callers who just have a ValueStore kinda NEED vs.Flush() to also Flush() the underlying ChunkStore. There may not really BE any of those outside of tests in the `types` package, though. So...figure that out, and see if we can get rid of the double flushing by making ValueStore.Flush() just roll its buffered chunks down to the ChunkStore.
3) Database should maybe have a public Flush() method for callers who want to make chunks durable without modifying the head of any dataset.

per our discussion this morning, @rafael-atticlabs

cmasone-attic:
Closed in favor of #3530

PR #3406, opened 2017-04-19
Make MemoryStore come correct

### LINKED ISSUE #3400: Enforce new ChunkStore.Put() semantics on MemoryStore
Once #2945 lands, ChunkStore.Put() will have new semantics: new chunks aren't durable until Flush/Commit. Memory store should really provide these semantics.

FYI @rafael-atticlabs

cmasone-attic:
It's important that MemoryStore (and, by extension TestStore)
correctly implement the new ChunkStore semantics before we go
shifting around the Flush semantics like we want to do in #3404

In order to make this a reality, I introduced a "persistence"
layer for MemoryStore called MemoryStorage, which can vend
MemoryStoreView objects that represent a snapshot of the
persistent storage and implement the ChunkStore contract.

Fixes #3400

### DIFF
--- a/go/chunks/chunk_store_common_test.go
+++ b/go/chunks/chunk_store_common_test.go
@@ -23,14 +23,7 @@ func (suite *ChunkStoreTestSuite) TestChunkStorePut() {
 	store.Put(c)
 	h := c.Hash()
 
-	// See http://www.di-mgt.com.au/sha_testvectors.html
-	suite.Equal("rmnjb8cjc5tblj21ed4qs821649eduie", h.String())
-
-	// Reading it via the API should work...
-	assertInputInStore(input, h, store, suite.Assert())
-
-	store.Commit(h, store.Root()) // Commit writes
-	// ...as well as after commit
+	// Reading it via the API should work.
 	assertInputInStore(input, h, store, suite.Assert())
 }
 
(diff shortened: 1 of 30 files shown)


cmasone-attic:
PTAL

cmasone-attic:
nvm, needs moar work

cmasone-attic:
Ok! PTAL. I have an open question in the comments I left, so if you want to skim and then answer the that (and/or discuss it) before doing a full review, that'd be fine with me. Or you can just dive in. The big change is the TestStorage and MemoryStorage thing we talked about yesterday, and the rest is mostly just fallout from that change.

go/datas/remote_database_handlers.go
```diff
@@ -341,7 +341,7 @@ func handleRootGet(w http.ResponseWriter, req *http.Request, ps URLParams, rt ch
 	if req.Method != "GET" {
 		d.Panic("Expected get method.")
 	}
-
+	rt.Rebase()
```
rafael-atticlabs: yeah, let's not do this. i think right now we should depend on ChunkStores being opened newly for each request.  I think there's some cool work to be done here (making serving ChunkStores efficient), but I think it needs to be designed.

go/chunks/chunk_store_common_test.go
```diff
@@ -13,81 +13,88 @@ import (
 
 type ChunkStoreTestSuite struct {
 	suite.Suite
-	Store      ChunkStore
-	putCountFn func() int
+	Factory Factory
 }
 
 func (suite *ChunkStoreTestSuite) TestChunkStorePut() {
+	store := suite.Factory.CreateStore("ns")
```
cmasone-attic: doon

go/datas/remote_database_handlers.go
cmasone-attic: Ok, I took this out, but minimized the disturbance to the testing code by having `newHTTPChunkStoreForTest()` calling `Rebase()` each time it's about to forward a request to a handler function. It provides a good approximation of proper behavior, and Making It Right is tracked in #3415

PR #3408, opened 2017-04-19
Merge chunks.RootTracker interface into chunks.ChunkStore

### LINKED ISSUE #3402: Consider whether RootTracker makes sense as a separate interface
All ChunkStores are RootTrackers, but not vice-versa. This is a little weird, because `UpdateRoot()` (soon to be renamed `Commit()`) implies making a bunch of Chunks durable -- though a RootTracker has no way to put in chunks.

On one hand, it's nice to narrow the interface sometimes when handing out a ChunkStore if you don't want the caller to be able to get or put chunks. On the other hand, it's weird because you can't fully describe the semantics of RootTracker without referencing stuff you can only do through the full ChunkStore interface.

cmasone-attic:
You can't fully specify RootTracker without referring to the
ChunkStore interface, so they should just merge.

Fixes #3402

### DIFF
--- a/cmd/noms/noms_root.go
+++ b/cmd/noms/noms_root.go
@@ -43,10 +43,10 @@ func runRoot(args []string) int {
 	}
 
 	cfg := config.NewResolver()
-	rt, err := cfg.GetRootTracker(args[0])
+	cs, err := cfg.GetChunkStore(args[0])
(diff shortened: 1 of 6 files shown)


cmasone-attic:
PTAL

cmasone-attic:
Whoops, tests build now. PTAL

PR #3409, opened 2017-04-20
implement "rollUp" minor compaction

### LINKED ISSUE #3378: NBS: Implement logarithmic minor compaction.
Right now NBS tries to avoid expiring OS file descriptors and keep the size of the manifest below 400k but limiting to total number of tables in a store. It does this by occasionally taking the smallest 1/2 of tables and compacting them into a one.

This was simple and effective, but it results in re-writing the same data more than necessary. The next step here would be to compact into sized levels.

rafael-atticlabs:
Fixes #3378 

This patch changes the strategy for which tables to compact. Choose 2 or more of the N smallest tables such that the resulting table will the smallest (or tied for smallest) table in the store.

### DIFF
--- a/go/nbs/table.go
+++ b/go/nbs/table.go
@@ -30,9 +30,6 @@ import (
    | (Chunk Length) Chunk Data | (Uint32) CRC32 |
    +---------------------------+----------------+
 
-     -Address suffix is the 4 least-significant bytes of the Chunk's address. Used (e.g. in place
-      of CRC32) as a checksum and a filter against false positive reads costing more than one IOP.
-
    Index:
    +------------+---------+----------+
    | Prefix Map | Lengths | Suffixes |
--- a/go/nbs/table_set_test.go
+++ b/go/nbs/table_set_test.go
@@ -10,6 +10,8 @@ import (
 	"sort"
 	"testing"
 
+	"encoding/binary"
+
 	"github.com/attic-labs/testify/assert"
 )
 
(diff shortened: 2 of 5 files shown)


go/nbs/table.go
```diff
@@ -30,9 +30,6 @@ import (
    | (Chunk Length) Chunk Data | (Uint32) CRC32 |
    +---------------------------+----------------+
 
-     -Address suffix is the 4 least-significant bytes of the Chunk's address. Used (e.g. in place
```
rafael-atticlabs: Just noticed that this comment is out of date.

go/nbs/table_set_test.go
```diff
@@ -116,58 +118,68 @@ func TestTableSetExtract(t *testing.T) {
...
+		{[]uint32{2, 3, 4, 8, 16, 32, 64}, []uint32{129}},
+	}
 
-	assert.Equal(4, ts.Size())
-	assert.Contains(ts.upstream, bigTable)
-	assertChunksInReader(moreChunks, ts, assert)
-	assertChunksInReader(extraChunks, ts, assert)
+	for _, c := range tc {
+		ts := makeTestTableSet(c.precompact)
+		ts2, _ := ts.Flatten().Compact()
```
cmasone-attic: you should Close `ts` and `ts2`

Also, it'd be really nice to use the `t.Run` subtest stuff to make each of these cases a named test that can fail individually. e.g.

```golang
tc := []struct {
  name string
  precompact []uint32
  postcompact []uint32
}
...

for _, c := range tc {
  t.Run(c.name, func(t *testing.T) {
    assert := assert.New(t)
    your code
  })
}
```

cmasone-attic:
LGTM w/test comments

go/nbs/table_set_test.go
rafael-atticlabs: neat. done.

PR #3412, opened 2017-04-20
Add types.MakeStructTemplate

### LINKED ISSUE #3347: `marshall` should use `types.MakeStructTemplate`
We should look at usages of MakeStruct to see if we need to provide a function that does not need to use a map and sort the fields.

For struct types we have two different versions and we should fix the naming inconsistencies.

rafael-atticlabs:
`types.MakeStructTemplate` creates a "template" whose struct & field names have been validated. It allows cases of rapidly creating struct with the same names/fieldNames to avoid the cost of allocating a map for passing field data and validating names.

Fixes #3347

### DIFF
--- a/go/types/struct.go
+++ b/go/types/struct.go
@@ -74,14 +74,14 @@ type StructTemplate struct {
 func MakeStructTemplate(name string, fieldNames []string) (t StructTemplate) {
 	t = StructTemplate{name, fieldNames}
 
-	verifyStructName(t.name)
-	if len(t.fieldNames) == 0 {
+	verifyStructName(name)
+	if len(fieldNames) == 0 {
 		return
 	}
-	verifyFieldName(t.fieldNames[0])
-	for i := 1; i < len(t.fieldNames); i++ {
-		verifyFieldName(t.fieldNames[i])
-		d.PanicIfFalse(t.fieldNames[i] > t.fieldNames[i-1])
+	verifyFieldName(fieldNames[0])
+	for i := 1; i < len(fieldNames); i++ {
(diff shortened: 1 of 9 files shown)


go/types/struct.go
```diff
@@ -66,6 +66,24 @@ func NewStruct(name string, data StructData) Struct {
 	return validateStruct(newStruct(name, fieldNames, values))
 }
 
+type StructTemplate struct {
+	name       string
+	fieldNames []string
+}
+
+func MakeStructTemplate(name string, fieldNames []string) StructTemplate {
+	st := StructTemplate{name, fieldNames}
```
arv: Also verify the name here.

go/types/struct.go
```diff
@@ -66,6 +66,24 @@ func NewStruct(name string, data StructData) Struct {
...
+	for i := 0; i < len(st.fieldNames); i++ {
+		verifyFieldName(st.fieldNames[i])
```
arv: Also verify the order. There is code for all of this in there. Maybe you need to refactor it a bit?

go/types/struct.go
```diff
@@ -66,6 +66,31 @@ func NewStruct(name string, data StructData) Struct {
...
 
+type StructTemplate struct {
+	name       string
+	fieldNames []string
+}
+
+func MakeStructTemplate(name string, fieldNames []string) (t StructTemplate) {
+	t = StructTemplate{name, fieldNames}
+
+	verifyStructName(t.name)
```
arv: Just name here... and just fieldNames below

rafael-atticlabs:
PTAL

PR #3413, opened 2017-04-21
Make Marshal use StructTemplate

arv:


go/marshal/encode.go
```diff
@@ -266,16 +266,14 @@ func structEncoder(t reflect.Type, seenStructs map[string]reflect.Type) encoderF
 	}
 
 	seenStructs[t.Name()] = t
-	fields, structType, originalFieldIndex := typeFields(t, seenStructs, encodeTypeOptions{})
+	fields, structType, structTemplate, originalFieldIndex := typeFields(t, seenStructs, encodeTypeOptions{})
```
arv: It is tempting to refactor typeFields so that we do not even ask for types.

arv:
PTAL

I refactored to not compute the type unless you are doing MarshalType

PR #3416, opened 2017-04-21
ValueStore.Flush() no longer persists Chunks

cmasone-attic:
ValueStore.Flush() now Puts all Chunks buffered in the ValueStore
layer into the underlying ChunkStore. The Chunks are not persistent
at this point, not until and unless the caller calls Commit() on
the ChunkStore.

This patch also removes ChunkStore.Flush(). The same effect can be
achieved by calling ChunkStore.Commit() with the current Root for both
last and current.

NB: newTestValueStore is now private to the types package.
The logic is that, now, outside the types package, callers
need to hold onto the underlying ChunkStore if they want to
persist Chunks.

Toward #3404

### DIFF
--- a/go/datas/database_common.go
+++ b/go/datas/database_common.go
@@ -211,7 +211,6 @@ func (dbc *databaseCommon) tryCommitChunks(currentDatasets types.Map, currentRoo
 	// TODO: This Map will be orphaned if the Commit below fails
 	newRootHash := dbc.WriteValue(currentDatasets).TargetHash()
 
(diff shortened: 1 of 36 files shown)


cmasone-attic:
Ok, PTAL

go/datas/database_common.go
```diff
@@ -211,7 +211,6 @@ func (dbc *databaseCommon) tryCommitChunks(currentDatasets types.Map, currentRoo
 	// TODO: This Map will be orphaned if the Commit below fails
 	newRootHash := dbc.WriteValue(currentDatasets).TargetHash()
 
-	// TODO: We've always been sorta sad that we Flush() the embedded ValueStore here, and then Commit() the composed ChunkStore below. That leads to two consecutive make-Chunks-durable operations on the underlying ChunkStore, and the latter won't actually have any novel Chunks. The problem arises from the fact that most users of ValueStore don't have access to the underlying ChunkStore, but Database _does_. ValueStore buffers Chunks, and most users who call Flush() need it to dump those Chunks into the ChunkStore and then make those Chunks durable. At this callsite, though, we only want to get those down into the ChunkStore, because we know that we're going to call Commit() later, which will make those Chunks durable. It's a conundrum :-/
```
rafael-atticlabs: \o/

rafael-atticlabs:
Landing for you because I want to roll noms.

PR #3417, opened 2017-04-22
make MemoryStoreFactory public

rafael-atticlabs:
Woops. Attic was using this. Making public again (and added mutex).

TBR @cmasone-attic


cmasone-attic:
LGTM

PR #3418, opened 2017-04-24
Rev Noms version

cmasone-attic:
Should've done this last week when I tweaked the HTTP
protocol to return the current root hash when a POST
to the root/ endpoint fails.


cmasone-attic:
FYI @rafael-atticlabs TBR

Issue #3410, opened 2017-04-20
NBS: Multiple writers may attempt the same compaction concurrently


rafael-atticlabs:
Thankfully, this is not incorrect behavior, but it can be very wasteful, especially for a large compaction -- and worse the larger the compaction, the more likely that multiple writers will attempt it because the need for the big compaction will be visible for alot longer.

We need some way to help writers avoid doing alot of pointless read/write work that is already underway.

rafael-atticlabs:
Summary of meatspace discussion w/ @cmasone-attic:

Near-term tactic: When writers go to commit, compute a threshold value for compaction, something: `numTables/maxTables`. If `Rand() > thresh`, compact. This will mitigate the probability of multiple writers computing the same compaction, but will result in far higher write amplification.

Long-term strategy: *prevent* (or something close to it) writers from attempting the same compaction, but allow concurrent writers to continue to commit. This has two steps:

-Effectively *remove* maxTables (basically set is to something very high). 
-Create a mechanism for writers to "lock" a compaction (prevent other writers from attempting it). for File:, this will simply be flock, for AWS, this will likely be setting an attribute on the manifest table which must be kept alive by the active writer.

rafael-atticlabs:
Closing in favor of more specific bugs.

PR #3419, opened 2017-04-24
Trigger compaction probabilistically to avoid lots of concurrent attempts

rafael-atticlabs:
Towards #3410


go/nbs/store.go
```diff
@@ -421,7 +425,14 @@ func (nbs *NomsBlockStore) updateManifest(current, last hash.Hash) error {
 
 	candidate := nbs.tables
 	var compactees chunkSources
-	if candidate.Size() > nbs.maxTables {
+
+	shouldCompact := func() bool {
+		// As the number of tables grows from 1 to maxTables, the probability of compacting, grows from 0 to 1
```
cmasone-attic: My only question is whether we should make it so that we never compact if there's just one table, because that's nothing but wasted time

cmasone-attic:
LGTM w/one question

PR #3420, opened 2017-04-24
Add Database.Rebase()

cmasone-attic:
All this really does is tell the underlying ChunkStore
to go fetch the current root from persistent storage
and drops the now-out-of-date Dataset map on the floor.

Fixes https://github.com/attic-labs/attic/issues/1157


cmasone-attic:
PTAL

Issue #3411, opened 2017-04-20
NBS: Implement streaming compaction


rafael-atticlabs:
The current implementation for minor (non-reorganizing) compaction is to read into RAM the input tables, create a compacted table in memory and then persist. It also has the behavior of de-duplicating chunks.

However, especially for local (i.e. desktop/laptop) use, this becomes a big drag because as the resulting table no longer fits in physical RAM, the os is forced to swap.

rafael-atticlabs:
I feel like a good step forward would be to have a compaction strategy that was

-Is mostly shared (code) between AWS and File
-Rather than extracting chunk-by-chunk, `io.Copy`s from each source's chunks region to the destination (essentially concatinating the payload section of each table), and the recreating a new index at the end as a final step
-The AWS implementation can use Multipart Upload Copy semantics for tables > 5MB to avoid moving alot of data to/from S3.

PR #3421, opened 2017-04-24
NBS table names now just hash of suffix block

cmasone-attic:
Used to be that an NBS table was named by hashing the hashes
of every chunk present in the table, in hash order. That means
that to generate the name of a table you'd need to iterate
the prefix map and load every associated suffix. That would
be expensive when e.g. compacting multiple tables. This is
waaay cheaper and only slightly more likely to wind up with a
name collision.

Toward #3411


cmasone-attic:
PTAL

cmasone-attic:
jenkins: test this

PR #3431, opened 2017-04-26
s3TablePersister must pass rate limiter to compacted tables

cmasone-attic:
Don't know how much this has been an issue in practice, but
it was definitely wrong.


cmasone-attic:
PTAL

PR #3435, opened 2017-04-27
Use one ldb instance per graph builder

willhite:
Remove unused opCache method from virtual store.
Also remove unused setMutator and mapMutator classes.

Fixes https://github.com/attic-labs/attic/issues/1562


go/types/graph_builder.go
```diff
@@ -313,5 +318,5 @@ func commonPrefixCount(stack graphStack, keys ValueSlice) int {
 }
 
 func (e *graphStackElem) String() string {
-	return fmt.Sprintf("key: %s, kind: %s, seq: %p", EncodedValue(e.key), e.kind, e.ch)
+	return fmt.Sprintf("key: %s, kind: %s, seq: %p", EncodedValue(e.key), KindToString[e.kind], e.ch)
```
rafael-atticlabs: `e.kind.String()`

go/types/graph_builder.go
```diff
@@ -112,19 +113,23 @@ type graphOpContainer struct {
...
+	var opcStore opCacheStore
 
-	checkFirstCall := func() {
+	defer func() {
+		opcStore.destroy()
+	}()
+
+	// Use function here to take advantage fo the deferred call to mutex.Unlock()
+	func() {
 		b.mutex.Lock()
```
rafael-atticlabs: i'm confused about the threadsafety of this object. This mutex doesn't appear to be used anywhere else. Is the assumption that this will only be used from a single goroutine? Is that a *true* assumption? 

Seems weird to lock here and nowhere else?

willhite: Yeah, I guess it is a little surprising but most of this just ends up making local changes to the data and then passing it down to ldb.Put() which is thread safe. I believe this is correct. It gets more complicated when we're pulling stuff out of ldb but that's all protected by this single check.

go/types/graph_builder.go
willhite: I see, I guess this was correct as it was. I'll revert this change.

willhite:
jenkins: build this

willhite:
jenkins: build this

willhite:
This had the same spurious test failure when I built it again which I haven't been able to repeat a second time. I modified the test to print out more info if it fails again. PTAL

PR #3441, opened 2017-04-28
Subtype: Make sure we continue to test after optional

arv:
In the case where we have an optional field that is not present we still
need to check the remaining fields.

### DIFF
--- a/go/types/subtype.go
+++ b/go/types/subtype.go
@@ -159,10 +159,7 @@ func IsValueSubtypeOf(v Value, t *Type) bool {
 
 		for _, f := range desc.fields {
 			fv, ok := s.MaybeGet(f.Name)
-			if !ok {
-				return f.Optional
-			}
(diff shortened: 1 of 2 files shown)


arv:
I ran the test before and after the fix to verify that it does test the right thing 🥇

go/types/subtype.go
```diff
@@ -159,10 +159,7 @@ func IsValueSubtypeOf(v Value, t *Type) bool {
 
 		for _, f := range desc.fields {
 			fv, ok := s.MaybeGet(f.Name)
-			if !ok {
-				return f.Optional
-			}
-			if !IsValueSubtypeOf(fv, f.Type) {
+			if !ok && !f.Optional || ok && !IsValueSubtypeOf(fv, f.Type) {
```
rafael-atticlabs: can you just make me happy and put parens here?

PR #3445, opened 2017-05-02
Is subtype disallow extra struct fields

rafael-atticlabs:
It bothered me too much to copy this code in Attic, so I added another kinda weird type function to noms.

Still need to add tests, but let me know if you are ok with this approach.

FYI, I refactored the struct section because the existing code was a bit too hard to reason about. I know my version is longer but it's easier to convince myself that it's correct.

### DIFF
--- a/go/types/subtype.go
+++ b/go/types/subtype.go
@@ -73,38 +73,43 @@ func isSubtype(requiredType, concreteType *Type, allowExtraStructFields bool, pa
 			return true
 		}
 
-		j := 0
-	outer:
-		for _, requiredField := range requiredDesc.fields {
-			for {
-				if j >= concreteDesc.Len() {
-					if requiredField.Optional {
-						continue outer
-					}
+		i, j := 0, 0
+		for i < requiredDesc.Len() && j < concreteDesc.Len() {
+			requiredField := requiredDesc.fields[i]
+			concreteField := concreteDesc.fields[j]
+			if requiredField.Name == concreteField.Name {
+				if !requiredField.Optional && concreteField.Optional {
 					return false
 				}
-				concreteField := concreteDesc.fields[j]
 
-				if concreteField.Name == requiredField.Name {
-					if requiredField.Optional {
-						break
-					}
-					if concreteField.Optional {
-						return false
-					}
-					break
+				if !isSubtype(requiredField.Type, concreteField.Type, allowExtraStructFields, append(parentStructTypes, requiredType)) {
+					return false
 				}
-				if requiredField.Optional && concreteField.Name > requiredField.Name {
-					continue outer
-				}
+
+				i++
 				j++
+				continue
 			}
 
-			concreteField := concreteDesc.fields[j]
-			if !isSubtype(requiredField.Type, concreteField.Type, allowExtraStructFields, append(parentStructTypes, requiredType)) {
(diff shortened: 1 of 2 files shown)


rafael-atticlabs:
Tests added

arv:
Another way to factor this might be to pass in a callback that is called before (and another one for after?) issubtype is called at a level. Then it would be easy to add custom logic for attic.

go/types/subtype.go
```diff
@@ -16,18 +16,23 @@ func assertSubtype(t *Type, v Value) {
 
 // IsSubtype determines whether concreteType is a subtype of requiredType. For example, `Number` is a subtype of `Number | String`.
 func IsSubtype(requiredType, concreteType *Type) bool {
-	return isSubtype(requiredType, concreteType, nil)
+	return isSubtype(requiredType, concreteType, true, nil)
 }
 
-func isSubtype(requiredType, concreteType *Type, parentStructTypes []*Type) bool {
+// IsSubtypeDisallowExtraFields is a slightly wiered variant of IsSubtype. It returns true IFF IsSubtype(requiredType, concreteType) AND Structs in concreteType CANNOT have field names absent in requiredType
```
arv: typo... wiered -> weird

go/types/subtype.go
```diff
@@ -215,7 +236,7 @@ func IsValueSubtypeOf(v Value, t *Type) bool {
 func isMetaSequenceSubtypeOf(ms metaSequence, t *Type) bool {
 	for _, mt := range ms.tuples {
 		// Each prolly tree is also a List<T> where T needs to be a subtype.
-		if !isSubtype(t, mt.ref.TargetType(), nil) {
+		if !isSubtype(t, mt.ref.TargetType(), true, nil) {
```
arv: Maybe add a third isSubtype without the bool and nil?

go/types/subtype_test.go
```diff
@@ -262,8 +262,8 @@ func TestAssertTypeCycleUnion(tt *testing.T) {
 		StructField{"y", MakeUnionType(NumberType, StringType), false},
 	)
 
-	assert.True(tt, isSubtype(t2, t1, nil))
-	assert.False(tt, isSubtype(t1, t2, nil))
+	assert.True(tt, isSubtype(t2, t1, true, nil))
```
arv: Yeah, I don't like the extra params. It is too hard to reason about if you are not aware of the motivation of this weird case.

go/types/subtype_test.go
```diff
@@ -438,6 +420,63 @@ func TestIsSubtypeOptionalFields(tt *testing.T) {
 	assert.False(IsSubtype(t2, t1))
 }
 
+func makeType(s string) *Type {
```
arv: maybe rename now that all tests see this.

go/types/subtype.go
rafael-atticlabs: done

go/types/subtype.go
rafael-atticlabs: done

go/types/subtype_test.go
rafael-atticlabs: done

rafael-atticlabs:
All done. TODO'd suggestion of adding a delegation method.

PR #3447, opened 2017-05-03
add sucky/shameless chromium histogram knock-off

rafael-atticlabs:
I'm generally worried about the cost of locking in metrics libraries and I'd like low-level code to be free to record high-frequency events without worrying about drain on perf. For instance, the next patch here is the NBS ChunkStore exposing a `Stats` member which surfaces a bunch of low-level metrics as `Histogram`s

My idea is that these histograms can be rolled up and used as input to more full-featured metrics abstractions when necessary.

Also, having a basic infrastructure in place, allows for low-activation-energy reporting without dealing with the stack of more sophisticated infrastructure which may need to be operating across larger scale work

cc @aboodman @ehalpern

### DIFF
--- a/go/metrics/histogram.go
+++ b/go/metrics/histogram.go
@@ -1,0 +1,184 @@
+// Copyright 2017 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package metrics
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/attic-labs/noms/go/d"
+	humanize "github.com/dustin/go-humanize"
+)
+
+// Histogram is a shameless and low-rent knock of the chromium project's
+// histogram:
+//   https://chromium.googlesource.com/chromium/src/base/+/master/metrics/histogram.h
+//
+// It logically stores a running histogram of uint64 values and shares some
+// important features of its inspiration:
+//   * It acccepts a correctness deficit in return for not needing to lock.
+//     IOW, concurrent calls to Sample may clobber each other.
+//   * It trades compactness and ease of arithmatic across histograms for
+//     precision. Samples lose precision up to the range of the values which
+//     are stored in a bucket
+//
+// Only implemented: Log2-based histogram
+type Histogram struct {
+	buckets [bucketCount]uint64
+}
+
+const bucketCount = 63
+
+// Sample adds a uint64 data point to the histogram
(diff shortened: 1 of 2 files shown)


go/metrics/histogram.go
```diff
@@ -0,0 +1,163 @@
...
+//     are stored in a bucket
+//
+// Only implemented: Log2-based histogram
+type Histogram struct {
+	buckets [bucketCount]uint64
+}
+
+const bucketCount = 63
+
+func (h *Histogram) Sample(v uint64) {
```
cmasone-attic: Can you provide Godoc-compliant comments on all these public methods?
```
// Sample blah blah blah

// Add combines two Histograms by blah blah blah
```

cmasone-attic:
LGTM w/comment about comments

ehalpern:
I see. Building a lock-free collector makes good sense, especially for hot paths. 

But I'm not sure I understand your suggestion about tooling for instrumenting more of the the code and exposing the metrics via a dashboard.

There are 3 components that we discussed: 
1. A system/api for metrics collection (like go-metrics) 
2. Remotely accessible metrics storage  (like influx-db)
3. Dashboard  (like grafana) 

This patch provides a low level primitive for metrics collection. It does not provide a simple way to report metrics (associated with a name and type) or anything above. Are you suggesting that we start here and not worry about the rest?

If so, I'd like to understand your concerns more about this kind of stack being a burden rather than a help. As I see it, this is very similar to logging. The decision to log in the app doesn't tie you into an aggregation and indexing infrastructure but having that infrastructure makes the logs a lot more useful. 

Similarly, the decision to instrument the code for perf measurement doesn't have to be invasive and is orthogonal to the infrastructure for aggregating stats and showing exploring them in a dashboard.

I don't want to waste time on this if there's disagreement on the goals, and if so, I'm not sure how to proceed on preparing the appropriate instrumentation and tooling for running servers with live users.  

@rafael-atticlabs @aboodman

rafael-atticlabs:
Like I said yesterday, I'm a bit concerned about putting in place too much complexity & infrastructure, but I'm deferring to your judgement here.

I didn't mean to suggest that we don't do 1, 2 or 3. The intent here is that there's a way to easily collect low-level metrics without being a drag on performance. I'm kind of expecting that some code will be instrumented to pass some of these metrics as a `Sum` to go metrics.

The comment about low-activation energy was more that there's kind of a poor mans way to look at aggregated stats even if just locally for development, without needing to pipe it through the full stack of 1, 2 & 3.

ehalpern:
Great. Thanks for the clarifications.

PR #3448, opened 2017-05-03
Remove PutMany

rafael-atticlabs:
No one was doing anything more useful than turning around and calling Put on themselves for each chunk.


cmasone-attic:
lgtm

PR #3449, opened 2017-05-03
Instrument NBS with perf metrics

rafael-atticlabs:
I haven't fixed up the tests. @cmasone-attic, can I get a pre-review before I do.

### DIFF
--- a/go/nbs/persisting_chunk_source.go
+++ b/go/nbs/persisting_chunk_source.go
@@ -8,24 +8,31 @@ import (
 	"bytes"
 	"io"
 	"sync"
+	"time"
 
 	"github.com/attic-labs/noms/go/chunks"
 	"github.com/attic-labs/noms/go/d"
 )
 
-func newPersistingChunkSource(mt *memTable, haver chunkReader, p tablePersister, rl chan struct{}) *persistingChunkSource {
+func newPersistingChunkSource(mt *memTable, haver chunkReader, p tablePersister, rl chan struct{}, stats *Stats) *persistingChunkSource {
+	t1 := time.Now()
+
 	ccs := &persistingChunkSource{mt: mt}
 	ccs.wg.Add(1)
 	rl <- struct{}{}
 	go func() {
 		defer ccs.wg.Done()
-		cs := p.Persist(mt, haver)
+		cs := p.Persist(mt, haver, stats)
 
 		ccs.mu.Lock()
 		defer ccs.mu.Unlock()
 		ccs.cs = cs
 		ccs.mt = nil
 		<-rl
+
+		if cs.count() > 0 {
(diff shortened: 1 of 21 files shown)


go/nbs/persisting_chunk_source.go
```diff
@@ -8,24 +8,31 @@ import (
...
+		cs := p.Persist(mt, haver, stats)
 
 		ccs.mu.Lock()
 		defer ccs.mu.Unlock()
 		ccs.cs = cs
 		ccs.mt = nil
 		<-rl
+
+		if cs.count() > 0 {
+			stats.PersistLatency.SampleTime(time.Since(t1))
```
cmasone-attic: the way this is right now, the time since `t1` will include any time spent waiting on the rate limiter. That may be desirable, or not. On one hand, it tells us how long we had to wait for this table to get done. On the other, we may be more interested in the time it takes to actually do the persisting (getting the chunks all compressed and written and yadda yadda)

go/nbs/stats.go
```diff
@@ -0,0 +1,94 @@
...
+		&metrics.Histogram{},
+	}
+}
+
+func (s *Stats) String() string {
+	return fmt.Sprintf(`---NBS Stats---
+GetLatency:       %s
+ChunksPerGet:     %s
+ReadLatency:      %s
+ChunksPerRead: .  %s
```
cmasone-attic: accidental `.`?

go/nbs/stats.go
```diff
@@ -0,0 +1,94 @@
...
+PutLatency:       %s
+PersistLatency:   %s
+ChunksPerPersist: %s
+BytesPerPersist:  %s
+ConjoinLatency:   %s
+ChunksPerConjoin: %s
+BytesPerConjoin:  %s
+TablesPerConjoin: %s
+`,
+		metrics.TimeHistogram(*s.GetLatency),
```
cmasone-attic: Wouldn't it work for these all to have the right type in the `Stats` struct?

go/nbs/table_reader.go
```diff
@@ -248,11 +258,20 @@ func (tr tableReader) readAtOffsets(readStart, readEnd uint64, reqs []getRecord,
 	}
 
 	wg.Done()
+
+	stats.BytesPerRead.Sample(readLength)
+	stats.ChunksPerRead.SampleLen(len(offsets))
```
cmasone-attic: I forget whether `offsets` or `reqs` is a better gauge of how many chunks are being read. Does one collapse multiple reads of the same chunk into a single item?

cmasone-attic:
in general, looks fine to me

rafael-atticlabs:
TBR @cmasone-attic

PR #3451, opened 2017-05-05
Clean up NBS cruft

cmasone-attic:
Clean up NBS cruft standing in the way of improvements:

- Unmap buffer in newMmapTableReader()
By the time this function exits, we're done with this buffer.
Hanging on to it complicates lifetime management for the file
backing the mmapTableReader, which is something I'm trying to
make simpler. So...ditch it!

- remove compactSourcesToBuffer
replace with simpler test-focused version


cmasone-attic:
PTAL

Issue #3456, opened 2017-05-07
`types.ValueChange` should contain K, OldV, NewV for ordered sequence diff reports


rafael-atticlabs:
1) There's no extra IO cost to deliver the old & new values given that both leaf sequences are already loaded.
2) Many uses turn around and fetch the values immediately, which risks IO
3) There's a bunch of unfortunate code trying to abstract Map/Set/Struct diff which has to deal with this. A much simpler thing is that for Map/Struct, K is the Key and OldV/NewV are the (possibly nil) old & new values. For Set, K is the Value and OldV/NewV are always nil.

Go clean up all uses in noms & attic which turn around and fetch values out of maps/structs and remove (then) needless wrapper code.

arv:
> For Set, K is the Value and OldV/NewV are always nil.

Doesn't a delete/remove get reported with a nil?

rafael-atticlabs:
fixed

PR #3458, opened 2017-05-08
have diff return old/new values

rafael-atticlabs:
Towards #3456

For whomever gets to this, here's a starter patch.

@kalman, @aboodman says you might have concerns with this approach?

cc @willhite


kalman:
modulo test failures obviously

Issue #3462, opened 2017-05-09
Stop abandoning compactions so much


cmasone-attic:
We waste a lot of time, memory and IO when one client abandons an (expensive) compaction because another client lands a manifest update before it finishes. Instead, make it so that a client who encounters an optimistic lock failure after compaction can still land it as long as it would be a benefit to the store

PR #3463, opened 2017-05-09
NBS: Remove need to manually manage chunkSource lifetime

cmasone-attic:
For file-backed NBS, it's important that chunkSources get closed at
some point, to free up limited resources like file descriptors.
Previously, whenever a caller compacted or rebased a tableSet, the
caller had to manually close the chunkSources that were no longer
needed. This complicates logic around retrying in the event of an
optimistic lock failure.

This patch introduces a cache for open file descriptors, which
the fsTablePersister uses to manage the lifetime of files
automatically. The fdCache ref-counts open file descriptors,
but doesn't keep a hard cap on the number of open files. Once
the cache's target size is exceeded, opening a new file causes
the cache to look for fds with zero refs to close. It closes
as many of those as needed to get back under the target size.

Towards #3462

### DIFF
--- a/go/nbs/fd_cache.go
+++ b/go/nbs/fd_cache.go
@@ -36,14 +36,25 @@ type fdCacheEntry struct {
 // indicating why the file could not be opened. If the cache already had an
 // entry for |path|, RefFile increments its refcount and returns the cached
 // pointer. If not, it opens the file and caches the pointer for others to
-// use.
+// use. This is intended for clients that hold fds for extremely short
+// periods.
 // If reffing a currently unopened file causes the cache to grow beyond
-// |fc.targetSize|, RefFile makes a best effort to shrink the
-// cache by dumping entries with a zero refcount. If there aren't enough zero
-// refcount entries to drop to get the cache back to |fc.targetSize|, the
-// cache will remain over |fc.targetSize| until the next call to RefFile().
+// |fc.targetSize|, RefFile makes a best effort to shrink the cache by dumping
+// entries with a zero refcount. If there aren't enough zero refcount entries
+// to drop to get the cache back to |fc.targetSize|, the cache will remain
+// over |fc.targetSize| until the next call to RefFile().
 func (fc *fdCache) RefFile(path string) (f *os.File, err error) {
-	f, err = fc.checkCache(path)
+	// Anonymous function to scope locking
+	f, err = func() (f *os.File, err error) {
+		fc.mu.Lock()
+		defer fc.mu.Unlock()
+		if ce, present := fc.cache[path]; present {
+			ce.refCount++
+			fc.cache[path] = ce
+			return ce.f, nil
+		}
+		return nil, nil
+	}()
 	if f != nil || err != nil {
 		return f, err
 	}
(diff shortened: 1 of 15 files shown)


cmasone-attic:
PTAL

go/nbs/fd_cache.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, err
+	}
+	fc.cache[path] = fdCacheEntry{f: f, refCount: 1}
+	return f, nil
+}
+
+func (fc *fdCache) checkCache(path string) (f *os.File, err error) {
+	fc.mu.Lock()
+	defer fc.mu.Unlock()
```
rafael-atticlabs: put a comment here that this ordering is critical

go/nbs/fd_cache.go
```diff
@@ -0,0 +1,131 @@
...
+// RefFile returns an opened *os.File for the file at |path|, or an error
+// indicating why the file could not be opened. If the cache already had an
+// entry for |path|, RefFile increments its refcount and returns the cached
+// pointer. If not, it opens the file and caches the pointer for others to
+// use.
+// If reffing a currently unopened file causes the cache to grow beyond
+// |fc.targetSize|, RefFile makes a best effort to shrink the
+// cache by dumping entries with a zero refcount. If there aren't enough zero
+// refcount entries to drop to get the cache back to |fc.targetSize|, the
+// cache will remain over |fc.targetSize| until the next call to RefFile().
```
rafael-atticlabs: I think it's worth noting here that this is designed for clients which have fd refs on their behalf for extremely short time periods.

go/nbs/fd_cache.go
```diff
@@ -0,0 +1,131 @@
...
+	defer fc.mu.Unlock()
+	defer fc.maybeShrinkCache()
+	if err != nil {
+		return nil, err
+	}
+	fc.cache[path] = fdCacheEntry{f: f, refCount: 1}
+	return f, nil
+}
+
+func (fc *fdCache) checkCache(path string) (f *os.File, err error) {
```
rafael-atticlabs: check cache isn't really a great name for this because its incrementing ref count on a file. there's only one caller for this. what about just inlining this in `RefFile` and move GCing open fds if neccesary to Unref()?

go/nbs/store.go
```diff
@@ -177,6 +204,16 @@ func (nbs *NomsBlockStore) Put(c chunks.Chunk) {
 	nbs.stats.PutLatency.SampleTime(time.Since(t1))
 }
 
+func (nbs *NomsBlockStore) SchedulePut(c chunks.Chunk) {
```
rafael-atticlabs: rebase cruft. remove

go/nbs/fd_cache.go
cmasone-attic: done

go/nbs/fd_cache.go
cmasone-attic: Moved to `Unref()` and, in that context, `maybeShrinkCache()` no longer needs to be deferred

go/nbs/fd_cache.go
cmasone-attic: done

go/nbs/store.go

PR #3465, opened 2017-05-09
expose isLeaf() as a more direct signal on sequence

rafael-atticlabs:


go/types/meta_sequence.go
```diff
@@ -278,7 +282,7 @@ func metaHashValueBytes(item sequenceItem, rv *rollingValueHasher) {
 	hashValueBytes(v, rv)
 }
 
-type emptySequence struct{}
+type emptySequence bool
```
kalman: Why did you change it? `struct{}` seems more idiomatic.

PR #3470, opened 2017-05-10
EmptySequenceDiff

rafael-atticlabs:
You are the winner today

### DIFF
--- a/go/types/meta_sequence.go
+++ b/go/types/meta_sequence.go
@@ -30,11 +30,6 @@ type metaTuple struct {
 	child     Collection // may be nil
 }
 
-type metaOrEmptySequence interface {
-	getCompositeChildSequence(start uint64, length uint64, height int) sequence
-	cumulativeNumberOfLeaves(idx int) uint64
-}
-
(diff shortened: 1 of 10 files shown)


go/types/meta_sequence.go
```diff
@@ -30,6 +30,11 @@ type metaTuple struct {
 	child     Collection // may be nil
 }
 
+type metaOrEmptySequence interface {
```
kalman: Can you comment that this is for diff? There are an overwhelming number of sequence types and this shouldn't be confused for one of them.

Though really this doesn't feel very Go-like. What if you just moved these to `sequence`? `cumulativeNumberOfLeaves` can be defined there (not that it's particularly useful), and `getCompositeChildSequence` can just panic, doesn't seem that bad.

Speaking of which, I didn't notice in your last patch, but you can just define these on `leafSequence` (leaf_sequence.go) and move the `isLeaf` function definition to there, remove fro the various leaf sequence types.

rafael-atticlabs: fair enough. all done. thanks for close review

PR #3471, opened 2017-05-10
NBS: Tolerate other clients writing novel tables during compaction

### LINKED ISSUE #3462: Stop abandoning compactions so much

cmasone-attic:
Currently, we handle NBS compactions by having clients detect that one
is necessary, perform the compaction locally, and then attempt to
persist the work. The old strategy, removed in this patch, was to
abandon the work if any other client had changed anything upstream
since the compaction began. Since compaction can be slow, this
happened frequently. The new strategy is to tolerate new tables
appearing upstream as long as all the tables the client compacted
still exist. In that case, the client will now rebase against upstream
successfully, droping the tables that got compacted, and then attempt
to land the compaction once again.

Fixes #3462

### DIFF
--- a/go/nbs/table_set.go
+++ b/go/nbs/table_set.go
@@ -21,9 +21,18 @@ func newTableSet(persister tablePersister) tableSet {
 
 // tableSet is an immutable set of persistable chunkSources.
 type tableSet struct {
-	novel, compacted, compactees, upstream chunkSources
-	p                                      tablePersister
-	rl                                     chan struct{}
+	// novel chunkSources contain chunks that have not yet been pushed upstream
+	novel chunkSources
+
+	// When Compact() conjoins upstream tables into larger tables, the results are stored in |compacted| until the next call to Flatten(). The tables that were conjoined are put into |compactees|, so they can be correctly handled during a Rebase()
+	compacted  chunkSources
+	compactees chunkSources
+
+	// upstream holds the set of already-persisted chunkSources that this tableSet references.
+	upstream chunkSources
+
+	p  tablePersister
+	rl chan struct{}
 }
 
 func (ts tableSet) has(h addr) bool {
@@ -248,17 +257,19 @@ func (ts tableSet) Rebase(specs []tableSpec) tableSet {
 	}
 
 	// Only keep locally-performed compactions if ALL compactees are still present upstream
-	keepCompactions := func() bool {
+	keepCompactions, compacteeSet := func() (bool, map[addr]struct{}) {
 		specsNames := map[addr]struct{}{}
 		for _, spec := range specs {
 			specsNames[spec.name] = struct{}{}
 		}
+		set := map[addr]struct{}{}
 		for _, compactee := range ts.compactees {
 			if _, present := specsNames[compactee.hash()]; !present {
-				return false
+				return false, nil
 			}
+			set[compactee.hash()] = struct{}{}
 		}
-		return true
+		return true, set
 	}()
 	if keepCompactions {
 		merged.compacted = make(chunkSources, len(ts.compacted))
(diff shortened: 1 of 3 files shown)


go/nbs/table_set.go
```diff
@@ -21,9 +21,9 @@ func newTableSet(persister tablePersister) tableSet {
 
 // tableSet is an immutable set of persistable chunkSources.
 type tableSet struct {
-	novel, upstream chunkSources
-	p               tablePersister
-	rl              chan struct{}
+	novel, compacted, compactees, upstream chunkSources
```
cmasone-attic: I initially tried not to keep all the `compacted` and `compactee` state here inside `tableSet`. Instead, I had `Compact()` return that info to the caller. This was pretty nice until I got to the part where I needed that state to persist across multiple invocations to `updateManifest`, in the event that the client encountered multiple optimistic lock failures while trying to land a compaction. Essentially, you need to remember which tables are the result of a compaction, and which tables are now rolled up into a larger table allll the way until you successfully update the manifest. Trying to maintain that state outside `tableSet` turned out to be hairier than just keeping the complexity in here, though it initially seemed like it'd be cleaner and simpler.

cmasone-attic:
PTAL

cmasone-attic:
jenkins: test this, guy

go/nbs/table_set.go
```diff
@@ -228,20 +276,19 @@ func (ts tableSet) Rebase(specs []tableSpec) tableSet {
 	}
 
 	// Open all the new upstream tables concurrently
-	openedTables := make(chunkSources, len(tablesToOpen))
+	merged.upstream = make(chunkSources, len(tablesToOpen))
```
rafael-atticlabs: it seems like when we compact, we remove compactees from upstream. am wrong, or will they re-appear here even if keepCompactions is true?

go/nbs/table_set.go
```diff
@@ -142,15 +154,19 @@ func (ts tableSet) Prepend(mt *memTable, stats *Stats) tableSet {
...
-		rl:    ts.rl,
+	ns := tableSet{
+		novel:      make(chunkSources, len(ts.novel)),
+		compacted:  make(chunkSources, len(ts.compacted)+1),
+		compactees: make(chunkSources, len(ts.compactees)),
+		p:          ts.p,
+		rl:         ts.rl,
 	}
 	copy(ns.novel, ts.novel)
+	copy(ns.compacted[1:], ts.compacted) // leave the first slot for the newly compacted table
```
cmasone-attic: My primary thing was that, just looking at the `tableSet` interface, only being able to call `Compact` once felt like a weird and artificial restriction.
I started coding it that way anyway and discovered that having `compacted` be a `chunkSource` instead of `chunkSources` made the code, in my opinion, a little clunkier to read. Like, you have to do explicit `nil` checks instead of just ranging over a thing that might happen to be nil. Or, instead of being able to re-use a little helper function that works on `novel` and `upstream`, you have to do something slightly different. See, for example `has()` and `hasMany()`.

That's a minor consideration, certainly, but given that restricting it to just one `Compact()` call already felt kind of weird, I decided to do it this way. There seemed to be little downside and it made the code and API better.

go/nbs/table_set.go
cmasone-attic: you're right! I dropped the code that did that when I changed approaches this morning. Fixed

go/nbs/table_set.go

cmasone-attic:
PTAL

go/nbs/table_set.go
```diff
@@ -21,9 +21,18 @@ func newTableSet(persister tablePersister) tableSet {
 
 // tableSet is an immutable set of persistable chunkSources.
 type tableSet struct {
-	novel, compacted, compactees, upstream chunkSources
-	p                                      tablePersister
-	rl                                     chan struct{}
+	// novel chunkSources contain chunks that have not yet been pushed upstream
+	novel chunkSources
+
+	// When Compact() conjoins upstream tables into larger tables, the results are stored in |compacted| until the next call to Flatten(). The tables that were conjoined are put into |compactees|, so they can be correctly handled during a Rebase()
```
rafael-atticlabs: I was kind of more looking for a description of the relationship between compacted, compactees and upstream. kind of explaining that if compacted is non-empty, then compactees will have a corresponded set of tables which comprised tables in compacted and upstream will "lack" the tables in compactees (if one actually "looked" upstream).

cmasone-attic: ok, I tried again.

PR #3472, opened 2017-05-10
add .Now() to datetime package

jesseditson:
Mainly putting this up for consideration, could easily put this abstraction in the consuming code.

Also not sure if there's a way to test this. Is there a way to mock a package that isn't injected in go?

### DIFF
--- a/go/util/datetime/date_time.go
+++ b/go/util/datetime/date_time.go
@@ -32,6 +32,11 @@ var dateTimeTemplate = types.MakeStructTemplate("DateTime", []string{"secSinceEp
 // which makes it useful for testing or checking for uninitialized values
 var Epoch = DateTime{time.Unix(0, 0)}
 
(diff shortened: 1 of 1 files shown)


jesseditson:
Reasoning behind this: `vet` doesn't like `DateTime{time.Now()}` (composite literal uses unkeyed fields)

aboodman:
Is the test failure real? Otherwise lgtm.

jesseditson:
Failures look unrelated to me, it's in go/types (TestStreamingMap2 and TestMapSuite4KStructs). They don't fail locally for me, so not really sure what's going on here. Perhaps intermittent?

willhite:
jenkins: build this

willhite:
I haven't seen this test failure before. The weird thing is that from the output, the look like they're equal.

PR #3475, opened 2017-05-11
NBS: Move mmapTableReader I/O into indexCache miss path

cmasone-attic:
Now that we might open/close a file at any time for reading, there's
no value in acutally reffing a file while building a mmapTableReader
in the case where the index is found in the indexCache.


cmasone-attic:
PTAL. Figured we should do this no matter what.

PR #3478, opened 2017-05-11
NBS: Don't add 0-sample to histogram in Put()

### LINKED ISSUE #1787: Remove "Q-encoding"
csv-export writes the escaped header names into the exported csv-file. csv-import reads in the escaped names and escapes them again. This causes field names and ultimately the row type to change.

cmasone-attic:
Turns out sometimes Put() is so fast it gets reported as
taking 0ns. It's nonsensical to put a 0 in a Histogram, so
don't. Just change it to 1.

Fixes #1787


cmasone-attic:
PTAL

PR #3480, opened 2017-05-15
NBS: Sharing one persister at the LocalStoreFactory level is bad

### LINKED ISSUE #3479: NBS file factory is writing table files into the top directory
seems bad

cmasone-attic:
When minting NBS stores from a LocalStoreFactory, each gets a
directory with the DB name in the path, to keep the tables nicely
separated.  That means that the tablePersister for each store needs to
know the right directory to use. It can't know that until the DB's
name is provided, which means that we can't share a single persister
object across all stores vended by a LocalStoreFactory. That's fine,
as long as they all share an indexCache and an fdCache.

Fixes #3479


cmasone-attic:
PTAL

ghost (approved):
lgtm

PR #3481, opened 2017-05-17
Stop using graphbuilder in NewStreamingMap and NewStreamingSet

willhite:


go/types/map.go
```diff
@@ -38,22 +38,34 @@ func NewMap(kv ...Value) Map {
...
 func NewStreamingMap(vrw ValueReadWriter, kvs <-chan Value) <-chan Map {
 	var k Value
 	outChan := make(chan Map, 1)
 	go func() {
 		defer close(outChan)
-		gb := NewGraphBuilder(vrw, MapKind, false)
+		ch := newEmptyMapSequenceChunker(vrw, vrw)
+		var lastK Value
 		for v := range kvs {
 			if k == nil {
```
rafael-atticlabs: let's just be a bit more direct here and maintain a local variable, like `nextIsKey` that we toggle each pass.

go/types/map.go
```diff
@@ -38,22 +38,34 @@ func NewMap(kv ...Value) Map {
...
+// graph_builder.go for building collections with values that are not in order.
 func NewStreamingMap(vrw ValueReadWriter, kvs <-chan Value) <-chan Map {
 	var k Value
 	outChan := make(chan Map, 1)
 	go func() {
 		defer close(outChan)
-		gb := NewGraphBuilder(vrw, MapKind, false)
+		ch := newEmptyMapSequenceChunker(vrw, vrw)
+		var lastK Value
 		for v := range kvs {
```
rafael-atticlabs: d.PanicIfNil(v)

go/types/map.go
```diff
@@ -38,22 +38,34 @@ func NewMap(kv ...Value) Map {
...
 			if k == nil {
 				k = v
+				if lastK != nil {
```
rafael-atticlabs: d.PanicIfFalse(lastK == nil || lastK.Less(k))

go/types/set.go
```diff
@@ -31,15 +32,27 @@ func NewSet(v ...Value) Set {
...
 func NewStreamingSet(vrw ValueReadWriter, vals <-chan Value) <-chan Set {
 	outChan := make(chan Set, 1)
 	go func() {
 		defer close(outChan)
-		gb := NewGraphBuilder(vrw, SetKind, false)
+		ch := newEmptySetSequenceChunker(vrw, vrw)
+		var lastV Value
 		for v := range vals {
-			gb.SetInsert(nil, v)
+			if lastV != nil {
```
rafael-atticlabs: d.PanicIfNil(v)
d.PanicIffalse(lastV == nil || lastV.Less(v))

go/types/set.go
```diff
@@ -31,15 +32,27 @@ func NewSet(v ...Value) Set {
...
+				d.PanicIfFalse(lastV.Less(v))
+			}
+			if v != nil {
```
rafael-atticlabs: why are we tolerating nil values here?

go/types/map.go
willhite: Done. Had to use `d.PanicIfTrue(v == nil)` to avoid boxing.

go/types/map.go
willhite: Done.

go/types/set.go
willhite: Done.

go/types/set.go
willhite: Good catch. This was leftover from testing.

PR #3482, opened 2017-05-17
Remove vestigial 'verbose' argument to NewGraphBuilder

willhite:


willhite:
Super-simple cleanup of an argument that is no longer necessary.

PR #3483, opened 2017-05-18
Add hasExtra return value to IsValueSubtypeOf()

willhite:
@rafael-atticlabs This is my change to enable ComputeView to determine whether a value can be returned as-is or whether it needs to be narrowed. I ended up with the idea that we needed to bits of info: whether a value was a subtype of something and whether it had extra fields. It seems like the easiest thing to do was to add a return value to IsValueSubtypeOf() to return that additional piece of information.

### DIFF
--- a/go/types/subtype.go
+++ b/go/types/subtype.go
@@ -9,70 +9,83 @@ import (
 )
 
 func assertSubtype(t *Type, v Value) {
-	if !IsValueSubtypeOf(v, t) {
+	if isSub, _ := IsValueSubtypeOf(v, t); !isSub {
 		d.Panic("Invalid type. %s is not a subtype of %s", TypeOf(v).Describe(), t.Describe())
 	}
 }
 
 // IsSubtype determines whether concreteType is a subtype of requiredType. For example, `Number` is a subtype of `Number | String`.
 func IsSubtype(requiredType, concreteType *Type) bool {
-	return isSubtypeTopLevel(requiredType, concreteType)
+	isSub, _ := isSubtypeTopLevel(requiredType, concreteType)
+	return isSub
 }
 
 // IsSubtypeDisallowExtraFields is a slightly weird variant of IsSubtype. It returns true IFF IsSubtype(requiredType, concreteType) AND Structs in concreteType CANNOT have field names absent in requiredType
 // ISSUE: https://github.com/attic-labs/noms/issues/3446
 func IsSubtypeDisallowExtraStructFields(requiredType, concreteType *Type) bool {
+	isSub, hasExtra := isSubtype(requiredType, concreteType, false, nil)
+	if hasExtra {
+		return false
+	}
+	return isSub
+}
+
+// isSubtypeTopLevel returns two values: IsSub and hasExtra. See IsValueSubtypeOf()
+// below for an explanation.
+func isSubtypeTopLevel(requiredType, concreteType *Type) (isSub bool, hasExtra bool) {
 	return isSubtype(requiredType, concreteType, false, nil)
 }
 
-func isSubtypeTopLevel(requiredType, concreteType *Type) bool {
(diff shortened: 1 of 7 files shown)


willhite:
@arv fyi

rafael-atticlabs:
I'm going to defer to @arv, here but WRT API, I feel like the public signature for IsValueSubtype should not change here but rather add another public call like IsValueSubtypeDetailed() which returns two bools and IsValueSubtype just calls. That way the existing call sites which don't care don't look so strange.

go/types/subtype.go
```diff
@@ -105,73 +120,108 @@ func isSubtype(requiredType, concreteType *Type, allowExtraStructFields bool, pa
...
+// IsValueSubtypeOf(v, type1) would return isSub == true, and hasExtra == false
+//
+// And given these types:
+//   type2 := struct S {               v := Struct S1 {
+//       a Number | string                 a: "hello"
+//       b ?int                            b: 2
+//   }                                     c: "hello again"
+//                                     }
+// IsValueSubtypeOf(v, type1) would return isSub == true, and hasExtra == true
+func IsValueSubtypeOf(v Value, t *Type) (bool, bool) {
```
arv: Don't return two booleans here. Just add another function that does that. It is very strange to ever want to care about this strange hasExtra thing. It is not part of subtype it is just a strange thing that most people should never have to run into.

willhite:
@arv PTAL

go/types/subtype.go
```diff
@@ -105,73 +124,111 @@ func isSubtype(requiredType, concreteType *Type, allowExtraStructFields bool, pa
...
+			}
+			isSub, hasMore := isValueSubtypeOfDetails(v, et, hasExtra)
+			if isSub {
+				hasExtra = hasExtra || hasMore
+				return isSub, hasExtra
 			}
 		}
-		return false
+
+		if anonStruct != nil {
```
arv: What if there are more than one anon struct?

go/types/subtype.go
```diff
@@ -180,16 +237,29 @@ func IsValueSubtypeOf(v Value, t *Type) bool {
...
 		}
-
+		missingOptionalFieldCnt := 0
 		for _, f := range desc.fields {
 			fv, ok := s.MaybeGet(f.Name)
-			if (!ok && !f.Optional) || (ok && !IsValueSubtypeOf(fv, f.Type)) {
-				return false
+			if !ok {
+				if f.Optional {
+					missingOptionalFieldCnt += 1
```
arv: ++

go/types/subtype_test.go
```diff
@@ -758,3 +758,46 @@ func TestIsValueSubtypeOf(tt *testing.T) {
...
+		true, true,
+	)
+	test(
+		NewStruct("Struct", StructData{"x": Bool(true)}),
+		MakeStructType("Struct",
+			StructField{"x", BoolType, false},
+			StructField{"y", NumberType, false},
+		),
+		false, false,
+	)
```
arv: No test for `false, true`?

Maybe structure this as in TestIsSubtypeDisallowExtraStructFields so that we can easily covefr more cases?

go/types/subtype.go
willhite: Copied a comment from elsewhere in the source that explains what is going on here. Also, we verified that there can only be one anon struct per type.

go/types/subtype.go
willhite: Done.

go/types/subtype_test.go
willhite: It doesn't really make sense to check hasExtra if isSubType is false. I'm not sure what that means. Excellent suggestion to model tests after TestIsSubtypeDisallowExtraStructFields. That's done.

PR #3486, opened 2017-05-18
add noms show --stats

rafael-atticlabs:


cmd/noms/noms_show.go
```diff
@@ -29,12 +29,14 @@ var nomsShow = &util.Command{
...
 
 var showRaw = false
+var showStats = false
 
 func setupShowFlags() *flag.FlagSet {
 	showFlagSet := flag.NewFlagSet("show", flag.ExitOnError)
 	outputpager.RegisterOutputpagerFlags(showFlagSet)
 	verbose.RegisterVerboseFlags(showFlagSet)
 	showFlagSet.BoolVar(&showRaw, "raw", false, "If true, dumps the raw binary version of the data")
+	showFlagSet.BoolVar(&showStats, "stats", false, "If true, reports statistics relatived to the value")
```
cmasone-attic: realtived?

cmd/noms/noms_show.go
```diff
@@ -49,6 +51,10 @@ func runShow(args []string) int {
 		return 0
 	}
 
+	if showRaw && showStats {
+		fmt.Fprintln(os.Stderr, "--raw and --stats are mutually exclusive")
```
cmasone-attic: Shouldn't this return 1 or otherwise error out?

go/types/value_stats.go
```diff
@@ -0,0 +1,123 @@
...
+		nodes = loadNextLevel(children, vr)
+		level--
+		totalCompressedSize += byteSize
+		totalChunks += chunkCount
+	}
+}
+
+func printTreeLevel(w io.Writer, level, values, chunks, byteSize uint64) {
+	avgItems := float64(values) / float64(chunks)
+	avgSize := float64(byteSize) / float64(chunks)
```
cmasone-attic: WHy do this as a float64 if you're just going to turn it back into a uint64 below?

go/types/value_stats.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	// Fetch committed child sequences in a single batch
+	valueChan := make(chan Value, len(hs))
+	go func() {
+		vr.ReadManyValues(hs, valueChan)
+		close(valueChan)
+	}()
+
+	vm := make(map[hash.Hash]Value, len(hs))
```
cmasone-attic: Do you actually care that `values` is in the same order as `refs`? If not, you could skip this mapping-back

cmd/noms/noms_show.go
rafael-atticlabs: you got a problem with that?

cmd/noms/noms_show.go
rafael-atticlabs: hmm.. does kinda seem like it should return non-zero, but i'm gonna be lazy and consistent with surrounding code.

go/types/value_stats.go
rafael-atticlabs: done.

go/types/value_stats.go
rafael-atticlabs: i care about order, per se, right now (although I was thinking that eventually this would report the NBS locality, in which case it would be useful to keep them in order), but I do want to return a slice here because Blobs & Lists can repeat nodes within a level and I wanted to retain that.

PR #3489, opened 2017-05-19
maxTable nit-picking

rafael-atticlabs:
Thinking more about it, there's no current need for a consumer to *set* max tables for the file backend (one could imagine it for tuning purposes, but we're not there). However, we now have a more meaningful control over max open file handles, so that's now configured.

Also, while we're picking numbers of out thin air, 128 tends to keep the manifest < 1 read on dynamo, but seems a tad small for file, which has no such limit. setting file 2x as large.

### DIFF
--- a/go/nbs/factory.go
+++ b/go/nbs/factory.go
@@ -18,6 +18,7 @@ import (
 
 const (
 	defaultAWSReadLimit = 1024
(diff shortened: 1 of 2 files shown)


rafael-atticlabs:
TBR @cmasone-attic

Issue #3384, opened 2017-04-13
Idea: Simplify Sync


rafael-atticlabs:
In light of https://github.com/attic-labs/noms/issues/3177, I was thinking about the basic approach to syncing, which has worked reasonably well but was designed with some assumptions which aren't a true any more.

The general form of the sync algorithm is:

```
for len(refsToSync) > 0 {
  for r in refsToSync {
    if !sink.has(r) {
      c := src.get(r)
      sink.put(r)
      r.walkRefs((sr) refsToSync.Insert(sr))
    }
  }
}
```

However, there's a heuristic optimization that essentially tries to reduce the number of `has` checks which are performed. That optimization tries to "walk" the "old" value more or less in parallel with the "new" one, using height to keep things in sync.

The optimization made sense in a world where each Has() check is basically a full block read. In the simple case of updating the spine of a deep/wide prolly tree, the trade off is reading Log(N) blocks from "sink" instead of doing Log(N)*~200 Has() checks, or something like 200x more reads.

However, with NBS, bulk Has() checks are dramatically less expensive and basically never hit disk. Even in the case of a very large index, the suffix index is generally in locality order, so there's a good chance that the same N refs contained in the block that would have been read, can be read from the index on disk in a single block.

I think this means that for "pulling" (the sink is local), the chunk-sync heuristic optimization can't ever really be a win any more.

In the case of a "push" (the sink is remote), it's a bit fuzzier because the bulk `has` need to go across the wire and in the case of a thin pipe, it might be preferable to do N local reads rather than 200N `has` because the number of hashes that need to be sent might be large.

One option to address this is this: when we implement batched writing, you can imaging the `writeValue` endpoint returning the set of dangling refs which aren't already contained in the target database after the write value. Given this, "push" can just `writeValue` each "level" and field the set of dangling pointers which are returned and write them in the next `writeValue`. This would effectively achieve the thing we observed in the past: that the kind of ideal thing for "push" is to ask the other side to "pull".

An additional benefit of changing things to work this way is it gets rid of the writing pattern for sync of writing all chunks of a given height in lockstep. The problem with this is it doesn't generate the locality that we'd like. The "correct" locality arises out of doing a BFS from the root value. Removing the height-order processing would make it correct and straight forward to have sync basically be:

```
for len(refsToWrite) > 0 {
  chunks := source.GetMany(refsToWrite)
  refsToWrite = sink.PutMany(chunks) // returns unresolved refs
}
```

[...]

PR #3490, opened 2017-05-22
Simplify Pull()

### LINKED ISSUE #3182: Sync should read "batches" of chunks at each level using `ReadManyValues`
So as to maximize read perf from nbs (by maximizing usage of chunk locality)

cmasone-attic:
In an NBS world, bulk 'has' checks are waaaay cheaper than they used
to be. In light of this, we can toss out the complex logic we were
using in Pull() -- which basically existed for no reason other than to
avoid doing 'has' checks. Now, the code basically just descends down a
tree of chunks breadth-first, using HasMany() at each level to figure
out which chunks are not yet in the sink all at once, and GetMany() to
pull them from the source in bulk.

Fixes #3182, Towards #3384

### DIFF
--- a/go/datas/pull.go
+++ b/go/datas/pull.go
@@ -7,9 +7,8 @@ package datas
 import (
 	"math"
 	"math/rand"
-	"sort"
-	"sync"
 
+	"github.com/attic-labs/noms/go/chunks"
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
 	"github.com/attic-labs/noms/go/types"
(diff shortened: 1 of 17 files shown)


cmasone-attic:
PTAL. This doesn't to anything to mitigate the fact that HasMany over the wire is still a bit expensive, but I think it's way better than what we were doing. As a part of future work we could leverage whatever we do for batched-writing to make that better, though.

go/nbs/reader_test.go
```diff
@@ -0,0 +1,19 @@
+// Copyright 2016 Attic Labs, Inc. All rights reserved.
```
rafael-atticlabs: ?

go/nbs/frag/frag_test.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"testing"
+
+	"github.com/attic-labs/noms/go/chunks"
+	"github.com/attic-labs/noms/go/types"
+	"github.com/attic-labs/testify/assert"
+)
+
+func TestFrag(t *testing.T) {
+	assert := assert.New(t)
```
rafael-atticlabs: what is this testing?

go/datas/pull.go
```diff
@@ -29,259 +28,63 @@ const bytesWrittenSampleRate = .10
...
-					sinkQ.PushBack(reachable)
-					if !isHeadOfSink {
-						srcQ.PushBack(reachable)
-					}
-				}
-				comWork--
-				updateProgress(1, 0, uint64(res.readBytes), 0)
+	var sampleSize, sampleCount uint64
+
+	cc := newCompletenessChecker()
```
rafael-atticlabs: i think we're in danger of having two completeness checkers operating. if the chunkStore is an httpChunkStore, then it will already have been wrapped in a validatingChunkStore

go/datas/pull.go
```diff
@@ -29,259 +28,63 @@ const bytesWrittenSampleRate = .10
 // validation, which triggers sinkDB reads, which means that the code can no
 // longer tell which reads were caused by Pull() and which by Flush().
 // TODO: Get rid of this (BUG 2982)
```
cmasone-attic: I had to tweak some stuff in the tests, but it's gone now :-)

PR #3497, opened 2017-05-23
Add Stats() to Database

### LINKED ISSUE #3493: Add a way to get NBS `Stats` from `Database`
I'm fine if its just

`func (db Database) Stats() interface{}`

cmasone-attic:
This just returns interface{}, allowing underlying ChunkStore
implementations to return whatever kind of stats struct they want.

Fixes #3493


cmasone-attic:
PTAL. Should be quick

PR #3499, opened 2017-05-25
Dataset caches HeadValue rather than HeadRef

rafael-atticlabs:
This is kind of minor, but it's silly that db.GetDataset was reading the head value then dropping it on the floor. Instead, cache it inside Dataset rather than the ref. As a side-benefit, this also creates the `HeadRef` lazily, which is nice because it's rarely needed and the type is presently expensive to compute.


rafael-atticlabs:
TBR @cmasone-attic

PR #3500, opened 2017-05-25
Modify roll.py to handle multiple --incl directories

ehalpern:
This makes it possible to include/exclude multiple directories
when rolling src into a vendor directory.

For example, to roll the `cmd` and `go` directories from github.com/attic-labs/noms
into your vendor directory, you could run the following:

```
cd $MY_REPO_HOME
$NOMS/tools/roll.py --incl cmd --incl go https://github.com/attic-labs/noms
```

This patch also strips '.git' if it's present on the end of the repo url.

### DIFF
--- a/tools/roll.py
+++ b/tools/roll.py
@@ -5,6 +5,7 @@
 # http://www.apache.org/licenses/LICENSE-2.0
 
 import argparse, os, os.path, subprocess, sys, shutil, urlparse
+from collections import defaultdict
 
 
 def main():
@@ -15,9 +16,6 @@ def main():
     parser.add_argument('--incl', action='append', help=(
         'subdirectories of the dependency to check out, relative to the path. '
         'Defaults to root. Evaluated before --excl.'))
-    parser.add_argument('--excl', action='append', help=(
-        'subdirectories of the dependency to delete after checking out, relative to the path. '
-        'Defaults to nothing. Evaluated after --incl.'))
     parser.add_argument('--version', default='HEAD', help=(
         'version of the dependency to snapshot, defaults to HEAD'))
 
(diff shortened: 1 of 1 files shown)


ehalpern:
Apologies for the diff. I reset tabs to the standard 4 spaces.

kalman:
Ummmmm can you un-reset it?

kalman:
It's basically reviewing the script from scratch which I guess I can do.

ehalpern:
Split change in two to aid review

tools/roll.py
```diff
@@ -77,28 +85,41 @@ def rel(subdir):
...
+      if dirpath in inclParentToName:
+        inclName = inclParentToName[dirpath]
+        # Don't descend into the included subdirectory.
+        dirnames.remove(inclName)
+      elif not inclPathStartsWith(dirpath):
+        # Remove directories that aren't an ancestor of the included.
+        print 'rm subdirectory: %s' % dirpath
+        shutil.rmtree(dirpath)
 
   if excl is not None:
```
kalman: Maybe we should just delete the excl stuff. AFAIK nobody uses it.

tools/roll.py
```diff
@@ -77,28 +85,41 @@ def rel(subdir):
...
+        path = os.path.abspath(dir)
+        parent, name = os.path.split(path)
+        inclPaths.append(path)
+        inclParentToName[parent] = name
+
+    def inclPathStartsWith(dir):
+      for p in inclPaths:
+        if p.startswith(dir):
+          return True
+      return False
```
kalman: How about:
```python
def inclPathStartsWith(dir):
  return any(p.startswith(dir) for p in inclPaths)
```
in fact at that point it might be more readable to just inline it when you check it.

tools/roll.py
ehalpern: Yeah. Let's do that.

tools/roll.py
ehalpern: Thx for the tip.

PR #3501, opened 2017-05-26
tablePersister refactor

rafael-atticlabs:


go/nbs/table_persister.go
```diff
@@ -69,26 +69,21 @@ func (csbc chunkSourcesByAscendingCount) Less(i, j int) bool {
...
-	swsI, swsJ := csbds[i], csbds[j]
-	if swsI.dataLen == swsJ.dataLen {
-		hi, hj := swsI.source.hash(), swsJ.source.hash()
+	si, sj := csbds[i], csbds[j]
+	if si.chunkDataLen() == sj.chunkDataLen() {
+		hi, hj := si.hash(), sj.hash()
 		return bytes.Compare(hi[:], hj[:]) < 0
 	}
-	return swsI.dataLen > swsJ.dataLen
+	return si.chunkDataLen() > sj.chunkDataLen()
```
cmasone-attic: This winds up recomputing the chunk data length over and over and over as you sort, which is why I didn't do it this way.

go/nbs/table_persister.go
```diff
@@ -69,26 +69,21 @@ func (csbc chunkSourcesByAscendingCount) Less(i, j int) bool {
...
 func (csbds chunkSourcesByDescendingDataSize) Swap(i, j int) { csbds[i], csbds[j] = csbds[j], csbds[i] }
 
-type sourceWithSize struct {
-	source  chunkSource
-	dataLen uint64
-}
-
 type compactionPlan struct {
-	sources             chunkSourcesByDescendingDataSize
+	sources             chunkSources
```
cmasone-attic: What was the issue with having the type of this field express the requirement that this set of chunkSources by sorted in a specific way?

go/nbs/store.go
```diff
@@ -129,15 +149,70 @@ func (nbs *NomsBlockStore) Put(c chunks.Chunk) {
 	nbs.stats.PutLatency.SampleTime(dur)
 }
 
+type byteTableReader struct {
+	tableReader
+	h    addr
+	data []byte
```
cmasone-attic: you can just compose this with a `*bytes.Reader`

```
type byteTableReader struct {
  tableReader
  io.ReaderAt
  h addr
}

br := byteTableReader{h: h, ReaderAt: bytes.NewReader(data)}
br.tableReader = newTableReader(index, br, 1)
```

go/nbs/store.go
```diff
@@ -129,15 +149,70 @@ func (nbs *NomsBlockStore) Put(c chunks.Chunk) {
...
+}
+
+func (br byteTableReader) hash() addr {
+	return br.h
+}
+
+// MUST be called already holding the store lock. Makes durable the existing
+// memtable. Will block if there are already |nbs.persistLimit| concurrent
+// persist operations underway until one has completed.
+func (nbs *NomsBlockStore) persistMemTable() {
```
cmasone-attic: Every time I find myself writing a method like this that relies on someone reading the comment and doing the right thing with locking, I die a little on the inside because I know that someone will hold it wrong and we'll have a horrible, insidious bug that takes us days to track down.

go/nbs/store.go
```diff
@@ -129,15 +149,70 @@ func (nbs *NomsBlockStore) Put(c chunks.Chunk) {
...
+	if nbs.mt == nil || nbs.mt.count() == 0 {
+		return // nothing to do
+	}
+
+	mt := nbs.mt
+	nbs.mt = nil
+
+	h, data, chunkCount := mt.write(nbs.tables, nbs.stats)
```
cmasone-attic: One of the design constraints for `persistingChunkSource` was to do the work of serializing a memTable outside of the storewide lock and, ideally, without holding up the `Put()` that put us over the memTable size limit. Does this new approach make that impossible? That seems like a meaningful degradation to me.

rafael-atticlabs:
After some thought, I'm gonna back away from this change. I think you're right that it doesn't accomplish enough to justify the scope of the change. I'll leave it to you to do this kind of refactoring in this code.

I'll try to post more focused patch to parallelize reads amongst different tables.

Issue #3377, opened 2017-04-11
NBS: Implement caching


rafael-atticlabs:
We are already caching table indices. We can think about some additional directions of caching:

For AWS
- [ ] cache tables (to disk) on write. Especially for boot strapping of processing pipelines, theres a good chance that some reader will immediately read a just-written table
- [ ] cache table indices (to disk) on read.
- [ ] cache chunks (to disk) on read. I don't know if this is possible, but it'd be sweet if we could do something like write the table index to disk, write chunks to the correct table location as they are read, keep a bit-map of preset chunks within the table and have sparse-file support make this not waste too much file space.

For File/AWS
- [ ] keep an in-memory cache of small chunks. Especially blobs tend to share a bunch of very small chunks. It sucks having to go get these repeatedly from the table manager. We'd need to experiment with the correct threshold

PR #3507, opened 2017-05-31
NBS: s3TablePersister caches tables locally on write

### LINKED ISSUE #3506: NBS on AWS: Cache tables to disk on write
Toward #3377, we want to cache tables on disk when writing them to S3.

cmasone-attic:
When running a noms server, it's pretty common for a client to write
data and then have another client want to read it presently. To
optimize for this, s3TablePersister can now cache tables on local disk
up to a certain amount of data.

Toward #3377
Fixes #3506


cmasone-attic:
PTAL. I do want to add more holistic tests at the s3TablePersister layer, but it'll take a bit more time. We could land as-is if we want, and then I can add those tests in a follow-on patch

PR #3509, opened 2017-05-31
NBS: Add more tests for table cache usage in s3TablePersister

cmasone-attic:
These tests verify that tables get cached on Perist(), and that
tableReaders vended by s3TablePersister correctly use the cache.


cmasone-attic:
PTAL

cmasone-attic:
Ping

PR #3510, opened 2017-05-31
NBS: fs table cache should tolerate its own temp files

cmasone-attic:
If one of these got left behind in a cache directory, e.g.
due to a crash, subsequent runs should tolerate and remove it
instead of panicing


cmasone-attic:
Whoops. PTAL

PR #3513, opened 2017-06-02
Marshal: Add support for embedded structs

arv:
The main difference is that we now have to keep track of the index path
into the embedded structs.

Towards #2376


aboodman:
@phritz will be happy to hear about this.

PR #3515, opened 2017-06-02
Noms Prolly Tree Chunking V2

### LINKED ISSUE #2762: Sufficiently long strings as keys in Maps & Sets can cause prolly tree construction to fail
@willhite Discovered this bug constructing maps of `String<String, T>` where the keys are URLs.

The can problem can occur because internal nodes of the prolly tree chunk over the stream of bytes which is the full encoding of each `MetaTuple`, which for Maps & Sets contains the largest value in the referenced sub-tree. In the case of a long string which happens to chunk, a new `MetaTuple` is created for a parent level, and will chunk again on the same encoding of the string, and so on, until the stack explodes.

### LINKED ISSUE #2763: Consider chunking over the compressed stream of value byte serialization
When we moved to chunking over the value byte-serialization for all chunks, we observed that our realized average chunk size can vary quite a bit from the desired size because compression occurs at a layer above chunking. Thus for blob types which are already compressed (e.g. jpg), we observed average chunk size close to the target (4k), but for collections of uncompressed data (e.g. a CSV imported to a List or Map), we tend to see average sizes closer to 2k. This obviously has implications for read performance as quite a bit more chunks would be well below the disk block read size.
#2762 Raises another aspect of this issue, which is that for keyed collections (Map, Set), very long key values will essentially result in a dramatic reduction in the number of values stored in each chunk and thus a much _narrower_ prolly tree.

I ran an experiment of constructing a large map of `String(fmt.Sprintf("%d", i))` = > Number(i) and compared that to prefixing the string with a large repeated substring (similar to a sequence of long urls which vary only in the last few bytes).

As expected, the average chunk size gets roughly cut in half for every 150-200 bytes of string length.

Separately I tested compressing a file of thousands of repeats of a 200 character URL with both zlib and snappy. Surprisingly, snappy fairs well (it compresses to a factor of 10 of zlib, but it does so proportionally).

I'm curious how common b-tree sql engines handle this. I know that leveldb run-length encodes keys and forces a re-start at the block-read-size-boundary.

Chunking over the compressed stream of value bytes is an appealing potential solution because snappy appears to have a sufficiently long "memory" to handle this well enough and its a general solution that would address this problem and also get us back to a more uniform observed average chunk size.

This requires investigation along two directions:

-Understanding snappy enough to know if its safe to chunk over the current output of the compressed stream (i.e. whether it's maintaining an internal buffer of input bytes which _would_ be flushed on close).

-We'd need to re-start the compression context after each boundary, and thus there may be implications for re-chunking correctly.

### LINKED ISSUE #3369: Pass all encoded bytes to rolling value hasher
In order to compensate for non-compressed bytes in the rolling hash, we presently don't pass Ref types. This is a big hack. It was done in https://github.com/attic-labs/noms/commit/fbfdd317fc014cd2a6cc9973f45dd90bf316291a and should be removed if/when #2763 lands

### LINKED ISSUE #3464: Consider encoding prolly trees to directly encode their tree height
This is distinct from Ref height which in some cases will be the same, but cannot be relied upon.

There are number of places we need to dig down into the tree to discover its height (depth), incurring pretty expensive IO.

rafael-atticlabs:
The story of this patch:

Commit 1: Bumps the Noms format version

Commit 2: Reset hash at boundaries

Prior to this patch, a chunk boundary was a function of the N bytes which came before it. Now, it is a function of the N previous bytes *within the current chunk*. This means that when ptrees are rechunked, we don't need to rewind beyond the first byte of the current chunk in order to boot strap the rolling hash.

Separately, when "finalizing" an edit, we now check that the new & old boundaries line up (and we can stop), but observing whether the cursor is in the last item of the current sequence, as opposed to stepping forward and checking whether we're on the first item in next sequence. This change, reduces the number of reads required to mutate by ~1/2.

Commit 3: Encode Tree Level Fixes #3464 

Rather than encode sequences with a single `bool` byte which denotes whether the sequence is leaf or meta, encode the tree level. In practice, since this is varint, it will never be more than a byte (not sure what'd half to happen to end up with a 128 level tree). This saves a bunch of serialized reads when doing things like diffing prolly trees where we need to "level" the two trees.

Commit 4: Add a test for the "infinite chunk" bug

Commit 5: Add a "salt" to the rolling value hash which is provided by the tree level. Fixes #2762 

This reduces (hopefully "enough") the degenerate case of a map key which is longer than the chunk window and happens not to chunk from creating an infinite stack of sequence chunkers. It's still possible with a sufficiently long key, but far less likely.

Commit 6: Raise the rolling hash window to 256. 

This increases the "cascading" probability a bit, but greatly reduces the likelihood that structs with lots of fields cause average chunk size to WAY too big.

Commit 7: Introduce Sloppy

Evil twin of snappy.

Commit 8: Passing sequence bytes into rolling value hash. Fixes #2763, Fixes #3369

Sequences bytes are now passed into sloppy before hitting the rolling hash. The effect is that the *eventual* size on disk after snappy compression will be alot closer to 4k. In practice, this finally gets us something close to LevelDB's run length encoding for keys of kv pairs within a storage chunk. With structs, it means that we get far greater density of instances within a chunk. (SFCrime dataset went from ~15 items/leaf chunk (2k compressed) => ~35 items/leaf chunk (4k compressed).

There is some additional CPU overhead to do this estimation, but the trade off is storing far fewer chunks (SFCrime: ~45%) which saves alot of work in many other places in the stack. Right now, it seems like roughly a wash. I suspect that when we get to lazy value decoding, the net effect will be savings of CPU cycles).

Lastly, the total space on disk is reduced (again, SF Crime dataset, by about %15)

Commit 9 & 10: Fix tests.

Note that sloppy is presently NOT applied to blobs, although I think that's a reasonable thing to investigate: https://github.com/attic-labs/noms/issues/3516

### DIFF
--- a/go/sloppy/sloppy.go
+++ b/go/sloppy/sloppy.go
@@ -1,0 +1,166 @@
+// Copyright 2017 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package types
+
+import "github.com/attic-labs/noms/go/d"
+
+// Sloppy is similar to, and in all ways worse than Google's snappy compression
+// algorithm (https://github.com/google/snappy). However, it has one important
+// property: it prefers copies to closer byte sequences.
+//
+// For example, snappy would logically encode ABCDxABCDyABCDz as:
+//
+//  ["ABCDx", copy(5, 4), "y", copy(10, 4), "z"]
+//
+// while sloppy will encode it as
+//
+//  ["ABCDx", copy(5, 4), "y", copy(5, 4), "z"]
+//
+// This property is useful as sloppy's purpose is to roughly estimate the
+// effect of snappy compressing a given byte sequence, while allowing a
+// change in any byte of the input stream to only effect bytes in the output
+// stream within an N byte window. This allows a byte stream to be fed to
+// sloppy, then into a rolling hash and result in chunk sizes AFTER compression
+// with snappy be to close to a choosen target average.
+//
+// Sloppy is slower and less effective than snappy and its output cannot be
+// decoded. If you are using sloppy for anything other than the purpose above,
+// you are almost certainly doing it wrong.
+
+const (
+	maxOffset    = 1<<12 - 1
+	maxTableSize = 1 << 14
+	maxLength    = 1<<12 - 1
+	tableMask    = maxTableSize - 1
+	shift        = uint32(20)
+)
+
+type Sloppy struct {
+	enc                      encoder
+	idx, maxOffset           int
+	matching                 bool
+	matchOffset, matchLength int
+	table                    [maxTableSize]uint16
+}
+
+// New returns a new sloppy encoder which will encode to |f|. If |f| ever
+// returns false, then encoding ends immediately. |max| is the maximum offset
+// with which a copy will refer back in the input stream.
+func New(f func(b byte) bool, max uint16) *Sloppy {
+	d.PanicIfTrue(max > maxOffset)
+
+	return &Sloppy{
+		binaryEncoder{f},
+		0, int(max),
+		false,
+		0, 0,
+		[maxTableSize]uint16{},
+	}
+}
+
+// Update continues the encoding of a given input stream. The caller is expected
+// to call update after having (ONLY) appended bytes to |src|. When |Update|
+// returns, sloppy will have emitted 0 or more literals or copies by calling
+// the |sf.f|. Note that sloppy will ALWAYS buffer the final three bytes of
+// input.
+func (sl *Sloppy) Update(src []byte) {
+	for ; sl.idx < len(src)-3; sl.idx++ {
+		// TODO: Implement skip heuristic
+
+		// If there are at least four bytes left, compute hash
+		if sl.idx < len(src)-3 {
+			nextHash := fbhash(load32(src, sl.idx), shift)
+			if !sl.matching && sl.idx > 0 {
+				// Look for new match
+				matchPos := int(sl.table[nextHash&tableMask])
+				if sl.idx-matchPos <= sl.maxOffset && src[sl.idx] == src[matchPos] {
+					sl.matching = true
+					sl.matchOffset = matchPos
+					sl.matchLength = 0
+				}
+			}
+
+			// Store new hashed offset
+			sl.table[nextHash&tableMask] = uint16(sl.idx)
+		}
+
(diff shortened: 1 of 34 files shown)


go/sloppy/sloppy.go
```diff
@@ -0,0 +1,166 @@
...
+//
+//  ["ABCDx", copy(5, 4), "y", copy(10, 4), "z"]
+//
+// while sloppy will encode it as
+//
+//  ["ABCDx", copy(5, 4), "y", copy(5, 4), "z"]
+//
+// This property is useful as sloppy's purpose is to roughly estimate the
+// effect of snappy compressing a given byte sequence, while allowing a
+// change in any byte of the input stream to only effect bytes in the output
```
aboodman: effect -> affect

go/sloppy/sloppy.go
```diff
@@ -0,0 +1,166 @@
...
+// stream within an N byte window. This allows a byte stream to be fed to
+// sloppy, then into a rolling hash and result in chunk sizes AFTER compression
+// with snappy be to close to a choosen target average.
```
aboodman: choosen -> chosen

go/types/rolling_value_hasher.go
```diff
@@ -5,18 +5,31 @@
...
+	//    than the window do not chunk and cause average chunk size to be way
+	//    larger than the target
+	// The most likely source of repeated sequences larger than the chunk window
+	// is structs, all of whose field names are encoded together. A window size
+	// of 256 gives a roughly 6% chance that a change within a chunk will cascade
+	// into the next chunk while being sufficiently large to be larger than the
+	// field encodings of a "reasonable" struct.
+	defaultChunkWindow = uint32(256)
+
+	defaultCompressionWindow = uint16(1<<12) - 1 // 4k
```
rafael-atticlabs: Why 4k?

Recall that sloppy prefers the most *recent* possible copy. Consider two cases: a map leaf containing a bunch of small structs: 

S1, S2, S3, S4, S5, S6.... 

and one containing one or two large structs

L1, (L2).

In the first case, the choice of 4k isn't likely to have a significant impact, because each struct will tend to refer to bytes in the struct before it (common field names). Obviously, it's possible to refer farther back, but more likely because of random common "content".

In the second case, the choice of 4k means that there's a much great chance that common content in L2 can refer to bytes in L1 and thus "fit" into the chunk. In practice, this means that larger value stores in maps/sets will still benefit from compression (greater density within the chunk).

PR #3517, opened 2017-06-02
Marshal: Add a way to define the struct name

### LINKED ISSUE #2972: Marshal: Rename Noms struct generated from a Go struct.

arv:
This is done by implementing StructNameMarshaler interface. For example

```go
type TestStruct struct {
  ...
}
func (ts TestStruct) MarshalNomsStructName() string {
  return "HelloWorld"
}
```

Fixes #2972

### DIFF
--- a/go/marshal/encode.go
+++ b/go/marshal/encode.go
@@ -75,7 +75,9 @@ import (
 //   Field int `noms:",omitempty"
 //
 // The name of the Noms struct is the name of the Go struct where the first
-// character is changed to upper case.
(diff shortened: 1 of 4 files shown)


arv:
This diverted a little bit from the plan but I wanted to try it out.

I think it fits better than putting a tag on a random field.

WDYT?

PR #3518, opened 2017-06-05
Add two small features to marshaling

aboodman:
1. In MarshalType() allow Noms collections without element types, e.g.:

type Foo struct {
  A types.List
}

marshal.MustMarshalType(Foo{}) => struct Foo { A List<Value> }

2. There were cases where we would fail to marshal structs containing
non-comparable Go types (e.g. slices). Fix those using DeepEqual().

### DIFF
--- a/go/marshal/encode.go
+++ b/go/marshal/encode.go
@@ -352,7 +352,7 @@ func isEmptyValue(v reflect.Value) bool {
 		return v.Float() == 0
 	case reflect.Struct:
 		z := reflect.Zero(v.Type())
-		return z.Interface() == v.Interface()
+		return reflect.DeepEqual(z.Interface(), v.Interface())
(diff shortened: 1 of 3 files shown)


aboodman:
This makes it possible to use only Go types (no nomdl) in photoindex and many other of our bots.

aboodman:
TBR

arv:
> There were cases where we would fail to marshal structs containing
non-comparable Go types (e.g. slices). Fix those using DeepEqual().

Can you give an example? Maybe add a test too?

aboodman:
Sadly I didn't capture the example. I should have, sorry.

I think it was something like:

```
type Foo struct {
  Bar types.Struct
}
```

Internally, `Bar` contains slices and so this struct couldn't marshal before.

PR #3521, opened 2017-06-06
Chunking V2

### LINKED ISSUE #2762: Sufficiently long strings as keys in Maps & Sets can cause prolly tree construction to fail

### LINKED ISSUE #3464: Consider encoding prolly trees to directly encode their tree height

rafael-atticlabs:
Far various reasons, including both @arv and @kalman being on vacation, and I'm going to hold off on landing all of https://github.com/attic-labs/noms/pull/3515, but I want to land parts of it.

Aside from fixes #3464, fixes #2762, this patch resets the hash context at the beginning of each chunk boundary. This dramatically simplifies the re-chunking logic and thus paves the way implementing streaming updates to prolly trees.

Also, a subtle implementation change in detecting where re-chunking can stop, reduces the number of reads required to re-chunk by ~50%.

Please review, @aboodman, as best you can.

TBR @kalman & @arv 

Note: about the commit marked "struct field encoding". We used to encode structs as a series of field names, then a series of values. Working on #3515, it became clear to me that the encoding needs to prioritize entropy. Encoding all struct fields together does compress better but risks dramatically reducing entropy.


go/types/blob.go
```diff
@@ -295,6 +295,7 @@ func readBlob(r io.Reader, vrw ValueReadWriter) Blob {
 	mtChan := make(chan chan metaTuple, runtime.NumCPU())
 
 	makeChunk := func() {
+		rv.Reset()
```
aboodman: Thinking about this more, I thought the reason for the old behavior was so that changes in one chunk don't propagate very far into downstream chunks.

rafael-atticlabs: It's a long time ago, but I my memory is that we did have that concern, but I think it wasn't well understood. Logically,  this change *reduces* the degree to which changes can cause cascade chunks because a boundary is only dependent on the N previous bytes *since the last boundary. One important change here is that if a change occurs near the end of the boundary, but doesn't move it, we dont *need* to enter the next chunk to see if it changed because it can't have.

The thing to be worried about with this change is the opposite -- that it reduces entropy. The first bytes of a chunk are suspectible to chunking in the same way if a small number of bytes are repeated because the "previous" bytes are effectively all zeros. That's partially why the change to struct encoding is included in this patch (IOW, to reduce the risk that a struct with field name bytes > 64) happens to chunk and causes a leaf level of only 1 element nodes.

In either case, I instrumented attic to report the rate of chunk cascades before & after this change. Since attic isn't deterministic, the numbers aren't stable, but using 1000 photos going through the pipeline, before & after, the cascade rate was ~.8%.

rafael-atticlabs: FWIW, Compressing the rolling hash using the snappy variants, increased the cascade rate (unsurprisingly), which is another part of why I think it makes sense to hold off on that.

I suspect that when we can apply bulk updates to ptrees (and have lazily decoding), the win will be big -- because it increases the value density more than 2x. IOW, we trade a modestly larger cascade rate for dramatically lower read IO when doing bulk edits to ptrees. IOW, I'm waiting until it's an evident win on the work loads we have.

PR #3523, opened 2017-06-06
narrow semantics of value_store value_cache

rafael-atticlabs:


cmasone-attic:
If there's a better name for `valueCache`, now, that indicates it's just there to allow short-circuiting of chunk decoding (as opposed to providing stronger indications of chunk-persistence) that would be ideal.

LGTM

PR #3524, opened 2017-06-07
move flushing of uncommitted ptree nodes out of EncodeValue

rafael-atticlabs:
Ok, here's step one. 

Separate the concern of flushing uncommitted (linked in-memory) ptree nodes from encoding.

Towards #3525

### DIFF
--- a/go/types/meta_sequence.go
+++ b/go/types/meta_sequence.go
@@ -342,7 +342,7 @@ func (es emptySequence) isLeaf() bool {
 
 // Invokes |cb| on all "uncommitted" ptree nodes reachable from |v| in
 // "bottom-up" order. It will only follow refs in the case of an in-memory
-// reference to an uncommitted child. Note that |v| maybe not neccesariby be
+// reference to an uncommitted child. Note that |v| maybe not necessarily be
 // a collection, but a collection with uncommitted children maybe be reachable
 // from it (e.g. a Struct).
 func iterateUncommittedChildren(v Value, cb func(sv Value)) {
--- a/go/types/meta_sequence.go
+++ b/go/types/meta_sequence.go
@@ -339,3 +339,35 @@ func (es emptySequence) getCompositeChildSequence(start uint64, length uint64, h
 func (es emptySequence) isLeaf() bool {
 	return es.height == 1
 }
+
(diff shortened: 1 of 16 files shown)


go/types/meta_sequence.go
```diff
@@ -339,3 +339,31 @@ func (es emptySequence) getCompositeChildSequence(start uint64, length uint64, h
...
+					cb(mt.child)
+				}
+			}
+			return
+		}
+	}
+
+	v.WalkValues(func(v Value) {
+		iterateUncommittedChildren(v, cb)
+	})
```
cmasone-attic: It's surprising that you do this here at the end. The comment, and function name, lead me to believe this function is all about visiting uncommitted children in prolly trees, which you explicitly handle above. What's this here for? I'm sure it's to cover some necessary case, (maybe explicit Refs in collections or structs...?) but an example in a comment would help me understand.

go/types/opcache.go
```diff
@@ -335,10 +335,10 @@ func decodeValue(bs []byte, asValue bool, vr ValueReader) ([]byte, Value) {
 }
 
 // Note that, if 'v' are prolly trees, any in-memory child chunks will be written to vw at this time.
```
cmasone-attic: Is this comment still true? I'm not entirely sure how it could be, and that would be a problem, no?

go/types/meta_sequence.go
rafael-atticlabs: Yeah. I didn't at first and it didn't work. The issue is that You can nest collections that have uncommitted values within other values (i.e. a struct) and you need to find those. So you need to recursively walk the value, but only "leave" the current chunk when traversing an uncommitted ref. I'll add a comment.

go/types/opcache.go
rafael-atticlabs: good catch. i think that probably does need to flush uncommitted child nodes as well.

rafael-atticlabs:
Ok. PTAL. I basically needed to undo edits to a few files here, so I flattened the commits again. The only changes since the last commit are in opcache and comments on iterateUncommittedChildren

cmasone-attic:
lgtm

Issue #3519, opened 2017-06-05
Remove `last` arg from ChunkStore.Commit()


rafael-atticlabs:
There's a comment in chunk_store.go asking if the last arg is necessary any more. It seems like it isn't.

rafael-atticlabs:
fixed in https://github.com/attic-labs/noms/commit/214054986baa0bdea0db48e62740da932c82fb3c

PR #3527, opened 2017-06-08
Enforce clearer concurrency semantics of ValueStore

### LINKED ISSUE #3436: Consider adding `ValueStore.Root()/Commit()`
Backing up,

`ChunkStore`
-deals in chunks
-is versioned
-an instance implements snapshot isolation for its client
-maybe it should enforce that the `root` hash references a contained chunk

`ValueStore`
-depends on ChunkStore for its versioning/isolation
-deals in values (knows how to convert chunks<->values>
-enforces chunk & value validity (i.e. chunks are properly encoded values, no dangling references, prolly trees are canonical)
-attempts to write chunks to the lower level with "good" locality, for efficient read
-*doesn't care what value the root references*

`Database`
-depends on ValueStore for versioning/isolation/validity/locality
-Enforces a particular data model, i.e.
  -The root references a `Map<String, Ref<Commit>>`
  -Commits are a particular shape
  -Manages concurrent writes to commit lineages

Put another way: You can imagine someone using noms only at the `ValueStore` level via values and possibly enforcing a different root data model or history model.

To me this implies that ValueStore should have `Root()` and `Commit()`, rather than `Flush()`. Database would simply validate the root object and then call its ValueStore.Commit().

From where we are, this isn't a big change.

cc: @cmasone-attic, @aboodman

### LINKED ISSUE #3525: Database's threadsafety is difficult to reason about
Largely because it's built out of a stack of 3 or 4 layers below, all of which are doing their own locking and it's depending on that.

The dangling references issues we are seeing are likely the result of a race resulting from concurrent use of a single database.

The rough shape of a solution here is: https://github.com/attic-labs/noms/issues/3436

rafael-atticlabs:
In this patch, ValueStore becomes a rootTracker and completeness checking becomes an internal concern of it.

Fixes #3525, Fixes #3436, #3519


rafael-atticlabs:
While I was in this code, I went ahead and fixed #3519 in commit 3.

rafael-atticlabs:
@cmasone-attic Probably easiest to review commit-wise

rafael-atticlabs:
Also, note that `pull` no longer has a completeness checker in the case of a local chunk store, but it occurred to me that the current pull algorithm is, *itself* a completeness checker, so if it finishes and commits, the resulting value will be complete. If it fails, the ChunkStore atomicity ensures that nothing will committed to the store.

go/datas/database_common.go
```diff
@@ -28,13 +28,18 @@ var (
...
+	vs := types.NewValueStore(cs)
+	if _, ok := cs.(*httpChunkStore); ok {
+		vs.SetEnforceCompleteness(false)
+	}
+
 	return &database{
-		ValueStore: types.NewValueStore(cs), // ValueStore is responsible for closing |cs|
-		rt:         cs,
+		ValueStore: vs, // ValueStore is responsible for closing |cs|
+		rt:         vs,
```
cmasone-attic: Since `vs` is now a rootTracker, we can just get rid of `rt`, yes?

go/nbs/store.go
```diff
@@ -375,7 +375,7 @@ func (nbs *NomsBlockStore) Commit(current, last hash.Hash) bool {
 		Jitter: true,
 	}
 	for {
-		if err := nbs.updateManifest(current, last); err == nil {
+		if err := nbs.updateManifest(current); err == nil {
 			return true
 		} else if err == errOptimisticLockFailedRoot || err == errLastRootMismatch {
```
cmasone-attic: Can probably get rid of `errLastRootMismatch` now?

go/types/value_store.go
```diff
@@ -284,14 +303,29 @@ func (lvs *ValueStore) bufferChunk(v Value, c chunks.Chunk, height uint64) {
...
+func (lvs *ValueStore) Rebase() {
+	lvs.cs.Rebase()
+}
+
 // Flush() puts all bufferedChunks into the ChunkStore, with best-effort
 // locality. NB: The Chunks will not be made durable unless the caller also
 // Commits to the underlying ChunkStore.
-func (lvs *ValueStore) Flush() {
-	func() {
+func (lvs *ValueStore) Commit(current hash.Hash) bool {
```
cmasone-attic: Update the comment on this method

go/types/value_store.go
```diff
@@ -284,14 +303,29 @@ func (lvs *ValueStore) bufferChunk(v Value, c chunks.Chunk, height uint64) {
...
+	return func() bool {
 		lvs.bufferMu.Lock()
 		defer lvs.bufferMu.Unlock()
 
+		checkCurrent := false
```
cmasone-attic: it's a little weird to frontload this check when you don't need to know this until way down below. Indeed, you only need to know this if you're checking completeness.

PR #3528, opened 2017-06-08
Remove WalkDifferentStructs

rafael-atticlabs:
The one user we had for this is now doing its work different. Remove dead code.


rafael-atticlabs:
TBR @cmasone-attic

Issue #3491, opened 2017-05-22
Add `root` hash to `/getRefs` wire API


rafael-atticlabs:
Right now, for every request to `/getRefs`, we are forced to re-open or rebase a database. For NBS/AWS, this means a strongly consistent read against dynamo which costs money and adds ~10ms to the total latency.

There are still code paths that do lots of sequential reads of 1 chunk. For those paths, this is probably doubling the total latency for the AWS backend.

cmasone-attic:
per discussion, I think this is actually going to be the root hash

PR #3529, opened 2017-06-08
Add chunks.Factory::CreateStoreFromCache()

cmasone-attic:
Add a method to chunks.Factory that allows the caller to
signal that it's willing to try to make forward progress
using an out-of-date ChunkStore. This allows AWSStoreFactory
and LocalStoreFactory to vend NBS instances without hammering
persistent storage every time.

Towards #3491

### DIFF
--- a/go/chunks/chunk_store.go
+++ b/go/chunks/chunk_store.go
@@ -67,6 +67,10 @@ type ChunkStore interface {
 type Factory interface {
 	CreateStore(ns string) ChunkStore
 
+	// CreateStoreFromCache allows caller to signal to the factory that it's
+	// willing to tolerate an out-of-date ChunkStore.
+	CreateStoreFromCache(ns string) ChunkStore
+
 	// Shutter shuts down the factory. Subsequent calls to CreateStore() will fail.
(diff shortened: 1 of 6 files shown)


go/chunks/chunk_store.go
```diff
@@ -67,6 +67,10 @@ type ChunkStore interface {
 type Factory interface {
 	CreateStore(ns string) ChunkStore
 
+	// CreateStoreFromCache allows caller to signal to the factory that it's
+	// willing to tolerate an out-of-date ChunkStore.
+	CreateStoreFromCache(ns string) ChunkStore
```
rafael-atticlabs: I think this makes sense for now. It's kind of interesting to think about this contract. The caller of this always has the ability to force a read from "head" via Rebase. You could imagine there being one call which returns you a chunk store and if you care about having reflect the underlying truth as of now, you need to call Rebase() on it yourself.

cmasone-attic: Yeah. This isn't particularly elegant, but I figure we aren't yet firm on exactly what contract we want here yet.

PR #3533, opened 2017-06-09
Remove some uses of Backoff

### LINKED ISSUE #3520: Audit uses of Backoff
I see at least one that looks correct (backing off when we get a throttle event from s3 or aws), but others now seem a bit fishy. I'm sure this is my fault, but looking at them now, I think maybe I was thinking it was addressing a worry about "liveness" (one writers failing to make progress), but I'm no longer convinced that's the case.

rafael-atticlabs:
Fixes #3520

### DIFF
--- a/go/nbs/store.go
+++ b/go/nbs/store.go
@@ -14,7 +14,6 @@ import (
 	"github.com/attic-labs/noms/go/constants"
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
-	"github.com/jpillora/backoff"
 )
 
 // The root of a Noms Chunk Store is stored in a 'manifest', along with the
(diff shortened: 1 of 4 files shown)


rafael-atticlabs:
I think going forward, we should document at the use-site of Backoff what problem we think it's solving.

go/nbs/store.go
```diff
@@ -381,19 +380,12 @@ func (nbs *NomsBlockStore) Commit(current, last hash.Hash) bool {
 		return true
 	}
 
-	b := &backoff.Backoff{
```
cmasone-attic: Should we do something here where we somehow pipe up a backoff error from Dynamo if we get one, and respond appropriately?

rafael-atticlabs: I'm not sure what happens in that case. My impression is that we get transparently throttled. If we also get errors, then it may be reasonable to back off.

https://github.com/attic-labs/noms/issues/3535

cmasone-attic: No, we get back errors saying that we got throttled

cmasone-attic: at least I think so...that definitely happens with batch gets. I thought the same happened with GetItem or whatever

cmasone-attic: Yeah, you can get a "ProvisionedThroughputExceededException"

rafael-atticlabs: Well, i think #3535 is a good place to address that. It was also kind of unfortunate that we'd backoff in the case of getting a manifest-cache oplock failure.

PR #3538, opened 2017-06-09
Put root in httpChunkStore /getRefs query string

### LINKED ISSUE #3491: Add `root` hash to `/getRefs` wire API

cmasone-attic:
Fixes #3491


cmasone-attic:
PTAL

PR #3545, opened 2017-06-15
Streaming p tree updates + Map Editor

rafael-atticlabs:
@kalman, please TBR the re-sequencing part of this patch

@aboodman, please review now MapEditor

### DIFF
--- a/go/types/map_editor.go
+++ b/go/types/map_editor.go
@@ -1,0 +1,168 @@
+// Copyright 2017 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package types
+
+import (
+	"sort"
+
+	"github.com/attic-labs/noms/go/d"
+)
+
+// MapEditor allows for efficient editing of Map-typed prolly trees. Edits
+// are buffered to memory and can be applied via Build(), which returns a new
+// Map. Prior to Build(), Get() & Has() will return the value that the resulting
+// Map would return if it were built immediately prior to the respective call.
+// Note: The implementation biases performance towards a usage which applys
+// edits in key-order.
+type MapEditor struct {
+	m          Map
+	edits      mapEntrySlice // edits may contain duplicate key values, in which case, the last edit of a given key is used
+	normalized bool
+}
+
+func NewMapEditor(m Map) *MapEditor {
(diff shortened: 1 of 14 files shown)


go/types/map_editor.go
```diff
@@ -0,0 +1,168 @@
...
+	"sort"
+
+	"github.com/attic-labs/noms/go/d"
+)
+
+// MapEditor allows for efficient editing of Map-typed prolly trees. Edits
+// are buffered to memory and can be applied via Build(), which returns a new
+// Map. Prior to Build(), Get() & Has() will return the value that the resulting
+// Map would return if it were built immediately prior to the respective call.
+// Note: The implementation biases performance towards a usage which applys
```
aboodman: applies

go/types/map_editor.go
```diff
@@ -0,0 +1,168 @@
...
+		return me.m // no edits
+	}
+
+	vr := me.m.sequence().valueReader()
+
+	me.normalize()
+
+	var ch *sequenceChunker
+	for i, kv := range me.edits {
+		if i+1 < len(me.edits) && me.edits[i+1].key.Equals(kv.key) {
```
aboodman: Did you consider keeping the edits sorted and only keeping the last of each entry? It seems like same amount of work for the computer, less memory, and reduce a lot of the code complexity.

aboodman:
As long as you're in here can you also please add an `Edit() MapEditor` convenience to `types.Map`.

rafael-atticlabs:
Comment fixed. `Edit()` added (good call, much nicer). 

Re: duplicate entries: My goal was that in- and out-of-order edits would be constant time. Keeping the edits sorted would make out-of-order edits Log(N). Additional complexity feels worth it to me. The algorithmic trade-off is that reads *can* be N Log (N), but I think the editor should bias towards editing =-).

PTAL. I'm gonna add a bit more testing, but I'd like to land quickly. Also, note that I *removed* the existing mutating API on Map.

aboodman:
- It's log-n in an in-memory array, meaning it's ~instant. The only cost here is the serialization of the noms object so that you can use `Less()` and that's happening in your approach too.

- Although the interface is called `Editor`, in practice it will be used for reading quite frequently ... probably just as frequently as editing (perhaps a better name would be `MutableMap`, though it doesn't quite roll off the tongue the same).

PR #3552, opened 2017-06-20
Add fsTableCache tests missed in previous patch

cmasone-attic:


cmasone-attic:
I'm pretty sure this is a file of tests I forgot to `git add` at some point. PTAL

PR #3553, opened 2017-06-20
sequenceChunker.advanceTo avoid resuming over items it already consumed

rafael-atticlabs:
When sequence chunker advances to a new position for streaming edits, it needs to ensure that it doesn't `resume()` over items which have already been consumed at this level.

The specific case that trigger this was making two edits to map, both of which took place within the same leaf sequence. this case can only arise when the final item in the map is an "explicit" boundary. If the first edit doesn't "change" the boundary (i.e. it will still cause the rolling hash to hit a target pattern), and the second edit is trying to append beyond the last existing element, then what happened was the code *thought* that the `next` cursor hadn't been reached and thus "resumed" at that level, but that caused it to re-consume items that were already consumed.

### DIFF
--- a/go/types/map.go
+++ b/go/types/map.go
@@ -308,8 +308,12 @@ func makeMapLeafChunkFn(vr ValueReader) makeChunkFn {
 		d.PanicIfFalse(level == 0)
 		mapData := make([]mapEntry, len(items), len(items))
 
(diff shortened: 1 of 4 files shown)


rafael-atticlabs:
TBR @kalman

PR #3556, opened 2017-06-21
NBS: Fix usage of manifestCache

### LINKED ISSUE #3551: NBS: manifestCache usage is racy
Reading the manifest from persistent storage and putting the parsed contents into the manifestCache is not currently atomic. This is leading to problems where ParseIfExists() read manifest contents that are out-of-date by the time they're put into the cache. This means they clobber more up-to-date information and can lead to tables "disappearing" from the POV of databases in process.

FYI @rafael-atticlabs

cmasone-attic:
There were races in both manifest implementations, between cache usage
in ParseIfExists() and Update(). It was possible for Update() calls to
occur between the time that ParseIfExists() read the manifest data and
the time it updated the cache. This led to newer manifest content
getting clobbered in the cache by out-of-date manifest content.

The new approach takes caching out of the hands of the manifest impls,
instead wrapping a shared caching layer around the outside.

Fixes #3551

### DIFF
--- a/go/nbs/store.go
+++ b/go/nbs/store.go
@@ -32,18 +32,21 @@ const (
 )
 
 var (
-	cacheOnce           = sync.Once{}
-	globalIndexCache    *indexCache
-	globalManifestCache *manifestCache
-	globalFDCache       *fdCache
-	globalConjoiner     conjoiner
+	cacheOnce        = sync.Once{}
+	globalIndexCache *indexCache
+	globalAddCache   func(manifest) manifest
+	globalFDCache    *fdCache
+	globalConjoiner  conjoiner
 )
 
 func makeGlobalCaches() {
 	globalIndexCache = newIndexCache(defaultIndexCacheSize)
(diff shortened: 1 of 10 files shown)


cmasone-attic:
PTAL

go/nbs/store.go
```diff
@@ -32,18 +32,21 @@ const (
...
 
 func makeGlobalCaches() {
 	globalIndexCache = newIndexCache(defaultIndexCacheSize)
-	globalManifestCache = newManifestCache(defaultManifestCacheSize)
 	globalFDCache = newFDCache(defaultMaxTables)
 	globalConjoiner = newAsyncConjoiner(defaultMaxTables)
+
+	cacheMu := &sync.Mutex{}
+	manifestCache := newManifestCache(defaultManifestCacheSize)
+	globalAddCache = func(mm manifest) manifest { return cachingManifest{mm, cacheMu, manifestCache} }
```
rafael-atticlabs: globalAddCache => createCachingManifest()?

PR #3557, opened 2017-06-21
Introduce SetEditor, Nestable Editors

### LINKED ISSUE #3561: Need a mechanism to defer edits to graphs of noms data and commit when edits are complete.

rafael-atticlabs:
Tests for nested editors coming. Please review now.

@aboodman 

cc @kalman 

Fixes #3561

### DIFF
--- a/go/types/map_editor.go
+++ b/go/types/map_editor.go
@@ -32,10 +32,10 @@ func (me *MapEditor) Kind() NomsKind {
 }
 
 func (me *MapEditor) Value(vrw ValueReadWriter) Value {
-	return me.Build(vrw)
+	return me.Map(vrw)
 }
 
-func (me *MapEditor) Build(vrw ValueReadWriter) Map {
+func (me *MapEditor) Map(vrw ValueReadWriter) Map {
 	if len(me.edits) == 0 {
 		return me.m // no edits
 	}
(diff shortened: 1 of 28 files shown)


aboodman:
Title of PR is wrong.

go/types/set_editor.go
```diff
@@ -0,0 +1,207 @@
...
+
+func (se *SetEditor) Kind() NomsKind {
+	return SetKind
+}
+
+func (se *SetEditor) build(vrw ValueReadWriter) Value {
+	return se.Build(vrw)
+}
+
+func (se *SetEditor) Build(vrw ValueReadWriter) Set {
```
aboodman: What do you think about calling this `Set()` (and `Map()` in the case of `MapEditor`). It is more parallel with the `Value()` method which does the same thing.

go/types/map_editor.go
```diff
@@ -38,14 +47,56 @@ func (me *MapEditor) Build(vrw ValueReadWriter) Map {
...
+				continue // next edit supercedes this one
+			}
+
+			// TODO: Use ReadMany
+			cc := make(chan *sequenceCursor, 1)
+			cursChan <- cc
+
+			go func(k Value, cc chan *sequenceCursor) {
+				cc <- newCursorAtValue(me.m.seq, k, true, false, false)
+			}(edit.key, cc)
```
aboodman: There is no need to pass `cc` -> it's already local to each time around the loop. And wrt `edit.key`, is more idiomatic in Go to reassign to a local:

```
for .... {
  edit := edit
  ...
}
```

rafael-atticlabs: done

go/types/set_editor.go
rafael-atticlabs: ok

PR #3571, opened 2017-06-27
Fix httpChunkStore race between Close() and Has/Get

### LINKED ISSUE #3566: Race in HttpChunkStore
I *think* that a race exists which looks like

-Threads A and B are reading using the same chunk store
-Thread A calls Close() while there are still outstanding reads
-It closes `hcs.finishedChan` and the blocks on `requestWg`
-Before existing requests can finish, thread B calls `Get()` which increments `requestWg` and then sends a request to `getQueue`
-The worker thread in `batchReadRequests` exits on `finishedChan` having been closed

Then no one is home to decrement `requestWg` Thread A blocks forever on Close()

cmasone-attic:
In httpChunkStore, calls to Get/Has and friends put a request object
with a 'return channel' onto a queue channel, and then block on the
return channel. The queue channel was buffered, which made it
impossible to cause Get, Has et al to terminate reliably when the
store was closed.

This patch removes the buffering on the channel so we can
deterministically bail from Get/Has et al when closing the store. I
don't think we were actually seeing any benefit from the buffer on the
queue channels, because everywhere we write to one of them we
immediately block on another channel, waiting for the result of the
request.

Fixes #3566

### DIFF
--- a/go/datas/http_chunk_store.go
+++ b/go/datas/http_chunk_store.go
@@ -235,15 +235,7 @@ func (hcs *httpChunkStore) batchReadRequests(queue <-chan chunks.ReadRequest, ge
 			case <-hcs.finishedChan:
 				done = true
 			}
-			/*	// Drain queue before returning
-				select {
-				case req := <-queue:
-					hcs.sendReadRequests(req, queue, getter)
-				default:
-					//drained!
-				}*/
 		}
-
 	}()
 }
 
--- a/go/datas/http_chunk_store.go
+++ b/go/datas/http_chunk_store.go
@@ -29,7 +29,6 @@ import (
 
 const (
 	httpChunkSinkConcurrency = 6
-	writeBufferSize          = 1 << 12 // 4K
 	readBufferSize           = 1 << 12 // 4K
 )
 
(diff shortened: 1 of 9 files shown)


cmasone-attic:
PTAL

go/datas/http_chunk_store.go
```diff
@@ -29,7 +29,6 @@ import (
 
 const (
 	httpChunkSinkConcurrency = 6
-	writeBufferSize          = 1 << 12 // 4K
 	readBufferSize           = 1 << 12 // 4K
```
rafael-atticlabs: looks like writeBufferSize *wasn't* used. Is *readBufferSize* still?

go/datas/http_chunk_store.go
```diff
@@ -215,46 +222,44 @@ func (hcs *httpChunkStore) batchHasRequests() {
...
 				done = true
 			}
-			// Drain queue before returning
-			select {
-			case req := <-queue:
-				hcs.sendReadRequests(req, queue, getter)
-			default:
-				//drained!
-			}
+			/*	// Drain queue before returning
```
rafael-atticlabs: ?

cmasone-attic: done

go/datas/http_chunk_store.go
cmasone-attic: done

cmasone-attic:
Put/Commit/Close race bug https://github.com/attic-labs/noms/issues/3572

PR #3575, opened 2017-06-28
NBS: updateManifest() fails fast if Update is DOOOOOOMED

### LINKED ISSUE #3574: NBS: Don't persist memTable if manifest.Update() is going to fail
Obvs we can only do this by loading the manifest again (slow and expensive) or getting a less-certain answer by querying the in-process manifest cache

cmasone-attic:
If NomsBlockStore can assume that its manifest is a cachingManifest,
it can pre-emptively check to see if someone else in-process has
already moved the manifest forward and, if so, fail early.

Fixes #3574

### DIFF
--- a/go/nbs/store.go
+++ b/go/nbs/store.go
@@ -34,7 +34,7 @@ const (
 var (
 	cacheOnce           = sync.Once{}
 	globalIndexCache    *indexCache
-	makeCachingManifest func(manifest) manifest
+	makeCachingManifest func(manifest) cachingManifest
 	globalFDCache       *fdCache
 	globalConjoiner     conjoiner
 )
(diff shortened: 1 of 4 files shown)


cmasone-attic:
PTAL

go/nbs/store.go
```diff
@@ -405,6 +405,18 @@ func (nbs *NomsBlockStore) updateManifest(current, last hash.Hash) error {
 		return errLastRootMismatch
 	}
 
+	if upstream, doomed := nbs.mm.updateWillFail(nbs.manifestLock); doomed {
+		// Pre-emptive optimistic lock failure. Someone else in-process moved to the root, the set of tables, or both out from under us.
```
rafael-atticlabs: maybe lift this into a local `handleOptimisticLockFailure := func() {...}` that can be used below as well?

PR #3576, opened 2017-06-29
add store-granularity locking to manifest cache

rafael-atticlabs:
Prior to this patch, the manifest cache was effectively serializing all IO to the underlying manifest across a factory (i.e. multiple calls to Rebase() of different stores). This allows IO to different stores to proceed concurrently.

### DIFF
--- a/go/nbs/manifest.go
+++ b/go/nbs/manifest.go
@@ -7,7 +7,6 @@ package nbs
 import (
 	"crypto/sha512"
 	"strconv"
-	"sync"
 
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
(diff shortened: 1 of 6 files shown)


go/nbs/manifest.go
```diff
@@ -67,13 +66,10 @@ func (mc manifestContents) size() (size uint64) {
 
 type cachingManifest struct {
 	mm    manifest
-	mu    *sync.Mutex
 	cache *manifestCache
 }
 
 func (cm cachingManifest) updateWillFail(lastLock addr) (cached manifestContents, doomed bool) {
```
cmasone-attic: Why is there no locking on this, now?

go/nbs/manifest_cache.go
```diff
@@ -27,18 +31,49 @@ type manifestCache struct {
...
+
+func (mc *manifestCache) Unlock(db string) {
+	mc.cond.L.Lock()
+	defer mc.cond.L.Unlock()
+
+	_, ok := mc.locked[db]
+	d.PanicIfFalse(ok)
+	delete(mc.locked, db)
+
+	mc.cond.Broadcast()
```
cmasone-attic: It's unfortunate that we wake up everyone who's waiting on this every time when we know that only goroutines that were waiting on the lock for |db| actually have a chance of making progress :-/

go/nbs/manifest.go
rafael-atticlabs: It's protected within `cache.Get`

go/nbs/manifest_cache.go
rafael-atticlabs: yup. don't have a better idea, though

cmasone-attic: One condition var per DB?

rafael-atticlabs: yeah, but then you need a master mutex to protect the map of condition vars.

PR #3579, opened 2017-06-29
fix manifest cache race

rafael-atticlabs:
It occurred to me that the timestamp here needs to be *pessimistic* and use the time prior to initiating the IO call in order to avoid a race.


cmasone-attic:
LGTM

PR #3584, opened 2017-07-12
NBS: Prevent concurrent loading of the same table index

### LINKED ISSUE #3581: NBS: Modify indexCache usage to prevent concurrent loading of indices
This can lead to extra memory usage when the system is under memory pressure. Indices get tossed out of the cache, then multiple clients call in about the same DB, and the same indices get loaded concurrently. That means we try to keep multiple copies of some indices in memory at the same time, which can be a problem if the DB (and thus the indices) is very large

cmasone-attic:
By adding per-entry locking to indexCache, this patch allows both NBS
storage backends to ensure that only one goroutine is loading a given
table index at a time. Previoudly, if multiple readers tried to open
the same table at the same time, it was possible for both to get a
cache-miss and then read in the table index at the same time, which
wastes memory.  Now, one will win the race, load the index, and then
the other will see a cache hit.

Fixes #3581

### DIFF
--- a/go/nbs/file_table_persister.go
+++ b/go/nbs/file_table_persister.go
@@ -46,6 +46,8 @@ func (ftp *fsTablePersister) persistTable(name addr, data []byte, chunkCount uin
 		io.Copy(temp, bytes.NewReader(data))
 		index := parseTableIndex(data)
 		if ftp.indexCache != nil {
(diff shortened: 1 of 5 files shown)


cmasone-attic:
PTAL

Issue #3562, opened 2017-06-21
Implement StructEditor


rafael-atticlabs:

rafael-atticlabs:
I'm going to declare this bug closed since implementing ListEditor is really an optimization from here.

PR #3586, opened 2017-07-12
List Editor

rafael-atticlabs:
This should look somewhat familiar =-).

Towards #3562. This removes the mutative API from List and adds a ListEditor which is logically consistent with the existing Set/MapEditor. BlobEditor is next and should look similar to ListEditor.

### DIFF
--- a/go/types/list_editor.go
+++ b/go/types/list_editor.go
@@ -1,0 +1,293 @@
+// Copyright 2017 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package types
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/attic-labs/noms/go/d"
+)
+
+type ListEditor struct {
+	l     List
+	edits *listEdit
+}
+
+func NewListEditor(l List) *ListEditor {
+	return &ListEditor{l, nil}
+}
+
+func (le *ListEditor) Kind() NomsKind {
+	return ListKind
+}
+
+func (le *ListEditor) Value(vrw ValueReadWriter) Value {
+	return le.List(vrw)
+}
+
+func (le *ListEditor) List(vrw ValueReadWriter) List {
+	if le.edits == nil {
+		return le.l // no edits
+	}
+
+	seq := le.l.sequence()
+	vr := seq.valueReader()
+
+	if vrw != nil {
+		vr = vrw
+	}
+
+	cursChan := make(chan chan *sequenceCursor)
+	spliceChan := make(chan chan listEdit)
+
+	go func() {
+		for edit := le.edits; edit != nil; edit = edit.next {
+			edit := edit
+
+			// TODO: Use ReadMany
+			cc := make(chan *sequenceCursor, 1)
+			cursChan <- cc
+
+			go func() {
+				cc <- newCursorAtIndex(seq, edit.idx, false)
+			}()
+
+			sc := make(chan listEdit, 1)
+			spliceChan <- sc
+
+			wg := sync.WaitGroup{}
+			subEditors := false
+			for i, v := range edit.inserted {
+				if _, ok := v.(Value); ok {
+					continue
+				}
+
+				subEditors = true
+				idx := i
+				wg.Add(1)
+				go func() {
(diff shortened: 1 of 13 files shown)


go/types/list_editor.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+
+		dc := sp.removed
+		for dc > 0 {
+			ch.Skip()
+			dc--
+		}
+
+		for _, v := range sp.inserted {
+			if emp, ok := v.(Emptyable); ok && emp.Empty() {
```
arv: Why `Emptyable` instead of `emptyable`. Does not seem like a lot of value exporting it.

go/types/list_editor.go
```diff
@@ -0,0 +1,293 @@
...
+	vr := seq.valueReader()
+
+	if vrw != nil {
+		vr = vrw
+	}
+
+	cursChan := make(chan chan *sequenceCursor)
+	spliceChan := make(chan chan listEdit)
+
+	go func() {
```
arv: To be honest... this channel juggling is too hard for me. I trust you are getting it right.

go/types/list_editor.go
```diff
@@ -0,0 +1,293 @@
...
+				cc <- newCursorAtIndex(seq, edit.idx, false)
+			}()
+
+			sc := make(chan listEdit, 1)
+			spliceChan <- sc
+
+			wg := sync.WaitGroup{}
+			subEditors := false
+			for i, v := range edit.inserted {
+				if _, ok := v.(Value); ok {
```
arv: When can `ok` be `false` here? `edit.inserted` is `[]Valuable`

go/types/list_editor.go
```diff
@@ -0,0 +1,293 @@
...
+					continue
+				}
+
+				subEditors = true
+				idx := i
+				wg.Add(1)
+				go func() {
+					edit.inserted[idx] = v.Value(vrw)
```
arv: This is strange. We continued the loop if we had `.Value`

go/types/list_editor.go
arv: Not clear why this is here. What happens if I try to add an empty list to a list?

go/types/list_editor.go
rafael-atticlabs: yup.

go/types/list_editor.go
rafael-atticlabs: here, we have a `Valuable` so we need to ask it to produce a `Value`

go/types/list_editor.go
rafael-atticlabs: Editors are *nestable*. We decided to start with the most common behavior which is that if a nested thing is empty, you probably want to remove it. In Attic, this was how all clients behaved. There's an abstract plan to make this behavior configurable per-editor, but for now it's just *the behavior*.

rafael-atticlabs: This comes up when you are mutating graphs and you discover that some collection *has become empty*. All of examples we had, removed the empty container in that case. Very strange bugs could arise if you didn't do that (at least in those cases).

go/types/list_editor.go
arv: So isn't this always true?

go/types/list_editor.go
arv: I think I don't understand what edit.inserted is. I thought it was a Value but I guess it is a `Valuable`?

arv: Also, should `Valuable` be `valueble`?

arv: How can we get here when we continue the loop on line 65?

go/types/list_editor.go
rafael-atticlabs: No. Basically, if it's a Valuable, then it's an Editor and we need to build it in its own thread, if it's a Value, then we can just pass it along (because it's already built).

go/types/list_editor.go
rafael-atticlabs: Essentially if `v` is an Editor. I realize is confusing. The thing that's strange is that `Value` is a `Valuable`, but it kind of had to be that way.

arv:
I'm out today. Feel free to merge. I still have questions but we can deal
with them tomorrow.

rafael-atticlabs:
Ok. Merging. Let's continue to discuss -- maybe Monday.

Issue #3590, opened 2017-07-15
Consider using log.Printf rather than fmt.Printf for verbose messages


ehalpern:
Noms code uses fmt.Printf (by way of verbose.Log) to print notable events.

Unfortunately, this makes it difficult for a program that embeds noms to control how these messages are logged.

Using log.Printf rather than fmt.Printf routes the message through the go standard logger which solves this problem.

ehalpern:
cc: @rafael-atticlabs, @cmasone-attic

cmasone-attic:
I vaguely remember wanting to do that, but we didn't for some reason. Can't
remember if it was some actual concern or just a "eh, we have no reason to
change and we're already doing the other..."

PR #3591, opened 2017-07-17
Use log.PrintXXX rather than fmt.PrintXXX for informational messages

ehalpern:
All messages emitted from non-cli code are now output through the standard
logger rather than directly to stdout. This gives code that embeds noms control
over how messages are logged (using log.SetFlags and log.SetOutput).

When the noms cli is used, the logger is initialized so that it only prints the
string logged. In other cases, the log setup is left to the embedding code.

toward: #3590

### DIFF
--- a/go/datas/database_server.go
+++ b/go/datas/database_server.go
@@ -6,6 +6,7 @@ package datas
 
 import (
 	"fmt"
+	"log"
 	"net"
 	"net/http"
 	"strconv"
@@ -56,7 +57,7 @@ func (s *RemoteDatabaseServer) Run() {
 	d.Chk.NoError(err)
 	s.port, err = strconv.Atoi(port)
 	d.Chk.NoError(err)
-	fmt.Printf("Listening on port %d...\n", s.port)
+	log.Printf("Listening on port %d...\n", s.port)
 
 	router := httprouter.New()
 
--- a/go/datas/remote_database_handlers.go
+++ b/go/datas/remote_database_handlers.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"net/url"
 	"runtime"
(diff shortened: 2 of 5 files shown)


ehalpern:
cc: @rafael-atticlabs

go/datas/remote_database_handlers.go
```diff
@@ -97,7 +98,7 @@ func createHandler(hndlr Handler, versionCheck bool) Handler {
 		w.Header().Set(NomsVersionHeader, constants.NomsVersion)
 
 		if versionCheck && req.Header.Get(NomsVersionHeader) != constants.NomsVersion {
-			verbose.Log("Returning version mismatch error")
+			log.Printf("returning version mismatch error")
```
ehalpern: Assuming that request errors are important enough to always log.

go/datas/remote_database_handlers.go
```diff
@@ -109,7 +110,7 @@ func createHandler(hndlr Handler, versionCheck bool) Handler {
 		err := d.Try(func() { hndlr(w, req, ps, cs) })
 		if err != nil {
 			err = d.Unwrap(err)
-			verbose.Log("Returning bad request:\n%v\n", err)
+			log.Printf("returning bad request error: %v", err)
```
ehalpern: Assuming that request errors are important enough to always log.

go/datas/database_server.go
```diff
@@ -56,7 +57,7 @@ func (s *RemoteDatabaseServer) Run() {
 	d.Chk.NoError(err)
 	s.port, err = strconv.Atoi(port)
 	d.Chk.NoError(err)
-	fmt.Printf("Listening on port %d...\n", s.port)
+	log.Printf("Listening on port %d...\n", s.port)
```
cmasone-attic: I'm a bit confused...above, you changed `fmt.Printf` to `verbose.Log`, while here you changed it to `log.Printf`

go/datas/remote_database_handlers.go
cmasone-attic: ...and then here you changed AWAY from `verbose.Log` to use `log.Printf` directly. What's the right thing for code to do?

cmasone-attic: ...ah

go/datas/database_server.go
ehalpern: This was never protected by the verbose flag, so I'm assuming that it should always be printed. Just making sure it goes through the logger.

go/datas/remote_database_handlers.go
ehalpern: There's a trade-off to make here. --verbose is all or nothing. 

One approach would be to hide all messages behind verbose and require the embedding program to log all or none.

Another is to identify a small, useful set of messages that will always be printed. I'm proposing the later. Specifically, that request errors should always be printed. 

FWIW, this is exactly why finer grained logging systems are the norm outside of the go universe. When a server embeds software, it's nice for it to have fine control over what to log when.

cmasone-attic:
LGTM

PR #3593, opened 2017-07-19
Refactor Blob reading

rafael-atticlabs:
Towards #3562 (Next patch will be BlobReadWriter)

This patch refactors Blob reading generally:

-Add `LoadLeafSequences`, which eagerly loads exactly the set of leaf sequences which contain a given idx range
-`Blob` now supports the golang `ReaderAt` interface. The implementation uses `LoadLeafSequences` to eagerly load exactly the needed byte range
-`BlobReader` is now basically just an `int` (pos) which turns around and uses `Blob.ReadAt`
-`BlobReader.Copy` is moved to `Blob` and uses a read ahead strategy of N concurrent forward reads, but each uses `ReadAt` (rather than the less precise `readAheadLeafCursors`).

This wasn't the goal here, but a happy side-effect is that blob reading perf has generally increased:

-Reading large blob via http: +40% (now about 250MB/s on my mac)
-Reading large blob via nbs: +40% (now about 650MB/s on my mac)
-`getBlob` http endpoint can now service many "image" sized blobs (e.g. 1-20MB) is ~3 physical reads, rather than 8-10.

Also, this `LoadLeafSequences` lays the groundwork for the graphql endpoint to use precise eager reading, since it always knows the exact item range it needs when it's reading ranges of ptrees.

### DIFF
--- a/go/types/blob.go
+++ b/go/types/blob.go
@@ -52,23 +52,24 @@ func (b Blob) ReadAt(p []byte, off int64) (n int, err error) {
 		return
 	}
 
-	leaves, localStart := LoadLeafSequences(b, startIdx, endIdx)
+	root := b.sequence()
+	leaves, localStart := loadLeafSequences(root.valueReader(), []sequence{root}, startIdx, endIdx)
 	endIdx = localStart + endIdx - startIdx
 	startIdx = localStart
 
 	for _, s := range leaves {
 		bl := s.(blobLeafSequence)
 
-		le := endIdx
-		ll := uint64(len(bl.data))
-		if le > ll {
-			le = ll
+		localEnd := endIdx
+		leafLength := uint64(len(bl.data))
+		if localEnd > leafLength {
+			localEnd = leafLength
 		}
-		src := bl.data[startIdx:le]
+		src := bl.data[startIdx:localEnd]
 
 		copy(p[n:], src)
 		n += len(src)
-		endIdx -= le
+		endIdx -= localEnd
 		startIdx = 0
 	}
 
@@ -83,6 +84,9 @@ func (b Blob) Copy(w io.Writer) (n int64) {
 	return b.CopyReadAhead(w, 1<<23 /* 8MB */, 6)
 }
 
+// CopyReadAhead copies the entire contents of |b| to |w|, and attempts to stay
+// |concurrency| |chunkSize| blocks of bytes ahead of the last byte written to
+// |w|.
 func (b Blob) CopyReadAhead(w io.Writer, chunkSize uint64, concurrency int) (n int64) {
 	bChan := make(chan chan []byte, concurrency)
 
@@ -93,14 +97,14 @@ func (b Blob) CopyReadAhead(w io.Writer, chunkSize uint64, concurrency int) (n i
 			bChan <- bc
 
 			start := idx
-			len := b.Len() - start
-			if len > chunkSize {
-				len = chunkSize
+			blockLength := b.Len() - start
+			if blockLength > chunkSize {
+				blockLength = chunkSize
 			}
-			idx += len
+			idx += blockLength
 
 			go func() {
-				buff := make([]byte, len)
+				buff := make([]byte, blockLength)
 				b.ReadAt(buff, int64(start))
 				bc <- buff
 			}()
@@ -110,7 +114,7 @@ func (b Blob) CopyReadAhead(w io.Writer, chunkSize uint64, concurrency int) (n i
 
 	// Ensure read-ahead goroutines can exit
 	defer func() {
-		for _ = range bChan {
+		for range bChan {
 		}
 	}()
 
--- a/go/types/blob.go
+++ b/go/types/blob.go
@@ -5,7 +5,6 @@
 package types
 
 import (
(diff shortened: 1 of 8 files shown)


rafael-atticlabs:
cc @arv, @aboodman

rafael-atticlabs:
@kalman, investigating nomsfs test failures. please review anyway.

rafael-atticlabs:
ok, all tests passing. PTAL

kalman:
because it's taking me a while to understand everything :-)

go/types/indexed_sequences.go
```diff
@@ -77,3 +80,74 @@ func orderedKeyFromSum(msd []metaTuple) orderedKey {
 	}
 	return orderedKeyFromUint64(sum)
 }
+
+func LoadLeafSequences(c Collection, startIdx, endIdx uint64) (leaves []sequence, firstIdx uint64) {
```
kalman: it would be useful to comment on inclusivity of startIdx/endIdx and what firstIdx means.
From reading the code it looks like it's inclusive of startIdx and exclusive of endIdx.

go/types/blob.go
```diff
@@ -30,10 +29,99 @@ func NewEmptyBlob() Blob {
...
+	}
+
+	leaves, localStart := LoadLeafSequences(b, startIdx, endIdx)
+	endIdx = localStart + endIdx - startIdx
+	startIdx = localStart
+
+	for _, s := range leaves {
+		bl := s.(blobLeafSequence)
+
+		le := endIdx
```
kalman: finding this code a bit hard to follow (all the math involving le's and ll's and so forth) but I'm sure it's correct :-)

maybe comments or something would help

go/types/blob.go
```diff
@@ -30,10 +29,99 @@ func NewEmptyBlob() Blob {
...
+				b.ReadAt(buff, int64(start))
+				bc <- buff
+			}()
+		}
+		close(bChan)
+	}()
+
+	// Ensure read-ahead goroutines can exit
+	defer func() {
+		for _ = range bChan {
```
kalman: You can just do `for range bChan {}`

go/types/blob.go
```diff
@@ -30,10 +29,99 @@ func NewEmptyBlob() Blob {
...
-	return &BlobReader{b.seq, cursor, nil, 0}
+	return &BlobReader{b, 0}
+}
+
+func (b Blob) Copy(w io.Writer) (n int64) {
+	return b.CopyReadAhead(w, 1<<23 /* 8MB */, 6)
+}
+
+func (b Blob) CopyReadAhead(w io.Writer, chunkSize uint64, concurrency int) (n int64) {
+	bChan := make(chan chan []byte, concurrency)
```
kalman: huh, chan of chans. Neat trick.
but why not use an array of chans, and take out the outer goroutine? it doesn't seem like it adds a whole lot. like:
```go
chans := make([]chan []byte, concurrency, 0)

idx := uint(0)
for idx < b.Len() {
  bc := make(chan []byte)
  chans = append(chans, bc)

  start := dx
  // etc
}
```

go/types/blob.go
```diff
@@ -30,10 +29,99 @@ func NewEmptyBlob() Blob {
...
+
+	go func() {
+		idx := uint64(0)
+		for idx < b.Len() {
+			bc := make(chan []byte)
+			bChan <- bc
+
+			start := idx
+			len := b.Len() - start
```
kalman: avoid using len as a variable name cos it's a keyword? my brain isn't used to that.

go/types/indexed_sequences.go
```diff
@@ -77,3 +80,74 @@ func orderedKeyFromSum(msd []metaTuple) orderedKey {
...
+	level := seqs[0].treeLevel()
+	childTuples := []metaTuple{}
+
+	cum := uint64(0)
+	for _, s := range seqs {
+		d.PanicIfFalse(s.treeLevel() == level)
+		ms := s.(metaSequence)
+
+		for _, mt := range ms.tuples {
+			if cum == 0 && mt.numLeaves <= startIdx {
```
kalman: it would have helped to have had a comment above here like "skip tuples up to starting index"

go/types/indexed_sequences.go
rafael-atticlabs: done.

go/types/indexed_sequences.go
rafael-atticlabs: done

go/types/blob.go
rafael-atticlabs: renamed local vars. hopefully that helps.

go/types/blob.go
rafael-atticlabs: The outer chan is the mechanism by read-ahead is actually accomplished. The result is that the code tries to stay |concurrency| blocks of |chunkSize| ahead of the current consumption point.

go/types/blob.go
rafael-atticlabs: done

go/types/blob.go
rafael-atticlabs: done

PR #3596, opened 2017-07-20
Allow server to re-try commits if it can

### LINKED ISSUE #3582: Allow servers to retry some optimistic lock failures
When multiple clients are committing to different datasets in the same database, they will experience optimistic lock failures even though there is nothing preventing the commits from merging cleanly. In the httpChunkStore case, this means another round-trip between the client and server even though the server has all the information it needs to try again.

The one piece of fallout here is that a successful retry results in the root of persistent storage winding up at a _different_ hash that the one proposed by the client. So, the server must return the new root hash in all cases, where it used to do this only in the case of an optimistic lock failure.

FYI @rafael-atticlabs

cmasone-attic:
Right now, the only kinds of Commits that the server will retry are those to different datasets.
That is, if another client concurrently landed a change to some other dataset, and that is the _only_ thing that causes your Commit attempt to fail, the server will retry.

Fixes #3582

### DIFF
--- a/go/datas/remote_database_handlers.go
+++ b/go/datas/remote_database_handlers.go
@@ -367,46 +367,186 @@ func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, cs c
 		d.Panic(`Expected "last" query param value`)
 	}
 	last := hash.Parse(tokens[0])
+	// "current" should really, really be called "proposed" or something in the wire API
 	tokens = params["current"]
 	if len(tokens) != 1 {
 		d.Panic(`Expected "current" query param value`)
 	}
-	current := hash.Parse(tokens[0])
+	proposed := hash.Parse(tokens[0])
 
 	vs := types.NewValueStore(cs)
 
 	// Even though the Root is actually a Map<String, Ref<Commit>>, its Noms Type is Map<String, Ref<Value>> in order to prevent the root chunk from getting bloated with type info. That means that the Value of the proposed new Root needs to be manually type-checked. The simplest way to do that would be to iterate over the whole thing and pull the target of each Ref from |cs|. That's a lot of reads, though, and it's more efficient to just read the Value indicated by |last|, diff the proposed new root against it, and validate whatever new entries appear.
-	datasets := types.NewMap()
-	if !last.IsEmpty() {
-		lastVal := vs.ReadValue(last)
-		if lastVal == nil {
-			d.Panic("Can't Commit from a non-present Chunk")
-		}
+	lastMap := validateLast(last, vs)
 
-		datasets = lastVal.(types.Map)
+	proposedMap := validateProposed(proposed, last, vs)
+	if !proposedMap.Empty() {
+		assertMapOfStringToRefOfCommit(proposedMap, lastMap, vs)
 	}
 
-	// Only allowed to skip this check if both last and current are empty, because that represents the special case of someone flushing chunks into an empty store.
-	if !last.IsEmpty() || !current.IsEmpty() {
-		// Ensure that proposed new Root is present in cs, is a Map and, if it has anything in it, that it's <String, <Ref<Commit>>
-		proposed := vs.ReadValue(current)
-		if proposed == nil {
-			d.Panic("Can't set Root to a non-present Chunk")
+	// If some other client has committed to |vs| since it had |from| at the
+	// root, this call to vs.Commit() will fail. Used to be that we'd always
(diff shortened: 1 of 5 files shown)


cmasone-attic:
PTAL

go/datas/remote_database_handlers.go
```diff
@@ -367,46 +367,175 @@ func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, cs c
...
+	proposed := hash.Parse(tokens[0])
 
 	vs := types.NewValueStore(cs)
 
 	// Even though the Root is actually a Map<String, Ref<Commit>>, its Noms Type is Map<String, Ref<Value>> in order to prevent the root chunk from getting bloated with type info. That means that the Value of the proposed new Root needs to be manually type-checked. The simplest way to do that would be to iterate over the whole thing and pull the target of each Ref from |cs|. That's a lot of reads, though, and it's more efficient to just read the Value indicated by |last|, diff the proposed new root against it, and validate whatever new entries appear.
-	datasets := types.NewMap()
+	lastMap := validateLast(last, vs)
+
+	proposedMap := validateProposed(proposed, last, vs)
+	if !proposedMap.Empty() {
```
rafael-atticlabs: are you sure you need this check? kinda looks like the assert will succeed on an empty map

go/datas/remote_database_handlers.go
```diff
@@ -367,46 +367,175 @@ func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, cs c
...
+
+		d.PanicIfFalse(aChange.Key.Equals(bChange.Key))
+		// If the two diffs generate different kinds of changes at the same key, conflict.
+		if aChange.ChangeType != bChange.ChangeType {
+			return parent, errors.New("Incompatible changes at " + types.EncodedValue(aChange.Key))
+		}
+
+		// Otherwise, we're OK IFF the two diffs made exactly the same change
+		aValue := a.Get(aChange.Key)
+		// fmt.Println("Both changed at key", types.EncodedValue(aChange.Key))
```
rafael-atticlabs: was it intention to keep these commented-out printlns?

go/datas/remote_database_handlers.go
cmasone-attic: I actually don't know, but we did it before so I just kept it. I don't know if it's a correctness thing, or if it's like..."hey, an empty map is fine and spinning up the diff machinery to figure out something we alreayd know is dumb."

I suppose I could just put this check down in the assert function, and we can take it out if we're certain it's not needed.

go/datas/remote_database_handlers.go
```diff
@@ -367,46 +367,175 @@ func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, cs c
...
-	datasets := types.NewMap()
+	lastMap := validateLast(last, vs)
+
+	proposedMap := validateProposed(proposed, last, vs)
+	if !proposedMap.Empty() {
+		assertMapOfStringToRefOfCommit(proposedMap, lastMap, vs)
+	}
+
+	// If some other client has committed to |vs| since it had |from| at the root, this call to vs.Commit() will fail. Used to be that we'd always propagate that failure back to the client and let them try again. This made one very common operation annoyingly expensive, though, as clients simultaneously committing to different Datasets would cause conflicts with this vs.Commit() right here. In this common case, the server already knows everything it needs to try again, so now we cut out the round trip to the client and just retry inline.
```
cmasone-attic: done

go/datas/remote_database_handlers.go
```diff
@@ -367,46 +367,175 @@ func handleRootPost(w http.ResponseWriter, req *http.Request, ps URLParams, cs c
...
+
+	// If committing succeeded, the root of the store might be |proposed|...or it might be some result of the merge performed above. So, we need to tell the client what the new root is. If the commit failed, obviously we need to inform the client of the actual current root.
+	w.Header().Add("content-type", "text/plain")
+	fmt.Fprintf(w, "%v", vs.Root().String())
+}
+
+func validateLast(last hash.Hash, vr types.ValueReader) types.Map {
 	if !last.IsEmpty() {
-		lastVal := vs.ReadValue(last)
+		lastVal := vr.ReadValue(last)
```
cmasone-attic: word

go/datas/remote_database_handlers.go
cmasone-attic: negatory, batman

PR #3598, opened 2017-07-21
NBS: Commit locks out Rebasers for less time

rafael-atticlabs:
Persisting tables now tacks place outside the manifest fetch lock so rebasers don't spend time blocking on writers persisting.


cmasone-attic:
lgtm

PR #3599, opened 2017-07-24
BlobEditor

rafael-atticlabs:
Towards #3562

Adds `BlobEditor` (which implements `ReadWriteSeeker`) and removes mutative method from Blob.

### DIFF
--- a/go/types/blob_editor.go
+++ b/go/types/blob_editor.go
@@ -1,0 +1,319 @@
+// Copyright 2017 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package types
+
+import (
+	"errors"
+
+	"io"
+
+	"sync"
+
+	"github.com/attic-labs/noms/go/d"
+)
+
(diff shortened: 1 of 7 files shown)


rafael-atticlabs:
@kalman, for reviewing, BlobEditor is about a 60% of ListEditor which has already been review by @arv. In particular,

-`be.Blob(vrw)` was copied and had some irrelevant code review (blobs cant have sub-editors),
-`collapseBlobEdit`, `Len()` and `Splice` are copied with types adjusted.

It's a shame to have so much code basically the same, but `byte` isn't a noms value and making it so would be problematic, perf wise.

The interesting new code here is `be.Read()`, `be.Write()` as well as the fuzz test

go/types/blob_editor.go
```diff
@@ -0,0 +1,319 @@
...
+	vr := seq.valueReader()
+
+	if vrw != nil {
+		vr = vrw
+	}
+
+	cursChan := make(chan chan *sequenceCursor)
+	spliceChan := make(chan blobEdit)
+
+	go func() {
```
rafael-atticlabs: Good point. Simplified.

PR #3603, opened 2017-07-26
NBS: Store small tables in DynamoDB when persisting to AWS

### LINKED ISSUE #3559: NBS: Store small tables in Dynamo
In certain kinds of usage, we wind up with lots of small tables. These will eventually get Conjoined, but they tend to be pretty "hot" early in their life. It might be a big perf win for these tables to be stored somewhere with lower latency than S3 until they do wind up getting conjoined.

cmasone-attic:
There seems to be a floor on the amount of time required to persist
small objects to S3. For workloads that generate lots of small tables,
this can really add up. DynamoDB is much faster to read/write, and can
hold items of up to 400k. This patch stores tables with < 64 chunks
that are < 400k in DynamoDB, caching them in memory on persist and
open to reduce load on the back end.

Fixes #3559

### DIFF
--- a/go/nbs/aws_table_persister.go
+++ b/go/nbs/aws_table_persister.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/util/sizecache"
 	"github.com/attic-labs/noms/go/util/verbose"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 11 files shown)


cmasone-attic:
OK, PTAL

go/nbs/dynamo_table_reader.go
```diff
@@ -0,0 +1,144 @@
...
+	return source
+}
+
+func (dtr *dynamoTableReader) hash() addr {
+	return dtr.h
+}
+
+func (dtr *dynamoTableReader) ReadAtWithStats(p []byte, off int64, stats *Stats) (n int, err error) {
+	if data, present := dynamoTableCacheMaybeGet(dtr.tc, dtr.hash()); present {
+		n = copy(p, data[off:])
```
rafael-atticlabs: since this is essentially the start of the L1 (mem) portion of the table cache, can we add `MemBytesPerRead` and `MemReadLatency` stats to start collecting. That way we can still compare apples-to-apples when reasoning about total read load, regardless of where the reads ended up

go/nbs/dynamo_table_reader.go
```diff
@@ -0,0 +1,144 @@
...
+
+func dynamoTableCacheMaybeAdd(tc *sizecache.SizeCache, name addr, data []byte) {
+	if tc != nil {
+		tc.Add(name, uint64(len(data)), data)
+	}
+}
+
+func tryDynamoTableRead(ddb ddbsvc, table string, name addr) (data []byte, err error) {
+	result, rerr := ddb.GetItem(&dynamodb.GetItemInput{
+		ConsistentRead: aws.Bool(true), // This doubles the cost :-(
```
rafael-atticlabs: we dont need a consistent read, do we? these tables will never change once they are written, right?

go/nbs/dynamo_table_reader.go
```diff
@@ -0,0 +1,144 @@
...
+
+	"github.com/attic-labs/noms/go/d"
+	"github.com/attic-labs/noms/go/util/sizecache"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+)
+
+const (
+	dataAttr    = "data"
+	tablePrefix = "*" // I want to use NBS table names as keys when they are written to DynamoDB, but a bare table name is a legal Noms Database name as well. To avoid collisions, dynamoTableReader prepends this prefix (which is not a legal character in a Noms Database name).
```
rafael-atticlabs: another option would be to namespace both key kinds. e.g. 

`m-/p/rafael@rafw.com` vs `t-03939aiei92ia`

rafael-atticlabs: don't feel strongly

go/nbs/aws_table_persister.go
```diff
@@ -44,22 +80,29 @@ type s3UploadedPart struct {
...
-func (s3p s3TablePersister) Persist(mt *memTable, haver chunkReader, stats *Stats) chunkSource {
+func (s3p awsTablePersister) Persist(mt *memTable, haver chunkReader, stats *Stats) chunkSource {
 	name, data, chunkCount := mt.write(haver, stats)
 	if chunkCount == 0 {
 		return emptyChunkSource{}
 	}
 	t1 := time.Now()
+	if s3p.limits.tableFitsInDynamo(name, data, chunkCount) {
+		dynamoTableWrite(s3p.ddb, s3p.table, name, data)
+		verbose.Log("Persisted table of %d Kb to DynamoDB in %s", len(data)/1024, time.Since(t1))
```
cmasone-attic: ...Probably not? I would guess they've been obsoleted by Stats{}

go/nbs/dynamo_table_reader.go
cmasone-attic: Done

go/nbs/dynamo_table_reader.go
cmasone-attic: They won't change, no, but DynamoDB doesn't provide consistency guarantees on creation, unless they've changed that recently. S3 does, but dynamo doesn't. So if we don't read consistently, we could fail to find a table that's been pushed to Dynamo recently.

rafael-atticlabs: It's a pretty big bummer to double read load for this. One thing we could do is this:

1) Do an initial eventually consistent read. Seems like this will basically always succeed as long as the table was written to dynamo
2) If it fails, race a read from s3 and a strongly consistent read from Dynamo.

In practice, what this will means it that:
-The code is correct, in general the read cost is 1/2.
-In rare cases someone is trying to open a table under the chunk threshold but over the size threshold, we'll needless burn some more dynamo reads.
-In excedingly rare cases where we read the same so fast after writing it that the eventually consistent read fails, it cost us 3 reads rather than 2

rafael-atticlabs: If you prefer to open a bug for this, that's totally fine.

rafael-atticlabs:
LGTM

PR #3605, opened 2017-07-28
NBS: increase size of small-table memory cache

cmasone-attic:
Looking at metrics on staging today, there are frequent spikes of
tens of thousands of throttled DynamoDB reads. One explanation is
that we're constantly evicting 'hot' tables from the in-memory
cache because the working set is larger than the space we've
allotted for the cache.


cmasone-attic:
TBR

PR #3606, opened 2017-07-28
NBS: Read small tables with eventually consistent reads first

### LINKED ISSUE #3604: NBS: try to read small tables with eventually-consistent reads
As of https://github.com/attic-labs/noms/pull/3603 NBS stores small tables in DynamoDB. For small, frequently-read tables, this is a huuuuge win. One downside, though, is that Dynamo doesn't provide perfect read-after-create consistency (while S3 does). This means that, if we do eventually-consistent Dynamo reads, there's a nonzero chance that the response will fail to find a table that was recently written.

We don't know what the average lag is between item creation and visibility to eventually consistent reads in Dynamo, but presumably it's short enough that we'll _usually_ be OK. To handle the case where we get an inconsistent result, we can just retry with a consistent read (while also checking S3)

cmasone-attic:
Fail over to fully-consistent reads if no result. This means that
misses will get more expensive, but hits will cost us half what they
were costing in the initial version of the code.

Fixes #3604

### DIFF
--- a/go/nbs/dynamo_table_reader.go
+++ b/go/nbs/dynamo_table_reader.go
@@ -7,6 +7,7 @@ package nbs
 import (
 	"fmt"
 	"io"
+	"log"
 	"time"
 
 	"github.com/attic-labs/noms/go/d"
@@ -135,6 +136,7 @@ func tryDynamoTableRead(ddb ddbsvc, table string, name addr) (data []byte, err e
 	}
 	data, err = try(&input)
 	if _, isNotFound := err.(tableNotInDynamoErr); isNotFound {
+		log.Printf("Eventually consistent read for %s failed; trying fully-consistent", name)
 		input.ConsistentRead = aws.Bool(true)
 		return try(&input)
 	}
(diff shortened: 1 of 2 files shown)


cmasone-attic:
PTAL

go/nbs/dynamo_table_reader.go
```diff
@@ -115,21 +115,30 @@ func dynamoTableCacheMaybeAdd(tc *sizecache.SizeCache, name addr, data []byte) {
...
-	} else if len(result.Item) == 0 {
-		return nil, tableNotInDynamoErr{name.String(), table}
-	} else if result.Item[dataAttr] == nil || result.Item[dataAttr].B == nil {
-		return nil, fmt.Errorf("NBS table %s in DynamoDB table %s is malformed", name, table)
 	}
-	return result.Item[dataAttr].B, nil
+	data, err = try(&input)
+	if _, isNotFound := err.(tableNotInDynamoErr); isNotFound {
+		input.ConsistentRead = aws.Bool(true)
+		return try(&input)
```
rafael-atticlabs: let's log this backoff so that we can use the log processing to get a sense as to how often (if ever this is happening). seems like it should basically be never, so a loud/annoying log level/message might be appropriate to start.

ghost (approved):
LGTM w/ nit

go/nbs/dynamo_table_reader.go
cmasone-attic: done

Issue #3607, opened 2017-07-28
NBS: Read Dynamo lazily when opening tables


cmasone-attic:
The current code eagerly checks dynamo to see if the small table is in there, even though there are many cases when we probably don't ever need to read the table at all (when the index is cached already, for example)

This is a source of significant inefficiency and load. fix it fixit FixitFixItFIXIT

FYI @rafael-atticlabs

PR #3608, opened 2017-07-28
NBS: populate small table cache on Open

cmasone-attic:
This will likely bloat the cache with tables no one's going to read
data from, BUT doing this also means that most checks to see if a table
is in Dynamo at all can proceed locally. Stopgap until fix for #3607 lands


cmasone-attic:
PTAL

ghost (approved):
LGTM. FWIW, I think this is subject to a similar problem we had with lots of concurrent rebasers to the same store concurrently loading the same manifest.

When we move to the aws_heisen_table reader, let's undo this.

cmasone-attic:
yah, I have a TODO in the code to take this out as a part of that work

PR #3609, opened 2017-07-31
Format DateTime objects as strings in human readable encoding

willhite:


cmd/noms/noms_show.go
```diff
@@ -30,13 +30,15 @@ var nomsShow = &util.Command{
...
 var showStats = false
+var tzName string
 
 func setupShowFlags() *flag.FlagSet {
 	showFlagSet := flag.NewFlagSet("show", flag.ExitOnError)
 	outputpager.RegisterOutputpagerFlags(showFlagSet)
 	verbose.RegisterVerboseFlags(showFlagSet)
 	showFlagSet.BoolVar(&showRaw, "raw", false, "If true, dumps the raw binary version of the data")
 	showFlagSet.BoolVar(&showStats, "stats", false, "If true, reports statistics related to the value")
+	showFlagSet.StringVar(&tzName, "tz", "", "timezone that dates or displayed in, can be: local or utc")
```
aboodman: Why not provide a default?

go/types/encode_human_readable.go
```diff
@@ -195,6 +204,23 @@ func (w *hrsWriter) writeStruct(v Struct, printStructName bool) {
 	w.write("}")
 }
 
+var dateTimeType = MakeStructTypeFromFields("DateTime", FieldMap{
+	"secSinceEpoch": NumberType,
+})
+
+func (w *hrsWriter) FormatDateTime(v Value) string {
```
aboodman: I thought that there was a bug for this, and that we decided to do it with comments, like so:

```
struct DateTime {  // 2017-07-31T20:36:00
  msSinceEpoch: 1501558457708,
}
```

... it's not as beautiful as it could be, but it's nice because it's general (other things could have similar comments)

go/types/encode_human_readable.go
```diff
@@ -8,20 +8,25 @@ import (
...
+	"math"
 	"strconv"
+	"time"
 
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/util/writers"
 	humanize "github.com/dustin/go-humanize"
 )
 
+const ReadableEncodingDateFormat = "2006-01-02T15:04:05-0700"
```
aboodman: This is basically `time. RFC3339`. Why not use that?

go/types/encode_human_readable.go
```diff
@@ -195,6 +204,23 @@ func (w *hrsWriter) writeStruct(v Struct, printStructName bool) {
 	w.write("}")
 }
 
+var dateTimeType = MakeStructTypeFromFields("DateTime", FieldMap{
+	"secSinceEpoch": NumberType,
+})
+
+func (w *hrsWriter) FormatDateTime(v Value) string {
+	if w.tz == nil {
```
aboodman: I see. Another advantage of using a comment is that no flag is necessary. We can just always output the useful string.

cmd/noms/noms_show.go
willhite: the default is now 'local'.

go/types/encode_human_readable.go
willhite: The one we had was ISO 8601, I'm not sure what that conforms to in the golang definitions or why it's not there but I made the switch to RFC3339.

go/types/encode_human_readable.go
willhite: Yes, this is better, I've modified it so it works this way now.

go/types/encode_human_readable.go
willhite: Agreed.

willhite:
@aboodman Added test, and fixed bug that test revealed. 👍

cmd/noms/noms_log.go
```diff
@@ -57,6 +59,7 @@ func setupLogFlags() *flag.FlagSet {
 	logFlagSet.BoolVar(&oneline, "oneline", false, "show a summary of each commit on a single line")
 	logFlagSet.BoolVar(&showGraph, "graph", false, "show ascii-based commit hierarchy on left side of output")
 	logFlagSet.BoolVar(&showValue, "show-value", false, "show commit value rather than diff information")
+	logFlagSet.StringVar(&tzName, "tz", "local", "display dates in this timezone, must be: local or utc ")
```
aboodman: This is nice feature.

willhite:
Implemented by allowing clients to register HRSCommenter functions. I decided to squash the changes because this implementation was different enough from the previous commits that it didn't warrant keeping them around. @aboodman PTAL

cmd/noms/noms_show.go
```diff
@@ -69,6 +74,13 @@ func runShow(args []string) int {
 		return 0
 	}
 
+	tz, err := locationFromTimezoneArg(tzName, nil)
```
arv: I'm wondering if this can be shared too?

go/types/encode_human_readable.go
```diff
@@ -9,12 +9,73 @@ import (
...
+//    lastRefresh: DateTime { // 2017-08-03T16:07:06-07:00
+//      secSinceEpoch: 1.501801626877e+09,
+//    }
+
+// Function type for commenter functions
+type HRSCommentFunc func(Struct) string
+
+var (
+	commenterRegistry = map[string]map[string]HRSCommentFunc{}
+	registryLock      sync.Mutex
```
arv: Can this be a RWMutex?

go/types/encode_human_readable.go
```diff
@@ -178,6 +239,14 @@ func (w *hrsWriter) writeStruct(v Struct, printStructName bool) {
 		w.write(" ")
 	}
 	w.write("{")
+	commenters := GetHRSCommenters(v.name)
+	for _, commentFunc := range commenters {
+		if comment := commentFunc(v); comment != "" {
+			w.write(" // " + comment)
```
arv: This seems to print `// a // b // c` if three commenters are registered that returns non empty strings. Not sure what we want to do here though?

go/types/encode_human_readable.go
```diff
@@ -178,6 +239,14 @@ func (w *hrsWriter) writeStruct(v Struct, printStructName bool) {
...
+			break
+		}
+
```
arv: gofmt is ok with this?

go/util/datetime/date_time.go
```diff
@@ -65,3 +74,16 @@ func (dt *DateTime) UnmarshalNoms(v types.Value) error {
 	*dt = DateTime{time.Unix(int64(s), int64(frac*1e9))}
 	return nil
 }
+
+func RegisterHRSCommenter(tz *time.Location) {
+	types.UnregisterHRSCommenter(datetypename, hrsEncodingName)
```
arv: Do you really need to unregister here?

aboodman:
@arv do you want to take over this pr? (welcome back :))

go/util/datetime/date_time.go
willhite: You're right. There's no need. I removed that line and modified the test to check that.

go/types/encode_human_readable.go
willhite: Done. Also, made a copy of the list of functions that I'm returning. The map could have been modified by someone after it was returned.

cmd/noms/noms_show.go
willhite: I did make the messages consistent but I'm not sure if it's worth sharing since they would still have to check the err and return.

go/types/encode_human_readable.go
```diff
@@ -9,12 +9,73 @@ import (
...
+//    }
+//
+// By registering a commenter function that returns a nicely formatted date,
+// the struct will be coded with a comment:
+//    lastRefresh: DateTime { // 2017-08-03T16:07:06-07:00
+//      secSinceEpoch: 1.501801626877e+09,
+//    }
+
+// Function type for commenter functions
+type HRSCommentFunc func(Struct) string
```
aboodman: I'm just saying:

```
type HRSCommenter interface {
  func(Struct) string
}

RegisterCommenter(c Commenter)
```

go/types/encode_human_readable.go
aboodman: No you'll need it as soon as multiple things want to register from separate goroutines.

aboodman: ... but it's fine with me.

go/types/encode_human_readable.go
```diff
@@ -9,12 +9,77 @@ import (
...
+
+// RegisterHRSCommenter is called to with three arguments:
+//  typename: the name of the struct this function will be applied to
+//  unique: an arbitrary string to differentiate functions that should be applied
+//    to different structs that have the same name (e.g. two implementations of
+//    the "Employee" type.
+//  f: a function that gets called for all structs with this name. The function
+//    should verify the type of the struct and, if appropriate, return a non-empty
+//    string to be appended as the comment
+func RegisterHRSCommenter(typename, unique string, f HRSCommentFunc) {
```
aboodman: I think that we should register by noms type, not typename, for a few reasons:

- you should be able to annotate things other than structs
- struct names are not unique/special in noms, only types are

aboodman: Just pass in a pointer (or if you follow my other comment) interface reference, and use that for identity, rather than `unique`.

aboodman: e.g.:

```
RegisterHRSCommenter(t *types.Type, c HRSCommenter)
```

arv:
If we are doing type we should be able to use the same for  List, Map & Set.

willhite:
@aboodman and @arv There were two comments that were unresolved when we last looked at this. Aaron suggested that instead of registering a function, that we register an interface with a Comment() function. I did that.

Aaron also suggested that instead of only working with Structs, that we registerer Commenters with *Type's as keys. Then we could make comment work on Value rather than just Struct. I haven't done that yet because it seems to me that the cost of generating the Type for each value that we print out would be cost prohibitive. Thoughts?

arv:
We talked about this in RL. Using `*Type` would be too expensive.

If we wanted to allow using the same mechanism for Maps we could use NomsKind+Name is the key but I'm not sure it is worth doing at this point.

cmd/noms/noms_log.go
```diff
@@ -66,6 +69,13 @@ func runLog(args []string) int {
 	useColor = shouldUseColor()
 	cfg := config.NewResolver()
 
+	tz, err := locationFromTimezoneArg(tzName, nil)
```
arv: Doesn't our new flag system handle this for us?

willhite: I don't believe it does, but I did need to add a flag definition for "tz" in the kingpin flag definition code in noms.go.

PR #3610, opened 2017-08-01
NBS: Defer AWS reads when opening a table with cached index

### LINKED ISSUE #3607: NBS: Read Dynamo lazily when opening tables

cmasone-attic:
Prior to this patch, whenever we created a chunkSource for a table
persisted to AWS, awsTablePersister::Open() would hit DynamoDB to
check whether the table data was stored there. That's how it knew
whether to create a dynamoTableReader or an s3TableReader. This
results in consulting Dynamo (or the in-memory small-table cache)
every time we go to open a table. Most of the time, this isn't
necessary, as we separately cache table indices and really only
need that data at Open() time.

This patch defers reading table data if possible.

Fixes #3607

### DIFF
--- a/go/nbs/dynamo_table_reader.go
+++ b/go/nbs/dynamo_table_reader.go
@@ -21,13 +21,10 @@ const (
 	tablePrefix = "*" // I want to use NBS table names as keys when they are written to DynamoDB, but a bare table name is a legal Noms Database name as well. To avoid collisions, dynamoTableReader prepends this prefix (which is not a legal character in a Noms Database name).
 )
 
-// dynamoTableReader assumes the existence of a DynamoDB table whose primary partition key is in String format and named `db`.
-type dynamoTableReader struct {
-	tableReader
-	ddb   ddbsvc
-	table string
-	h     addr
-	tc    *sizecache.SizeCache
+// dynamoTableReaderAt assumes the existence of a DynamoDB table whose primary partition key is in String format and named `db`.
+type dynamoTableReaderAt struct {
+	ddb *ddbTableStore
+	h   addr
 }
 
 type tableNotInDynamoErr struct {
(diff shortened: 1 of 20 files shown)


cmasone-attic:
Uff. A bunch of the files are changed just because I added `stats *Stats` to the signature of `Open()`. Reading indices out of storage wasn't getting tracked properly in our stats, which should be mostly fixed now for AWS.

PTAL.

cmasone-attic:
Ok! I fixed all the extra index data reading and simplified some stuff. PTAL

go/nbs/dynamo_table_reader.go
```diff
@@ -38,98 +35,71 @@ func (t tableNotInDynamoErr) Error() string {
...
-		if i, present := tc.Get(name); present {
-			return i.([]byte), true
+type ddbTableStore struct {
+	ddb    ddbsvc
+	table  string
+	readRl chan struct{}
+	cache  *sizecache.SizeCache // TODO: merge this with tableCache as part of BUG 3601
+}
+
+func (dts *ddbTableStore) ReadTable(name addr, stats *Stats) (data []byte, err error) {
```
rafael-atticlabs: Did you consider making this `MaybeReadTable(name addr, chunkCount uint64, stats *Stats)` and the first thing it does is `tableMayBeInDynamo`. if it *can't* be dynamo, then just return `tableNotInDynamoErr` without checking?

cmasone-attic: `tableMayBeInDynamo()` is provided by `awsLimits`, and chunkMax and the other things in there are subject to change in future patches. As you pointed out, it's conceivable that we might find ourselves in a situation where the current version of the code decides a table "can't" be in Dynamo according to current policy, but a _previous_ version of the code had a different policy. In that case we'd go try to read it from S3 and discover it's not there (because an earlier version of the server had written the table to Dynamo) and so we'd want to fail over to checking in Dynamo in that case.

Given that I kind of expect that situation in the future, and that baking the `awsLimits` into this little object here seemed like it wouldn't make the current code THAT much simpler, I decided to do it this way :-)

PR #3617, opened 2017-08-15
Fix CORS redirects in noms serve

kalman:
The httprouter was HTTP-redirecting paths like graphql to graphql/ which
isn't CORS-compatible. This patch just hard-codes the trailing and
non-trailing slash versions of each handler path.

### DIFF
--- a/go/datas/database_server.go
+++ b/go/datas/database_server.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/attic-labs/noms/go/chunks"
 	"github.com/attic-labs/noms/go/constants"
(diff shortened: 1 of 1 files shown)


kalman:
@rafael-atticlabs ptal
@ehalpern fyi

I looked at httprouter and I didn't see a better way of doing this.

kalman:
I.e. before patch:

screen shot 2017-08-15 at 12 38 42 pm

kalman:
(adding the trailing / i.e. `graphql` => `graphql/` fixes it)

rafael-atticlabs:
I think @arv is a better person to review.

go/datas/database_server.go
```diff
@@ -59,23 +60,32 @@ func (s *RemoteDatabaseServer) Run() {
...
+
+	post := func(path string, hndlr Handler) {
+		router.OPTIONS(path, s.corsHandle(noopHandle))
+		router.POST(path, s.corsHandle(s.makeHandle(hndlr)))
+		if path != "/" {
+			router.OPTIONS(strings.TrimSuffix(path, "/"), s.corsHandle(noopHandle))
+			router.POST(strings.TrimSuffix(path, "/"), s.corsHandle(s.makeHandle(hndlr)))
+		}
+	}
+	post(constants.BasePath, HandleBaseGet)
```
arv: Why did HandleBaseGet change to POST?

arv:
Maybe disable RedirectTrailingSlash instead?

https://godoc.org/github.com/julienschmidt/httprouter#Router.RedirectTrailingSlash

kalman:
This patch does also disable RedirectTrailingSlash. But without adding it back manually, that would be backwards incompatible.

arv:
Do we really want to support both?

go/datas/database_server.go
```diff
@@ -59,23 +60,32 @@ func (s *RemoteDatabaseServer) Run() {
...
-	router.POST(constants.WriteValuePath, s.corsHandle(s.makeHandle(HandleWriteValue)))
-	router.OPTIONS(constants.WriteValuePath, s.corsHandle(noopHandle))
-	router.GET(constants.BasePath, s.corsHandle(s.makeHandle(HandleBaseGet)))
-
-	router.GET(constants.GraphQLPath, s.corsHandle(s.makeHandle(HandleGraphQL)))
-	router.POST(constants.GraphQLPath, s.corsHandle(s.makeHandle(HandleGraphQL)))
-	router.OPTIONS(constants.GraphQLPath, s.corsHandle(noopHandle))
+	// Note: we implement trailing trailing-slash removal ourselves because
+	// httprouter doesn't handle the redirects properly with CORS.
+	router := httprouter.Router{}
```
arv: This changes more things than RedirectTrailingSlash. Maybe use New and then set RedirectTrailingSlash instead

kalman: I checked the options and having them all seems correct to me.

kalman:
> Do we really want to support both?

Yes I believe so.

go/datas/database_server.go
kalman: Oops...

go/datas/database_server.go
kalman: All false, I mean.

See https://github.com/julienschmidt/httprouter/blob/master/router.go#L170
* We don't want any of the Redirect options (they're all CORS incompatible). 
* HandleMethodNotAllowed is weird
* HandleOPTIONS is irrelevant since we implement our own

Issue #2157, opened 2016-07-25
proposal: edit noms from cli: `noms struct` and friends


aboodman:
It would be really useful to make minor edits to data from the command line. Basically, to be able to do the Noms equivalent of `UPDATE stuffs SET foo=bar WHERE ...`.

Proposal:

```
noms struct [command] [args]

`noms struct` creates or modifies struct values.

Commands

    new <db> [--name=<name>] [key, value, key, value, ...]
    set <source> [--at=<rel-path>] <key> <val> [key, value, key, value, ...]
    del <source> [--at=<rel-path>] <key> [key, key, ...]
    name <source> [--at=<rel-path>] <name>

    <db>     : Database spec to create new struct in
    <name> : Name to assign to struct
    <source>  : Path to the root of the tree to modify. If [rel-path] is not specified, this will be the path to a struct to be modified. Otherwise, it will be one of that struct's parents.
    <rel-path> : An optional relative path inside <source> to the struct to be modified.
    <key>     : Key to add or update
    <val>     : Value to set key to

Example

    # Mark crime 14235 as resolved - this returns a new commit object with the specified
    change.

    > noms struct set http://localhost:8000::sf-crime.value [14235] STATUS "RESOLVED"
    Result: #kenik8esfarr4r0im46gl76o0edjkm4o

    > noms show http://localhost:8000::#kenik8esfarr4r0im46gl76o0edjkm4o
    struct Commit {

      ...

      14235: {
        TYPE: "stolen property",
        DESCRIPTION: "Missing bicycle",
        ...
        STATUS: "RESOLVED",
      }

     ...
    })

Formatting values

    Values are formatted as in the Human-Readable Serialization:
    - If a value is prefixed with @, it is a absolute path in this database to a value
    - If a value is wrapped in double-quotes, it is a string
    - If a value is prefixed with #, it's a hash
    - If the value is a bare `true` or `false`, it's a boolean
    - Otherwise the value is a number
```

```
noms map [command] [args]

`noms map` creates or modifies map values.

Commands

    new <kv...>
    set <source> [--at=<rel-path>] <kv...>
    del <source> [--at=<rel-path>] <k...>
```

```
noms list [command] [args]

`noms list` creates or modifies list values.

Commands

    new [val...]
    insert <source> [--at=<rel-path>] [pos] <val...>
    del <source> [--at=<rel-path>] pos [len]
```

```
noms set [command] [args]

`noms set` creates or modifies set values.

Commands

    new [val...]
[...]

PR #3618, opened 2017-08-16
refactor noms to use kingpin instead of hand-rolled flags parser

jesseditson:
Initial work on #2157:

Summary
---

This PR switches the CLI to use [kingpin](https://github.com/alecthomas/kingpin), which has a number of advantages, namely:

- arbitrary depth contextual commands and help. This will remove weird hacks like `noms ds --delete <ds>`, which should be `noms ds delete <ds>`, and allow for creating "suites" of CLI tools, like `noms blob [command]`.
- don't hide basic usage docs under `thing help` commands, show it at the root and provide details when `thing --help`'d
- no need to check `args.length` anymore, can just mark args required
- everything is named, so stuff is a bit easier to parse
- non-defined args are `nil` when not provided by the user now that they are pointers. Could simplify things like using `-1` as a proxy for `nil`
- [easier CLI doc customization](https://github.com/alecthomas/kingpin#custom-help)
- [free shell completion and manpage generation](https://github.com/alecthomas/kingpin#bashzsh-shell-completion)

in general, it's also nice to have a lib handle deciding how to represent things like optional params and short flags, so it'll be consistent.

I opened this PR for initial eyes, but I'll at minimum break out the dataset command on this PR, and possibly start adding commands from #2157 (e.g. `blob`) on here depending on the temperature of the room.

TODO
---

- [ ] `@<file>` stuff for file paths?
- [ ] Restore or intentionally remove profiling
- [ ] Fix any broken tests
- [ ] Simplify verbosity and remove old flags libs

examples
---

**output from `noms`**:
```
usage: noms [<flags>] <command> [<args> ...]

Noms is a tool for showing off Noms data.

Flags:
  --help  Show context-sensitive help (also try --help-long and --help-man).

Commands:
  help [<command>...]
    Show help.

  commit [<flags>] <database> [<absolute-path>]
    Commits a specified value as head of the dataset

    If absolute-path is not provided, then it is read from stdin. See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the dataset and absolute-path arguments.

  config
    Prints the active configuration if a .nomsconfig file is present

  diff [<flags>] <object1> <object2>
    Shows the difference between two objects

    See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the object arguments.

  ds [<flags>] [<database>]
    Noms dataset management

    See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the database argument.

  log [<flags>] <path-spec>
    Displays the history of a path

    See Spelling Values at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the <path-spec> parameter.

  merge [<flags>] <database> <left-dataset-name> <right-dataset-name> <output-dataset-name>
    Merges and commits the head values of two named datasets

    See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the database argument.

    You must provide a working database and the names of two Datasets you want to merge. The values at the heads of these Datasets will be merged, put into a new Commit object, and set as the Head of the third provided Dataset name.

  root [<flags>] <database>
    Get or set the current root hash of the entire database

    See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the database argument.

  serve [<flags>] <database>
    Serves a Noms database over HTTP

    See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the database argument.

  show [<flags>] <object>
    Shows a serialization of a Noms object

    See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the object argument.

  sync [<flags>] <source-object> <dest-dataset>
    Moves datasets between or within databases

    See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the object and dataset arguments.

  version
    Print the noms version
```

**output of `noms merge --help`**
```
usage: noms merge [<flags>] <database> <left-dataset-name> <right-dataset-name> <output-dataset-name>

Merges and commits the head values of two named datasets

See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the database argument.

You must provide a working database and the names of two Datasets you want to merge. The values at the heads of these Datasets will be merged, put into a new Commit object, and set as the Head of the third provided Dataset name.

Flags:
      --help      Show context-sensitive help (also try --help-long and --help-man).
      --policy=n  conflict resolution policy for merging. Defaults to 'n', which means no resolution strategy will be applied. Supported values are 'l' (left), 'r' (right) and 'p' (prompt). 'prompt' will bring up a simple command-line prompt allowing you to resolve conflicts
                  by choosing between 'l' or 'r' on a case-by-case basis.
  -v, --verbose   show more
  -q, --quiet     show less

Args:
  <database>             a noms database path
  <left-dataset-name>    a dataset
  <right-dataset-name>   a dataset
  <output-dataset-name>  a dataset
```

### DIFF
--- a/cmd/noms/noms.go
+++ b/cmd/noms/noms.go
@@ -29,7 +29,6 @@ var actions = []string{
 	"nomming on",
 }
 
-// CommandHandler - a callback passed to commands
 type CommandHandler func() (exitCode int)
 
 var (
@@ -37,15 +36,15 @@ var (
 	quietFlag   *bool
 )
 
-// AddVerboseFlags adds --verbose and --quiet flags to the passed command
-func AddVerboseFlags(cmd *kingpin.CmdClause) (verboseFlag *bool, quietFlag *bool) {
+// addVerboseFlags adds --verbose and --quiet flags to the passed command
+func addVerboseFlags(cmd *kingpin.CmdClause) (verboseFlag *bool, quietFlag *bool) {
 	verboseFlag = cmd.Flag("verbose", "show more").Short('v').Bool()
 	quietFlag = cmd.Flag("quiet", "show less").Short('q').Bool()
 	return
 }
 
-// ApplyVerbosity - run when commands are invoked to apply the verbosity arguments configured in AddVerboseFlags
-func ApplyVerbosity() {
+// applyVerbosity - run when commands are invoked to apply the verbosity arguments configured in addVerboseFlags
+func applyVerbosity() {
 	verbose.SetVerbose(*verboseFlag)
 	verbose.SetQuiet(*quietFlag)
 }
@@ -55,30 +54,31 @@ func AddDatabaseArg(cmd *kingpin.CmdClause) (arg *string) {
 	return cmd.Arg("database", "a noms database path").Required().String() // TODO: custom parser for noms db URL?
 }
 
+type NomsCommand func(*kingpin.Application) (*kingpin.CmdClause, CommandHandler)
+
 // Commands, in order of preference
-var commands = []func(*kingpin.Application) (*kingpin.CmdClause, CommandHandler){
-	NomsCommit,
-	NomsConfig,
-	NomsDiff,
-	NomsDs,
-	NomsLog,
-	NomsMerge,
-	NomsRoot,
-	NomsServe,
-	NomsShow,
-	NomsSync,
-	NomsVersion,
+var commands = []NomsCommand{
+	nomsCommit,
+	nomsConfig,
+	nomsDiff,
+	nomsDs,
+	nomsLog,
+	nomsMerge,
+	nomsRoot,
+	nomsServe,
+	nomsShow,
+	nomsSync,
+	nomsVersion,
 }
 
 func main() {
 	// allow short (-h) help
 	kingpin.CommandLine.HelpFlag.Short('h')
 
-	// TODO: is there a way to dynamically generate help text other than re-initializing this every time?
 	i := rand.New(rand.NewSource(time.Now().UnixNano())).Intn(len(actions))
 	noms := kingpin.New("noms", fmt.Sprintf(`Noms is a tool for %s Noms data.`, actions[i]))
 
-	handlers := make(map[string]CommandHandler)
+	handlers := map[string]CommandHandler{}
 
 	// install handlers
 	for _, cmdFunction := range commands {
--- a/cmd/noms/noms.go
+++ b/cmd/noms/noms.go
@@ -9,27 +9,12 @@ import (
 	"log"
 	"math/rand"
 	"os"
-	"path"
 	"time"
 
-	"github.com/attic-labs/noms/cmd/util"
-	"github.com/attic-labs/noms/go/util/exit"
-	flag "github.com/juju/gnuflag"
-)
+	"github.com/attic-labs/noms/go/util/verbose"
 
-var commands = []*util.Command{
-	nomsCommit,
-	nomsConfig,
-	nomsDiff,
-	nomsDs,
-	nomsLog,
-	nomsMerge,
-	nomsRoot,
-	nomsServe,
-	nomsShow,
-	nomsSync,
-	nomsVersion,
-}
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
+)
 
 var actions = []string{
 	"interacting with",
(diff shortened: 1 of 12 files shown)


aboodman:
Wow, fast work. One comment on the example output -- can it wrap the lines?

cmd/noms/noms.go
```diff
@@ -9,27 +9,12 @@ import (
...
-	nomsDs,
-	nomsLog,
-	nomsMerge,
-	nomsRoot,
-	nomsServe,
-	nomsShow,
-	nomsSync,
-	nomsVersion,
-}
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
```
aboodman: You'll need to vendor this before landing.

cmd/noms/noms.go
```diff
@@ -44,49 +29,70 @@ var actions = []string{
...
+	verbose.SetQuiet(*quietFlag)
+}
+
+// AddDatabaseArg adds a "database" arg to the passed command
+func AddDatabaseArg(cmd *kingpin.CmdClause) (arg *string) {
+	return cmd.Arg("database", "a noms database path").Required().String() // TODO: custom parser for noms db URL?
+}
+
+// Commands, in order of preference
+var commands = []func(*kingpin.Application) (*kingpin.CmdClause, CommandHandler){
```
aboodman: This line is not formatted correctly. Install `gofmt` and set it to run on save automatically in your editor.

cmd/noms/noms.go
```diff
@@ -44,49 +29,70 @@ var actions = []string{
...
+	NomsCommit,
```
aboodman: These names probably shouldn't be public.

cmd/noms/noms.go
```diff
@@ -44,49 +29,70 @@ var actions = []string{
...
 }
 
 func main() {
-	util.InitHelp(path.Base(os.Args[0]), commands, usageString())
+	// allow short (-h) help
+	kingpin.CommandLine.HelpFlag.Short('h')
 
-	flag.Usage = util.Usage
-	flag.Parse(false)
+	// TODO: is there a way to dynamically generate help text other than re-initializing this every time?
```
aboodman: I don't see the problem with how this is working? `main()` will only get called once per run of the program.

cmd/noms/noms.go
```diff
@@ -44,49 +29,70 @@ var actions = []string{
...
+	i := rand.New(rand.NewSource(time.Now().UnixNano())).Intn(len(actions))
+	noms := kingpin.New("noms", fmt.Sprintf(`Noms is a tool for %s Noms data.`, actions[i]))
 
-	args := flag.Args()
-	if len(args) < 1 {
-		util.Usage()
-		return
-	}
+	handlers := make(map[string]CommandHandler)
```
aboodman: It's more idiomatic to say `map[string]CommandHandler{}`. Basically it's more idiomatic to use literals when possible, rather than `make`. `make` must be used for certain uncommon use cases (e.g., allocating a slice with a particular capacity to start), and thus, when a reader sees `make`, it's a flag that something interesting is about to happen. But nothing interesting is happening here :).

jesseditson:
I _think_ it might wrap based on terminal width? These are all literals so can easily manually wrap them. I think a lot of these longer messages could be made shorter by documenting more rather than linking, or by moving links to a more global place, there's quite a lot of redundancy.

cmd/noms/noms.go
```diff
@@ -44,49 +29,70 @@ var actions = []string{
 	"nomming on",
 }
 
-func usageString() string {
-	i := rand.New(rand.NewSource(time.Now().UnixNano())).Intn(len(actions))
-	return fmt.Sprintf(`Noms is a tool for %s Noms data.`, actions[i])
+// CommandHandler - a callback passed to commands
```
aboodman: This comment doesn't look correct... I think this type actually is the type that command handlers must implement. You probably don't even need a comment since that's what the name says.

jesseditson: Fair enough! vet complains at me when I don't put comments on public methods, so sometimes I feel like my go comments are redundant because the name is descriptive enough.

cmd/noms/noms.go
jesseditson: oops, for some reason my editor was using `goreturns` instead.

cmd/noms/noms.go
jesseditson: oops, not vet, golint. I can probably just turn that off, doesn't seem like we follow it.

cmd/noms/noms.go
jesseditson: Interestingly, running gofmt seems to leave this alone. What were you expecting it to change?

aboodman: You are right! I misread what this was doing.

cmd/noms/noms.go
jesseditson: Wow, duh - yeah I'll make them private

cmd/noms/noms.go
jesseditson: it's fine, it's just different than the way it worked before, where the flags were not created at runtime but at build time. Not sure the performance effect but assumedly negligible.

cmd/noms/noms.go
aboodman: To make it more clear you might consider defining a type for this function that returns a command handler, and then making this slice a slice-of-that-new-type. I'm not sure what to call that new type though though... maybe CommandDispatcher? CommandParser? CommandOption? CommandCase?

aboodman: I don't feel strongly about that last suggestion tnough.

arv:
This looks great. I didn't look closely at the code but based on the PR summary this is clearly an improvement.

jesseditson:
Closing in favor of #3621

PR #3621, opened 2017-08-16
kingpin docs, `noms blob [put | get]`

### LINKED ISSUE #2157: proposal: edit noms from cli: `noms struct` and friends

jesseditson:
Fixes #2157:

Alternative/path to #3618

Summary
---

- removes `util/help.go` in favor of [kingpin](https://github.com/alecthomas/kingpin), which now handles printing help
- adds `noms blob import` and `noms blob export` commands, who's functionality is identical to the prior `blob-put` and `blob-get` samples, which they supercede
- adds [shell completion](https://github.com/alecthomas/kingpin#bashzsh-shell-completion) to noms (needs docs)

Unlike #3618, this only adds kingpin documentation and argument validation, rather than refactoring the commands.

That may still be too much of a change, as the validation will cause things like `--bool-flag=false` to fail. See the updated tests to see where this changed.

The reasoning behind this approach rather than trying to keep both around is that the `help` command would be missing the blob command without switching over - alternatively I could hardcode the blob docs in for now, but this seemed like a good path forward to me.

examples
---

output from `noms` and `noms help merge` are about the same as #3618

**output of `noms help blob`**
```
usage: noms blob <command> [<args> ...]

interact with blobs in a dataset

Flags:
  --help  Show context-sensitive help (also try --help-long and --help-man).

Subcommands:
  blob import [<flags>] <url-or-file> <dataset>
    imports a blob to a dataset

  blob export [<flags>] <dataset> [<file>]
    exports a blob from a dataset
```

**output of `noms help blob import`**
```
usage: noms blob import [<flags>] <url-or-file> <dataset>

imports a blob to a dataset

Flags:
      --help                   Show context-sensitive help (also try --help-long and --help-man).
  -v, --verbose                show more
  -q, --quiet                  show less
      --cpuprofile=CPUPROFILE  write cpu profile to file
      --memprofile=MEMPROFILE  write memory profile to file
      --blockprofile=BLOCKPROFILE
                               write block profile to file
      --concurrency=8          number of concurrent HTTP calls to retrieve remote resources

Args:
  <url-or-file>  a url or file to import
  <dataset>      the path to import to
```

**output of `noms help blob export`**
```
usage: noms blob export [<flags>] <dataset> [<file>]

exports a blob from a dataset

Flags:
      --help                   Show context-sensitive help (also try --help-long and --help-man).
  -v, --verbose                show more
  -q, --quiet                  show less
      --cpuprofile=CPUPROFILE  write cpu profile to file
      --memprofile=MEMPROFILE  write memory profile to file
      --blockprofile=BLOCKPROFILE
                               write block profile to file

Args:
  <dataset>  the dataset to export
  [<file>]   an optional file to save the blob to
```


aboodman:
WDYT about `noms blob get|put` instead of `import|export`?

jesseditson:
@aboodman piece of cake to change if you like, they both make sense to me!

cmd/noms/noms.go
```diff
@@ -31,6 +33,13 @@ var commands = []*util.Command{
 	nomsVersion,
 }
 
+type kCommandHandler func(input string) (exitCode int)
```
aboodman: How about `command` and `commandHandler`. The `k` prefix is frequently used as a nemonic to mean 'constant', so it's confusing to see it here.

go/util/profile/profile.go
```diff
@@ -30,6 +35,25 @@ func RegisterProfileFlags(flags *flag.FlagSet) {
 	}
 }
 
+func KAddProfileFlags(cmd *kingpin.CmdClause) {
```
aboodman: Same thing with the naming here, the `k` prefix is confusing. Can we just drop it?

cmd/noms/noms.go
```diff
@@ -86,7 +105,128 @@ func main() {
 			return
 		}
 	}
+}
+
+var (
```
aboodman: The globals here skeeve me out. How about having caller just track these pointers itself?

cmd/noms/noms_blob.go
```diff
@@ -0,0 +1,43 @@
...
+	return blob, func(input string) int {
+		applyVerbosity()
+		profile.KApplyProfileFlags()
+		switch input {
+		case blobImport.FullCommand():
+			return nomsBlobImport(*importFile, *importDs, *concurrency)
+		case blobExport.FullCommand():
+			return nomsBlobExport(*exportDs, *exportPath)
+		}
+		// unreachable
```
aboodman: We usually do `d.Panic("notreached")` in this case to make sure it doesn't happen.

aboodman:
OK let's do get/put instead. Few characters shorter :).

jesseditson:
done!

cmd/noms/noms.go
jesseditson: yeah was wondering if this was gonna look too hungarian. Was planning on removing once the kingpin stuff no longer needed to be differentiated, but I think this is a simple enough file to not need that.

cmd/noms/noms.go
jesseditson: Sounds good!

jesseditson:
@aboodman feedback addressed, ptal!

jesseditson:
The potentially controversial things I can think of are:

- can no longer do `--bool-flag=true`, must use `--bool-flag` instead (only an issue because it could break scripts)
- adds some dependencies (which is why this PR is 14k lines)
- kingpin uses a fluent API, which I haven't seen a ton of in go land (but I enjoy elsewhere)

but I think we've addressed those! I'll leave open for a bit and merge later or tomorrow

aboodman:
No need to wait. It passes tests, land it!

aboodman:
🚢 :shipit: 🚀 🎸 ▶️

PR #3623, opened 2017-08-18
Introduce noms-splore

kalman:
A lot of the JS code is taken from the old splore sample, but in
particular main.js is completely different - much simpler, because the
architecture of noms-splore uses a specialised {path => node} HTTP API,
implemented in Go, which does the noms graph traversal.

noms-splore also improves on the old splore sample by making it more
obvious what the key/value pairs are for maps and structs, but regresses
slightly in what it can say about prollytrees.


kalman:
@arv ptal, sorry for the nightmare
@aboodman you might want to look at this as the author of splore
@jesseditson fyi

kalman:
(Screenshot, but you really need to see it in action)

screen shot 2017-08-17 at 1 53 23 pm

kalman:
@jesseditson ptal

(I just wrote some more tests and made the chunk detection a little better)

cmd/noms/noms.go
```diff
@@ -203,13 +205,22 @@ See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spell
...
 `)
 	sync.Flag("parallelism", "").Short('p').Default("512").Int()
 	sync.Arg("source-object", "a noms source object").Required().String()
 	sync.Arg("dest-dataset", "a noms dataset").Required().String()
 
+	// splore
+	splore := noms.Command("splore", `Interactively explore a Noms database using a web browser.
+See Spelling Objects at https://github.com/attic-labs/noms/blob/master/doc/spelling.md for details on the path argument.
+`)
+	splore.Flag("port", "Server port. Defaults to a random port.").Short('p').Int()
```
jesseditson: random seems like an odd decision - why not something consistent?

cmd/noms/splore/js2go.py
```diff
@@ -0,0 +1,23 @@
...
+
+    sys.stdout.write('package %s\n\n' % (pkg,))
+    sys.stdout.write('const %s = "' % (varname,))
+    with open(jsf, 'r') as js:
+        sys.stdout.write(
+                js.read()
+                .replace('\\', '\\\\')
+                .replace('"', '\\"')
+                .replace('\n', '" +\n"'))
+    sys.stdout.write('"')
```
jesseditson: Seems like this could just as easily be done in a language we already require for this repo (js or go), rather than adding python as a new build dep - unless we already require it somewhere?

jesseditson: FWIW a build script for escaping also feels a bit smelly to me - perhaps something like rollup could do a better job packaging? Not understanding why this script is necessary

cmd/noms/splore/noms_splore.go
```diff
@@ -0,0 +1,307 @@
...
+
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		req(w, "text/html")
+		fmt.Fprintf(w, indexHtml, sp.String())
+	})
+
+	mux.HandleFunc("/out.js", func(w http.ResponseWriter, r *http.Request) {
+		req(w, "application/javascript")
+		// To develop JS, uncomment this line and run `yarn start`:
+		//http.ServeFile(w, r, "splore/out.js")
```
jesseditson: this is fine, but I believe webpack dev server was made to serve this purpose, might be worth considering if we actively develop splore in the future.

cmd/noms/noms.go
kalman: So that you can go
```
> noms splore -b /path/to/some/db &
> noms splore -b /path/to/some/other/db &
```
and it'll work.

jesseditson: why not lsof the port and increment, or require a port arg if not 8000 or whatever default is? Seems like it would be annoying to script this if the port is not dependable, maybe it's a moot point because of the `--browser` flag though

cmd/noms/splore/js2go.py
kalman: > unless we already require it somewhere

Yeah, we have Python elsewhere in the code (for most scrips).

> Not understanding why this script is necessary

I did spend a small amount of time researching how to do this. Basically I need to build resources into a .go binary, because distributing it as a separate resource is a non-starter. AFAIK there is no general JS -> Go way of doing this (and besides this script is super simple).

Perhaps a more idiomatic way of doing these things is to use a `go generate` script but given we need to use npm/yarn for building the JS anyway, it seemed better to stick with only a single build system.

cmd/noms/splore/noms_splore.go
kalman: Hm, interesting. Something like https://github.com/go-webpack/webpack?

Worth looking into for sure.

(though given the amount of time it took for me to get webpack to behave, I'm kinda scared of trying to introduce something else into the mix)

kalman:
Using 0 as the port is pretty standard I think. And yeah - the point is
that `noms splore` should be used with `--browser` like `noms show`. If
that ends up being annoying we can revisit, though going `-p9000` or
whatever isn't so bad.

kalman:
Thanks for the review @jesseditson. Regarding your comment about d3 - yeah, that's an interesting observation, the code here was taken from the original code that @aboodman wrote for the graph layout. Did you look into d3?

Jesse what kind of stuff does d3 get us (besides a maintained and familiar code base)? Can it do weirdo graph layout stuff like this>?

jesseditson:
I've used d3 extensively - it can literally draw anything, it outputs svg. The thing that makes it fit so well here is that its api is super well suited for adding/removing nodes and drawing trees quickly (it even outperforms react in this area). It also has a much easier to user animation/interaction API IMO for rich graphs like this. I believe the chart you're drawing is not too dissimilar from a [dendogram](https://bl.ocks.org/mbostock/4063570), but the same data could be represented using a [treemap](http://mbostock.github.io/d3/talk/20111018/treemap.html) or maybe even a [partition layout](http://mbostock.github.io/d3/talk/20111018/partition.html) - all of these can be drawn and made interactive trivially in d3.

kalman:
Nice! I love the idea of transitioning to d3 (agree on dendogram) and perhaps having different visualisations controlled by a command line flag?

jesseditson:
In fact I think this tree layout demo is almost exactly the same as splore, just with curvier lines: http://mbostock.github.io/d3/talk/20111018/tree.html

jesseditson:
For changing layout, yeah could do command line or just put a select dropdown in the UI, d3 is very fast at redraws and the data shape will be the same regardless

kalman:
Except noms isn't a tree, it's a DAG. I'd need to make sure it can handle DAGs.

aboodman:
I stole the layout from that d3 demo. There are very few dag layout
packages that are at all attractive and scale to very many nodes. So I
modified the tree layout to support dags.

aboodman:
Personally I don't find d3 pleasant to work in but if you all do that's
fine.

jesseditson:
I'd be interested in at least experimenting with it. D3's API is different than others but really good for exactly this kind of thing. I can't think of any reason why it would have issues with a DAG, it's totally free form. Would also be curious about something like this: https://github.com/attic-labs/noms/network

cmd/noms/splore/.flowconfig
```diff
@@ -0,0 +1,58 @@
+[ignore]
+; We fork some components by platform
```
arv: I know this is just a copy from our React Native app but this is not really relevant.

cmd/noms/splore/.flowconfig
```diff
@@ -0,0 +1,58 @@
...
+
+suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe\\($\\|[^(]\\|(\\(>=0\\.\\(4[0-5]\\|[1-3][0-9]\\|[0-9]\\).[0-9]\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)
+suppress_comment=\\(.\\|\n\\)*\\$FlowIssue\\((\\(>=0\\.\\(4[0-5]\\|[1-3][0-9]\\|[0-9]\\).[0-9]\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)?:? #[0-9]+
+suppress_comment=\\(.\\|\n\\)*\\$FlowFixedInNextDeploy
+suppress_comment=\\(.\\|\n\\)*\\$FlowExpectedError
+
+unsafe.enable_getters_and_setters=true
+
+[version]
+^0.45.0
```
arv: Can we use a newer version?

cmd/noms/splore/js2go.py
```diff
@@ -0,0 +1,23 @@
...
+
+def main():
+    if len(sys.argv) != 4:
+        print('usage: %s <source> <package> <varname>' % (sys.argv[0]))
+        return
+
+    jsf, pkg, varname = sys.argv[1:]
+
+    sys.stdout.write('package %s\n\n' % (pkg,))
+    sys.stdout.write('const %s = "' % (varname,))
```
arv: Why not use a backtick go string instead. Less escaping needed.

cmd/noms/splore/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": ["stage-3", "react"]
```
arv: I guess this works and you don't want to spend time on it?

The only transformers we really need at this point are Flow+JSX. Everything else is supported natively in Chrome.

cmd/noms/splore/noms_splore.go
```diff
@@ -0,0 +1,307 @@
...
+		req(w, "application/json")
+		r.ParseForm()
+		id := r.Form.Get("id")
+
+		var v types.Value
+		switch {
+		case id == "":
+			v = getValue()
+		case id[0] == '#':
+			abspath, err := spec.NewAbsolutePath(id)
```
arv: Just curious. Should we change the API to handle this?

cmd/noms/splore/noms_splore.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	return
+}
+
+func nodeName(v types.Value) string {
+	typeName := func(iface interface{}) string {
+		return reflect.TypeOf(iface).Name()
+	}
+
+	switch v := v.(type) {
```
arv: This is usually done by checking `Kind()`. You can also do `Kind().String()` to get the human readable name of the `NomsKind`.

cmd/noms/splore/noms_splore.go
```diff
@@ -0,0 +1,307 @@
...
+//   collection of mixed types, but heuristically assume that's it's chunked.
+func getMetaChildren(v types.Value) (children []nodeChild) {
+	var l uint64
+	if col, ok := v.(types.Collection); ok {
+		l = col.Len()
+	} else {
+		l = v.(types.Blob).Len()
+	}
+
+	vKind := types.TypeOf(v).Desc.Kind()
```
arv: This is doing a lot of extra work that is not needed. It has to create the `types.Type`. Just do `v.Kind()`.

cmd/noms/splore/noms_splore.go
```diff
@@ -0,0 +1,307 @@
...
+	})
+
+	if uint64(len(children)) >= l {
+		children = nil
+	}
+	return
+}
+
+func nodeHasChildren(v types.Value) bool {
+	switch k := types.TypeOf(v).Desc.Kind(); k {
```
arv: Same here.

cmd/noms/splore/.babelrc
kalman: I spent some time trying to make this less specialised, as in hours, but I couldn't make the combination of flow/eslint/webpack work, so I gave up. I.e. webpack kept crashing and/or eslint - I can't remember (besides which: given this calls `open` there is a fair chance it will open Safari - does that have the features too?).

cmd/noms/splore/.flowconfig
kalman: Yeah this was related to giving up trying to make this build sensible :-)

cmd/noms/splore/.flowconfig
kalman: Probably. Didn't want to rock the boat. This is copied from our RN app.

cmd/noms/splore/js2go.py
kalman: Because there are backticks in the JS source code, and in Go you can't have a backtick inside a backtick.

cmd/noms/splore/noms_splore.go
kalman: We might want to add a `spec.FigureOutPath` or something, but that would also need to give a clue into what it figured out (since we handle value vs absolute differently here, obviously).

cmd/noms/splore/noms_splore.go
kalman: I wanted to use the actual methods on the types, like `struct.Name()` etc.

cmd/noms/splore/noms_splore.go
kalman: Ah, thanks.

cmd/noms/splore/noms_splore.go
kalman: Will fix in a minute.

cmd/noms/splore/js2go.py
arv: Wow. That is pretty lame.

PR #3633, opened 2017-08-23
Pipe through ValueReader in ngql

arv:
This is in preparation for passing in a ValueReader to collection
constructors.

### DIFF
--- a/go/ngql/query_test.go
+++ b/go/ngql/query_test.go
@@ -1110,8 +1110,9 @@ func (suite *QueryGraphQLSuite) TestSetWithComplexKeys() {
 }
 
 func (suite *QueryGraphQLSuite) TestInputToNomsValue() {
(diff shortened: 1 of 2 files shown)


go/ngql/query_test.go
```diff
@@ -1110,8 +1110,9 @@ func (suite *QueryGraphQLSuite) TestSetWithComplexKeys() {
 }
 
 func (suite *QueryGraphQLSuite) TestInputToNomsValue() {
+	var vr types.ValueReader
```
rafael-atticlabs: can you just go ahead and create a test one here. Once my patch lands, it's going to be bad practice to ever hand in a null VRW.

arv: Oops... missed this comment. Will TBR a fix

PR #3637, opened 2017-08-24
Introduce ipfs-chat

willhite:


samples/go/ipfs-chat/ipfs-chatd/main.go
```diff
@@ -0,0 +1,75 @@
+package main
```
aboodman: I forgot the license header on some of these, can you add it?

PR #3639, opened 2017-08-24
Make field and index paths work for types

kalman:
Struct type field support field paths, like `.fieldName`
Compound types (List/Set/Map/Union) support index paths, like `[0]`

### DIFF
--- a/go/types/path_test.go
+++ b/go/types/path_test.go
@@ -129,7 +129,7 @@ func TestPathIndexType(t *testing.T) {
 	st := MakeSetType(NumberType)
 	lt := MakeListType(st)
 	mt := MakeMapType(st, lt)
-	ut := MakeUnionType(st, lt, mt)
+	ut := MakeUnionType(lt, mt, st)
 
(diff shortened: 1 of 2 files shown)


go/types/path_test.go
```diff
@@ -97,6 +123,50 @@ func TestPathIndex(t *testing.T) {
...
+	assertResolvesTo(assert, st, mt, "[0]")
+	assertResolvesTo(assert, st, mt, "[-2]")
+	assertResolvesTo(assert, lt, mt, "[1]")
+	assertResolvesTo(assert, lt, mt, "[-1]")
+	assertResolvesTo(assert, NumberType, mt, "[1][0][0]")
+	assertResolvesTo(assert, NumberType, mt, "@at(1)@at(0)@at(0)")
+	assertResolvesTo(assert, nil, mt, "[2]")
+	assertResolvesTo(assert, nil, mt, "[-3]")
+
+	// simplifyType is reordering the types from (st, lt, mt) to (lt, mt, st), so
```
arv: This is the order I expect 😛

PR #3641, opened 2017-08-25
Implement type sploring

kalman:
Display names are relatively similar to `noms show`.
I've also changed the display names for struct values to be more similar
to `noms show`.

### DIFF
--- a/cmd/noms/splore/noms_splore.go
+++ b/cmd/noms/splore/noms_splore.go
@@ -149,8 +149,8 @@ func run(mux *http.ServeMux, port int, browser bool, spStr string) int {
 }
 
 func getNodeChildren(v types.Value, parentPath string) (children []nodeChild) {
-	atPath := func(i int, suffix string) string {
-		return fmt.Sprintf("%s@at(%d)%s", parentPath, i, suffix)
+	childPath := func(f string, args ...interface{}) string {
+		return parentPath + fmt.Sprintf(f, args...)
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/noms/splore/noms_splore.go
```diff
@@ -191,25 +191,47 @@ func getNodeChildren(v types.Value, parentPath string) (children []nodeChild) {
...
+			i := 0
+			d.IterFields(func(name string, t *types.Type, optional bool) {
+				children[i] = nodeChild{
+					Label: name,
+					Value: info(t, childPath(".%s", name)),
+				}
+				i++
+			})
+		default:
+			children = []nodeChild{} // TODO: cycles?
```
kalman: I would test this but the place I usually get cycles - from the photo data - is no longer compatible with this noms version. Will fix later.

arv: Cycles are represented as pointer cycles in memory so you would keep getting back to the same struct type over and over.

arv: It is easy enough to create a cyclic type programmatically of you want to test this.

kalman: > you would keep getting back to the same struct 

Yep I see this with Commit. It's not a big deal. I think the graph library we use would get confused trying to draw a cyclic graph, so shrug

> want to test this 

Yeah, I'm just not sure what I want the behavior to be yet, I guess.

PR #3644, opened 2017-08-29
Print ref with pound

arv:
Print Ref values as `#123` instead of `123`

Since our hashes are SHA-512 and we write them using Base32 there are a lot of overlaps with other parts of NomDL. This makes them unambiguous.

Towards #1466


arv:
First commit is #3643

PR #3647, opened 2017-08-29
HRS: Cleanup struct head

arv:
For struct values we always prefix with `struct `.

For struct types we always prefix with `Struct `.

Towards #1466


arv:
Only last commit

PR #3658, opened 2017-08-31
NBS: Fragmentation tool using new estimate of locality

cmasone-attic:
The new version of this tool now estimates the locality of a DB
written using the "grandchild" strategy implemented by
types.ValueStore. It does do by dividing each level of the graph
up into groups that are roughly the size of the branching factor
of that level, and then calculating how many physical reads are
needed to read each group.

In the case of perfect locality, each group could be read in a
single physical read, so that's what the tool uses as its estimate
of the optimal case.

Toward #2968

### DIFF
--- a/go/nbs/frag/main.go
+++ b/go/nbs/frag/main.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"sync"
 
-	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/datas"
 	"github.com/attic-labs/noms/go/hash"
 	"github.com/attic-labs/noms/go/nbs"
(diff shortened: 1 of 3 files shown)


go/nbs/table_set.go
```diff
@@ -86,16 +86,17 @@ func (ts tableSet) getMany(reqs []getRecord, foundChunks chan *chunks.Chunk, wg
 }
 
 func (ts tableSet) calcReads(reqs []getRecord, blockSize uint64) (reads int, split, remaining bool) {
-	f := func(css chunkSources) (reads int, split, remaining bool) {
+	f := func(css chunkSources) (int, bool, bool) {
```
cmasone-attic: This was busted, so I fixed it as a part of this work

cmasone-attic:
PTAL. New sync strategy in a separate patch. Seems to do WAY better than the current version, at least by the metric provided by this tool :-)

go/nbs/frag/main.go
```diff
@@ -63,72 +62,85 @@ func main() {
...
-						for _, r := range children {
-							refs <- r
-						}
-					}()
+		// Iterate all the Values at the current level of the graph IN ORDER (as specified in |current|) and gather up their embedded refs. We'll build two different lists of hash.Hashes during this process:
+		// 1) An ordered list of ALL the children of the current level.
+		// 2) An ordered list of the child nodes that contain refs to chunks we haven't yet visited. This *excludes* already-visted nodes and nodes without children.
+		// We'll use 1) to get an estimate of how good the locality is among the children of the current level, and then 2) to descend to the next level of the graph.
+		orderedChildren := hash.HashSlice{}
+		nextLevel := hash.HashSlice{} // TODO: re-use |current| to save allocations?
```
rafael-atticlabs: nah.

go/nbs/table_reader.go
```diff
@@ -389,6 +389,7 @@ func canReadAhead(fRec offsetRec, fLength uint32, readStart, readEnd, blockSize
 		return readEnd, true
 	}
 
+	// Is this right? This seems like it means that we're willing to keep reading ahead as long as the next chunk is < blockSize away from the end of the last chunk. That may be what we want, but then it's not really a "block size". It's more like...the amount of overhead we're willing to tolerate per chunk read.
```
rafael-atticlabs: real comment

PR #3659, opened 2017-08-31
Stop noms sync from destroying locality

### LINKED ISSUE #2968: (NBS) *Prevent* noms sync from actively *hurting* chunk locality

cmasone-attic:
This patch implements a new strategy for Pull() that pulls the chunks
from a given level of the graph over in the order they'll be
encountered by clients reading the graph.

Fixes #2968

### DIFF
--- a/go/datas/pull.go
+++ b/go/datas/pull.go
@@ -40,18 +40,16 @@ func Pull(srcDB, sinkDB Database, sourceRef types.Ref, progressCh chan PullProgr
 	}
 	var sampleSize, sampleCount uint64
 
-	absent := hash.NewHashSet(sourceRef.TargetHash())
+	absent := hash.HashSlice{sourceRef.TargetHash()}
 	for len(absent) != 0 {
 		updateProgress(0, uint64(len(absent)), 0)
 
 		// Concurrently pull all the chunks the sink is missing out of the source
-		// For each chunk, put it into the sink, then decode it into a value so we can later iterate all the refs in the chunk.
-		absentValues := types.ValueSlice{}
-		foundChunks := make(chan *chunks.Chunk)
-		go func() { defer close(foundChunks); srcDB.chunkStore().GetMany(absent, foundChunks) }()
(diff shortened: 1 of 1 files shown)


cmasone-attic:
PTAL.

The workload I tried this on was using `csv-import` to import good ol' `sfcrime.csv`, and then syncing it to a new DB using ToT and the new code. In the former case, the frag tool at https://github.com/attic-labs/noms/pull/3658 reported that one would need almost 18x the optimal number of reads to get the data out. With the new code, the new frag tool reported near-optimal results!

rafael-atticlabs:
Awesome to see this finally closed, BTW. \o/

Issue #3619, opened 2017-08-16
GraphQL resolution of collection ranges should use exact loading


rafael-atticlabs:
Right now, when requesting a range of values from a List, Set or Map, an iterator is created at the start position and advanced until the end position is reached. E.g. 

https://github.com/attic-labs/noms/blob/master/go/ngql/types.go#L442

Iterators perform no read ahead, so this effectively serializes loading of all chunks needed to service this request (and is slow, obviously).

`loadLeafSequences` was recently added (https://github.com/attic-labs/noms/blob/master/go/types/blob.go#L60) and allows eagerly loading an exact set of leaf sequences which contain a range of indexed values.

It should be straightforward to directly use this to load indexed ranges of values for List/Map/Set, and an analagous mechanism can work for loading keyed ranges of values for Map/Set.

phritz:
@rafael-atticlabs I assigned this to me but if it is really P0 then someone else should probably take it. I'm doing some research today and tomorrow for aaron and won't get to it until next week. (Or feel free to tell me it's super important to prioritize this bug over everything else -- P0 in other teams I've worked with means drop everything and fix it.)

rafael-atticlabs:
There's no fire. Right now `P0` just means "candidate for inclusion in Noms V8"

PR #3660, opened 2017-09-01
request set & list elements in batch

phritz:
Load all leaf nodes for sets and lists before iterating them. Previously it loaded them on demand while iterating, meaning access to the underlying chunks was serialized. 

@rafael-atticlabs there are some questions in comments for you

This the first part of what's required for toward https://github.com/attic-labs/noms/issues/3619


rafael-atticlabs:
Well done, @phritz. In general, this looks good. I'm going to defer review to @arv, since this is his code area, and let him answer most of the questions.

About tree level, there are tests in types that keep adding elements until `col.sequence().treeLevel() > 0`, but that's not public API and available from this package. One route would be to add some public API. Lacking public API to directly return the tree level, you can also tell when a collection is more than one level deep by calling `WalkRefs` (which will start visiting leaf-nodes as soon as it's more than one level).

phritz:
> About tree level, there are tests in types that keep adding elements until
> col.sequence().treeLevel() > 0, but that's not public API and available 
> from this package. One route would be to add some public API

That's fine I don't need an API, was just wondering if there was a different way than brute force.

go/types/indexed_sequences.go
```diff
@@ -83,7 +83,7 @@ func orderedKeyFromSum(msd []metaTuple) orderedKey {
 
 // loads the set of leaf nodes which contain the items [startIdx -> endIdx).
```
arv: Can you make this a valid GoDoc comment?

go/ngql/types.go
```diff
@@ -419,32 +419,41 @@ func getListElements(vrw types.ValueReadWriter, v types.Value, args map[string]i
...
 	}
 
 	values := make([]interface{}, count)
 
-	// This is a workaround for allowing types.Set as types.List.
-	type listOrSetIterator interface {
-		Next() types.Value
-	}
+	// @raf a few questions:
+	//  - why do we do MaybeGetScalar below?
```
arv: This is so that the GraphQL query/request bottoms out. The scalars (bool, float64, string) are the base types that the graphql library uses to know that no more resolving is needed.

go/ngql/types.go
```diff
@@ -419,32 +419,41 @@ func getListElements(vrw types.ValueReadWriter, v types.Value, args map[string]i
...
+	//  - in the unittests how could you force the collection to have tree height > 1?
+	//  - where does the rejiggering of the tree height happen?
```
arv: Not fully sure what you mean here? The sequence chunker builds and mutates prollytrees... it keeps track of the tree level.

go/ngql/types.go
```diff
@@ -419,32 +419,41 @@ func getListElements(vrw types.ValueReadWriter, v types.Value, args map[string]i
...
+	//  - i just ran the unit tests; what else should i do to make sure this change is WAI?
```
arv: There are some tests that keep track of number of reads and writes to a chunk store. Maybe you could keep track of the number of reads to see that we read all the chunks even if exit early (can we exit early?)

Checking that we do not read sequentially seems flaky since even with random order sequential order is possible so that does not seem viable...

arv: Oh! You could use a ValueReadWriter that logs the number of calls to `ReadManyValues`. That used to be 0 in this scenario but now should be 1.

go/ngql/types.go
```diff
@@ -419,32 +419,41 @@ func getListElements(vrw types.ValueReadWriter, v types.Value, args map[string]i
...
+		if elementsSeen >= offset && elementsSeen-offset < uint64(count) {
+			values[elementsSeen-offset] = MaybeGetScalar(v)
+		}
+		elementsSeen++
 	}
-	for i := uint64(0); i < uint64(count); i++ {
-		values[i] = MaybeGetScalar(iter.Next())
+	for _, c := range cols {
+		v := c.(types.Value)
+		v.WalkValues(maybeAddElement)
```
arv: This is fine for now but can you add a `TODO: Use a cursor so we do not have to instantiate all values in the future`. This is because I'm working on a change that only creates Values as needed (the collection will be backed by a `[]byte`)

go/types/indexed_sequences.go
phritz: done

go/ngql/types.go
phritz: done

PR #3661, opened 2017-09-03
Rename --summarize flag to --stat

### LINKED ISSUE #3526: Rename the --summarize flag to --stat
It should clearly be `--stat` - that is what the `git` flag that does the same thing is called and what I always think ours is called.

iand:
Fixes #3526 

My first PR here, so go easy :) 

I have signed the contributor agreement.


aboodman:
Looks great. Thanks for the contribution.

PR #3663, opened 2017-09-03
Bring noms-tour up to date with recent changes

iand:
Some of the Go examples in the tour were broken so I've updated them to working versions. The one with the largest change is appending to a list where I've used a list editor which I think is the replacement for the Append method which has been removed from lists.


aboodman:
Thanks!

PR #3664, opened 2017-09-03
Use ETag/If-None-Match in url-fetch

### LINKED ISSUE #2086: Use ETag/If-None-Match in url-fetch
Socrata sets the ETag header, so we can save downloading the full CSV file if it hasn't changed (which seems to usually be the case).

Generally speaking, url-fetch should store any `ETag` header - presumably in the commit metadata - then set the `If-None-Match` header if it's present.

In bizarro pseudocode:

```
path, url := ...
head, meta := GetPath(path)
headers = [];
if etag, ok := meta.Get("ETag"); ok {
  headers = append(headers, "If-None-Match: \"%s\"" % etag)
}
resp := RequestURL(headers)
if resp.Code == 304 {
  print("Cache hit!")
  return
}
...
meta = meta.Set("ETag", resp.GetHeader("ETag"))
Commit(NewBlob(resp.Value), meta)
```

iand:
Fixes #2086

### DIFF
--- a/samples/go/url-fetch/fetch.go
+++ b/samples/go/url-fetch/fetch.go
@@ -15,6 +15,7 @@ import (
 	"github.com/attic-labs/noms/go/config"
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/datas"
+	"github.com/attic-labs/noms/go/marshal"
 	"github.com/attic-labs/noms/go/spec"
 	"github.com/attic-labs/noms/go/types"
 	"github.com/attic-labs/noms/go/util/exit"
@@ -58,7 +59,22 @@ func main() {
 	var r io.Reader
 	var contentLength int64
 
-	additionalMetaInfo := make(map[string]string)
+	var root = struct {
+		Meta struct {
+			Etag string `noms:"etag,omitempty"`
+			File string `noms:"file,omitempty"`
+			URL  string `noms:"url,omitempty"`
+		}
+	}{}
+	if ds.HasHead() {
+		err = marshal.Unmarshal(ds.Head(), &root)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Could not unmarshal head: %s\n", err)
+			return
+		}
+	}
+
+	additionalMetaInfo := map[string]string{}
 	if *stdin {
 		r = os.Stdin
 		contentLength = -1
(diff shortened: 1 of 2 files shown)


samples/go/url-fetch/fetch.go
```diff
@@ -57,18 +57,36 @@ func main() {
...
 		contentLength = -1
 	} else if url := flag.Arg(0); strings.HasPrefix(url, "http") {
-		resp, err := http.Get(url)
+		req, err := http.NewRequest(http.MethodGet, url, nil)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Could not build http request for url %s, error: %s\n", url, err)
+			return
+		}
+
+		if head, ok := ds.MaybeHead(); ok {
```
aboodman: A common pattern in our code is something like:

```
var root = struct {
  Meta struct {
    Etag string `noms:"etag,omitempty"`
    File string `noms:"file,omitempty"`
    URL string `noms:"url,omitempty"`
  }
}{}
if ds.HasHead() {
  err = marshal.Unmarshal(db.Head(), &root)
  ...
}
```

samples/go/url-fetch/fetch.go
```diff
@@ -77,7 +95,10 @@ func main() {
 
 		r = resp.Body
 		contentLength = resp.ContentLength
-		sourceType, sourceVal = "url", url
+		additionalMetaInfo["url"] = url
```
aboodman: Here you could just use `root.Meta` from above suggestion.

samples/go/url-fetch/fetch.go
```diff
@@ -103,13 +124,9 @@ func main() {
...
 
 	if *performCommit {
-		var additionalMetaInfo map[string]string
-		if sourceType != "" {
-			additionalMetaInfo = map[string]string{sourceType: sourceVal}
-		}
 		meta, err := spec.CreateCommitMetaStruct(db, "", "", additionalMetaInfo, nil)
 		d.CheckErrorNoUsage(err)
-		ds, err = db.Commit(ds, b, datas.CommitOptions{Meta: meta})
+		_, err = db.Commit(ds, b, datas.CommitOptions{Meta: meta})
```
aboodman: ... and here do `Meta: marshal.MustMarshal(root.Meta)` ... i think?

samples/go/url-fetch/fetch.go
iand: I just tried that out and found I'd have to blank out some fields. For example if you've already fetched a file then fetch a url, root.Meta.File will contain a value that needs to be removed.

Also spec.CreateCommitMetaStruct adds some other info such as the commit date which I'd have to replicate. 

Because of this I decided to leave the additionalMetaInfo map in place. Let me know if I'm misunderstanding.

iand:
PTAL

samples/go/url-fetch/fetch.go
aboodman: Makes sense. It's a shame it's not easier to use this pattern with the meta fields. We should go back through and clean up the API.

aboodman:
Thanks for all the patches @iand. I'd love to know what you're up to with Noms and if there's any way we can help. Find me in our slack (http://slack.noms.io/) if you have time, or email me: aa at attic dot it.

PR #3666, opened 2017-09-04
Modifications to ipfs-chat can ipfs chunkstore

willhite:
* Change ipfs paths to include directory where ipfs repo is stored.
 * Rework ipfs-chat to create ipfs chunkstores manually rather than
   relying on Spec.ForDataset. This enables creating two chunkstores
   (one local and one network) using the same IpfsNode (ipfs repo).
 * Create separate replicate function for daemon and mergeMessage
   function for client to experiment with slightly different behaviors
   for each.

### DIFF
--- a/go/ipfs/cs.go
+++ b/go/ipfs/cs.go
@@ -17,7 +17,6 @@ import (
 	"github.com/attic-labs/noms/go/chunks"
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
-	"github.com/ipfs/go-ipfs/Godeps/_workspace/src/github.com/mitchellh/go-homedir"
 	"github.com/ipfs/go-ipfs/blocks/blockstore"
 	"github.com/ipfs/go-ipfs/blockservice"
 	"github.com/ipfs/go-ipfs/core"
@@ -26,37 +25,50 @@ import (
 	"github.com/ipfs/go-ipfs/repo/fsrepo"
 )
 
-var (
-	CurrentNode *core.IpfsNode
-
-	// Ugg, such a hack. Setting this to an integer >= 0 ** <= 9 will cause
-	// NewChunkStore to alter the api, http, and swarm ports for this IPFS node
-	// to get reset to a port ending in that digit.
-	NodeIndex int = -1
-)
-
 // NewChunkStore creates a new ChunkStore backed by IPFS.
 //
 // Noms chunks written to this ChunkStore are converted to IPFS blocks and
 // stored in an IPFS BlockStore.
 //
-// The name distinguishes this chunkstore on the local machine. A corresponding
-// file is created in ~/.noms/ipfs which stores the root of the noms database.
-// This should ideally be done with IPNS, but that is currently too slow to be
-// practical.
+// Noms repositories that are build on top of IPFS are identified by paths of
+// the following form:
+//   ipfs://<path-to-ipfs-dir>
+// where 'ipfs' indicates the noms protocol and the path indicates the path to
+// the directory where the ipfs repo resides. The chunkstore creates two files
+// in the ipfs directory called 'noms' and 'noms-local' which stores the root
+// of the noms database. This should ideally be done with IPNS, but that is
+// currently too slow to be practical.
 //
 // This function creates an IPFS repo at the appropriate path if one doesn't
-// already exist. The default location is ~/.ipfs, but this can be configured with
-// the IPFS_PATH environment variable, or IPFS_LOCAL_PATH if "local" is true. If
-// the global NodeIndex variable has been set to a number between 0 and 9
-// inclusive, the api, http, and swarm ports will be modified to end in with
-// that digit. (e.g. if NodeIndex == 3, API port will be set to 5003)
+// already exist. If the global NodeIndex variable has been set to a number
+// between 0 and 9 inclusive, the api, http, and swarm ports will be modified to
+// end in with that digit. (e.g. if NodeIndex == 3, API port will be set to 5003)
 //
 // If local is true, only the local IPFS blockstore is used for both reads and
 // write. If local is false, then reads will fall through to the network and
 // blocks stored will be exposed to the entire IPFS network.
-func NewChunkStore(name string, local bool) *chunkStore {
-	p := getIPFSDir(local)
+func NewChunkStore(p string, local bool) *chunkStore {
+	node := OpenIPFSRepo(p, -1)
+	return NewChunkStorePrimitive(p, local, node)
+}
+
+// Creates a new chunchStore using a pre-existing IpfsNode. This is currently
+// used to create a second 'local' chunkStore using the same IpfsNode as another
+// non-local chunkStore.
+func NewChunkStorePrimitive(p string, local bool, node *core.IpfsNode) *chunkStore {
+	return &chunkStore{
+		node:      node,
+		name:      p,
+		local:     local,
+		rateLimit: make(chan struct{}, 1024),
+	}
+}
+
+// Opens a pre-existing ipfs repo for use as a noms store. This function will
+// create a new IPFS repos at this indictated path if one doesn't already exist.
+// Also if portIdx is a number between 0 and 1 inclusive, the config file will
+// be modified to change external facing port numbers to end in 'portIdx'.
+func OpenIPFSRepo(p string, portIdx int) *core.IpfsNode {
 	r, err := fsrepo.Open(p)
 	if _, ok := err.(fsrepo.NoRepoError); ok {
 		var conf *config.Config
@@ -68,7 +80,7 @@ func NewChunkStore(name string, local bool) *chunkStore {
 	}
 	d.CheckError(err)
 
-	resetRepoConfigPorts(r)
+	resetRepoConfigPorts(r, portIdx)
 
 	cfg := &core.BuildCfg{
 		Repo:   r,
@@ -78,15 +90,9 @@ func NewChunkStore(name string, local bool) *chunkStore {
 		},
 	}
 
-	CurrentNode, err = core.NewNode(context.Background(), cfg)
+	node, err := core.NewNode(context.Background(), cfg)
 	d.CheckError(err)
-
-	return &chunkStore{
-		node:      CurrentNode,
-		name:      name,
-		local:     local,
-		rateLimit: make(chan struct{}, 1024),
-	}
+	return node
 }
 
 type chunkStore struct {
@@ -242,7 +248,7 @@ func (cs *chunkStore) Version() string {
 func (cs *chunkStore) Rebase() {
 	h := hash.Hash{}
 	var sp string
-	f := cs.getLocalNameFile(cs.name)
+	f := cs.getLocalNameFile(cs.local)
 	b, err := ioutil.ReadFile(f)
 	if !os.IsNotExist(err) {
 		d.PanicIfError(err)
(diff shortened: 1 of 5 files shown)


go/ipfs/cs.go
```diff
@@ -26,37 +25,50 @@ import (
...
 // Creates a new ChunkStore backed by IPFS.
 //
 // Noms chunks written to this ChunkStore are converted to IPFS blocks and
 // stored in an IPFS BlockStore.
 //
-// The name distinguishes this chunkstore on the local machine. A corresponding
-// file is created in ~/.noms/ipfs which stores the root of the noms database.
-// This should ideally be done with IPNS, but that is currently too slow to be
-// practical.
+// Noms repositories that are build on top of IPFS are identified by paths of
```
aboodman: IPFS database specs have the form: `ipfs://...`

go/ipfs/cs.go
```diff
@@ -26,37 +25,50 @@ import (
...
-	p := getIPFSDir(local)
+func NewChunkStore(p string, local bool) *chunkStore {
+	node := OpenIPFSRepo(p, -1)
+	return NewChunkStorePrimitive(p, local, node)
+}
+
+// Creates a new chunchStore using a pre-existing IpfsNode. This is currently
+// used to create a second 'local' chunkStore using the same IpfsNode as another
+// non-local chunkStore.
+func NewChunkStorePrimitive(p string, local bool, node *core.IpfsNode) *chunkStore {
```
aboodman: How about `ChunkStoreFromIPFSNode`.

samples/go/ipfs-chat/main.go
```diff
@@ -55,10 +58,8 @@ func main() {
 	daemonCmd := kingpin.Command("daemon", "runs a daemon that simulates filecoin, eagerly storing all chunks for a chat")
 	daemonTopic := daemonCmd.Flag("topic", "IPFS pubsub topic to publish and subscribe to").Default("ipfs-chat").String()
 	daemonInterval := daemonCmd.Flag("interval", "amount of time to wait before publishing state to network").Default("5s").Duration()
-	netNodeIdx := daemonCmd.Flag("net-node-idx", "a single digit to be used as last digit in all port values: api, gateway and swarm (must be 0-9 inclusive)").Default("-1").Int()
-	localNodeIdx := daemonCmd.Flag("local-node-idx", "a single digit to be used as last digit in all port values: api, gateway and swarm (must be 0-9 inclusive)").Default("-1").Int()
-	daemonNetworkDS := daemonCmd.Arg("network-dataset", "the dataset spec to use to read and write data to the IPFS network").Required().String()
-	daemonLocalDS := daemonCmd.Arg("local-dataset", "the dataset spec to use to read and write data locally").Required().String()
+	daemonNodeIdx := daemonCmd.Flag("node-idx", "a single digit to be used as last digit in all port values: api, gateway and swarm (must be 0-9 inclusive)").Default("-1").Int()
```
aboodman: Why not just default it to zero so that you can use its default value, rather than testing whether it has the default value?

samples/go/ipfs-chat/main.go
```diff
@@ -69,19 +70,27 @@ func main() {
...
 }
 
 func runClient(username, topic, clientDS string, nodeIdx int) {
 	var displayingSearchResults = false
 
-	ipfs.NodeIndex = nodeIdx
 	dsSpec := clientDS
 	sp, err := spec.ForDataset(dsSpec)
 	d.CheckErrorNoUsage(err)
+	fmt.Println("sp.Path:", sp.Path)
```
aboodman: Remove debug prints

samples/go/ipfs-chat/main.go
```diff
@@ -491,3 +498,21 @@ func expandRLimit() {
 	d.Chk.NoError(err)
 	fmt.Println("current thread limit:", rLimit)
 }
+
+func reconfigureIPFSChunkStore(sp spec.Spec, topic string, nodeIdx int) (*core.IpfsNode, chunks.ChunkStore) {
+	ds := sp.GetDataset()
+	sp.Close()
+
+	// recreate database so that we can have control of chunkstore's ipfs node
+	sp.Close()
```
aboodman: Duplicate line

samples/go/ipfs-chat/main.go
```diff
@@ -491,3 +498,21 @@ func expandRLimit() {
 	d.Chk.NoError(err)
 	fmt.Println("current thread limit:", rLimit)
 }
+
+func reconfigureIPFSChunkStore(sp spec.Spec, topic string, nodeIdx int) (*core.IpfsNode, chunks.ChunkStore) {
+	ds := sp.GetDataset()
```
aboodman: Why bother getting the dataset if you're just going to close it immediately? `sp` already has `DatabaseName` parsed, which is all you need.

samples/go/ipfs-chat/main.go
```diff
@@ -491,3 +498,21 @@ func expandRLimit() {
 	d.Chk.NoError(err)
 	fmt.Println("current thread limit:", rLimit)
 }
+
+func reconfigureIPFSChunkStore(sp spec.Spec, topic string, nodeIdx int) (*core.IpfsNode, chunks.ChunkStore) {
```
aboodman: how about `initChunkStore`.

samples/go/ipfs-chat/main.go
```diff
@@ -491,3 +498,21 @@ func expandRLimit() {
...
+	ds := sp.GetDataset()
+	sp.Close()
+
+	// recreate database so that we can have control of chunkstore's ipfs node
+	sp.Close()
+	node := ipfs.OpenIPFSRepo(sp.DatabaseName, nodeIdx)
+
+	cs := ipfs.NewChunkStorePrimitive(sp.DatabaseName, sp.Protocol == "ipfs-local", node)
+	db := datas.NewDatabase(cs)
```
aboodman: `db` and `ds` are not used.

samples/go/ipfs-chat/pubsub.go
```diff
@@ -7,46 +7,55 @@ package main
...
-	"gx/ipfs/QmddZj8gx1Ceisj3QEWeAzPEjCPWpXzneN1ANdkZdwUdc3/floodsub"
+	"github.com/ipfs/go-ipfs/core"
 )
 
-func Replicate(sub *floodsub.Subscription, source, dest datas.Dataset, didChange func(ds datas.Dataset)) {
-	lastHash := ""
+// MergeMessages continually listens for commit hashes published by ipfs-chat. It
+// merges new messages into it's existing dataset when necessary and if an actual
+// merge was necessary, it re-publishes the new commit.
+func MergeMessages(node *core.IpfsNode, topic string, ds datas.Dataset, didChange func(ds datas.Dataset)) {
```
aboodman: This function doesn't need to be public.

samples/go/ipfs-chat/daemon.go
```diff
@@ -5,49 +5,136 @@
...
-		Publish(sub, topic, s)
+		Publish(node, topic, destDS.HeadRef().TargetHash())
 		time.Sleep(interval)
 	}
 }
+
+// replicate continually listens for commit hashes published by ipfs-chat nodes,
+// ensures that all nodes are replicated locally, and merges new data into it's
+// dataset when necessary.
+func replicate(node *core.IpfsNode, topic string, sourceDS, destDS datas.Dataset, didChange func(ds datas.Dataset)) {
```
aboodman: What's the difference between this and `MergeMessages`? It looks like one calls `Pull` and the other doesn't. If that's it, or if they are very close they should probably share code so that they don't drift apart.

samples/go/ipfs-chat/daemon.go
```diff
@@ -5,49 +5,136 @@
...
+
+		destDS, err = destDB.SetHead(destDS, destDB.WriteValue(datas.NewCommit(merged, types.NewSet(destDB, destDS.HeadRef(), sourceRef), types.EmptyStruct)))
+		if err != nil {
+			dbg.Debug("call failed to SetHead on destDB, err: %s", err)
+		}
+		didChange(destDS)
+	}
+}
+
+func pullCommits(h hash.Hash, sourceDB, destDB datas.Database) {
```
aboodman: Does this seem to work?

aboodman:
Also see https://docs.google.com/document/d/1n_-qg1k7yE_OwEeoVFaKH3P1fDT3yldpqcwxNg3GrFo/edit#heading=h.tp4acv9ag0jo. After some thought, I don't think it's worth pushing harder on this direction. It seems like it will continue to be flakey.

go/ipfs/cs.go
willhite: Done

go/ipfs/cs.go
willhite: Done

samples/go/ipfs-chat/daemon.go
```diff
@@ -5,10 +5,16 @@
 package main
 
 import (
+    "bytes"
```
aboodman: formatting

samples/go/ipfs-chat/daemon.go
willhite: I agree, they're pretty similar but at this point I think it's helpful to be able to work with them independently until we find the magic sauce. Then we can merge them again.

samples/go/ipfs-chat/daemon.go
willhite: It hasn't been tested thoroughly and still seems flakey. A work in progress.

samples/go/ipfs-chat/main.go
willhite: I would do that if the default configuration had all ports ending in 0. The default config has a mixture of 0's and 1's and so I wanted to leave the defaults unchanged unless a portIdx was specified.

samples/go/ipfs-chat/main.go
willhite: Done. I added more dbg.Debug() calls though.

samples/go/ipfs-chat/main.go
willhite: I like it, done.

samples/go/ipfs-chat/main.go
willhite: Done. Much better.

samples/go/ipfs-chat/main.go
willhite: Removed.

samples/go/ipfs-chat/main.go
willhite: Cleaned up.

samples/go/ipfs-chat/pubsub.go
willhite: Changed.

samples/go/ipfs-chat/daemon.go
```diff
@@ -20,121 +26,186 @@ import (
...
+    nwDB := datas.NewDatabase(nwCS)
+
+    // Use the same ipfs node to create a second chunkstore that restricts data
+    // access to it's local blockstore
+	csLocal := ipfs.ChunkStoreFromIPFSNode(sp.DatabaseName, true, node)
+	localDB := datas.NewDatabase(csLocal)
+	localDS := localDB.GetDataset(sp.Path.Dataset)
+    
+    // Initial the database, if necessary, to an empty commit. Committing to
+    // the local database will also reset the head for the network database.
```
aboodman: Really? I thought they used different files to track the heads.

samples/go/ipfs-chat/daemon.go
```diff
@@ -20,121 +26,186 @@ import (
...
 
-func pullCommits(h hash.Hash, sourceDB, destDB datas.Database) {
-	dbg.Debug("pullCommits, h: %s", h)
-
-	// read
-	v := destDB.ReadValue(h)
-	if v != nil {
-		dbg.Debug("pullCommits, found h: %s, commit:", h, types.EncodedValueMaxLines(v, 10))
+// One of the problem that this daemon currently has is that it hangs which
+// effectively stops are 'replicate' loop. I can't figure out why that is
```
aboodman: Grammar here doesn't make sense.

samples/go/ipfs-chat/daemon.go
```diff
@@ -20,121 +26,186 @@ import (
...
-	types.EncodedValue(v)
-	dbg.Debug("pullCommits, read h: %s, commit:", h, types.EncodedValueMaxLines(v, 10))
+    }
+    
+    dbg.Debug("pullCommits, not found in local, reading from net h: %s", h)
+    v := netDB.ReadValue(h)
+    d.Chk.NotNil(v)
+    
+    dbg.Debug("pullCommits, encoding value from net, h: %s", h)
+    s1 := types.EncodedValue(v)
```
aboodman: EncodedValue is pretty heavy (it generates a huge string). You could accomplish same thing by calling `value.WalkRefs() -> ref.TargetValue() -> value.WalkRefs()` recursively.

aboodman: I also continue to feel like you could generalize this and optimize it at the same time by using pinned-ness rather than present-in-local-db-ness.

So basically this would recurse an arbitrary noms ref (typically starting at a commit) reading and pinning all its child refs. If it ever finds a ref that is already pinned it can stop.

Then I think you might not even need `ipfs-local` at all?

aboodman:
I'm just going to land it and land another PR w/ my comments after.

PR #3671, opened 2017-09-06
gofmt

aboodman:
TBR


aboodman:
I didn't get to my comments about using pinned-ness rather than present-in-local-db-ness, and removing ipfs-local. I'll look into that and do some testing tonight.

PR #3673, opened 2017-09-07
Update vision.md

aboodman:


aboodman:
OK I don't know how to do that in the PR editor, so will do it separately.

PR #3675, opened 2017-09-07
Add --lowercase option to map column names to lowercase struct names

ehalpern:
By default, each column name maps to a struct field preserving the original case.
If --lowercase is specified the resulting struct fields will always be lowercase.


ehalpern:
This feature is motivated by the work to import taxi data (http://www.nyc.gov/html/tlc/html/about/trip_record_data.shtml) into noms. The case conventions for csv columns are inconsistent across files. The best way to avoid complication and false duplicate fields is to normalize to lowercase.

cc: @cmasone-attic

cmasone-attic:
lgtm

PR #3676, opened 2017-09-07
Add instructions for running ipfs-chat demo

willhite:


willhite:
@phritz changed this to say 'Clone the noms repository' PTAL

PR #3677, opened 2017-09-07
Use stretchr/testify not attic-labs/testify

kalman:
stretchr has fixed a bug with the -count flag. I could merge these
changes into attic-labs, but it's easier to just use strechr.

We forked stretchr a long time ago so that we didn't link in the HTTP
testing libraries into the noms binaries (because we were using d.Chk in
production code). The HTTP issue doesn't seem to happen anymore, even
though we're still using d.Chk.


kalman:
CC @arv but might as well TBR this once Jenkins is happy

kalman:
jenkins: test this

kalman:
:-\

kalman:
Well, this passes, which was basically the point of this patch:
```
go test ./cmd/... ./go/... ./samples/... ./tools/... -count 2
```

arv:
LGTM

Issue #3668, opened 2017-09-05
Add noms.To() and noms.From() helpers


aboodman:
It is incredibly common to marshal to and from Go types. Common enough that `marshal.(Must)?(Un)?Marshal is cumbersome to type. We should put these at the top level and make them super short.

PR #3687, opened 2017-09-08
remove no-longer-needed extraneous subdirs

wardn:


arv:
Thanks for the PR.

Can this wait? I have a huge patch in progress which touches a lot of files. It would be better to do the reorg when we do not have large patches in flight.

cmasone-attic:
Was there an issue filed to do this cleanup? I'm not against deleting the
JS stuff if it's no longer in use, but moving almost every file doesn't
seem important right now

aboodman:
@cmasone-attic yes there was, it me: #3668.

Thanks @wardn for the patch. I'm in favor of this. The useless `/go` subdir has been bugging me for awhile.

aboodman:
PS @wardn in order to accept any patch from you we'll need you to complete the contributor agreement first: https://github.com/attic-labs/noms/blob/master/CONTRIBUTING.md#contributing-code

wardn:
fixed merge conflicts

PR #3689, opened 2017-09-11
`csv-invert` command line tool

cmasone-attic:
Takes the output of a CSV file imported as a List of Struct and
"inverts" it so that it's now a Struct of Lists.

Example:

List<Struct Row {
  Base?: String,
  DOLocationID?: String,
}>

becomes

Struct Columnar {
  base: List<String>,
  dolocationid: List<String>,
}

### DIFF
--- a/samples/go/csv/csv-invert/main.go
+++ b/samples/go/csv/csv-invert/main.go
@@ -49,8 +49,27 @@ func main() {
 	outDB, outDS, err := cfg.GetDataset(flag.Arg(1))
 	defer outDB.Close()
 
+	// I don't want to allocate a new types.Value every time someone calls zeroVal(), so instead have a map of canned Values to reference.
+	zeroVals := map[types.NomsKind]types.Value{
+		types.BoolKind:   types.Bool(false),
+		types.NumberKind: types.Number(0),
+		types.StringKind: types.String(""),
+	}
+
+	zeroVal := func(t *types.Type) types.Value {
+		v, present := zeroVals[t.TargetKind()]
+		if !present {
+			d.CheckErrorNoUsage(fmt.Errorf("csv-invert doesn't support values of type %s", t.Describe()))
+		}
+		return v
+	}
+
 	defer profile.MaybeStartProfile().Stop()
-	streams := map[string]chan types.Value{}
+	type stream struct {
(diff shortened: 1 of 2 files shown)


cmasone-attic:
Takes about 6 mins to invert ~1GB of imported CSV data here on my lappytop

If we think that's fast enough for right now, then PTAL

cmasone-attic:
ok, @ehalpern PTAL

samples/go/csv/csv-invert/main.go
```diff
@@ -0,0 +1,95 @@
...
+	d.CheckError(err)
+	defer inDB.Close()
+
+	head, present := inDS.MaybeHead()
+	if !present {
+		d.CheckErrorNoUsage(fmt.Errorf("The dataset %s has no head", flag.Arg(0)))
+	}
+	v := head.Get(datas.ValueField)
+	l, isList := v.(types.List)
+	if !isList {
```
ehalpern: We'll need to also handle maps to support index by primary key. This isn't necessary for the taxi benchmark, so it can wait.

samples/go/csv/csv-invert/main.go
```diff
@@ -0,0 +1,95 @@
...
+			lists[lowerName] = types.NewStreamingList(outDB, streams[lowerName])
+		}
+		lowers[name] = lowerName
+	})
+
+	columnVals := make(map[string]types.Value, len(streams))
+	emptyString := types.String("")
+	l.IterAll(func(v types.Value, index uint64) {
+		for lowerName := range streams {
+			columnVals[lowerName] = emptyString
```
ehalpern: Don't you need to use a type specific empty value here (rather than always emptyString) to ensure the list element type is not a union?

ehalpern: For now, we're supporting three types: String, Number, Bool.

cmasone-attic: We are? I thought in our chat yesterday we were just doing String at this point.

cmasone-attic: Slacked and decided to add support for these types

cmasone-attic:
Ok, PTAL

PR #3691, opened 2017-09-11
Modifications to ipfs-chat and ipfs chunkstore

willhite:
* Change ipfs paths to include directory where ipfs repo is stored.
 * Rework ipfs-chat to create ipfs chunkstores manually rather than
   relying on Spec.ForDataset. This enables creating two chunkstores
   (one local and one network) using the same IpfsNode (ipfs repo).
 * Create separate replicate function for daemon and mergeMessage
   function for client to experiment with slightly different behaviors
   for each.


willhite:
The PR contains the latest changes to the IPFS portions of the chat program. There's lots of code duplication between the ipfs and p2p parts of the program.

samples/go/ipfs-chat/lib/termui.go
```diff
@@ -0,0 +1,368 @@
...
+	Gui      *gocui.Gui
+	InSearch bool
+}
+
+var (
+	viewNames   = []string{usersView, messageView, inputView}
+	firstLayout = true
+)
+
+func CreateTermUI(events chan ChatEvent) TermUI {
```
aboodman: Don't you need to return `*TermUI`? Otherwise `TermUI.InSearch` will always be by value.

PR #3697, opened 2017-09-12
NOMSFS: ensure clean path parsing

wardn:


wardn:
The code is straightforward so I'm guessing there's a test case that needs updating.  hanwen/go-fuse doesn't support bsd, so I'll look into it when I get access to a linux/darwin box.

Issue #3690, opened 2017-09-11
Grab low-hanging-fruit in `csv-invert`


cmasone-attic:
Once #3689 lands, profile moar and see if there's any low-hanging fruit to pick. There should be at least a little in sequence_chunker.go

PR #3698, opened 2017-09-12
csv-invert: Perf improvments!

cmasone-attic:
Tweaking the main loop that processes list entries to avoid some
map assignments, lookups, and allocations saves 15% or so, resulting
in an overall savings of about 1m on the 6m runtime of our test
workload (as run on my laptop).

Towards #3690

### DIFF
--- a/samples/go/csv/csv-invert/main.go
+++ b/samples/go/csv/csv-invert/main.go
@@ -69,7 +69,7 @@ func main() {
 		ch      chan types.Value
 		zeroVal types.Value
 	}
-	streams := map[string]stream{}
+	streams := map[string]*stream{}
 	lists := map[string]<-chan types.List{}
 	lowers := map[string]string{}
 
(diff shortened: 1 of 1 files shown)


go/types/sequence_chunker.go
```diff
@@ -173,7 +173,7 @@ func (sc *sequenceChunker) createSequence() (sequence, metaTuple) {
 	seq := col.sequence()
 	mt := newMetaTuple(sc.vrw.WriteValue(col), key, numLeaves)
 
-	sc.current = []sequenceItem{}
+	sc.current = sc.current[:0]
```
cmasone-attic: Since this is a more general-purpose perf win, I wonder if it should be in a separate PR? @rafael-atticlabs, thoughts?

cmasone-attic:
PTAL

go/types/sequence_chunker.go
rafael-atticlabs: yeah, kalman has a patch outstanding in this area. I'll ask him to do this as a part of that.

cmasone-attic: ok, removing

cmasone-attic:
well, the win will be less, now, that the sequence_chunker.go change is put off to another patch. But, still we should get about a 15% improvement

Issue #3645, opened 2017-08-29
Sequence chunker should avoid eagerly persisting root chunk


rafael-atticlabs:
Now that prolly trees are eagerly written on creation, the sequence chunker chunker should avoid persisting the root chunk (which is returned). This is important, because the root chunk may not be directly addressable in the resulting value (i.e. if it is embedded into another value).

kalman:
Started looking at this. It isn't trivial when you consider all of the special cases listed at https://github.com/attic-labs/noms/blob/759cf4fe589dd8a0af4e08983637c8b10cff29cc/go/types/sequence_chunker.go#L204 (`sequenceChunker.Done`), namely that we don't really know what the root chunk is until `Done` is called.

Oh well, here goes...

PR #3699, opened 2017-09-13
Don't write a sequence chunk if there is no parent

kalman:
In most cases this will avoid writing the root chunk of a prolly tree,
which is the behavior we're aiming for: a prolly tree might be used
inline in which case the root never needs to be written.

The solution in this patch is imperfect because it may unnecessarily
write chunks, but this is rare.

Fixes https://github.com/attic-labs/noms/issues/3645

### DIFF
--- a/go/datas/http_chunk_store_test.go
+++ b/go/datas/http_chunk_store_test.go
@@ -151,7 +151,7 @@ func (suite *HTTPChunkStoreSuite) TestPutChunk() {
 	suite.True(suite.http.Has(c.Hash()))
 
 	suite.True(suite.http.Commit(hash.Hash{}, hash.Hash{}))
-	suite.Equal(2, suite.serverCS.Writes)
+	suite.Equal(1, suite.serverCS.Writes)
 }
 
 func (suite *HTTPChunkStoreSuite) TestPutChunksInOrder() {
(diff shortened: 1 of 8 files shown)


rafael-atticlabs:
Thinking about this solution this way actually makes me feel better about it. Basicallly there are two cases here to worry about:

1) Upon completion, we want to avoid persisting the resulting chunk from the root elements
2) It's possible that N levels below the top chunker will have randomly chunked on the first item and we'll normalize by walking down. In this case, we'd prefer to have avoided the *canonical* root chunk.

This patch completely handles (1), but doesn't address (2). let's just open a bug for (2) and fix it...probably never.

kalman:
OK, ptal. Comments addressed, plus I did a bit of cleanup from the previous attempt at solving this problem.

I'm not sure that passing in a bool to createSequence (new code) is better than checking for the presence of a parent (old code), because we still need to know when to write any unwritten collection, which is when createParent is called.

I.e. I find it more direct seeing "if there is a parent, write the sequence" and "when there is a new parent, write an unwritten sequence" vs looking at call sites of createSequence to make that connection.

Arguments either way.

kalman:
I'm going to do it separately so we can discuss separately.

cmasone-attic:
For what it's worth, I actually tried initializing `sc.current` to some non-zero size and it didn't change the profile much. Re-using the storage is the real win.

Issue #3667, opened 2017-09-05
Cleanup noms directory structure


aboodman:
Now that we no longer have a js implementation, we can get rid of the superfluous `go` dir at the root and just move everything into `/noms`.

Similarly, we can get rid of the `go` subdir` in `samples`.

arv:
What about the grapiql sample? It was pretty useful when using Noms+GraphQL ?

@rafael-atticlabs

arv:
cc @wardn

wardn:
looks like the graphiql sample is referencing the js implementations

PR #3708, opened 2017-09-14
revert graphiql sample removal

wardn:
see #3667


wardn:
Going to pull this out for now.  I can add it back depending on the outcome of #3667, but I don't want to clutter up the pull request queue.

PR #3719, opened 2017-09-18
Add source code for quickstart sample

willhite:
I'll add this source code into noms/samples/go/quickstart.go.


willhite:
I added a sample for the code that I'm showing in quickstart.md

samples/go/quickstart/quickstart.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2017 Attic Labs, Inc. All rights reserved.
```
aboodman: Can we make this show up directly in the doc somehow? If not, what is the benefit to checking it in?

willhite: That doesn't appear possible. I wanted to make sure the code compiled so I created a sample and thought it might be nice to have it as a convenience. But it's not that big of a win, I'll delete this PR.

PR #3720, opened 2017-09-18
Rough cut at bucketdb about.md

cmasone-attic:


cmasone-attic:
This is just a stab at this content. PTAL, and please feel free to either give comments -- or totally rewrite sections or the whole thing and just push a new patch :-)

cmasone-attic:
FYI @phritz

phritz:
Thanks, @cmasone-attic. I think this is a good start and you should check it in. 

I also think we (you? me? @ehalpern ?) should take another pass and tailor it even more to the analytics use case. For example explain how noms' sync plays out in this use case and why it is important, and eliminate stuff that isn't relevant to the use case such as rolling up to a 20 byte hash (unless that is important for OLAP for some reason).

If it'd be useful for me to make more, specific suggestions I'm happy to.

cmasone-attic:
The hash part is relevant, I believe, because it helps illuminate how versioning works, which in the analytics case lets you track the version of the data that generated a given output.

PR #3724, opened 2017-09-19
Change ReadMany to do HashSlice -> ValueSlice

arv:
The order of the values is the same as the order of the hashes.

This way we can easily map the hash to the value.

Motivation #3710

### DIFF
--- a/go/types/indexed_sequences.go
+++ b/go/types/indexed_sequences.go
@@ -132,17 +132,11 @@ func LoadLeafNodes(cols []Collection, startIdx, endIdx uint64) ([]Collection, ui
 	}
 
 	// Fetch committed child sequences in a single batch
-	fetched := make(map[hash.Hash]Collection, len(hs))
-	if len(hs) > 0 {
-		readValues := vrw.ReadManyValues(hs)
-		for i, v := range readValues {
-			fetched[hs[i]] = v.(Collection)
-		}
-	}
+	readValues := vrw.ReadManyValues(hs)
 
-	childCols := make([]Collection, len(childTuples))
-	for i, mt := range childTuples {
-		childCols[i] = fetched[mt.ref.TargetHash()]
+	childCols := make([]Collection, len(readValues))
+	for i, v := range readValues {
+		childCols[i] = v.(Collection)
 	}
 
 	return LoadLeafNodes(childCols, startIdx, endIdx)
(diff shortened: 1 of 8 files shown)


go/types/indexed_sequences.go
```diff
@@ -126,21 +126,17 @@ func LoadLeafNodes(cols []Collection, startIdx, endIdx uint64) ([]Collection, ui
...
-	hs := hash.HashSet{}
-	for _, mt := range childTuples {
-		hs.Insert(mt.ref.TargetHash())
+	hs := make(hash.HashSlice, len(childTuples))
+	for i, mt := range childTuples {
+		hs[i] = mt.ref.TargetHash()
 	}
 
 	// Fetch committed child sequences in a single batch
 	fetched := make(map[hash.Hash]Collection, len(hs))
```
rafael-atticlabs: You can ditch the indirection through `fetched` now. \o/

go/types/meta_sequence.go
```diff
@@ -383,27 +383,24 @@ func (ms metaSequence) getChildren(start, end uint64) (seqs []sequence) {
...
 
 	// Fetch committed child sequences in a single batch
-	valueChan := make(chan Value, len(hs))
-	go func() {
-		ms.vrw.ReadManyValues(hs, valueChan)
-		close(valueChan)
-	}()
+	readValues := ms.vrw.ReadManyValues(hs)
+
 	children := make(map[hash.Hash]sequence, len(hs))
```
rafael-atticlabs: same thing. you can ditch `children` here since it was just serving as a mapping from hash -> position (maybe a bit of math required as well).

go/types/value_stats.go
```diff
@@ -96,28 +96,11 @@ func compressedSize(v Value) uint64 {
...
-		vr.ReadManyValues(hs, valueChan)
-		close(valueChan)
-	}()
-
-	vm := make(map[hash.Hash]Value, len(hs))
-	for v := range valueChan {
-		vm[v.Hash()] = v
-	}
-
+	hs := make(hash.HashSlice, len(refs))
```
rafael-atticlabs: \o/

cmasone-attic:
Agree with raf on the nits. LGTM

cmasone-attic:
LGTM

PR #3725, opened 2017-09-19
Make EncodeValue return bytes directly

arv:
Also make the primitives allocate a more exact buffer instead of the
initialBufferSize.

Also, stack allocate binaryNomsWriter

### DIFF
--- a/go/types/string.go
+++ b/go/types/string.go
@@ -4,7 +4,11 @@
 
 package types
 
-import "github.com/attic-labs/noms/go/hash"
+import (
+	"encoding/binary"
+
+	"github.com/attic-labs/noms/go/hash"
+)
 
 // String is a Noms Value wrapper around the primitive string type.
 type String string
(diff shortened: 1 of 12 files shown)


go/types/string.go
```diff
@@ -51,3 +55,12 @@ func (s String) writeTo(w nomsWriter) {
 	StringKind.writeTo(w)
 	w.writeString(string(s))
 }
+
+func (s String) valueBytes() []byte {
+	// We know the size of the buffer here so allocate it once.
+	// StringKind, Length (UVarint), UTF-8 encoded string
+	buff := make([]byte, 1+binary.MaxVarintLen64+len(s))
```
rafael-atticlabs: doesn't this need to be `len([]byte(s))`?

arv: That is the same thing

https://play.golang.org/p/ulXSONDx1Y

rafael-atticlabs: fair enough

PR #3728, opened 2017-09-20
Add readTypeOfValue

arv:
This improves csv-import a bit since it allows us to skip allocation of
values in the struct when we only want the type.


arv:
@cmasone-attic since you are looking at this area today

cmasone-attic:
lgtm

rafael-atticlabs:
clever.

PR #3729, opened 2017-09-20
Change List.IterAll to read-ahead 8MB chunks over 6 threads

kalman:
This copies the approach that Blob.Copy uses, with some extra logic to
adjust the number of values to read as more values are read.

### DIFF
--- a/go/types/leaf_sequence.go
+++ b/go/types/leaf_sequence.go
@@ -6,6 +6,7 @@ package types
 
 import (
 	"bytes"
+	"math"
 
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
@@ -91,10 +92,20 @@ func (seq leafSequence) writeTo(w nomsWriter) {
 }
 
 func (seq leafSequence) values() []Value {
+	return seq.valuesSlice(0, math.MaxUint64)
+}
+
+func (seq leafSequence) valuesSlice(from, to uint64) []Value {
 	dec, count := seq.decoderSkipToValues()
-	vs := make([]Value, count)
-	for i := uint64(0); i < count; i++ {
-		vs[i] = dec.readValue()
+	if to > count {
+		to = count
+	}
+	for i := uint64(0); i < from; i++ {
+		dec.skipValue()
+	}
+	vs := make([]Value, to-from)
+	for i := from; i < to; i++ {
+		vs[i-from] = dec.readValue()
 	}
 	return vs
 }
--- a/go/types/leaf_sequence.go
+++ b/go/types/leaf_sequence.go
@@ -96,13 +96,10 @@ func (seq leafSequence) values() []Value {
 }
 
 func (seq leafSequence) valuesSlice(from, to uint64) []Value {
-	dec, count := seq.decoderSkipToValues()
-	if to > count {
-		to = count
+	if len := seq.Len(); to > len {
+		to = len
 	}
-	for i := uint64(0); i < from; i++ {
-		dec.skipValue()
-	}
+	dec := seq.decoderSkipToIndex(int(from))
 	vs := make([]Value, to-from)
 	for i := from; i < to; i++ {
 		vs[i-from] = dec.readValue()
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -211,10 +211,10 @@ func (l List) IterAll(f listIterAllFunc) {
 	}
 }
 
-func (l List) copyReadAhead(buff []Value, startIdx uint64) (numBytes uint64) {
+func (l List) copyReadAhead(out []Value, startIdx uint64) (numBytes uint64) {
 	d.PanicIfFalse(startIdx < l.Len())
 
-	endIdx := startIdx + uint64(len(buff))
+	endIdx := startIdx + uint64(len(out))
 	if endIdx > l.Len() {
 		endIdx = l.Len()
 	}
@@ -231,8 +231,8 @@ func (l List) copyReadAhead(buff []Value, startIdx uint64) (numBytes uint64) {
 		ls := leaf.sequence().(listLeafSequence)
 
 		values := ls.valuesSlice(startIdx, endIdx)
-		copy(buff, values)
-		buff = buff[len(values):]
+		copy(out, values)
+		out = out[len(values):]
 		endIdx = endIdx - uint64(len(values)) - startIdx
 		startIdx = 0
 
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -5,6 +5,8 @@
 package types
 
 import (
+	"sync/atomic"
+
 	"github.com/attic-labs/noms/go/d"
 	"github.com/attic-labs/noms/go/hash"
 )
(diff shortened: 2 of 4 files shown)


kalman:
This isn't to submit yet, because it hardly makes a difference in csv-export. @rafael-atticlabs could you check it's sensible?

Here is what the csv exported used to do (400MB of csv data):
```
> go install ./samples/go/csv/... && (time bin/csv-export db::csv) >/dev/null

real	0m14.539s
user	0m22.550s
sys	0m4.139s
```
and with my changes:
```
> go install ./samples/go/csv/... && (time bin/csv-export db::csv) >/dev/null
read 1000 chunks (345 kB)
adjusted chunk size from 1000 to 24301 (scaled by 24.301)
read 1000 chunks (355 kB)
adjusted chunk size from 1000 to 23600 (scaled by 23.600)
read 1000 chunks (371 kB)
adjusted chunk size from 1000 to 22620 (scaled by 22.621)
read 1000 chunks (351 kB)
adjusted chunk size from 1000 to 23899 (scaled by 23.899)
read 1000 chunks (364 kB)
adjusted chunk size from 1000 to 23054 (scaled by 23.054)
read 1000 chunks (318 kB)
adjusted chunk size from 1000 to 26354 (scaled by 26.354)
read 1000 chunks (338 kB)
adjusted chunk size from 1000 to 24821 (scaled by 24.821)
read 1000 chunks (346 kB)
adjusted chunk size from 1000 to 24280 (scaled by 24.280)
read 26354 chunks (8.4 MB)
read 24280 chunks (7.7 MB)
adjusted chunk size from 24280 to 26305 (scaled by 1.083)
read 24280 chunks (7.7 MB)
adjusted chunk size from 24280 to 26392 (scaled by 1.087)
read 24280 chunks (7.7 MB)
adjusted chunk size from 24280 to 26331 (scaled by 1.084)
read 24280 chunks (7.7 MB)
adjusted chunk size from 24280 to 26286 (scaled by 1.083)
read 24280 chunks (7.8 MB)
adjusted chunk size from 24280 to 26246 (scaled by 1.081)
read 24280 chunks (7.9 MB)
adjusted chunk size from 24280 to 25825 (scaled by 1.064)
read 24280 chunks (7.8 MB)
adjusted chunk size from 24280 to 26189 (scaled by 1.079)
read 25825 chunks (8.3 MB)
adjusted chunk size from 25825 to 26240 (scaled by 1.016)
read 26189 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.3 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.3 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.3 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.3 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.4 MB)
read 26240 chunks (8.3 MB)
read 23945 chunks (7.6 MB)

real	0m14.637s
user	0m22.263s
sys	0m4.393s
```

go/types/list.go
```diff
@@ -149,15 +155,107 @@ type listIterAllFunc func(v Value, index uint64)
...
+	}
+
+	leaves, localStart := LoadLeafNodes([]Collection{l}, startIdx, endIdx)
+	endIdx = localStart + endIdx - startIdx
+	startIdx = localStart
+	n := 0
+
+	for _, leaf := range leaves {
+		ls := leaf.sequence().(listLeafSequence)
+		numBytes += uint64(len(ls.buff))
```
kalman: @arv is this right?

kalman: background: @arv @rafael-atticlabs this is supposed to get the number of bytes that were read as a result of this copyReadAhead call. It would be most literal to know the actual Chunks that were read, but I don't have access to that. I am _presuming_ that `buff` is the raw bytes of the sequence as read from the value store.

kalman:
final observations before I did into this again:
* the codec benchmark is also identical
* maybe this needs to run on S3 to see the change?
* marshal test fails. I think it's something to do with IterAll and (un)marshalling using the same internal state (like values()?) and getting confused.

go/types/list.go
arv: This is the same size as the leaf chunk.

go/types/list.go
```diff
@@ -149,15 +151,101 @@ type listIterAllFunc func(v Value, index uint64)
...
+
+		localEnd := endIdx
+		values := ls.values()
```
arv: This is a bit too eager. It will read and decode all the values in the leaf sequence.

Also, it seems like you are just interested in the bytes below so it might make sense to get the offset of startIdx and localEnd and just copy that slice over.

go/types/list.go
kalman: Ok. So `offsets` isn't relevant?

arv: offsets is used to be able to jump to a certain position in a sequence in constant time

go/types/list.go
kalman: > This is a bit too eager. It will read and decode all the values in the leaf sequence.

To be clear: you're saying that if we're reading only a subset of this leaf chunk (say the first 10 values of a 100 value chunk) then we're going to unnecessarily decode 90 values.

Fixed in latest patch.

> Also, it seems like you are just interested in the bytes below

Not sure what you mean, I'm interested in the Values, `buff` is a `[]Value`.

kalman:
latest patch is simpler/better but marshal test still fails, and csv export is unaffected (in fact it might be slower).

go/types/leaf_sequence.go
```diff
@@ -91,10 +92,20 @@ func (seq leafSequence) writeTo(w nomsWriter) {
...
+
+func (seq leafSequence) valuesSlice(from, to uint64) []Value {
 	dec, count := seq.decoderSkipToValues()
-	vs := make([]Value, count)
-	for i := uint64(0); i < count; i++ {
-		vs[i] = dec.readValue()
+	if to > count {
+		to = count
+	}
+	for i := uint64(0); i < from; i++ {
```
arv: There is a decoderSkipToIndex already that uses the offset to jump to the right spot without seeking

go/types/list.go
```diff
@@ -149,15 +151,94 @@ type listIterAllFunc func(v Value, index uint64)
...
+	for _, leaf := range leaves {
+		ls := leaf.sequence().(listLeafSequence)
+
+		values := ls.valuesSlice(startIdx, endIdx)
+		copy(buff, values)
+		buff = buff[len(values):]
+		endIdx = endIdx - uint64(len(values)) - startIdx
+		startIdx = 0
+
+		numBytes += uint64(len(ls.buff))
```
arv: You are only using the buff here so it seems like, instead of getting the values you could just get the offsets directly.

kalman: There is `buff` and there is `ls.buff`. I am only using the length of `ls.buff` but it doesn't cost anything because it's just a slice. I do use `buff` in the `copy` call.

kalman:
There we go.

kalman:
OK, @rafael-atticlabs (and @arv) PTAL.
* The marshal tests are now passing - I guess the values changes that @arv suggested fixed that (though honestly it does still worry me a little bit why it was failing in the first place).
* I tried with a much larger dataset (~2GB), with a hacked-up csv-export to not actually construct the CSV, and the old code took 35s and the new code 30s, so that's quite encouraging.

go/types/list.go
```diff
@@ -149,15 +151,91 @@ type listIterAllFunc func(v Value, index uint64)
...
+	concurrency := 6
+	vcChan := make(chan chan []Value, concurrency)
+
+	// Target reading data in |targetBatchBytes| per thread. We don't know how
+	// many bytes each value is, so update |estimatedNumValues| as data is read.
+	targetBatchBytes := 1 << 23 // 8MB
+	estimatedNumValues := uint64(1000)
+
+	go func() {
+		for idx := uint64(0); idx < l.Len(); {
```
arv: Extract Len(). It is no longer trivial.

kalman: sweet, thanks. fixed in blob as well.

go/types/list.go
```diff
@@ -149,15 +151,91 @@ type listIterAllFunc func(v Value, index uint64)
...
+	// many bytes each value is, so update |estimatedNumValues| as data is read.
+	targetBatchBytes := 1 << 23 // 8MB
+	estimatedNumValues := uint64(1000)
+
+	go func() {
+		for idx, len := uint64(0), l.Len(); idx < len; {
+			numValues := atomic.LoadUint64(&estimatedNumValues)
+
+			start := idx
+			blockLength := l.Len() - start
```
arv: `l.Len()` -> `len`

go/types/list.go
```diff
@@ -149,15 +151,91 @@ type listIterAllFunc func(v Value, index uint64)
...
+			i++
+		}
+	}
+}
+
+func (l List) copyReadAhead(out []Value, startIdx uint64) (numBytes uint64) {
+	d.PanicIfFalse(startIdx < l.Len())
+
+	endIdx := startIdx + uint64(len(out))
+	if endIdx > l.Len() {
```
arv: precompute lengt here too

go/types/list.go
kalman: thanks

PR #3733, opened 2017-09-21
Implement StructEditor

### LINKED ISSUE #3602: Consider implementing a StructEditor
per meatspace discussion with @aboodman, the present decision is to see how working with editors generally feels.

in my view, the reason to add a list editor would be three fold:

-consistency of editing *graphs* of noms data.
-CPU perf: if Structs are midway along paths to mutated values, creating a new struct for each mutation is potentially alot of allocations and it means that editors which contain the struct undergo alot of churn as the same logical struct is repeatedly re-added
-IO perf: Right now, in order to "navigate" to a value at a path with editors, you need to progressively load each value (in order to know if what you're holding is a value or an editor), which presently requires sync IO. By adding StructEditor, we could allow these navigations to fully defer their IO so that the application of the edits is optimally parallel once all changes are described.

arv:
This is similar to MapEditor

Unlike a Map all the struct data is available in memory so the building
is slightly simpler.

Fixes #3602

### DIFF
--- a/go/types/struct_editor.go
+++ b/go/types/struct_editor.go
@@ -1,0 +1,206 @@
+// Copyright 2017 Attic Labs, Inc. All rights reserved.
+// Licensed under the Apache License, version 2.0:
+// http://www.apache.org/licenses/LICENSE-2.0
+
+package types
+
+import (
+	"sort"
+
+	"github.com/attic-labs/noms/go/d"
+)
+
+// StructEditor allows for efficient editing of Noms structs. Edits are
+// buffered to memory and can be applied via Struct(), which returns a new
+// Struct. Prior to Struct(), Get() will return the value that the resulting
+// Struct would return if it were built immediately prior to the respective call.
+// Note: The implementation biases performance towards a usage which applies
+// edits in name-order.
+type StructEditor struct {
+	s          Struct
+	edits      structEditSlice // edits may contain duplicate name values, in which case, the last edit of a given name is used
+	normalized bool
+}
+
+// NewStructEditor returns a new StructEditor starting with the fields in s.
+func NewStructEditor(s Struct) *StructEditor {
+	return &StructEditor{s, structEditSlice{}, true}
+}
+
+// Kind returns the kind of editor this is.
+func (se *StructEditor) Kind() NomsKind {
+	return StructKind
+}
+
+// Value builds a new Struct based on the edits done to the StructEditor.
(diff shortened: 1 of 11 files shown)


go/types/struct_editor.go
```diff
@@ -0,0 +1,206 @@
...
+func (se *StructEditor) Struct() Struct {
+	if len(se.edits) == 0 {
+		return se.s // no edits
+	}
+
+	se.normalize()
+
+	kvsChan := make(chan chan structEntry)
+
+	go func() {
```
aboodman: We generally make the assumption that structs are "memory-sized". Can we also make the assumption that edits to structs are "memory-sized" ?

Invoking a goroutine is not cheap. This code means that if you have a large collection of structs and you edits all of them you're going to launch a goroutine for every one.

aboodman: This also means that even trivial edits to structs will end up costing one goroutine. Where before they were a stack allocation.

aboodman: You also have a channel in here. Lots of machinery to do `Set("foo", "bar")`.

aboodman:
I'm really concerned about the ballooning of API complexity. My experience working on Attic was that the best way to work with structs was by way of marshaling. But now we are adding this whole other way to edit them which is "more efficient" but lots more code. People will have to decide which way they want to do it.

It doesn't feel right to me that this mechanism is completely ignorant of marshaling, which is otherwise the easiest way to work with structs in Go.

arv:
The motivation is in the bug.

I agree that we might want to get rid of the go-routine here though and keep the direct mutation methods on Map/Set/List/Struct for the one off cases.

arv:
@rafael-atticlabs ^^^

rafael-atticlabs:
@aboodman What would it look like for this mechanism to be aware of marshalling?

aboodman:
I don't have an idea right now, but I feel like this is an instance of "wait for good design". That's why I was advocating for us gaining experience with the editor interface before worrying about structs.

rafael-atticlabs:
I don't feel like dying on this hill. If @aboodman feels strongly, let's just leave this patch outstanding and demote the bug to P1. It doesn't affect format. Presumably, if someone hits memory problems trying to logically update a "column" of some tabular data, we'll hear about it and can revisit.

arv:
I'll extract the logic for mutating a struct since it is a lot cleaner than the old code.

PR #3734, opened 2017-09-22
Update code blocks to always specify syntax, fix ordered list

jesseditson:
General cleanup to finish up the last bit of formatting on https://github.com/attic-labs/decent-website/issues/13


jesseditson:
Intentionally outdented the code there because it seems like showdownjs doesn't support indented code blocks - I could preprocess the blocks (I currently do that but I don't like it), but my preference is to just use formatting it can handle. What do you think?

phritz:
Sgtm

phritz:
That is, outdenting is fine.

PR #3736, opened 2017-09-22
Replace regexp with loop for struct name/field validation

### LINKED ISSUE #2866: Optimize struct type validation. The regexp's show in profiles.
While doing something else, I noticed that sync appeared to have gotten slower. I bisected it and discovered that https://github.com/attic-labs/noms/commit/5e901b092401c14b36084fe9f140dc3e3a4f6a9e regressed syncing ~300MB of sf_scrime data from ldb to ldb by ~ 100%.

After some short digging, I'm fairly sure it's because of the addition of type validation to the validating_batch_sink.

Thinking about it some, I think it's an easy fix because of the type cache. Basically we're validating over and over again exactly the same in-memory type graph (the type cache ensures that identical types are interned). we just need to find a way to only do it *once* per type.

arv:
Fixes #2866

### DIFF
--- a/go/types/struct.go
+++ b/go/types/struct.go
@@ -510,7 +510,24 @@ func EscapeStructField(input string) string {
 // IsValidStructFieldName returns whether the name is valid as a field name in a struct.
 // Valid names must start with `a-zA-Z` and after that `a-zA-Z0-9_`.
 func IsValidStructFieldName(name string) bool {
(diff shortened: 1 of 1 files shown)


arv:
Verified. It showed up before and doesn't show up any more.

rafael-atticlabs:
Nice.

PR #3737, opened 2017-09-22
Simplify Struct Set/Delete

arv:
The old code was hard to follow and it ended up encoding twice when a
new field was added.

### DIFF
--- a/go/types/struct.go
+++ b/go/types/struct.go
@@ -334,13 +334,13 @@ func (s Struct) IsZeroValue() bool {
 // If name is not an existing field in the struct then the current struct is returned.
 func (s Struct) Delete(n string) Struct {
 	prolog, head, tail, count, found := s.splitFieldsAt(n)
+	if !found {
+		return s
+	}
 
 	w := binaryNomsWriter{make([]byte, len(s.buff)), 0}
 	w.writeRaw(prolog)
-	if found {
-		count--
-	}
-	w.writeCount(count)
+	w.writeCount(count - 1)
 	w.writeRaw(head)
 	w.writeRaw(tail)
 
--- a/go/types/struct.go
+++ b/go/types/struct.go
@@ -289,7 +289,7 @@ func (s Struct) Set(n string, v Value) Struct {
 	return Struct{s.vrw, w.data()}
 }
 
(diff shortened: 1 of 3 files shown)


go/types/struct.go
```diff
@@ -270,57 +343,37 @@ func (s Struct) Get(n string) Value {
...
-func (s Struct) set(w *binaryNomsWriter, n string, v Value, addedCount int) Struct {
-	// TODO: Reuse bytes if we end up adding a field
-	dec := s.decoder()
-	StructKind.writeTo(w)
-	dec.skipKind()
-	dec.copyString(w)
-	count := dec.readCount()
-	w.writeCount(count + uint64(addedCount))
+func (s Struct) set(name string, value Value) Struct {
+	entries := s.structEntries()
```
arv: This could be done without allocating this slice but this way is much cleaner.

arv:
Ping

go/types/struct.go
```diff
@@ -271,56 +271,58 @@ func (s Struct) Get(n string) Value {
...
+	w.writeString(n)
+	v.writeTo(&w)
+	w.writeRaw(tail)
+
+	return Struct{s.vrw, w.data()}
 }
 
-func (s Struct) set(w *binaryNomsWriter, n string, v Value, addedCount int) Struct {
-	// TODO: Reuse bytes if we end up adding a field
+// splitFieldsAt splits the buffer into two parts. The fields comming before the field we are looking for
```
rafael-atticlabs: `coming`

go/types/struct.go
```diff
@@ -331,32 +333,18 @@ func (s Struct) IsZeroValue() bool {
...
-		} else {
-			w.writeString(name)
-			dec.copyValue(&w)
-		}
-	}
+	prolog, head, tail, count, found := s.splitFieldsAt(n)
 
+	w := binaryNomsWriter{make([]byte, len(s.buff)), 0}
+	w.writeRaw(prolog)
 	if found {
```
rafael-atticlabs: shouldn't this just `return s if !found`?

arv: yup

PR #3742, opened 2017-09-27
Add ability to import csv in column major order

willhite:
Added --invert argument to indicate column major order
  Added --append argument to append imported data to current head of dataset
  Added --limit-records to import data for limited number of rows

### DIFF
--- a/samples/go/csv/read.go
+++ b/samples/go/csv/read.go
@@ -92,28 +92,93 @@ func MakeStructTemplateFromHeaders(headers []string, structName string, kinds Ki
 	return
 }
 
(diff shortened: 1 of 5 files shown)


samples/go/csv/read.go
```diff
@@ -116,6 +123,64 @@ func ReadToList(r *csv.Reader, structName string, headers []string, kinds KindSl
...
 }
 
+type column struct {
+	ch        chan types.Value
+	list      <-chan types.List
+	zeroValue types.Value
+	hdr       string
+}
+
+// ReadToInvertedList takes a CSV reader and reads data from each column into a
```
ehalpern: -> ReadToColumnar

PR #3744, opened 2017-09-28
Tiny cleanup of getCompositeChildSequence

arv:


arv:
This code was just really strange.

I think that there is something wrong here though... Is it not possible for a sequence to have both leafSequences and metaSequences as its direct children?

arv:
TBR

Issue #2270, opened 2016-08-04
Lazily decode values from sequences


rafael-atticlabs:
We have strong evidence that perf is presently bound by number of memory allocations. Operations like mutating prolly trees or streaming input/output are hobbled by use eagerly converting sequences into values in memory (and thus bouncing back & forth between values & bytes sequences for operations over prolly trees).

arv:
I've started on this.

When decoding a chunk containing a collection we store the offsets and the chunk in the sequence. This introduces a new sequence implementation that decodes the values in the sequence on demand.

When encoding this sequence we can just copy the bytes.

A few notes

- We can do this incrementally. We can do the leaf sequences first
- We need a type check because we currently have a testNomsReader and only if we have a binaryNomsReader can we use the bytes directly when computing the new chunk.

arv:
Once #3694 lands there is some follow up work that needs to be done:

- [ ] types.Type
- [ ] Primitives
- [ ] Make the containers not do decoding for reading but just "slice the bytes"

PR #3746, opened 2017-09-28
Make metaTuple backed by []byte

arv:
This is a version change since there was a bug in the old code; it
didn't include the numLeaves in the rolling hash.

Towards #2270


go/types/meta_sequence.go
```diff
@@ -336,16 +373,7 @@ func (ms metaSequence) getChildren(start, end uint64) (seqs []sequence) {
...
-	mt := item.(metaTuple)
-	v := mt.key.v
-	if !mt.key.isOrderedByValue {
-		// See https://github.com/attic-labs/noms/issues/1688#issuecomment-227528987
-		d.PanicIfTrue(mt.key.h.IsEmpty())
-		v = constructRef(mt.key.h, BoolType, 0)
-	}
-
-	hashValueBytes(mt.ref, rv)
-	hashValueBytes(v, rv)
```
arv: This old code did not include the numLeaves in the rolling hash.

arv:
It is below the fold (before and after) for csv-import and csv-export

The motivation was consistency

PR #3751, opened 2017-10-03
Sync: Batch up chunk processing during pull

### LINKED ISSUE #3692: Sync: Avoid exploding memory with chunk data
After #3659, we need to avoid exploding memory by attempting to load too many chunks of a given level at once.

cmasone-attic:
In Pull(), attempt to limit memory usage by capping the number
of chunks processed at the same time. If a given level of the
chunk graph has more than |batchSize| members, they will still
be processed in order.

Fixes #3692

### DIFF
--- a/go/datas/pull.go
+++ b/go/datas/pull.go
@@ -19,7 +19,10 @@ type PullProgress struct {
 	DoneCount, KnownCount, ApproxWrittenBytes uint64
 }
 
-const bytesWrittenSampleRate = .10
+const (
+	bytesWrittenSampleRate = .10
+	batchSize              = 1 << 12 // 4096 chunks
+)
 
 // Pull objects that descend from sourceRef from srcDB to sinkDB.
 func Pull(srcDB, sinkDB Database, sourceRef types.Ref, progressCh chan PullProgress) {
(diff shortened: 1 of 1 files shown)


cmasone-attic:
The first two patches are from #3738 
PTAL at that, and then this

go/datas/pull.go
```diff
@@ -19,7 +19,10 @@ type PullProgress struct {
 	DoneCount, KnownCount, ApproxWrittenBytes uint64
 }
 
-const bytesWrittenSampleRate = .10
+const (
+	bytesWrittenSampleRate = .10
+	batchSize              = 1 << 15 // 32k chunks
```
rafael-atticlabs: does this mean 128MB average batch size (assuming 4k chunks)? Did you try adjusting this up & down to see the impact on perf. I might have guessed a lower value (e.g. 10s of MB) would have been optimal.

cmasone-attic: It does mean that, yeah. I can play around with it a little to see if anything seems immediately better

cmasone-attic: I dropped it all the way down to ~16MB per batch (4096 chunks at avg 4k per) before the performance started to slowly drop off

PR #3752, opened 2017-10-04
Remove readAhead flags which are always false

kalman:


go/types/set_iterator.go
```diff
@@ -54,7 +54,7 @@ func (si *setIterator) SkipTo(v Value) Value {
 			return si.Next()
 		}
 
-		si.cursor, _ = si.s.getCursorAtValue(v, true)
+		si.cursor = newCursorAtValue(si.s.orderedSequence, v, true, false)
```
kalman: I believe this was a mistake.

kalman:
Yep, that's what I'm working towards.

PR #3758, opened 2017-10-09
Use refWalker for Value.WalkRefs

cmasone-attic:
For values that are currently backed by byte-slices in memory, use `refWalker` as the implementation for `WalkRefs`. This is preferable ATM because it avoids allocating (in particular, strings in structs).

### DIFF
--- a/go/types/leaf_sequence.go
+++ b/go/types/leaf_sequence.go
@@ -138,10 +138,7 @@ func (seq leafSequence) getItem(idx int) sequenceItem {
 }
 
 func (seq leafSequence) WalkRefs(cb RefCallback) {
-	dec, count := seq.decoderSkipToValues()
-	for i := uint64(0); i < count; i++ {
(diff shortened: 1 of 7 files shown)


go/types/leaf_sequence.go
```diff
@@ -138,10 +138,7 @@ func (seq leafSequence) getItem(idx int) sequenceItem {
 }
 
 func (seq leafSequence) WalkRefs(cb RefCallback) {
-	dec, count := seq.decoderSkipToValues()
-	for i := uint64(0); i < count; i++ {
-		dec.readValue().WalkRefs(cb)
-	}
+	walkRefs(seq.valueBytes(), cb)
```
arv: This can go on `valueImpl`

rafael-atticlabs: https://github.com/attic-labs/noms/pull/3762

PR #3760, opened 2017-10-09
Add List.IterRange

rafael-atticlabs:
Adds explicit API to list to iterate a range of indices. The implementation of this uses the most eager read of the required data.

### DIFF
--- a/go/types/list.go
+++ b/go/types/list.go
@@ -92,11 +92,9 @@ func (l List) Concat(other List) List {
 	return newList(seq)
 }
 
-type listIterFunc func(v Value, index uint64) (stop bool)
-
 // Iter iterates over the list and calls f for every element in the list. If f returns true then the
 // iteration stops.
-func (l List) Iter(f listIterFunc) {
+func (l List) Iter(f func(v Value, index uint64) (stop bool)) {
 	idx := uint64(0)
 	cur := newCursorAtIndex(l.sequence, idx)
 	cur.iter(func(v interface{}) bool {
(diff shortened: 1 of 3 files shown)


rafael-atticlabs:
Turns out to be a little complicated to add usage from ngql. PTAL as is.

go/types/list.go
```diff
@@ -108,11 +106,41 @@ func (l List) Iter(f listIterFunc) {
 	})
 }
 
-type listIterAllFunc func(v Value, index uint64)
+func (l List) IterRange(f func(v Value, idx uint64), startIdx, endIdx uint64) {
```
kalman: This is basically `copyReadAhead`, which makes sense, because "copy read ahead" is basically "get my all values in a range". It seems like it'd be relatively easy to share the bulk of this code; whereas `copyReadAhead` returns `[]Value`, this function runs a callback.

Though honestly it's almost like we should have `Slice` instead of `IterRange`, because returning a slice is more direct and flexible (for the caller), and it more accurately represents how this is implemented.

... Unless you wanted to evolve this function later to divide up the iteration range if it's large (for parallelism or to save memory), i.e. refactor `IterAll` into `IterRange` and implement the former with `IterRange(0, l.Len())`.

kalman: Or `GoSlice`, since we may want a prollytree-level `Slice` down the road.

Fun idea, but probably unnecessary at this point:
```go
func (l List) GoSlice(range ...uint64) []Value {
  var from, to uint64

  if len(range) > 0 {
    from = range[0]
    if len(range) > 1 {
      to = range[1]
      if len(range) > 2 {
        panic("too many arguments to GoRange")
      }
    } else {
      to = l.Len()
    }
  }
}
```

kalman: ^ updated this

rafael-atticlabs:
PTAL

Issue #3743, opened 2017-09-27
Unable to concat lists over 33 million entries long


willhite:
I'm building up taxi data and importing data in column major order. When lists get longer than ~33 million entries, I consistently get  a panic when calling db.WriteValue on a newly concatenated list.

The stack was produced by code in this #3742 which hasn't been landed into master yet.

        Error Trace:    try.go:99c:
	            	try.go:58
	            	store.go:137
	            	value_store.go:243
	            	value_store.go:297
	            	value_store.go:217
	            	importer.go:213
	            	struct.go:210
	            	importer.go:215
	            	proc.go:185
	            	asm_amd64.s:2197
	Error:      	Expected false

goroutine 1 [running]:
github.com/attic-labs/noms/go/d.PanicIfFalse(0xc420192500)
	/home/ec2-user/go/src/github.com/attic-labs/noms/go/d/try.go:58 +0xa0
github.com/attic-labs/noms/go/nbs.(*NomsBlockStore).Put(0xc4201925a0, 0xacc61c6e45a54e62, 0xf8e01704be5f5980, 0x5b6b469e, 0xc5bb29a000, 0x10b07606, 0x20000000)
	/home/ec2-user/go/src/github.com/attic-labs/noms/go/nbs/store.go:137 +0x137
github.com/attic-labs/noms/go/types.(*ValueStore).bufferChunk.func1(0xacc61c6e45a54e62, 0xf8e01704be5f5980, 0xc45b6b469e, 0xacc61c6e45a54e62, 0xf8e01704be5f5980, 0x5b6b469e, 0xc5bb29a000, 0x10b07606, 0x20000000)
	/home/ec2-user/go/src/github.com/attic-labs/noms/go/types/value_store.go:243 +0x69
github.com/attic-labs/noms/go/types.(*ValueStore).bufferChunk(0xc449416000, 0x179c900, 0xc4d9769700, 0xacc61c6e45a54e62, 0xf8e01704be5f5980, 0x5b6b469e, 0xc5bb29a000, 0x10b07606, 0x20000000, 0x1)
	/home/ec2-user/go/src/github.com/attic-labs/noms/go/types/value_store.go:297 +0x443
github.com/attic-labs/noms/go/types.(*ValueStore).WriteValue(0xc449416000, 0x179c900, 0xc4d9769700, 0x0, 0x0, 0x0, 0x0, 0x0)
[...]

PR #3765, opened 2017-10-19
Avoid creating gigantic chunks with Lists of very long repeated values

rafael-atticlabs:
Towards #3743 

Includes a set of changes aimed at better handling of Lists which contain long subsequences of repeated values (e.g. 1.1 Billion "yellow" strings).

Turns out a few things were happening here

1) The sloppy lookup table was only 16 bits and so when the input byte stream crossed 64k it stopped find matches. Fix: the lookup table now stores 32 bit offsets

2) "yellow" repeated encodes as patterns of 8 bytes, this (by chance) drove the rolling hash to zero from which it never recovers. Fix: make the hash window a prime value (67 bytes).

3) There already is a maxCopy in sloppy, but this effectively results in the rolling hash seeing long repeated sequences of 3 bytes (just repeats of the copy). The reduces entropy further and the input sequence never chunks (just by chance). Fix: add a `maxChunkSize` of 16MB.

4) Bump the noms version

Note that this patch leaves in place the problem that if you inserted another yellow at start of the 1.1 billion yellows, it'll cascade all the way to the end. I think the right solution to that *has* to be something like the repeat kind. For now, this will allow us to successfully import insanely large columns which have one value.

### DIFF
--- a/go/types/rolling_value_hasher.go
+++ b/go/types/rolling_value_hasher.go
@@ -15,9 +15,9 @@ import (
 const (
 	defaultChunkPattern = uint32(1<<12 - 1) // Avg Chunk Size of 4k
 
-	// The window size to use for computing the rolling hash. This is way more than necessary assuming random data (two bytes would be sufficient with a target chunk size of 4k). The benefit of a larger window is it allows for better distribution on input with lower entropy. At a target chunk size of 4k, any given byte changing has roughly a 1.5% chance of affecting an existing boundary, which seems like an acceptable trade-off.
+	// The window size to use for computing the rolling hash. This is way more than necessary assuming random data (two bytes would be sufficient with a target chunk size of 4k). The benefit of a larger window is it allows for better distribution on input with lower entropy. At a target chunk size of 4k, any given byte changing has roughly a 1.5% chance of affecting an existing boundary, which seems like an acceptable trade-off. The choice of a prime number provides better distribution for repeating input.
 	chunkWindow  = uint32(67)
-	maxChunkSize = 1 << 24
+	maxChunkSize = 1 << 24 // TODO: Remove when https://github.com/attic-labs/noms/issues/3743 is fixed.
 )
 
 // Only set by tests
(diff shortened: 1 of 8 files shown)


go/types/rolling_value_hasher.go
```diff
@@ -16,7 +16,8 @@ const (
 	defaultChunkPattern = uint32(1<<12 - 1) // Avg Chunk Size of 4k
 
 	// The window size to use for computing the rolling hash. This is way more than necessary assuming random data (two bytes would be sufficient with a target chunk size of 4k). The benefit of a larger window is it allows for better distribution on input with lower entropy. At a target chunk size of 4k, any given byte changing has roughly a 1.5% chance of affecting an existing boundary, which seems like an acceptable trade-off.
-	defaultChunkWindow = uint32(64)
+	defaultChunkWindow = uint32(67)
```
kalman: - Maybe update comment to mention why this is prime.
- You should also update the test chunk window.

go/types/rolling_value_hasher.go
```diff
@@ -84,6 +85,9 @@ func (rv *rollingValueHasher) HashByte(b byte) bool {
 	if !rv.crossedBoundary {
 		rv.bz.HashByte(b ^ rv.salt)
 		rv.crossedBoundary = (rv.bz.Sum32()&rv.pattern == rv.pattern)
+		if rv.bw.offset > maxChunkSize {
```
kalman: Seems unfortunate that we need to do this (if it's just to solve that problem). Worth coming back to I think?

go/types/rolling_value_hasher.go
rafael-atticlabs: Yup. I was thinking that we'd remove this when we add the repeat kind. I'll add a todo

go/types/rolling_value_hasher.go
rafael-atticlabs: yup. added a todo to remove

PR #3768, opened 2017-10-20
Add noms stats command

cmasone-attic:
Prints a summary of stats for the given DB


cmasone-attic:
PTAL

FYI @willhite @ehalpern

PR #3770, opened 2017-10-20
Whoops, make 'noms stats' work for HTTP

cmasone-attic:


cmasone-attic:
PTAL verified locally that this works against `noms serve` now

Issue #3747, opened 2017-09-29
SimplifyType is too expensive


rafael-atticlabs:
Now that the dust is starting to settle after changing values to be backed by bytes, SimplifyType is rising the top of alot of profiles.

The algorithm is unavoidably costly. I think a good kind of approach here is to try to avoid *recomputing* a simplified type that we've already computed a bunch (e.g. when creating large collections)

rafael-atticlabs:
Thinking about this more, I think I can imagine three approaches to mitigating the cost of simplify type:

1) Find some way to implement which is lost costly in terms of number of memory allocations
2) Memoize the work so that we don't keep recomputing the same simplified type for the same (or very similar) input types
3) Avoid doing it at all if the type doesn't need to be simplified

rafael-atticlabs:
I think maybe a good first place to start is (3) above. I think there's probably a 60-80% base case here which is that SimplifyType gets called everytime the sequence chunker produces a new node in a prolly tree. What gets simplified is all of the elements of the sequence. It's probably more often than not that

a) Each element is *exactly* the same type
b) Any given element is already Simplified

I think simply detecting this case and avoiding the call to SimplifyType is a good place to start and will help with the major of cases.

PR #3772, opened 2017-10-23
Optimize typeOf for sequences and simplifyType

arv:
When we compute the type of a sequence we used to compute the type of
every value in the sequence and then simplify the union. Now we check
if the type of the value is the same as the type of the last element, in
which case we do not need to include it in the union. This is the common
case and it allows us top skip allocating a lot of Types and it makes
the Type to simplify smaller which makes simplifying the type faster.

Also, handle the simple cases where the type is already simplified.

Optimize `*Type` `Equals` to reduce the number of times we have to
encode it.

Introducing `makeUnionType` that removes the `Union<T>` earlier so that
there is less work to do in `simplifyType`.

This halves the amount of time spent in TypeOf for csv import, going
from 18 MB/s to 27 MB/s

Towards #3710, #3747

### DIFF
--- a/go/types/make_type.go
+++ b/go/types/make_type.go
@@ -31,7 +31,7 @@ func MakePrimitiveType(k NomsKind) *Type {
 
 // MakeUnionType creates a new union type unless the elemTypes can be folded into a single non union type.
 func MakeUnionType(elemTypes ...*Type) *Type {
-	return simplifyType(makeCompoundType(UnionKind, elemTypes...), false)
+	return simplifyType(makeUnionType(elemTypes...), false)
 }
 
 func MakeListType(elemType *Type) *Type {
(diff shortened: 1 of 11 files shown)


go/types/make_type.go
```diff
@@ -84,6 +84,13 @@ func makeCompoundType(kind NomsKind, elemTypes ...*Type) *Type {
 	return newType(CompoundDesc{kind, elemTypes})
 }
 
+func makeUnionType(elemTypes ...*Type) *Type {
+	if len(elemTypes) == 1 {
+		return elemTypes[0]
+	}
+	return makeCompoundType(UnionKind, elemTypes...)
```
rafael-atticlabs: audit remaining usage of makeCompoundType

go/types/struct.go
```diff
@@ -41,6 +41,29 @@ func skipStruct(dec *valueDecoder) {
 	}
 }
 
+func isStructSameTypeForSure(dec *valueDecoder, t *Type) bool {
```
rafael-atticlabs: if this isn't going to return immediately in the false case, it should document why (i.e. to completely consume the struct from `dec`)

go/types/value_decoder.go
```diff
@@ -310,6 +310,67 @@ func (r *valueDecoder) readTypeOfValue() *Type {
 	panic("not reachable")
 }
 
+// isValueSameTypeForSure may return false even though the type of the value is
+// equal. We do that in cases wherer it would be too expensive to compute the
+// type.
+func (r *valueDecoder) isValueSameTypeForSure(t *Type) bool {
```
rafael-atticlabs: open bugs for the cases that aren't covered here?

PR #3775, opened 2017-10-24
Compute length of collection when creating it

### LINKED ISSUE #3759: `Collection.Len()` should be cheap
It's very likely that API users will assume it is. E.g. `for i := 0; i < l.Len(); i++`.

Right now it does some decoding for each call.

arv:
Also, clean up value decoder to reduce code duplication.

Fixes #3759

### DIFF
--- a/go/types/value_decoder.go
+++ b/go/types/value_decoder.go
@@ -41,169 +41,132 @@ func (r *valueDecoder) skipRef() {
 	skipRef(&(r.typedBinaryNomsReader))
 }
 
-func (r *valueDecoder) skipBlobLeafSequence() []uint32 {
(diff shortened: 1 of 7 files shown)


go/types/value_decoder.go
```diff
@@ -41,169 +41,132 @@ func (r *valueDecoder) skipRef() {
...
+		return setLeafSequence{seq.(leafSequence)}
 	}
-	end := r.pos()
+	return seq.(orderedSequence)
+}
 
-	if level > 0 {
-		return newMetaSequence(r.vrw, r.byteSlice(start, end), offsets)
+func (r *valueDecoder) readMapSequence() orderedSequence {
+	seq := r.readSequence(SetKind, r.skipMapLeafSequence)
```
kalman: MapKind?

PR #3777, opened 2017-10-27
 Remove hack for meta tuples of refs

### LINKED ISSUE #1688: Prolly trees: dont serialize full `Ref` of max value for ordered meta sequences
all we need is the `Hash`, and for sets/maps of inlined keys, the type information could be very large

arv:
Instead introduce a new NomsKind (called hashKind) which is internal
only.

Fixes #1688

Update version to 7.18

### DIFF
--- a/go/constants/version.go
+++ b/go/constants/version.go
@@ -10,7 +10,7 @@ import (
 	"os"
 )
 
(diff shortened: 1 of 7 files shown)


cmasone-attic:
LGTM

PR #3779, opened 2017-10-28
Update intro.md

aboodman:
This is more what I meant originally. Some of these overlap now though and it should probably be rewritten.


aboodman:
Whoops, sorry @arv, will update to take into account your comment.

PR #3780, opened 2017-10-28
ipfs-chat

willhite:
Updated go-ipfs and gx dependencies.
Bug fixes in ipfs chunkstore.
Make termio scrolling more efficient.


go/ipfs/cs.go
```diff
@@ -201,6 +200,7 @@ func (cs *chunkStore) HasMany(hashes hash.HashSet) hash.HashSet {
 		wg := sync.WaitGroup{}
 		wg.Add(len(hashes))
 		for h := range hashes {
+			h := h
```
aboodman: oh god ... what was it even doing before? I guess we got lucky and HasMany() wasn't frequently used with multiple chunks?

willhite: I think it was using the same h for most of the hashes. So many of them weren't actually checked.

Issue #3782, opened 2017-11-13
broken submodules (breaks glide and dep)


tobowers:
When trying to use glide or the "official" go dep manager I get error message similar to `76400c6b710112a3ec35cb963c3eea1bb13d9dd3: unexpected error while defensively updating submodules: : command failed: [git submodule update --init --recursive]: exit status 1` (glide is slightly different, but also fails on the submodule update)

Performing the same command in the directory:

```
$ git submodule update --init --recursive
No submodule mapping found in .gitmodules for path 'vendor/gx/ipfs/QmcRKRQjNc2JZPHApR32fbkZVd6WXG2Ch9Kcy7sPxuAJgd/cbor/test-vectors'
```

fische:
`go get` also fails because of submodules:
```
$ go get -u github.com/attic-labs/noms
# cd $GOPATH/src/github.com/attic-labs/noms; git submodule update --init --recursive
fatal: No url found for submodule path 'vendor/gx/ipfs/QmcRKRQjNc2JZPHApR32fbkZVd6WXG2Ch9Kcy7sPxuAJgd/cbor/test-vectors' in .gitmodules
package github.com/attic-labs/noms: exit status 128
```

tobowers:
@fische I just opened a PR to fix it #3784

senorsmile:
Just trying the project out for the first time.  This is hitting me too trying ```go get``` .

What's the work around until someone merges?

PR #3784, opened 2017-11-14
remove git link folder preventing submodule update

tobowers:
As mentioned in #3782, this folder makes it so you can't use noms in a vendor manager (at least deps, glide, and godeps). It's a simple fix ( read about it here: https://stackoverflow.com/questions/4185365/no-submodule-mapping-found-in-gitmodule-for-a-path-thats-not-a-submodule/13394710#13394710 ). 

Hopefully we can get this one merged quickly?


arv:
This probably needs to be upstreamed.

I'm looking for it in https://github.com/ipfs/go-ipfs but I cannot find it.

@willhite Maybe we just need to update our ipfs deps?

tobowers:
well - you might need to update the deps, but I think you need to merge this too... it's a "special mode" 16000 folder in the git repo, but there is no .gitmodule file... so it's definitely "broken" (checkout the stack overflow link)

arv:
I'm OK merging this as is if it unblocks you. The change will be clobbered if we update the deps again.

@tobowers I don't think you have signed the CLA. See https://github.com/attic-labs/noms/blob/master/CONTRIBUTING.md

podviaznikov:
I had the same problem when tryings to use `deps`. Mentioned it before in the slack channel

aboodman:
Closed in favor of #3794 since @tobowers doesn't have a CA on file.

PR #3785, opened 2017-11-15
Fix issue with marshal caching vrw

arv:
We were caching the encoder function enclosing over the passed in
ValueReadWriter. This causes problems when the encoder is used again but
with another ValueReadWriter.

Instead, make the internal encoder function take a ValueReadWriter.


arv:
FYI @tobowers

PR #3791, opened 2018-01-18
Update README.md

rafael-atticlabs:
The `bucket-db` use is aspirational enough that I think it's safe to remove it for now. In it's place, I've referenced the existence of NBS's ability to operate against an S3 bucket.


aboodman:
Hm, really? I thought it made a lot of sense, especially in that section on use cases.

rafaelweinstein:
If you feel strongly you want to keep it, I don't have a serious objection (I'm happy to revert).

That being said, for my taste, I think there's enough unfinished work in this direction (e.g. robust support for columnar storage) that we don't want to give the impression that it's presently practical to use it for this purpose.

PR #3790, opened 2018-01-10
WIP Introduce Dockerfile

stevepeak:
Introducing a Dockerfile for Noms. This will be the official image found in Docker Hub at [noms/noms](https://hub.docker.com/r/noms/noms).

- [ ] Need feedback fro Go expert to confirm this install strategy
- [x] Need to fill out the [Docker Hub listing](https://hub.docker.com/r/noms/noms/)
- [ ] Remove Sample projects
- [ ] Setup automated builds

> Close #3789

PR #3800, opened 2018-03-29
Add a multi-stage-build dockerfile

coriolinus:
This is substantially based on https://hub.docker.com/r/scjalliance/noms/
but has been updated in two ways:

 1. It uses a multi-stage build to substantially reduce the size of the
    resultant image. The naive build image clocks in at ~1gb; the output
    image here is ~30mb.
 2. It takes advantage of the fact that the build context is already
    present, so we can speed up the build process by simply copying
    the context instead of cloning the data again.

This is intended to be compatible with #3790, and does not address
any of the issues mentioned there except simple creation of a dockerfile.
If the source repository for that PR were publicly available, this PR
would have targeted that branch instead of master.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.git
+docs
+codecov.yml
+CONTRIBUTING.md
+LICENSE
+README.md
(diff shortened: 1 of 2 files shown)


aboodman:
LGTM, thanks for the contribution. You need to sign a contributor agreement as detailed here in order to accept the patch: https://github.com/attic-labs/noms/blob/master/CONTRIBUTING.md#contributing-code

coriolinus:
I signed one before submitting: corporate, for Oneiro.

Issue #1592, opened 2016-05-23
Allow API use of bignums in Go


ghost:
This is two stages:

1) Writing go bignum types
2) Reading numbers onto bignum types (which is probably blocked by https://github.com/attic-labs/noms/issues/1591

PR #3802, opened 2018-04-03
Add integer type

### LINKED ISSUE #3801: No integer type
The current noms numeric data type is backed by `float64`, so integers above 2<sup>53</sup> + 1 can't be represented

ORBAT:
Hi folks!

To fix #3801, I've started work on adding an integer type backed by `int64`.  This PR is more of an RFC, so the feature should not be considered finished.

What I'd like is some feedback on whether what I'm doing is

1. sensible
2. backwards-compatible
3. the correct way to do this

My current idea is to not do `uint*` support, at least for the time being. It might be a good idea to add that while I'm at it, but this has turned out to be a, uh, non-trivial change.

Still to do (at least):

- [ ] diff code & tests
- [ ] rest of the marshaling tests (+ code if needed)
- [ ] CSV schema-related code & tests
- [ ] GraphQL-related code & tests
- [ ] `TestMapSuite4KStructs/TestChunkCountAndType`. Why did chunk count go from expected 11 to actual 14 without any changes to the test‽
- [ ] `TestMapMutationReadWriteCount`. Why did read count go from expected 105 to 206 without any changes to the test‽
- [ ] conversions/coercions?
- [ ] what to do with `DateTimeType`?

### DIFF
--- a/cmd/noms/splore/noms_splore.go
+++ b/cmd/noms/splore/noms_splore.go
@@ -308,7 +308,7 @@ func getMetaChildren(v types.Value) (children []nodeChild) {
 
 func nodeHasChildren(v types.Value) bool {
 	switch k := v.Kind(); k {
-	case types.BlobKind, types.BoolKind, types.NumberKind, types.StringKind:
+	case types.BlobKind, types.BoolKind, types.NumberKind, types.IntegerKind, types.StringKind:
 		return false
 	case types.RefKind:
 		return true
--- a/go/marshal/decode.go
+++ b/go/marshal/decode.go
@@ -153,8 +153,12 @@ func (e *UnmarshalTypeMismatchError) Error() string {
 	return fmt.Sprintf("Cannot unmarshal %s into Go value of type %s%s", types.TypeOf(e.Value).Describe(), ts, e.details)
 }
 
-func overflowError(v types.Number, t reflect.Type) *UnmarshalTypeMismatchError {
+func overflowErrorNum(v types.Number, t reflect.Type) *UnmarshalTypeMismatchError {
 	return &UnmarshalTypeMismatchError{v, t, fmt.Sprintf(" (%g does not fit in %s)", v, t)}
+}
+
+func overflowErrorInt(v types.Integer, t reflect.Type) *UnmarshalTypeMismatchError {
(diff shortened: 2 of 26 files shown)


aboodman:
So, actually the serialization of Noms' number type is currently arbitrary-precision, so it can represent integers. The only thing is that the *API* uses float64. What we should do instead is change the API to use some bignum abstraction (maybe optionally?).

I think my original thought was that you could marshal a Noms `Number` on to any of the builtin Go numeric types, or some other bignum type.

See https://github.com/attic-labs/noms/issues/1592

ORBAT:
Thanks for the answer, not to mention the patience.

> So, actually the serialization of Noms' number type is currently arbitrary-precision, so it can represent integers. The only thing is that the _API_ uses float64. What we should do instead is change the API to use some bignum abstraction (maybe optionally?).

D'oh, actually reading `writeNumber`'s code makes that obvious. OK, so, I've got a few further questions and comments: looking at [`writeNumber`](https://github.com/attic-labs/noms/blob/master/go/types/codec.go#L217-L224), the way I take it is that the mantissa of `Number` is effectively limited only by `PutVarInt` (so `int64`) if the serialization is to be backwards-compatible.

- is that method what you meant with `Number`'s serialization and it being arbitrarily precise (for all intents and purposes anyhow, with a max mantissa of `math.MaxInt64`)?
- would a better way to approach this be having `Number` be backed by `big.Float` (plus modifying `float64ToIntExp` et al in `number_util.go`), + needed changes to the number-related decoders&encoders in `marshal`?

Personally I'd still like to have a separate integer type (although the serialization could be the same as with `Number`), as it'd allow you to be explicit about whether you expect floating point numbers or integers, leading to fewer surprises and runtime problems (with eg. truncation) or outright crashes as the types in noms would map more closely to types used by the application, and something that you decode to an integer will always be "backed" by an integer.

Having a separate type while allowing to have guarantees about truncation could be avoided somewhat with e.g. struct tags (unmarshaling could have`notrunc`,`integer` or the like to make truncation an error, for example).  What are your views on providing an explicit `Int` type vs. some other solution with just `Number`?

aboodman:
Looking at this all again I'm wondering why we didn't choose to serialize as a rational number, since it can represent more values accurately :-/. Bummer.

aboodman:
@ORBAT I'm going to go ahead and close this in favor of eventually implementing something like: #3806

PR #3806, opened 2018-06-03
Proposal for new numbers support in Noms

aboodman:
There have been a number of request for better support for numeric types in Noms (#3801, #3802). With the benefit of some time away, I had some inspiration about how this should work.

I've created this start at a proposal. If you have feedback, please leave it here on the PR.

Eventually I will commit this in the proposal directory, and then (given energy/motivation) implement it and move the doc up to he docs directory when it is complete.

### DIFF
--- a/doc/proposals/numbers.md
+++ b/doc/proposals/numbers.md
@@ -1,0 +1,91 @@
+# It's time to fix numbers in Noms
+
+Currently numbers in Noms are are stored as arbitrary precision decimal numbers.
+
+However, the API represents numbers as `float64`, so it is not possible to store even 64-bit integers in Noms accurately today.
+
+The last time we looked at this, we concluded we didn't have a clear enough idea of the desired properties for numbers, so we decided to wait for more information.
+
+It's now clear to me that desiderata include:
+
+## Use cases
+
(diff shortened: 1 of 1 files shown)


aboodman:
PS, in case you didn't know, you can view the PR in formatted form here:

https://github.com/attic-labs/noms/blob/ce02c04cfe986a09634054cf0ad706b62711a7a4/doc/proposals/numbers.md

doc/proposals/numbers.md
```diff
@@ -0,0 +1,183 @@
...
+Modify Noms to support arbitrarily large and precise rational numbers using the following conceptual representation:
+
+```
+(np * 2^ne) / (dp * 2^de)
+```
+
+Where:
+
+- `np`: numerator precision (signed integer)
+- `ne`: numerator exponent (unsigned integer)
```
aboodman: Shoot, `ne` needs to be able to go negative, for values smaller than 0.5.

aboodman:
I think there are a bunch of problems with this, so I'm going to back up and start at desired features in a bug.

aboodman:
Collecting requirements here: https://github.com/attic-labs/noms/issues/3807

PR #3810, opened 2018-08-07
Using Dockerhub as a basic CI

kav:
Regarding #3792 
Wanted to start the discussion on picking the low hanging fruit here. 

- Are there any other steps that should go into CI at this point? Linting, formatting, etc.
- Is it important to run on PRs ala Travis? If so can we just run docker build as our travis script?

If we'd like to do that I'm happy to write the `.travis.yml` if someone can enable travis for the repo

Note: tests have two actual failures at this point so this will fail to build at dockerhub. Accepting this pull as is though will show that failures do get correctly reported in the badge so should perhaps proceed before fixing the two issues.

Even if we move to travis and don't use docker build as the script, which is also fine, it might be worth considering accepting this change so that we don't publish broken code to the `latest` tag on dockerhub.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 .git
-docs
+doc
 codecov.yml
 CONTRIBUTING.md
 LICENSE
 README.md
+samples
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,9 @@ ENV GOOS=linux
 
 RUN mkdir -pv $NOMS_SRC
 COPY . ${NOMS_SRC}
+RUN cd $NOMS_SRC/cmd/noms; \
+  go build; \
+  go test
 RUN go install -v github.com/attic-labs/noms/cmd/noms
 RUN cp $GOPATH/bin/noms /bin/noms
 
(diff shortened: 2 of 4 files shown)


aboodman:
It'd be cool to test PRs. And it looks like Travis is free for open source projects.

Are there any downsides to using Travis that you know of? It's been awhile since I've had to think about it. If not, I'll set it up today/tonight.

I agree it still makes sense to land this for the reason you mention.

===

Side note: in order to land this, I need you to complete the contributors agreement: https://github.com/attic-labs/noms/blob/master/CONTRIBUTING.md#contributing-code

aboodman:
PS: Thanks for the contributions!

kav:
I've not used Travis a lot, been doing more with Drone lately, but Travis still seems pretty handy given the free for OSS and hosted nature of it all.

.dockerignore
```diff
@@ -1,6 +1,7 @@
 .git
-docs
+doc
 codecov.yml
 CONTRIBUTING.md
 LICENSE
 README.md
+samples
```
aboodman: Nit, but add the trailing newline.

Dockerfile
```diff
@@ -3,9 +3,11 @@ FROM golang:latest AS build
 ENV NOMS_SRC=$GOPATH/src/github.com/attic-labs/noms
 ENV CGO_ENABLED=0
 ENV GOOS=linux
+ENV NOMS_VERSION_NEXT=1
```
aboodman: Remove the dupe line further down.

aboodman:
Nits PR here: https://github.com/attic-labs/noms/pull/3811/files

PR #3811, opened 2018-08-09
nits from pull #3810

aboodman:


Dockerfile
```diff
@@ -18,7 +18,6 @@ COPY --from=build /bin/noms /bin/noms
 VOLUME /data
 EXPOSE 8000
 
-ENV NOMS_VERSION_NEXT=1
```
kav: This is adding that variable in the second container. They both need it

PR #3813, opened 2018-11-13
Enable to stop in struct's IterFields method

sairoutine:
As well as list and map, I would like to write the following the code by struct's iteration method.
```
var error err
str.IterFields(func(name string, value types.Value) bool {
    err = SomeMethod()
    if err != nil {
        return true
    }
    return false
})

if err != nil {
    return err
}
```

### DIFF
--- a/go/types/struct_test.go
+++ b/go/types/struct_test.go
@@ -330,3 +330,39 @@ func TestStructWithNil(t *testing.T) {
 		})
 	})
 }
+
+func TestStructIterFields(t *testing.T) {
+	assert := assert.New(t)
+
+	tstruct := NewStruct("A", StructData{
+		"a": String("aaa"),
+		"b": String("bbb"),
+		"c": String("ccc"),
+	})
+
+	// Iterate over all.
+	i := 0
+	tstruct.IterFields(func(k string, v Value) bool {
+		assert.True(tstruct.Get(k).Equals(v))
+
+		i += 1
+
+		return false
+	})
+
+	assert.Equal(3, i)
+
+	// Iterate and stop.
+	i = 0
+	tstruct.IterFields(func(k string, v Value) bool {
+		if k == "b" {
+			return true
+		}
+
+		i += 1
+
+		return false
+	})
+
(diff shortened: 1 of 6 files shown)


go/types/struct_test.go
```diff
@@ -330,3 +330,39 @@ func TestStructWithNil(t *testing.T) {
...
+	i := 0
+	tstruct.IterFields(func(k string, v Value) bool {
+		assert.True(tstruct.Get(k).Equals(v))
+
+		i += 1
+
+		return false
+	})
+
+	assert.True(i == 3)
```
aboodman:
```suggestion
	assert.Equal(3, i)
```

go/types/struct_test.go
```diff
@@ -330,3 +330,39 @@ func TestStructWithNil(t *testing.T) {
...
+		if k == "b" {
+			return true
+		}
+
+		i += 1
+
+		return false
+	})
+
+	assert.True(i == 1)
```
aboodman:
```suggestion
	assert.Equal(1, i)
```

sairoutine:
I finished the agreement.

go/types/struct_test.go
sairoutine: Fixed

go/types/struct_test.go
sairoutine: Fixed

sairoutine:
It seems that it is not the matter of this pull request that fails the test of Travis.

sairoutine:
@aboodman 
Thank you for your review. I finished fixing your pointed out.
Please review again.

arv:
Wasn't there some performance issues with assert.Equal?

aboodman:
@arv is right - I forgot about that perf issue. I will land the version of the change that you originally posted.

Thanks for the patch!

aboodman:
Actually the usage here is in a test file, so the perf is irrelevant. Landed as-is.

sairoutine:
Thank you for your merging!

PR #3818, opened 2018-12-24
The screencast does not seem to exist in youtube

eperez:


arv:
@aboodman Which video was this pointing at? 

https://www.youtube.com/watch?v=bfoEGdxseVg
https://www.youtube.com/watch?v=womWInzG- 
https://www.youtube.com/watch?v=ncITL4xdXh4

aboodman:
It wasn't the first link or the third link. It was a screencast that had did the same things listed on that doc page. The second link you pasted doesn't work. I think perhaps the account that had the screencast went away.

OTOH, I think the third link you pasted is pretty neat and has some new stuff, so maybe we should just use that?

aboodman:
The third link also has the advantage of not relying upon servers being alive to have data in them.

arv:
The second should be https://www.youtube.com/watch?v=womWInzG-00

aboodman:
Definitely not that one. I think the target was associated with an account
that went away with the acquisition of Attic Labs, but not sure tbh.

aboodman:
Thanks for the report @eperez - I'll take care of this.

PR #3869, opened 2019-12-04
Do not ignore empty collections in ListEditor

### LINKED ISSUE #3868: nomdl.Parse(): Incorrect result
`nomdl.Parse("[map {}]")` should return `types.NewList(vs, types.NewMap(vs)))`, but it actually returns an empty list.

arv:
We used to ignore empty collections inside a ListEditor which meant that
a list editor could not be used to create lists that contain empty
lists, maps or sets.

Fixes #3868

### DIFF
--- a/go/nomdl/parser_test.go
+++ b/go/nomdl/parser_test.go
@@ -282,11 +282,27 @@ func TestValueList(t *testing.T) {
 	assertParseError(t, "[,]", "Unexpected token \",\", example:1:3")
 
 	assertParse(t, vs, `[42,
-                Bool,
-        ]`, types.NewList(vs, types.Number(42), types.BoolType))
+	            Bool,
+	    ]`, types.NewList(vs, types.Number(42), types.BoolType))
 	assertParse(t, vs, `[42,
-                Bool
-        ]`, types.NewList(vs, types.Number(42), types.BoolType))
+	            Bool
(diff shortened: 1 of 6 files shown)


arv:
This passes on Travis... I'm seeing test failures on master locally though?

aboodman:
LGTM. What test failure are you seeing?

arv:
```
FAIL	github.com/attic-labs/noms/go/perf/suite
```

I have a slight memory of it being problematic 2 years ago

arv:
> Do we need `Emptyable` after this?

Not sure. Will look into it.

aboodman:
I think we can remove the `Emptyable` interface, but leave the `Empty()`
methods. See also:
http://neugierig.org/software/blog/2019/11/interface-pattern.html

arv:
I'm still curious to what might break in Attic with this change. If I understand correctly the attic processing pipeline depended on this behavior. To me that seems like a breach of abstraction and the attic pipeline should have been able to deal with this.

aboodman:
I mean attic doesn't exist anymore so it's irrelevant. But it sounds like Attic UI (or pipeline) was relying on the fact that removing the last item from a collection would also remove the collection itself from its parent, and so-on, recursively.

aboodman:
You can land this whenever you're ready, thanks!

PR #3872, opened 2020-04-09
fix mapeditor bug

phritz:
MapEditor was erroneously going to the underlying map if the edit value was nil. The edit value will be nil when it has been removed (and so we shouldn't go to the underlying map).

### DIFF
--- a/go/types/map_test.go
+++ b/go/types/map_test.go
@@ -547,6 +547,17 @@ func TestMapHas(t *testing.T) {
 	doTest(getTestRefToValueOrderMap, 2)
 }
 
+func TestMapRemoveMasksUnderlyingMap(t *testing.T) {
+	assert := assert.New(t)
+	vrw := newTestValueStore()
+
+	k := String("foo")
+	me := NewMap(vrw, k, String("bar")).Edit()
+	me.Remove(k)
+	assert.False(me.Has(k))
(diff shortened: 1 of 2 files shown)


aboodman:
Eep. It was the way this change makes it originally, then it was changed, here:

https://github.com/attic-labs/noms/pull/3557/files#diff-9eaeaf4074255abb6cfef108897c6261R163

... as part of introducing support for "nested" ediotrs (e.g., you can have a value in an editor itself be another editor).

Can you see if there was some reason to add this branch that is now broken (which apparently doesn't have a test) or if it was just a mistake to introduce it originally?

Thanks!

go/types/map_test.go
```diff
@@ -547,6 +547,17 @@ func TestMapHas(t *testing.T) {
...
 }
 
+func TestMapRemoveMasksUnderlyingMap(t *testing.T) {
+	assert := assert.New(t)
+	vrw := newTestValueStore()
+
+	k := String("foo")
+	me := NewMap(vrw, k, String("bar")).Edit()
+	me.Remove(k)
+	assert.False(me.Has(k))
```
arv, line 557: Can you also add 2 more cases:

 1. Add a key-value-pair to the map editor and remove that.
 2. Add a key-value-pair where the value is another editor and remove that.

PR #3873, opened 2020-04-10
fix mapeditor bug

arv:
Finishing #3872


arv:
FYI @phritz @aboodman

PR #3875, opened 2020-06-19
configurable listen address for db server

nustiueudinastea:
Hi folks, this is a small change that allows the user to configure the listening address for the db server, besides the port.

### DIFF
--- a/cmd/noms/noms_serve.go
+++ b/cmd/noms/noms_serve.go
@@ -19,6 +19,7 @@ import (
 
 func nomsServe(noms *kingpin.Application) (*kingpin.CmdClause, util.KingpinHandler) {
 	cmd := noms.Command("serve", "Serves a Noms database over HTTP.")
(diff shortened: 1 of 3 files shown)


aboodman:
Thank you for this contribution! Looks good.

You need to sign the contributor agreement, then I can land it:

https://github.com/attic-labs/noms/blob/master/CONTRIBUTING.md

nustiueudinastea:
> Thank you for this contribution! Looks good. 
> 

My pleasure! Thanks for this awesome piece of software.

> You need to sign the contributor agreement, then I can land it:
> 
> https://github.com/attic-labs/noms/blob/master/CONTRIBUTING.md

Done


I want to submit a couple of other PRs:
- update to the latest go and deps. 
- replace log.Printf lines with some configurable logger. When using noms as a library the logging is inconsistent because noms uses its own logger.

Do you think these types of PRs would be accepted?

aboodman:
Yes, they would be accepted. We want those changes too. @phritz

